Issue #173, opened 2021-06-23
As Poppy, I need to see a standard Innovation footer on all pages


skeeJay:
This is the "Innovation footer" that sits on all pages on the site. The legal message that sits just above it on certain pages is ticket #155.

Design for this Innovation footer is being provided by Drawing Board ticket #172. 

**Design**
[Landing & Dashboard laptop](https://www.figma.com/proto/Hg3ggwTj50wxdg4mjG4xRB/Dashboard-Exploration?page-id=2654%3A0&node-id=2787%3A6384&viewport=-2988%2C186%2C0.35706689953804016&scaling=min-zoom)
[Landing & Dashboard mobile](https://www.figma.com/proto/Hg3ggwTj50wxdg4mjG4xRB/Dashboard-Exploration?page-id=488%3A3970&node-id=2870%3A37568&viewport=1489%2C1536%2C0.2498604655265808&scaling=min-zoom)

**User Actions**
@skeeJay says: these actions have been validated:
- [x] **Give Us Feedback**: opens https://forms.business.nj.gov/feedback/ in a new tab - same behavior as "Let us know what you think!" in the beta bar
- [x] **Terms & Privacy Policy**: opens https://business.nj.gov/privacy-policy in a new tab
- [x] **Sign Up for Our Newsletter**: opens https://business.nj.gov/newsletter-signup in a new tab
- [x] **Office of Innovation**: opens https://innovation.nj.gov/ in a new tab
- [x] **Open Source**: opens https://business.nj.gov/license-and-copyright-notice in a new tab
- [x] **Github** icon: opens https://github.com/newjersey in a new tab
- [x] **Twitter** icon: opens https://twitter.com/i/lists/1148625562895826945 in a new tab
- [x] All footer links use the [Links](https://newjersey.github.io/njwds/components/detail/links.html) component but remove the underline in the default state.
- [x] Social media icon style: default color base-lighter (#E6E6E6); hover color white(#FFFFFF)

Implementation at the developer's discretion, but something to keep in mind is that we may someday want to treat this footer as an "include" that would also appear at the bottom of Business.NJ.gov, innovation.nj.gov, and other Innovation sites.

skeeJay:
@maitanavarro @aloverso This design is approved at this point, and the "needs design" label has been removed. It should be ready to implement. Let us know if there are any questions.

guy-nj:
I have added a note of the default and hover states for the Footer links. @maitanavarro @skeeJay

[...]

PR #185, opened 2021-07-12
[#173] add footer on all pages (WIP)

maitanavarro:


web/components/InnovFooter.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export const InnovFooter = (): ReactElement => {
+    return (
+        <div className="usa-identifier">
+        <div className="display-flex flex-column margin-x-auto fac padding-top-2 footer-custom-desktop">
+
+        <section className="usa-identifier__section usa-identifier__section--usagov " aria-label="U.S. government information and services">
+            <div className="usa-identifier__container">
+                
+        <p className="footer-container footer-container-mobile">
```
aloverso, line 14: instead of having both these classes, you should be able to just use a single class, and use css media queries to describe the mobile behavior. let me know if you have questions on this

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
+.footerLinkCustom {
```
aloverso, line 1: let's try to keep consistent on just using kebab case for classes (class-names-with-hyphens)

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
...
+    align-items: center;
+}
+
+.display-block-mobile {
+    display: block;
+}
+
+.footer-divider-top {
+             border-top:1px solid #757575;
+                ;
```
aloverso, line 45: a little wonky spacing here?

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
...
+             width: 416px;
```
aloverso, line 46: this seems strangely specific. why are we hardcoding it?

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
...
+
+.heartCustom {
+    color: #DD3534
+}
+
+.footerCustom {
+    max-width: 64rem;
+}
+
+.margin-x-auto {
```
aloverso, line 24: isn't this is njwds already?

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
...
+    margin-left: auto;
+    margin-right: auto;
+}
+
+.margin-right-none {
```
aloverso, line 29: seems like this isn't used?

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
...
+
+        .footer-divider-top {
+             border-top:1px solid #757575;
+                ;
+             width: 550px;
+            }    
+        
+        .footer-container {
+            display: flex;
+            flex-direction: row !important;
```
aloverso, line 66: all throughout this section - I wouldn't expect `!important` to be necessary for a class we define for ourselves. can we remove them?

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
...
+             border-top:1px solid #757575;
+                ;
+             width: 416px;
+            }    
+
+.footer-divider-bottom {
+    border-bottom:1px solid #757575;
+}
+
+@media only screen and (min-width: 1024px) {
```
aloverso, line 53: at the top of `global.scss` we have page size variables defined so that we don't ever reference by pixel value directly

web/styles/innov-footer.scss
```diff
@@ -0,0 +1,89 @@
...
+    font-weight: 400 !important;
+}
+
+.footerTextCustom {
+    font-weight: 400 !important;
+}
+
+.githubCustom {
+    color:#E6E6E6;
+    width:40px !important;
```
aloverso, line 12: same comment as below - I'm surprised by all the `important!` on our own classes. I wouldn't expect to need important except in being very intentional about parts of njwds to override. can they be removed?

aloverso:
Out of date, has been merged

Issue #141, opened 2021-05-18
As Poppy, the URL in the address bar should update as I change screens during onboarding


skeeJay:
@aloverso This is dealer's choice on whether the specific URL for each onboarding page should then allow you to access that onboarding page. It is unnecessary for this user story. But would probably enable #97 going forward.

Activities include:
- [ ]

## Dev notes
Use a query parameter (something like `onboarding?page=1`)

skeeJay:
I obviously wasn't specific at all about what the URLs would be, so this has been implemented totally in line with the (non-existent) acceptance criteria. I am just wondering if there was a specific reason that the URLs were implemented with a query string instead of giving each a path as all the Tasks have; was there an implementation reason, or just stylistic?

aloverso:
Good question. Not a _particular_ reason, I think it's just typical stylistic? I think of a full route like `/onboarding/1` or `/onboarding/2` as more of an independent page. Whereas for us, the "pages" are just pagination so it felt like query params on a single route was more appropriate. I think it'd be fine to implement as a path if that's what was wanted. Maybe as a separate story to address the change?

skeeJay:
I don't think we need to change unless there's a compelling reason; I sort of agree that as "numbered" steps they don't need a "full" URL like the named Task screens do. Certainly, this satisfies the need for the analytics. I assume that this _doesn't_ also implement direct access to each onboarding step with its unique URL? That would probably need to be implemented in the future as part of #97.

aloverso:
Oh! Oops, I didn't realize this was a separate story. No, this story includes being able to type a specific page into your browser and go to a specific page. We'll have to use #97 to actually implement links to specific pages according to design though

aloverso:
Although, this makes me realize it's actually kind of breaking #138 a little bit? What we implemented in #138 prevents the user, for example, from moving from step 3 to step 4 if they haven't selected a legal structure. But now due to this, they could just replace the `page=3` with `page=4` in their URL and get access to the next page and the roadmap without picking a legal structure.

I can think of a couple ways to address this but what do you think?

skeeJay:
We'll have to come up with a way to address that "hack" in a future user story. For "regular" users who we'll be conducting testing with, I think the UX solution of not letting them progress satisfies our requirements.

PR #187, opened 2021-07-13
[#141] onboarding url page change (WIP)

maitanavarro:


web/pages/onboarding.tsx
```diff
@@ -95,10 +132,20 @@ const OnboardingPage = (props: Props): ReactElement => {
 
   const onBack = () => {
     if (page.current + 1 > 0) {
+      const previousPage = page.current - 1;
       setPage({
-        current: page.current - 1,
+        current: previousPage,
         previous: page.current,
       });
+      router.push(
```
aloverso, line 140: we use this method so often in the code now, maybe we could extract this whole "router shallow push with query" into a function that we can reuse, to clean up/simplify the code a bit?

aloverso (changes requested):

web/test/pages/onboarding.test.tsx
```diff
@@ -44,6 +44,41 @@ describe("onboarding form", () => {
     useMockRouter({});
   });
 
+  it("changes url pathname every time a user goes to a different page", async () => {
+    subject = render(
+      <Onboarding displayContent={createEmptyOnboardingDisplayContent()} municipalities={[]} />
+    );
+    expect(subject.getByTestId("step-1"));
```
aloverso, line 51: for all these expects in all the tests, it looks like we're not expecting anything?
should have a `.toBeInTheDocument()` or something like that on them all

aloverso:
@maitanavarro This looks good to me! You can ship-it on the main branch. If you just change the commit message to remove the "WIP" before shipping, that'd be great.

Issue #182, opened 2021-07-08
As Poppy, I need to see additional status types on license Task screen


guy-nj:
Activities include:

- [x] We need to map all possible license statuses in the database:
  - Active
  - Pending
  - Expired
  - Barred
  - Out of Business
  - Reinstatement Pending
  - Closed
  - Deleted
  - Denied
  - Voluntary Surrender
  - Withdrawn

### Todo:

- [x] Backend should return the status of the _most recent_ entry that we get back from searching license statuses
- [x] Frontend should apply Grey Theme to any status that is not Active or Pending 
- [x] [Design QA] Change the green checkmarks to grey (same grey as the exclamation) for grey theme's application item list. 

### Dev notes:

Frontend:

- [x] We control all the theme components with the PermitThemes defined in `LicenseStatusReceipt`
- [x] The theme values are css classes. If they don't exist in njwds, we create them in `license-status.scss`
- [x] Most of the receipt component assumes the status is only either Active or Pending, this needs to be fixed all over the component
- [x] LicenseStatus type will now only be used on LicenseStatusItems (checklist items). A new type will be needed to represent the overall LicenseStatus, which will have to expand to include a dozen more statuses

Backend:

- [x] LicenseEntity type should allow any string as the license status
- [x] LicenseStatus type will now only be used on LicenseStatusItems (checklist items). A new type will be needed to represent the overall LicenseStatus, which will have to expand to include a dozen more statuses
- [x] Write a db migration for the new userData types (because of new license statuses) even though the migration won't have any logic in it
- [x] in `searchLicenseStatusFactory`:
  - [x] `const match = entities.find...` should find the MOST RECENT license in there, according to spec below
  - [x] `const match = entities.find...` allow finding an expired license if that's the most recent one
  - [x]  will need to use dayjs to parse the datestring in the weird format that we get from boomi
  - [x]  `status: match.licenseStatus === "Active" ? "ACTIVE" : "PENDING",` will now need to figure out from `match.licenseStatus` what the actual status should be (yknow one of like 12 options or something)
- [ ] "most recent" can be determined by the date of the record. If it has an `issueDate`, use that. If not, use `dateThisStatus` (which needs to be added to the Entity type). If it does not have that, use `expirationDate`.

**Design**
1. Use the exact status from the database to show in our Permit Status (for now).
2. For "Grey theme" statuses, the task status would display "NOT STARTED".
3. The task status would display "PENDING" for "Reinstatement Pending" application status.
[...]

PR #199, opened 2021-07-21
[#182] show additional license types on license Task screen

maitanavarro:


api/src/domain/license-status/searchLicenseStatusFactory.test.ts
```diff
@@ -97,13 +111,67 @@ describe("searchLicenseStatus", () => {
...
+        licenseStatus: "Expired",
+        issueDate: undefined,
+        dateThisStatus: "20200405 000000.000"
+      }),
+      generateLicenseEntity({
+        applicationNumber: "12345",
+        checklistItem: "Item 2",
+        checkoffStatus: "Completed",
+        issueDate: undefined,
+        dateThisStatus: "20200405 000000.000"
```
aloverso, line 151: having this entity stubbed seems irrelevant to the test, does it provide extra value or just noise? seems like the test is comparing being able to select "item 1" vs "active application...." right?

although, I'm a bit confused as to why "Item 1" winds up being the one that matches. even though the other does not have an issue Date, presumably it does have another date that it falls back on? The fact that its date is not defined in the test is a bit confusing

api/src/domain/license-status/searchLicenseStatusFactory.test.ts
```diff
@@ -97,13 +111,67 @@ describe("searchLicenseStatus", () => {
...
+        },
+        {
+          title: "Item 2",
+          status: "ACTIVE",
+        },
+      ])
+    ); 
+  } );
+
+  it("returns a license if it doesn't have an issue date", async () => {
```
aloverso, line 129: In the vein of "tests as documentation of what the code does", I think a test that is more along the lines of explicitly "it uses dateThisStatus as the date of an entity if issueDate is undefined, and it uses expirationDate as the date of an entity if the dateThisStatus is undefined" (maybe slightly less wordy and/or split into multiple tests if needed) but the idea is to show/construct a situation where the test _cannot_ pass unless that exact logic is coded.

api/src/domain/license-status/searchLicenseStatusFactory.ts
```diff
@@ -15,15 +18,50 @@ export const searchLicenseStatusFactory = (licenseStatusClient: LicenseStatusCli
...
+        return d.valueOf() == mostRecentDate.valueOf();
+      })[0];
 
-    const match = entities.find(
-      (it) =>
-        cleanAddress(it.addressLine1).startsWith(cleanAddress(nameAndAddress.addressLine1)) &&
-        it.licenseStatus !== "Expired"
-    );
 
+<<<<<<< HEAD
```
aloverso, line 58: looks like some rebase merge-conflicts here that need to be resolved

web/components/tasks/LicenseStatusReceipt.tsx
```diff
@@ -37,33 +37,63 @@ const activePermitTheme: PermitTheme = {
...
+    : props.status === "WITHDRAWN" ? LicenseScreenDefaults.withdrawnPermitStatusText
+    : ""
+  }
 
   const getIcon = (status: LicenseStatus): string => {
     if (status === "ACTIVE") {
       return "check";
     } else if (status === "PENDING") {
       return "schedule";
+    } else if (
```
aloverso, line 85: could this just be an `else`?

web/components/tasks/LicenseStatusReceipt.tsx
```diff
@@ -37,33 +37,63 @@ const activePermitTheme: PermitTheme = {
...
 
   const getText = (): string => {
-    if (props.status === "ACTIVE") {
-      return LicenseScreenDefaults.activePermitStatusText;
-    } else if (props.status === "PENDING") {
-      return LicenseScreenDefaults.pendingPermitStatusText;
-    }
-    return "";
-  };
+    return props.status === "ACTIVE" ? LicenseScreenDefaults.activePermitStatusText
```
aloverso, line 66: might be nice to have these set up in a lookup table (much like we do those `Lookup` objects for the error texts)

web/components/tasks/LicenseStatusReceipt.tsx
```diff
@@ -37,33 +37,63 @@ const activePermitTheme: PermitTheme = {
...
+};
+
 export const LicenseStatusReceipt = (props: Props): ReactElement => {
   const [theme, setTheme] = useState<PermitTheme>(pendingPermitTheme);
   const { userData } = useUserData();
 
   useEffect(() => {
     if (props.status === "ACTIVE") {
       setTheme(activePermitTheme);
+      console.log(props.status);
```
aloverso, line 56: stray console log :)

api/src/domain/license-status/searchLicenseStatusFactory.ts
```diff
@@ -15,15 +18,50 @@ export const searchLicenseStatusFactory = (licenseStatusClient: LicenseStatusCli
...
+          var d = dayjs(it.issueDate, "YYYYMMDD X")
+        }
+        else if (it.dateThisStatus) {
+          var d = dayjs(it.dateThisStatus, "YYYYMMDD X")
+        }
+        else {
+          var d = dayjs(it.expirationDate, "YYYYMMDD X")
+        }
+        return d.valueOf() == mostRecentDate.valueOf();
+      })[0];
```
aloverso, line 55: although this approach works, I think taking another pass at trying a different algorithm might be useful here. happy to talk more in-depth about the pros and cons of different approaches, but here's the thought on this one: from an efficiency stance, it takes a lot of "passes" through the array.  Here's what I mean:
- we start with `entities` which has, say 1000 entities in the list (as a random number)
- we create `allMatchingAddressesArray` by filtering `entities`. Maybe `allMatchingAddressesArray`. a filter loops through the whole list, so it visits all 1000 items. and at the end, we get the matching ones, let's say it's 100 items
- we create `datesArray` by mapping across `allMatchingAddressesArray`. a map loops through the whole list, so it visits 100 items
- we use `max` to find the most recent date. this function also loops through the whole list to find the highest value, so it visits 100 items
- we find the `match` by filtering `allMatchingAddressesArray`. again, filter loops the whole list, so it visits 100 items.

That's a total of 4 passes through the list - one through the "big" list and one through the "small" list. the numbers themselves are made-up, but I'd challenge you to find a solution that takes fewer passes to find the most-recent match

web/components/tasks/LicenseStatusReceipt.tsx
```diff
@@ -37,33 +37,63 @@ const activePermitTheme: PermitTheme = {
   headerIconColor: "text-success",
 };
 
+const grayPermitTheme: PermitTheme = {
```
aloverso, line 40: This is a nitpick and I'm sorry but can we be consistent about 'grey' vs 'gray'? Personally I prefer 'grey' but it looks like njwds uses 'gray' so 🤷

aloverso (changes requested):
LGTM - one small wording fix, then you can ship-it to main :)

PR #246, opened 2021-08-18
refactor: switch config files to typescript

dsueltenfuss:


aloverso:
@dsueltenfuss it looks like there's a few different fixes in this commit - including changes in the pipeline related to renovate, and linting fixes for the "win32 fix. Could you clean up the commit & separate those changes out (if they need to be included on main?)

dsueltenfuss:
@aloverso I've removed those additional changes

Issue #221, opened 2021-07-29
As Poppy, I need to see Profile actions consolidated under a dropdown


guy-nj:
BLOCKED ON #220 (work on after)

**User Actions**
As Poppy,
3. I see a dropdown menu when I hover or click Profile. There are 2 subnav items: myNewJersey Account and Log Out.
4. Clicking myNewJersey Account brings me to my myNerJersey's "My Account" section in a new tab. @aloverso  to confirm URL. 
6. Clicking Log Out logs me out from Navigator and brings me back to the Landing page (@aloverso to confirm the destination page)

Dev Task

- [x]  Perform end-to-end feature testing to ensure navbar works in mobile

skeeJay:
LGTM.

PR #295, opened 2021-08-23
220 added nav bar to the roadmap/task pages

FarazA22:


FarazA22:
@aloverso Maita made some changed the AuthButton component, can you please take a second look to see if we need to keep any of the CSS class names on the button.

web/components/NavBarLoggedInDesktop.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <div className="grid-row grid-gap">
+        <div className="desktop:grid-col-12 usa-prose">
+          <header className="flex fac fjb">
+            <div className="flex-custom">
+              <img className="padding-top-1" src="/img/Navigator-logo.svg" alt="Business.NJ.Gov Navigator" />
+            </div>
+            <div className="margin-left-auto flex fac">
+              <span className="white-space-no-wrap padding-right-1">
+                <span>
+                  <Link href={""}>
```
aloverso, line 30: Is this link supposed to navigate somewhere?

web/next-env.d.ts
```diff
@@ -1,6 +1,3 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
-
-// NOTE: This file should not be edited
-// see https://nextjs.org/docs/basic-features/typescript for more information.
```
aloverso, line 6: Looks like this removal was an accident?

web/components/NavBarLoggedInDesktop.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const NavBarLoggedInDesktop = (): ReactElement => {
+  const { userData } = useUserData();
+
+  const getUser = (): string => {
+    return userData?.user.name
+      ? userData?.user.name
+      : userData?.user.email
+      ? userData?.user.email
+      : NavDefaults.userNotSetText;
+  };
```
aloverso, line 17: A few notes here:
- I wonder if this would be more readable as a simple if - else if - else instead of a double ternary 
- Same as for the file below (maybe both should make use of a helper method?)
- This is a type of logic I'd write a test for (uses name if exists, email if not exists, and empty text of userData itself is undefined)

aloverso:
@FarazA22 a few comments above

web/next-env.d.ts
FarazA22: This issue appeared after running the 'ship-it.sh' script from my local machine; I didn't alter this particular file. I've created chore #298 to track this issue and added it to the backlog. I'll bring it up during today's stand-up.

web/components/NavBarLoggedInDesktop.tsx
FarazA22: The link component and button that is inside the link will be replaced with a drop down menu (#221). It's not suppose to go anywhere. I can remove it, comment it out, or leave as is in this story - do you have a preference?

web/components/NavBarLoggedInDesktop.tsx
FarazA22: Created test, and created helper method.

web/components/NavBarLoggedInDesktop.tsx
aloverso: 🤷 If it's not a link to anything, I'd suggest just remove the link component? It won't become a link when the dropdown happens, right, so it seems irrelevant? Maybe I'm misunderstanding? 

As a rule, I would definitely never want commented-out code on main 🙃

FarazA22: I've removed the Link and updated the button to a div. In the future we may need to create button components for primary, secondary and tertiary buttons. The AuthButton for logout is using two different sets of styling (one set in the Hero component and another set in the Nav Bar).

FarazA22:
> @aloverso Maita made some changed the AuthButton component, can you please take a second look to see if we need to keep any of the CSS class names on the button.

I've made changes to fix this.

FarazA22:
@aloverso I have a question for how we are using the AuthButton component with the Hero Component (Landing Page). Currently, if the user is logged in and navigates to the landing page, instead of a 'log in' button, they will see a 'log out'. After a few seconds, the browser navigates to the roadmap/onboarding page. 

Is the Log In button within the hero suppose to be dynamic or static?

aloverso:
@FarazA22 It's dynamic based on whether the user is signed in, but the call to get whether the user is signed in or not is asynchronous, so on initial render it will not know and will display "Log in"

At the same time it figures out the user is signed in, it executes the code that says "a signed in user should be redirected to the roadmap" so it never has a chance to switch to the "Log out" state of the button

PR #357, opened 2021-08-30
Feature: SignUp Modal on redirect

mikewolfd:


aloverso:
In addition to the 2 small code comments, can we re-push these commits as a single commit (instead of separate commits for code/test) and with the commit message including the story number, `[#xxx] message here` style

Issue #352, opened 2021-08-30
As Jess, I need to reference steps by their number, not ID


aloverso:
Follow up from #146 as discussed with Jess

This will make it easier to build/change roadmaps without having to look up a step ID

For example, this should be a valid input:
```
 {
    "step": 5,
    "weight": 8,
    "task": "apply-for-shop-license"
  },
```

Jesslax:
@aloverso Looks great, all set. In the future, want me to move this over?

aloverso:
@Jesslax yep! if it's good, feel free to move over

PR #372, opened 2021-09-02
[#287] refactor buildRoadmap test to separately test behavior and logic

aloverso:


aloverso:
@mikewolfd merged this because I want to do #352 on top of it, but would still appreciate a look over the changes if you can!

PR #382, opened 2021-09-03
feat: switch to using parameter store for AWS infrastucture config

dsueltenfuss:


.github/workflows/pipeline.yml
```diff
@@ -147,16 +147,23 @@ jobs:
         run: npm run serverless config credentials -- --provider aws --key ${{secrets.AWS_ACCESS_KEY_ID_INNOV}} --secret ${{secrets.AWS_SECRET_ACCESS_KEY_INNOV}}
         working-directory: api
 
+      - name: Retrieve AWS configuration
```
aloverso, line 150: Unfortunately, this breaks our existing pattern of having each CI step name start with an emoji 🙁

dsueltenfuss: Added  the emjoi

Issue #305, opened 2021-08-25
"Next" and "Prev" buttons on Task screen


skeeJay:
…so that users don't get confused about where to go next at the bottom of each Task screen.

Design ticket #209 

[Next/Previous Buttons for Desktop here](https://www.figma.com/file/Hg3ggwTj50wxdg4mjG4xRB/Dashboard-Exploration?node-id=4755%3A100637) 

[Next/Previous Buttons for Mobile here](https://www.figma.com/file/Hg3ggwTj50wxdg4mjG4xRB/Dashboard-Exploration?node-id=4577%3A95484) 

Note: similar buttons do not exist in USWDS or NJWDS or original Navigator assets, so were customized in similar style.

Acceptance Criteria:
- Clicking "Next" advances to next task in that phase of the Roadmap; if no tasks follow, then the "Next" button is not displayed
- Clicking "Previous" returns the user to the previous task in that phase of the Roadmap; if no task precede this one, then the "Previous" button is not displayed

aloverso:
@carye3 can you link to the most recent agreed-upon designs (per our design review today) and then remove the needs-design tag?

carye3:
@aloverso latest designs updated in links above ([EJ's first comment](https://github.com/newjersey/business.nj.gov/issues/305#issue-979449643))

skeeJay:
I'll continue QAing, but I am seeing weird behavior on the first Step on the Roadmap, where the chevron isn't even inside the button, and I suspect the button is supposed to be right-aligned.

<img width="595" alt="Screen Shot 2021-09-10 at 7 25 56 AM" src="https://user-images.githubusercontent.com/506096/132846594-68f71728-ddde-46f5-afc2-c039be3b219c.png">

skeeJay:
This looks like it is implemented in line with the design. Flagging @carye3 for a final eye as well. I still remain a little concerned that conceptually users may not exactly understand where they are going "Next" (e.g., if you're on the business name search Task, I hope a user doesn't type a query into the field and then click "Next") but I suppose user testing will really tell us for real if there is any problem.

carye3:
Will be testing this today @skeeJay 

Looks good on site

skeeJay:
@mikewolfd Is there a reason this got moved back to Review?

PR #387, opened 2021-09-07
[#305] next & previous task buttons

mikewolfd:


web/test/pages/[urlSlug].test.tsx
```diff
@@ -220,4 +224,50 @@ describe("task page", () => {
...
+    it("renders only Next Task button for first task", async () => {
+      subject = renderPage(taskOne);
+      expect(subject.getByText(TaskDefaults.nextTaskButtonText)).toBeInTheDocument();
+      expect(subject.queryByText(TaskDefaults.previousTaskButtonText)).not.toBeInTheDocument();
+    });
+    it("renders only Previous Task button at last task", async () => {
+      subject = renderPage(taskThree);
+      expect(subject.queryByText(TaskDefaults.nextTaskButtonText)).not.toBeInTheDocument();
+      expect(subject.getByText(TaskDefaults.previousTaskButtonText)).toBeInTheDocument();
+    });
```
Comment (aloverso) [inline web/test/pages/[urlSlug].test.tsx:257]: was this run through prettier? I'm surprised it passed through without adding newlines between tests. can we add that?

Comment (aloverso) [reply web/test/pages/[urlSlug].test.tsx:257]: also a newline directly above the beforeeach

web/test/pages/[urlSlug].test.tsx
```diff
@@ -3,13 +3,15 @@ import * as materialUi from "@material-ui/core";
...
-import { generateOnboardingData, generateTask, generateUserData } from "@/test/factories";
+import { generateOnboardingData, generateStep, generateTask, generateUserData } from "@/test/factories";
 import { useMockRoadmap, useMockRoadmapTask } from "@/test/mock/mockUseRoadmap";
 import {
   currentUserData,
   setupStatefulUserDataContext,
   WithStatefulUserData,
 } from "@/test/mock/withStatefulUserData";
+import { TaskDefaults } from "@/display-content/tasks/TaskDefaults";
+import { mockPush, useMockRouter } from "../mock/mockRouter";
```
Comment (aloverso) [inline web/test/pages/[urlSlug].test.tsx:14]: can we use the @ style import here instead?

web/pages/tasks/[urlSlug].tsx
```diff
@@ -23,7 +26,40 @@ interface Props {
 const TaskPage = (props: Props): ReactElement => {
   useAuthProtectedPage();
 
+  const router = useRouter();
+
   const { roadmap } = useRoadmap();
+  const arrayOfTasks = getUrlSlugs(roadmap);
+  const currentUrlSlugIndex = arrayOfTasks.indexOf(props.task.urlSlug);
```
Comment (aloverso) [inline web/pages/tasks/[urlSlug].tsx:33]: this isn't really a review comment (ie, it shouldn't block merging this story) but these lines make me wonder if we should be using `useMemo` for these 4 lines so that we're not redoing that reduce (and the indexOf) every time the page re-renders

Issue #362, opened 2021-08-31
As Poppy, I need to see the mobile navigation menu close with an animation


mikewolfd:
...so that it is visually pleasing

The navigation menu in mobile currently opens with a css animation but does not close with one.

skeeJay:
@mikewolfd If this is intended for sprint 18, moving into the Current Sprint column.

skeeJay:
Out of curiosity, what easing function and timing does this animation have? Feels like we could make it a little faster/springier, but that's one person's opinion. cc @carye3 as well.

FarazA22:
@skeeJay @carye3 The animation is currently set at .5 seconds. The contextual info panel (side bar that appears on the right) can be used for comparison, it has a duration of .3 seconds.

I think it would make sense to use the same duration across both sidebars. 
Note:  Ticket #414 will add the close animation to the contextual info panel.

carye3:
@FarazA22 i vote for the shorter length as well -- .3 for both. I can add that duration to #414

FarazA22:
@skeeJay @carye3 Sounds good, I'll makes the duration updates under #414

skeeJay:
Before signing off on deploying to staging, I'm still curious: what easing function is this animation using? The current one feels linear, but that's obviously just a guess, so I wanted to find out. cc @FarazA22

skeeJay:
We are going with the "ease" animation that is currently on dev, overriding the "ease-in-ease-out" animation currently on staging, so I'm moving this to Ready to Deploy so that we can update staging.

PR #397, opened 2021-09-07
[#362] added close animation to mobile nav sidebar

FarazA22:


web/components/FocusTrappedSidebar.tsx
```diff
@@ -1,14 +1,27 @@
-import React, { ReactElement } from "react";
+import React, { ReactElement, useState, useEffect } from "react";
 import { onEscape, useMountEffect } from "@/lib/utils/helpers";
 import FocusTrap from "focus-trap-react";
 
 interface Props {
   children: React.ReactNode;
   close: () => void;
   isOpen: boolean;
+  delayTime?: number;
```
aloverso, line 9: Wouldn't we want this to apply to all sidebars? I guess the other sidebar comes from the other side of the page so I don't know if that would make it more complicated though? @mikewolfd @FarazA22

Issue #386, opened 2021-09-07
Accessibility issues [ongoing]


mikewolfd:
## Issues
- [ ] when on a task page and navigating to the next task, the screen reader is reading the previous task title
- [ ] some issue with the intercom container and the "main" landmark element? (maybe an issue with how the Div element renders on the DOM tree)

## Completed
- [x] #913
- [x] FA - #780
- [x] #815
- [x] #718
- [x] FA - #689
- [x] AEL have to keyboard-tab through entire mini-roadmap before getting to main content
- [x] AEL beta-bar should be `<banner>` or some semantic HTML, not just a div
- [x] AEL delete 'aria-label` from "Follow us on" div in footer
- [x] FA - #720
- [x] Step 4 of onboarding, location dropdown, is not navigable through screen reader buttons but is through the keyboard. Requires further discussion (maybe an issue with how the Div element renders on the DOM tree)
- [x] AEL aria-label on the close button for contextual info panel
- [x] AEL remove aria-label=progress from vertical step indicator
- [x] AEL add aria-label for 1, 2, 3 etc
- [x] AEL legal banner on roadmap should have `<banner>` or some semantic html
- [x] FA - Task page does not have an h1 (https://dequeuniversity.com/rules/axe/3.5/page-has-heading-one?application=axeAPI)
- [x] #413
- [x] #440
- [x] - MD - page/tab title for onboarding steps
- [x] onboarding pages do not have an h1

## No Longer An Issue
~~- [ ] - FA - on the roadmap page, bullet icons are being read~~

aloverso:
@skeeJay @carye3 A couple of the items here we wanted to raise with you all to see how you wanted to fix them:

1. Because screenreaders read out the title of the page first, each page should have an informative title. Right now, the onboarding pages are titled as "Business.NJ.Gov Navigator", which is not distinct from the landing page / roadmap page. We should give onboarding a unique page title to give screenreader users information that they're on a different page. What would you like the page title to be for onboarding?
2. Both the task pages and the onboarding page do not have an `<h1>` main header. Mostly for you @carye3, do you want to think about giving the pages a "proper" h1 according to the styleguide (as in, bigger text) - or do you want us to make the header a semantic h1 tag while applying styling to make it continue to look like the h2 that it currently is?

carye3:
@aloverso 

1. I'm seeing "Tell Us a Bit" as the title of the page... am i missing something?
[...]

PR #421, opened 2021-09-09
[#386] accessibility - add keyboard option to skip to main content

aloverso:


aloverso:
@mikewolfd my "Show" PR :)

Issue #422, opened 2021-09-10
Introduce Sections to Roadmap for existing Steps


skeeJay:
**As** Poppy
**I need to** see my Roadmap Steps broken into "Plan" and "Start" sections
**So that** I can see my Roadmap broken into further "phases" that can be collapsed when I move onto another phase of my business's lifecycle.

## Acceptance criteria

All users should receive a plan and start section.

## Notes

- [x] Steps fall within Sections, and need to be editable by Jess in terms of which Section they are in and the Order within each Section.
- [x] Steps retain their "global" numbering within each Section. If the "Plan" Section has Steps 1 and 2, then the "Start" Section starts with Step 3. [updated by @skeeJay after design review discussion]

~~- [ ] Plan and Start should both be expanded by default. They should be collapsible by the user, and the Roadmap should remember the state of whether each is collapsed **in the home roadmap screens (desktop & mobile)** (not in sidebar roadmap).~~ Addressed in #439

- [x] These will need to be integrated into the mini-roadmap on Task screens as well. (@carye3 This may need design.)

~~- [ ] Initially, place the existing Step "Plan Your Business" In the Plan Section and the other 4 Steps in the Start Section.~~ Step 1 and 2 will fall in plan section as per the design

- [x] Steps 1 and 2 should fall in the plan section.
- [x] Steps 3-5 should fall in the start section.
- [x] @carye3 will attach needed designs.

carye3:
Jess: add onboarding step, ask them if they’re operating only

Jesslax:
@skeeJay right now we dont have information to put under operate. So I dont want it to be built, a user goes in and then sees no information in the operate section. For this tkt, should I write conditions for the current scenario, or the scenario where we have content under the operate section?

skeeJay:
@Jesslax This ticket doesn't introduce "Operate" at all. The description only refers to introducing "Plan" and "Start" sections. So let's stay limited to that: we're taking the existing Steps and splitting them between "Plan" and "Start" sections.

skeeJay:
@carye3 Since we're not introducing "Operate" at all in the scope of this ticket, this shouldn't require any new onboarding questions. I think adding to Onboarding is for a far future epic, after we've introduced "Operate" for Poppy.

carye3:
@skeeJay @mikewolfd @aloverso @FarazA22 for your review:

[...]

PR #454, opened 2021-09-15
[#422] Moved Steps into Sections

mikewolfd:


web/components/MiniRoadmap.test.tsx
```diff
@@ -37,4 +51,20 @@ describe("<MiniRoadmap />", () => {
     expect(subject.queryByText("task2")).not.toBeInTheDocument();
     expect(subject.queryByText("task1")).not.toBeInTheDocument();
   });
+
+  it("Displays each step under associated section", () => {
```
aloverso, line 55: can we keep the test name lowercase, ie `displays` not `Displays`

web/components/MiniRoadmap.tsx
```diff
@@ -39,58 +41,76 @@ export const MiniRoadmap = (props: Props): ReactElement => {
     }
   };
 
-  const isLast = (stepNumber: number) => {
-    return roadmap?.steps[roadmap?.steps.length - 1].step_number === stepNumber;
+  const getSection = (sectionName: SectionType) => {
+    const _sectionName = sectionName.toLowerCase();
```
aloverso, line 45: what is starting a variable name with an underscore intended to convey? Haven't seen that pattern

web/components/MiniRoadmap.tsx
```diff
@@ -39,58 +41,76 @@ export const MiniRoadmap = (props: Props): ReactElement => {
     }
   };
 
-  const isLast = (stepNumber: number) => {
-    return roadmap?.steps[roadmap?.steps.length - 1].step_number === stepNumber;
+  const getSection = (sectionName: SectionType) => {
```
aloverso, line 44: maybe to be more careful with variable naming... this isn't the section name but rather the type, right?

web/test/pages/roadmap.test.tsx
```diff
@@ -178,4 +179,43 @@ describe("roadmap page", () => {
     expect(subject.queryByText("Completed")).toBeInTheDocument();
     expect(subject.queryByText("Not started")).toBeInTheDocument();
   });
+
+  it("Displays each step under associated section", () => {
```
aloverso, line 183: can we keep the test name lowercase, ie `displays` not `Displays`

web/components/MiniRoadmap.test.tsx
```diff
@@ -1,22 +1,36 @@
...
   beforeEach(() => {
-    useMockRoadmap(
-      generateRoadmap({
-        steps: [
-          generateStep({ name: "step1", tasks: [generateTask({ name: "task1", id: "task1" })] }),
-          generateStep({ name: "step2", tasks: [generateTask({ name: "task2", id: "task2" })] }),
-        ],
-      })
-    );
+    useMockRoadmap({
```
aloverso, line 12: It's an opinion, but I have a preference as tests grow bigger to keep specific setup cases within the test block itself instead of the beforeEach. Like, so that each test contains a given-when-then when you read it, and you don't have to scroll all the way up to the beforeEach to understand the "given" setup

Used to get in heated debates about this opinion with a friend on a previous team though :) so I understand it's not shared by all. what do y'all think of having the useMockRoadmap call in your specific test? Since none of the rest of the tests care about sections, having them listed out overall at the top is clutter for everyone except our specific test. And then you have to _know_ which pieces are for which, ie that the `tasks` properties are meaningful to the first 3 tests but not the last one, and the `section` properties are meaningful only to the last test. So it might be more clear to just define what we need for each test/set

web/components/njwds-extended/VerticalStepIndicator.tsx
```diff
@@ -3,10 +3,12 @@ import { useMountEffect, useOnWindowResize } from "@/lib/utils/helpers";
 
 interface Props {
   number: number;
+  order: number;
```
aloverso: I (now) see the purpose of this prop re resetting the numbering for each section, but I think the naming is a little confusing to me? We're not using this prop to order anything. It seems that it represents the way that the step number displays to the user? Maybe we could rename for clarity?

web/components/njwds-extended/VerticalStepIndicator.tsx
```diff
@@ -3,10 +3,12 @@ import { useMountEffect, useOnWindowResize } from "@/lib/utils/helpers";
 
 interface Props {
   number: number;
+  order?: number;
   last: boolean;
   active?: boolean;
   small?: boolean;
   completed?: boolean;
+  openSteps?: number[];
```
aloverso, line 11: The thing about `openSteps` is that, now that we have both `order` and `number` props representing different ways of numbering steps, it's not obvious from looking here whether this is referencing the `number` or `order` as the list of open steps. I have to go look to see how it's used in the code to find out. Maybe a rename to something like `openStepNumbers` might be more descriptive to help here

web/components/njwds-extended/VerticalStepIndicator.tsx
```diff
@@ -3,10 +3,12 @@ import { useMountEffect, useOnWindowResize } from "@/lib/utils/helpers";
 
 interface Props {
   number: number;
+  order?: number;
```
aloverso, line 6: I see (now) why this prop is needed re the discussion about resetting step numbering for each section. But I think the name is confusing, this doesn't change anything for the step ordering? Instead it seems to represent the way we display the step number to the user. Can we rename the prop for clarity?

aloverso: Also, why optional?

aloverso:
before merge, can you rename the commit message to the pattern `[#xxx] descriptive message`

so, be sure to include the story number in the commit message, and then the message itself should describe the added functionality (ie, "display steps in associated sections") - no need for "checks whether" because hopefully it's not in question "whether" it works or not :)

web/components/njwds-extended/VerticalStepIndicator.tsx
```diff
@@ -3,12 +3,12 @@ import { useMountEffect, useOnWindowResize } from "@/lib/utils/helpers";
 
 interface Props {
   number: number;
-  order?: number;
+  visualIndex?: number;
```
aloverso, line 6: why optional?

mikewolfd: just in case we decide to reuse the step indicator without the sections, no reason for it not to be optional

web/components/MiniRoadmap.test.tsx
```diff
@@ -38,3 +46,48 @@ describe("<MiniRoadmap />", () => {
     expect(subject.queryByText("task1")).not.toBeInTheDocument();
   });
 });
+
+describe("<MiniRoadmap /> with sections", () => {
```
aloverso, line 50: i'd prefer to keep with the pattern of a single top-level describe per test, since that's sort of the expected assumption everywhere else

think this could be simpler, you can just re-run useMockRoadmap in your one test

web/components/njwds-extended/VerticalStepIndicator.tsx
aloverso: now I'm thinking more about that, is there any reason to have both of these props at all? we could just have a single prop representing "this is the number to show on the stepper" and the caller can decide whether that's the "step number" or the "visual index", and it would simplify this component's interface

aloverso: and we could replace `openStepNumbers` with an `isOpen` boolean or something instead, and similarly let the caller do that logical calculation, so that we don't need to have the list of numbers and `props.number` in this component

Issue #414, opened 2021-09-08
As Poppy, I need to see contextual awareness menu close with an animation


mikewolfd:
Utilize the work done by the previous animation ticket

Duration: .3 seconds

skeeJay:
I don't currently see this animation on desktop on dev.

skeeJay:
Seeing this now. All the content in the bar disappears during the animation. Is that unavoidable?

skeeJay:
Moving back to In Progress because @FarazA22 said he is still working on.

skeeJay:
@FarazA22 Minor thing I noticed: as the contextual sidebar opens, the rest of the screen darkens gradually; but as it closes, the darkness disappears all at once. Should the animations match? cc @carye3

FarazA22:
There is a .2 second transition added to the overlay when both the mobile menu/drawer and contextual info panel/drawer appear on the screen. The overlay immediately disappears when the draw and panel are closed. If we make a change, we should make it to both parts of the app - perhaps create a separate ticket? I like the current close experience because it immediately 'reenables' the background.

I was playing around the app within the browser and turned off the .2 second transition to the overlay and liked the experience. I feel its more snappier and pairs well with the opening and closing of either drawer. If either of you want to look at that scenario, we can jump on a zoom call. It wouldn't require any code changes, we can simply disable the functionality in the browser within Chrome Dev Tools.

skeeJay:
I sort of buy that explanation, that the content is immediately available again so the dark overlay should disappear in a snappy way. Interested to hear @carye3's take also, but I can buy that.

skeeJay:
@carye3 pinging on this as well, the above discussion needs your design eye.

skeeJay:
Okay @carye3. Then I guess this is okay to deploy.

PR #464, opened 2021-09-17
[#414] markdown content stays on contextual info panel until close animation completes

FarazA22:


web/components/FocusTrappedSidebar.tsx
```diff
@@ -10,17 +11,21 @@ interface Props {
...
 
   useEffect(() => {
     let timeoutId: ReturnType<typeof setTimeout>;
     if (isOpen && !showDiv) {
       setShowDiv(true);
     } else if (!isOpen && showDiv) {
-      timeoutId = setTimeout(() => setShowDiv(false), delayTime);
+      timeoutId = setTimeout(() => {
+        setShowDiv(false);
+        setContextualInfo({ ...contextualInfo, markdown: "" });
```
mikewolfd, line 24: should this be here? don't we use this component elsewhere, maybe do a callback or something instead?

FarazA22: @mikewolfd Thanks for pointing that out. I've made some changes to the code. The markdown property is not being updated within the contextualInfoPanel component.

PR #512, opened 2021-09-24
439 accordion in roadmap sections

a-afriyie:


web/components/njwds-extended/VerticalStepIndicator.tsx
```diff
@@ -24,14 +24,14 @@ export const VerticalStepIndicator = (props: Props): ReactElement => {
...
-      const newHeight = height - margin;
+      const newHeight = height - parseInt(marginStyle.marginTop);
       verticalBar.style.height = `${newHeight}px`;
     } else {
       verticalBar.style.height = `${height}px`;
     }
   };
-  useMountEffect(resizeVerticalBarToContent);
+
+  useEffect(resizeVerticalBarToContent, [props.isOpen, props.last, props.stepNumber]);
```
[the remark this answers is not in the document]
aloverso: I'm not sure if that's true... we want the resize to happen every time the isOpen changes, right?

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const toggleStep = (stepNumber: number): void => {
+    analytics.event.task_mini_roadmap_step.click.expand_contract();
+    if (openSteps.includes(stepNumber)) {
+      setOpenSteps(openSteps.filter((number) => number !== stepNumber));
+    } else {
+      setOpenSteps([...openSteps, stepNumber]);
+    }
+  };
+
+  const handleAccordionStateChange = async (
```
aloverso, line 47: this handler code is duplicated in RoadmapBySections, as is a lot of the HTML below. Their main difference is the contents of the AccordianDetails, it seems? This seems like an opportunity for refactoring some kind of Roadmap Accordian component so that this handler and all that code could only be written once, and used in both places

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -0,0 +1,167 @@
...
+    sectionType: SectionType,
+    expanded: boolean | undefined
+  ): Promise<void> => {
+    const openRoadMapSections = userData?.preferences.roadmapOpenSections;
+
+    if (expanded && openRoadMapSections) {
```
aloverso, line 53: stylistic preference is to have a guard clause. instead of `&& openRoadMapSections`, in both the if AND the else-if, you could instead have a line above this if that's just:
```
if (!openRoadMapSections) return;
``` 

that way, you get the following benefits:
- you can use an if/else instead of an if/else-if
- the if statements are easier to read. I had a hard time puzzle through "what does it mean, if it's expanded and the openroadmapsections exists... when would it not exist though? what happens then?" until I realized the second half was just a guard against undefined. a guard clause lets us move past that and then the if clause is descriptive "if it's expanded, do this, otherwise, do that". very straightforward and easier to read

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      setOpenSteps([...openSteps, stepNumber]);
+    }
+  };
+
+  const handleAccordionStateChange = async (
+    sectionType: SectionType,
+    expanded: boolean | undefined
+  ): Promise<void> => {
+    const openRoadMapSections = userData?.preferences.roadmapOpenSections;
```
aloverso, line 51: let's try to keep variable naming consistent. if it's spelled `roadmapOpenSections` , it's kind of confusing to rename that variable to `openRoadMapSections`
- the switching around "open" and "roadmap" is confusing, why are we changing order? is it supposed to convey something about the value of the variable?
- if it's spelled "Roadmap" we should keep the same spelling, not change to "RoadMap"

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -0,0 +1,167 @@
...
+          </AccordionDetails>
+        </Accordion>
+        <hr className="margin-y-2 bg-base-lighter" />
+      </div>
+    );
+  };
+
+  return (
+    <>
+      {roadmapSections.map((sectionName) => {
```
aloverso, line 161: same comment as above - this is shared with RoadmapBySections and could be refactored and shared

web/pages/_app.tsx
```diff
@@ -130,6 +130,25 @@ const theme = createTheme({
         elevation: 8,
       },
     },
+    MuiAccordionSummary: {
```
aloverso, line 133: If possible, I prefer to keep component-specific style overrides with the component that's using it. see `TaskProgressDropdown` as an example, of using the `makeStyles`/`useStyles` paradigm that comes with MUI

aloverso: the purpose is to help keep code close to what's using it. So if we ever delete the accordian, we can delete the accordian-specific style code because it's in the same place, so we know if it gets used or not. whereas here, it'd be much harder to remember to come in here and delete this when it's no longer used

web/test/pages/roadmap.test.tsx
```diff
@@ -228,4 +239,132 @@ describe("roadmap page", () => {
     expect(within(sectionStart).getByText("step2")).toBeInTheDocument();
     expect(within(sectionStart).getByText("step4")).toBeInTheDocument();
   });
+
+  it("displays sections based on default state", () => {
```
aloverso, line 243: small nitpick, but technically this test isn't testing a "default" state. it displays sections based on whatever are the open sections in the userData preferences. the fact that they happen to be "default" is really just something we know. But in the eyes of this test, there's nothing default about it. It's just displaying whatever sections are in the preferences array

web/test/pages/roadmap.test.tsx
```diff
@@ -228,4 +239,132 @@ describe("roadmap page", () => {
...
+        generateStep({
+          name: "step1",
+          section: "PLAN",
+          tasks: [generateTask({ id: "task1" })],
+        }),
+        generateStep({
+          name: "step2",
+          section: "START",
+          tasks: [generateTask({ id: "task2" })],
+        }),
```
aloverso, line 286: duplicating this task setup over multiple tests is a great opportunity to put them all in a describe block with a `beforeEach` to do shared setup

web/test/pages/roadmap.test.tsx
```diff
@@ -228,4 +239,132 @@ describe("roadmap page", () => {
...
+    const subject = renderRoadmapPage();
+
+    const sectionStart = subject.getByTestId("section-start");
+    const sectionPlan = subject.getByTestId("section-plan");
+
+    expect(within(sectionStart).getByText("step2")).not.toBeVisible();
+    expect(within(sectionPlan).getByText("step1")).toBeVisible();
+  });
+
+  it("updates accordion state", async () => {
```
aloverso, line 336: this could be more descriptive - "accordion" is a UI term, but what it's updating here is the userData preferences, right?

web/test/pages/roadmap.test.tsx
```diff
@@ -228,4 +239,132 @@ describe("roadmap page", () => {
...
+    const subject = render(
+      <WithStatefulUserData initialUserData={userData}>
+        <RoadmapPage displayContent={{ contentMd: "" }} />)
+      </WithStatefulUserData>
+    );
+
+    const sectionPlan = subject.container.querySelector("#plan-header");
+    expect(sectionPlan).toBeInTheDocument();
+    fireEvent.click(sectionPlan as Element);
+    expect(currentUserData().preferences.roadmapOpenSections).toEqual(["START"]);
```
aloverso, line 368: it seems like we're testing only half of that "if (expanded) else if (!expanded)" block in the handleAccordionStateChange function. We're only testing the removal of something from the roadmapOpenSections array. I could delete the entire contents of your "else if" block right now and the tests would still pass.

aloverso: As an aside, this is one of the reasons that I personally like to practice TDD. When you write the failing tests first and only write enough code to get them passing, then you can ensure that your tests cover all the possible paths of your code branches.

web/test/pages/roadmap.test.tsx
```diff
@@ -228,4 +239,132 @@ describe("roadmap page", () => {
...
+  });
```
aloverso, line 369: It seems like we're only testing the functionality in RoadmapBySections but not the logic in MiniRoadmap. Seems like we should have a similar set of tests in MiniRoadmap. 

That, or if you like the suggestion of refactoring the logic in RoadmapBySections and MiniRoadmap into a shared component that they both use, then you could potentially consolidate those tests into a single unit test for that new component.

aloverso: Again not to sound like a broken record (I'm sorry!!) but in a similar way as above, if I write the test first, I would not add any functionality to MiniRoadmap until I had a failing test to drive it out. There are many ways of writing tests and you don't have to do it the way I do, of course!!! But I am always excited to show the benefits

web/components/njwds-extended/VerticalStepIndicator.tsx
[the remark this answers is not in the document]
dsueltenfuss: Anne is correct. I withdraw my comment

api/src/db/migrations/v16_add_user_preferences.ts
```diff
@@ -0,0 +1,113 @@
...
+  nameAndAddress: v16NameAndAddress;
+  completedSearch: boolean;
+  lastCheckedStatus: string;
+  status: v16LicenseStatus;
+  items: v16LicenseStatusItem[];
+};
+
+type v16Preferences = {
+  roadmapOpenSections: v16SectionType[];
+  roadmapOpenSteps: number[];
```
aloverso, line 87: a nitpick, and no need to change this for this now but something to think about for the future. I know we'll need the roadmapOpenSteps eventually for an upcoming story, but since it's not used in this story, I would have waited to add this property of preferences until we actually need to use it, instead of having an empty array in there that we do nothing with until we get to that story later on

aloverso: my reasoning for this is twofold: 
- commits being a record of the work involved in each story. if we do a bit of work for another story in this one, then someone reading through this 5 months from now would be all "why are we adding this value if we don't use it??" because they aren't aware of the context of "oh we'll need it soon" so it feels out of place
- I know, and you know, that we are likely to do the openSteps story next. but we don't know. maybe our PMs will decide something else is more important. maybe it gets shuffled down the backlog for who-knows-how-long. maybe we never get around to it. and in the off-chance that we never get around to it, we'd have this extra field taking up space in our DB and in our brains for no reason. and we'd have to put work into cleaning it up! hence, preferring to keep it out until we actually need it, if it doesn't give us value now

api/src/domain/factories.ts
```diff
@@ -120,6 +122,13 @@ export const generateLicenseData = (overrides: Partial<LicenseData>): LicenseDat
   };
 };
 
+export const generatePreferences = (): Preferences => {
+  return {
+    roadmapOpenSections: ["PLAN", "START"],
+    roadmapOpenSteps: [],
```
aloverso, line 128: this isn't quite a generator function because we're not randomly generating these. which is maybe fine, depending on how it gets used, but worth noting

aloverso: sometimes it's too hard to randomly generate from a list when you have a string enum like this and I acknowledge that. maybe this is one of those times. but just wanted to think about it

web/cypress/integration/roadmap-mobile.spec.ts
```diff
@@ -30,38 +34,47 @@ describe("Roadmap [feature] [all]", () => {
...
 
     // step 4
     cy.get('[data-step="4"]').should("exist");
 
+    // step 5
+    cy.get('[data-step="5"]').should("exist");
+
     // tasks screen
-    cy.get('[data-task="register-trade-name"]').click();
+    cy.get('[data-task="register-trade-name"]').click({ force: true });
```
aloverso, line 52: why do we need all the force-true statements now?

web/cypress/integration/roadmap-mobile.spec.ts
```diff
@@ -22,6 +22,10 @@ describe("Roadmap [feature] [all]", () => {
     cy.get('[data-municipality="Absecon"]').should("exist");
 
     // step 1
+    cy.get('[id="plan-content"]').should("be.visible");
+    cy.get('[id="plan-header"]').click({ force: true });
+    cy.get('[id="plan-content"]').should("not.be.visible");
+    cy.get('[id="plan-header"]').click({ force: true });
```
aloverso, line 28: is there a reason for using a proper ID here instead of a data-testid?

a proper ID would be necessary if we are using aria-labelledby or something, I think. Or, if we are using specific CSS for those IDs (although, that would be a yellow flag to ask why...)

Whereas if we're only using it as a selector for testing, `data-testid` is a better choice because it is explicitly saying "hey this ID is not semantically / functionally meaningful, it is just for testing selector purposes"

api/src/db/migrations/v16_add_user_preferences.ts
dsueltenfuss: I think we added this property before we ended up agreeing to remove the acceptance criteria that had been added on. Not sure the best way to catch that after the fact, but something we may want to discuss (hopefully scope creep/revert isn't something that happens often)

aloverso: That makes total sense, and I don't want to suggest going back now and taking this out of the Preferences object, seems like more trouble than it's worth. More like something to keep in the back-of-mind regarding incremental changes for each story. But yeah I agree hopefully this is a unique case and most stories don't have as much shifting acceptance criteria anyways

web/components/roadmap/MiniRoadmap.tsx
a-afriyie: Refactored to a utility function called by both RoadmapBySections and MiniRoadmap.

web/components/roadmap/MiniRoadmap.tsx
a-afriyie: Addressed.

web/components/roadmap/MiniRoadmap.tsx
a-afriyie: Renamed openRoadMapSections to roadmapOpenSections for consistency.

web/components/roadmap/MiniRoadmap.tsx
aloverso: I think I was suggesting a component more than a utility function (so that the duplicated JSX can also be refactored) - if you think that makes sense, of course

a-afriyie: Sorry, should have been explicit, the utility function is just for the handler. Will address the JSX portion.

a-afriyie: Though the JSX are similar in some cases for both RoadmapBySections and MiniRoadmap, there are some key differences concerning their child elements and class names / styles. I'd suggest we leave them both as is.

web/components/roadmap/MiniRoadmap.tsx
a-afriyie: Addressed.

web/cypress/integration/roadmap-mobile.spec.ts
a-afriyie: Cypress seemed to be skipping some click events.

web/pages/_app.tsx
FarazA22: @aloverso 

The styling changes were applied at the theme level so we could have a baseline of the MUI Accordion component. This simplifies the process of applying styling for all future uses of that component. 

By resetting the padding/margin at the theme level we can apply the padding/margin within the JSX passed inside the component and no longer need to adjust for additional spacing that was previously there or and avoid using makeStyles/useStyles.

Additionally, MUI v.4 had issues like, when you clicked the 'AccordionSummary' to display the 'AccordionDetails', MUI would add additional padding to the 'AccordionSummary' element. The change in spacing caused a deviation in the design. MUI v.5 now has a disableSpacing prop that avoids this behavior - we might be able to set it at the theme level.

The theme changes need to be revisited as a whole since we upgraded to MUI v.5. For example, line 147 'display: block' can be deleted because its not the default for MuiAccordionDetails-root. I can look at it with Akwasi tomorrow. 

With the additional context, do you feel more comfortable with why we made the change?

web/cypress/integration/roadmap-mobile.spec.ts
FarazA22: @aloverso We are experiencing some weird behavior with Cypress. It will click a button, but the screen doesn't change and leads to a failed test. We did some research and think the option passed into the click method would address the instances were cypress is failing. 

I've also added a new Cypess command - forceClick, which would apply the click with the options.

web/pages/_app.tsx
aloverso: Sure, that seems reasonable to me

web/cypress/integration/roadmap-mobile.spec.ts
FarazA22: @aloverso Per MUI Doc, for optimal accessibility they recommend setting ```id``` and ```aria-controls``` on the ```AccordionSummary```. The Accordion will derive the necessary ```aria-labelledby``` and ```id``` for the content region of the accordion.

web/test/pages/roadmap.test.tsx
a-afriyie: Addressed.

web/test/pages/roadmap.test.tsx
a-afriyie: Addressed.

web/test/pages/roadmap.test.tsx
a-afriyie: Addressed.

web/test/pages/roadmap.test.tsx
a-afriyie: Addressed.

web/test/pages/roadmap.test.tsx
a-afriyie: Added tests for MiniRoadmap.

web/cypress/integration/roadmap-mobile.spec.ts
aloverso: Make sense to me

web/cypress/integration/roadmap-mobile.spec.ts
aloverso: Huh, strange, but yeah if this helps then that's fine

api/src/domain/factories.ts
FarazA22: Are you suggesting the name of the function doesn't align with what it actually does? If so, would renaming it to something like 'setDefaultPreferences' be more appropriate?

aloverso: Sorry, I wasn't very clear. I think of a "generator function" as all of these in the factories file that build test objects like this. Typically, they generate test objects with random values, which helps enforce good developer habits when writing assertions. 

For a string union type though, like `"PLAN" | "START"` or others we have like `"UNSTARTED" | "IN-PROGRESS" | "COMPLETED"` TypeScript makes it annoying difficult to generate a random value from the options available. Which is why the other functions in this file will generate random string values but use a static status like:

```export const generateLicenseStatusResult = .... {
  return {
    status: "PENDING",
    ...
```

I think the code for this story is fine, because I know it's hard to get a random one, but mostly I wanted to point out "hey here's another example of places we're not being random, and that's important that we're all aware of, because if we're not aware we might mistakenly write tests assuming it has a static value" and that's not ideal because it leads to brittle tests.

Sorry for being wordy. I think the initial version is fine, but I was just internally thinking through the implications of a lot of enums and their lack of randomness, so I added a comment noting it, i guess

Issue #97, opened 2021-04-19
As Poppy, I need to be able to change one of my onboarding inputs directly instead of going through the whole flow


skeeJay:
Activities include:

- [ ]

skeeJay:
Moved this to our sprint planning board @Jesslax

carye3:
Design changes by Guy [here](https://www.figma.com/file/Hg3ggwTj50wxdg4mjG4xRB/Dashboard-Exploration?node-id=4086%3A68994)

skeeJay:
@carye3 @Jesslax We'll take a look at function, but if you can both check this for design and business functionality, respectively. cc @aaronvux

Jesslax:
Noticed a few things here:

aaronvux:
This looks good in mobile and desktop. Agree with Jess that there could be a little more space between the buttons.  We should either have the save got back to the roadmap as Jess suggested or change the back button to "Back to Roadmap", leave the success message up with the link or add a separate button. @carye3 let's add the spacing to the Design chore list. Moving to ready for staging.

mikewolfd:
@Jesslax @skeeJay 

I can fix  1) re: the spacing but in reference to 2):  This design specifically is setup *not* return you back when you click save, hence why there's a success toast, which includes a link to Roadmap. I can do whatever, but this seems like something that was missed at the design review? How would you like to proceed?

skeeJay:
@Jesslax Outside of those two UX issues, does the rest of the profile screen function/behave as expected? cc @aaronvux

PR #574, opened 2021-10-06
[#97] Profile Page

mikewolfd:


web/components/MunicipalityDropdown.tsx
```diff
@@ -67,7 +67,9 @@ export const MunicipalityDropdown = (props: Props): ReactElement => {
             ...params.inputProps,
           }}
           value={searchText}
-          onChange={(event: ChangeEvent<HTMLInputElement>) => setSearchText(event.target.value)}
+          onChange={(event: ChangeEvent<HTMLInputElement>) => {
+            return setSearchText(event.target.value);
+          }}
```
aloverso, line 72: why this change? like, it's not a problem or anything, but do you want to transition towards this style?

web/components/onboarding/OnboardingHomeBasedBusiness.tsx
```diff
@@ -13,9 +14,13 @@ export const OnboardingHomeBasedBusiness = (): ReactElement => {
     });
   };
 
+  const header = setHeaderRole(3, "h3-element");
```
aloverso, line 17: is this line redundant?

web/test/mock/withStatefulUserData.tsx
```diff
@@ -3,7 +3,7 @@ import { UserData } from "@/lib/types/types";
 import * as useUserModule from "@/lib/data-hooks/useUserData";
 import { getLastCalledWith } from "@/test/helpers";
 
-const updateSpy = jest.fn();
+export const updateSpy = jest.fn();
```
aloverso, line 6: this was not exported on purpose - the idea was that in any test where we need to check this, we should be using the statefulUserData and asserting against the currentUserData instead

web/test/pages/profile.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+    fillText("Business name", "Cool Computers");
+    clickBack();
+    expect(subject.getByTestId("return")).toBeInTheDocument();
+  });
+
+  it("returns user to profile page from un-saved changes modal", async () => {
+    subject = renderPage({});
+    fillText("Business name", "Cool Computers");
+    clickBack();
+    fireEvent.click(subject.getByTestId("escape"));
```
aloverso, line 182: it's small nitpick but where possible, it'd be nicer to use the ProfileDefaults.escapeWhatever as a getByText instead

aloverso: same comment goes for all over this test

Issue #498, opened 2021-09-22
Remember which Steps are toggled open on mini-Roadmap


aloverso:
**As** poppy
**I need to** have a business roadmap that is customizable (read: step accordion states are remembered) 
**So that** my Navigator workflow is personalized to my business' needs.

## Acceptance criteria

**Given** I am in a task page and have toggled the arrow button for a Step to be open
**When** I come back to the page later
**Then** the application remembers, and keeps that step open

## Notes

- [ ]

aloverso:
@carye3 to fill in the outcome/need being filled by this story

aaronvux:
This looks good.  Moving to staging.

PR #831, opened 2021-10-29
feat: [#498] open step state management

mikewolfd:


web/components/roadmap/MiniRoadmap.test.tsx
```diff
@@ -28,31 +38,26 @@ describe("<MiniRoadmap />", () => {
   });
 
   it("expands the step that you are in by default", () => {
-    useMockUserData({});
```
aloverso, line 31: why switch everything to using stateful? do you think we should do that more generically? I always assumed it felt kinda like overkill for a lot of these tests to render statefully when the test contents don't care about that. but I could be wrong

web/components/roadmap/MiniRoadmap.test.tsx
```diff
@@ -91,18 +95,77 @@ describe("<MiniRoadmap />", () => {
...
 
-    const subject = renderMiniRoadMap("task3");
+    it("displays steps based on userData preferences", async () => {
+      const subject = renderMiniRoadMap("task1", userData);
+      expect(subject.getByText("task2")).toBeInTheDocument();
+    });
 
-    const sectionStart = subject.getByTestId("section-start");
-    const sectionPlan = subject.getByTestId("section-plan");
+    it("does not displays steps if not in userData preferences", async () => {
```
aloverso, line 145: "displays" -> "display"

aloverso: and not to nitpick for these test names but technically it _does_ display the step even if it's not in preferences. it just doesn't display the tasks within the step

web/components/roadmap/MiniRoadmap.test.tsx
```diff
@@ -91,18 +95,77 @@ describe("<MiniRoadmap />", () => {
...
+      const subject = renderMiniRoadMap("task2", userData);
+      expect(subject.queryByText("task1")).not.toBeInTheDocument();
+    });
 
-    expect(within(sectionStart).getByText("step2")).toBeVisible();
-    expect(within(sectionPlan).getByText("step1")).toBeVisible();
+    it("updates userData preferences on active step", async () => {
+      const subject = renderMiniRoadMap("task1", userData);
+      expect(subject.getByText("task1")).toBeInTheDocument();
+      await waitFor(() => expect(currentUserData().preferences.roadmapOpenSteps.sort()).toEqual([1, 2]));
```
aloverso, line 153: `sort()` is an inline mutation which could lead to unexpected behavior? idk. anyways, there's a jest function for what you want to do here:

`expect(currentUserData().preferences.roadmapOpenSteps).toEqual(expect.arrayContaining([1, 2])))`

web/components/roadmap/MiniRoadmap.test.tsx
```diff
@@ -91,18 +95,77 @@ describe("<MiniRoadmap />", () => {
...
+      fireEvent.click(subject.getByText("step1"));
+      expect(subject.getByText("task1")).toBeInTheDocument();
+      expect(subject.getByText("task2")).toBeInTheDocument();
+      await waitFor(() => expect(currentUserData().preferences.roadmapOpenSteps.sort()).toEqual([1, 2]));
+    });
+    it("updates userData on active step click ", async () => {
+      const subject = renderMiniRoadMap("task1", userData);
+      fireEvent.click(subject.getByText("step1"));
+      expect(subject.queryByText("task1")).not.toBeInTheDocument();
+      await waitFor(() => expect(currentUserData().preferences.roadmapOpenSteps.sort()).toEqual([2]));
```
aloverso, line 168: same sort -> arrayContaining comment for these two as above

web/components/roadmap/MiniRoadmap.test.tsx
```diff
@@ -91,18 +95,77 @@ describe("<MiniRoadmap />", () => {
...
+
+    it("updates userData on step click", async () => {
+      const subject = renderMiniRoadMap("task2", userData);
+      expect(subject.queryByText("task1")).not.toBeInTheDocument();
+      fireEvent.click(subject.getByText("step1"));
+      expect(subject.getByText("task1")).toBeInTheDocument();
+      expect(subject.getByText("task2")).toBeInTheDocument();
+      await waitFor(() => expect(currentUserData().preferences.roadmapOpenSteps.sort()).toEqual([1, 2]));
+    });
+    it("updates userData on active step click ", async () => {
```
aloverso, line 164: Thinking of the test as documentation, I'm looking at these test names and trying to figure out why "step click" and "active step click" would be different. But I see from the expectations (I think) that the difference is that an active step click CLOSES (removes from array) that step, whereas otherwise it OPENS (adds to array)

maybe we could make that clearer in test names to distinguish why the tests exist? for example `it('adds step to userData openSteps when non-active step is clicked")` and so on

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -49,54 +53,93 @@ export const MiniRoadmap = (props: Props): ReactElement => {
...
+  step: Step;
+  isLast: boolean;
+  activeTaskId?: string;
+  isOpen?: boolean;
+  completed: boolean;
+  onTaskClick?: () => void;
+  toggleStep: (number: number, isOpen: boolean) => void;
+}
+
+export const MiniRoadmapStep = (props: StepProps): ReactElement => {
```
aloverso, line 83: i do think these should be in separate files. even if it's only ever used here

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -49,54 +53,93 @@ export const MiniRoadmap = (props: Props): ReactElement => {
...
+    if (!props.activeTaskId) return undefined;
+    return !!props.step.tasks.find((task) => task.id === props.activeTaskId);
+  }, [props.activeTaskId, props.step]);
+
+  useEffect(() => {
+    if (isActive) {
+      setIsOpen(true);
+      props.toggleStep(stepNumber, true);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
aloverso, line 97: why? can we just add the extra deps?

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -15,32 +16,35 @@ interface Props {
...
 
-  const toggleStep = (stepNumber: number): void => {
-    analytics.event.task_mini_roadmap_step.click.expand_contract();
-    if (openSteps.includes(stepNumber)) {
-      setOpenSteps(openSteps.filter((number) => number !== stepNumber));
-    } else {
-      setOpenSteps([...openSteps, stepNumber]);
-    }
-  };
+  const toggleStep = useCallback(
```
aloverso, line 21: since this is a callback, can we name it `on...` as in `onToggleStep` or something? because technically this does not toggle the step open/closed (that happens in the step component), it just updates the userdata when that happens

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -15,32 +16,35 @@ interface Props {
...
+            }
+          : undefined;
+      };
+      if (!userData) return;
+      const openSteps = userData?.preferences.roadmapOpenSteps;
+      analytics.event.task_mini_roadmap_step.click.expand_contract();
+      if (openSteps.includes(stepNumber)) {
+        if (isOpen) {
+          return;
+        }
```
aloverso, line 40: 1. is there a refactoring for this nested if?
2. i'm actually confused by this isOpen check... why do we do nothing if the step is open?

web/components/roadmap/MiniRoadmap.tsx
```diff
@@ -49,54 +53,93 @@ export const MiniRoadmap = (props: Props): ReactElement => {
...
+  isLast: boolean;
+  activeTaskId?: string;
+  isOpen?: boolean;
+  completed: boolean;
+  onTaskClick?: () => void;
+  toggleStep: (number: number, isOpen: boolean) => void;
+}
+
+export const MiniRoadmapStep = (props: StepProps): ReactElement => {
+  const [isOpen, setIsOpen] = useState<boolean>(props.isOpen ?? false);
```
aloverso, line 84: why do we track this independently and duplicate state here? could it be tracked from just whether the step is contained in openSteps? since the userdata is essentially state, wouldn't the update to the userData trickle downward to change the component, if the component were just figuring out for itself if it should be open. and the set-is-open-if-active functionality, instead of setting state, the render could use a function that computes whether the step is open. unless I am missing something

PR #857, opened 2021-11-02
refactor: move industry config to a centralized location for roadmap …

dsueltenfuss:
…steps


web/lib/roadmap/roadmapBuilder.ts
```diff
@@ -25,6 +26,29 @@ export const buildRoadmap = async ({
...
+  if (process.env.NODE_ENV === "test") {
+    industryRoadmap = (await import(`@/lib/roadmap/fixtures/industries/${industryId}.json`))
+      .default as IndustryRoadmap;
+  } else {
+    industryRoadmap = (await import(`../../roadmaps/industries/${industryId}.json`))
+      .default as IndustryRoadmap;
+  }
+
+  addTasksFromAddOn(builder, industryRoadmap.addOns);
+  await applyAddOns(builder, ["generic-tasks", ...addOns]);
```
aloverso, line 45: Small feedback - can we just use the existing `applyAddOns` pattern and add the industry add-ons to the array instead of doing it separately?

Like, instead of adding line 44, change line 45 to be

```
  await applyAddOns(builder, ["generic-tasks", ...industryRoadmap.addOns, ...addOns]);
```

And same for modifications

web/lib/roadmap/roadmapBuilder.ts
```diff
@@ -25,6 +26,29 @@ export const buildRoadmap = async ({
...
+  modifications: string[]
+): Promise<RoadmapBuilder> => {
+  let industryRoadmap: IndustryRoadmap;
+  if (process.env.NODE_ENV === "test") {
+    industryRoadmap = (await import(`@/lib/roadmap/fixtures/industries/${industryId}.json`))
+      .default as IndustryRoadmap;
+  } else {
+    industryRoadmap = (await import(`../../roadmaps/industries/${industryId}.json`))
+      .default as IndustryRoadmap;
+  }
```
aloverso, line 42: Can we break this into a helper function like the other similar ones? The if/else necessity for testing really looks ugly which is why I hid them all in functions at the bottom of the file to make the logical flow as readable as possible

web/lib/roadmap/roadmapBuilder.ts
dsueltenfuss: Unfortunately no, because its a collection of addon tasks already, and doesn't need to be imported the same way the addons need to be

aloverso: Oh gotcha, I understand. Makes sense

web/lib/roadmap/roadmapBuilder.ts
dsueltenfuss: Refactored

Issue #764, opened 2021-10-26
Profile: Inline Location Entry Error Notification with Snackbar


carye3:
**As a** general user
**I need to** have a consistent experience
**In order to** use the tool with ease and minimal errors.

Consistent with in-line text field entry error message, accompanied by snackbar

## Acceptance Criteria

- Profile Page:
   - If format validation fails, on blur display inline error message below a red boundary
     - Error message: "This location is invalid. Please select one from the dropdown menu."
   - If the user attempts to save their changes, display error snackbar
 - Onboarding
   -  If format validation fails, on blur display inline error message below a red boundary
      - Error message: "This location is invalid. Please select one from the dropdown menu."
   -  If the user attempts to click "Next", display error snackbar

## Design

[Profile Page / Location error border, text and snackbar](https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=5763%3A107348)
![image](https://user-images.githubusercontent.com/62653834/139305157-3daf3140-52b9-42d7-82fe-701391ed7754.png)

[Onboarding/Location error](https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=5906%3A106847)
![image](https://user-images.githubusercontent.com/62653834/139314868-b2a75943-a74f-4978-8dd9-10f4872b5fb4.png)

mikewolfd:
Can you upgrade the acceptance criteria to include inline notifications for legal standing in onboarding, as it uses the same kind of logic

aaronvux:
@mikewolfd not seeing the inline error on the location field.
<img width="1099" alt="Screen Shot 2021-11-03 at 10 23 48 AM" src="https://user-images.githubusercontent.com/36163601/140078805-b1236a07-ace2-45c8-affe-fac9183a7ae9.png">

aaronvux:
@mikewolfd @carye3 should it say "Error" in the field.  I would think the field would just be blank
<img width="820" alt="Screen Shot 2021-11-04 at 11 14 07 AM" src="https://user-images.githubusercontent.com/36163601/140344263-d9a228b4-4517-45ee-af22-f557651d9107.png">
.

mikewolfd:
It wasn't in the design? And all the fields in the profile shows that field text

carye3:
[...]

PR #860, opened 2021-11-02
feat: [#764] Municipality inline validation

mikewolfd:


web/display-content/onboarding/OnboardingDefaults.ts
```diff
@@ -5,4 +5,6 @@ export const OnboardingDefaults: Record<string, string> = {
   stepXofYTemplate: "(Step ${currentPage} of ${totalPages})",
   errorTextRequiredLegal: "Please select a legal structure.",
   errorTextRequiredMunicipality: "Please select a location.",
+  errorTextMinimumEmployerId: "Must be 9 digits long.",
+  errorTextMinimumNumericField: "Must be ${length} digits long.",
```
aloverso, line 9: why are these 2 separate?

web/lib/types/types.ts
```diff
@@ -196,15 +196,16 @@ export const profileFields = [
...
 type ProfileFieldStatus = {
   invalid: boolean;
-  focus: boolean;
 };
 export type ProfileFieldErrorMap = Record<ProfileFields, ProfileFieldStatus>;
 
 export const createProfileFieldErrorMap = (): ProfileFieldErrorMap =>
   profileFields.reduce((p, c: ProfileFields) => {
-    p[c] = { invalid: false, focus: false };
+    p[c] = { invalid: false };
```
aloverso, line 208: ah,I missed these variable names in the last review

web/lib/utils/helpers.tsx
```diff
@@ -116,9 +118,29 @@ export const isStepCompleted = (step: Step, userData: UserData | undefined): boo
   });
 };
 
+export interface AlertProps {
+  variant: "success" | "warning" | "error";
+  body: string;
+  header?: string;
+  link?: string;
+}
```
aloverso, line 126: tbh, I feel weird that a "props" type is exported at all, but I see the reason. It still makes me think the type should like in the Alert file though? What do you think?

PR #1177, opened 2021-12-16
feat: cms setup & content reorg

mikewolfd:


content/build.js
```diff
@@ -0,0 +1,9 @@
+var fs = require("fs");
+function build() {
+  const getIndustries = fs.readdirSync("./src/roadmaps/industries/").reduce((pv, cv) => {
+    pv.push(require(`./src/roadmaps/industries/${cv}`));
+    return pv;
+  }, []);
+  fs.writeFileSync("./src/industry.json", JSON.stringify({ industries: getIndustries }));
```
dsueltenfuss, line 7: Can we write this to the lib folder instead of the src folder?  
I think keeping the generated code separate from the actual code is cleaner

content/package.json
```diff
@@ -5,6 +5,7 @@
   "private": true,
   "main": "src",
   "scripts": {
+    "build": "node ./build.js",
```
dsueltenfuss, line 8: If we are adding a build, we should also add a clean step

lerna.json
```diff
@@ -1,6 +1,6 @@
 {
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "packages": ["api", "content", "shared", "web"],
+  "packages": ["content", "api", "shared", "web"],
```
dsueltenfuss, line 4: Just curious, why did you change the order of these?

web/src/pages/mgmt/cms.tsx
```diff
@@ -11,8 +12,15 @@ const CMS = dynamic(
...
     import("netlify-cms-app").then((CMS) => {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
       // @ts-expect-error: No type definition available
       CMS.init({ CMS_CONFIG });
+      // @ts-expect-error: No type definition available
+      CMS.registerWidget("slug", SlugControl);
+      // CMS.registerEventListener({
+      //   name: 'preSave',
+      //   handler: ({ author, entry }) => console.log(JSON.stringify({ author, data: entry.get('data') })),
+      // });
```
aloverso, line 23: do we need this commented code?

content/build.js
```diff
@@ -0,0 +1,22 @@
...
+  var dirname = path.dirname(filePath);
+  if (fs.existsSync(dirname)) {
+    return true;
+  }
+  ensureDirectoryExistence(dirname);
+  fs.mkdirSync(dirname);
+}
+
+function build() {
+  const getIndustries = fs.readdirSync("./src/roadmaps/industries/").reduce((pv, cv) => {
```
aloverso, line 15: pv? cv?

PR #1240, opened 2021-12-23
feat: content splitting onboarding flows and profile

mikewolfd:


web/src/lib/types/types.ts
```diff
@@ -18,57 +18,63 @@ export type OperateDisplayContent = {
...
+export type OwningFlowContent = {
+  hasExistingBusiness: RadioFieldContent;
+  businessName: TextFieldContent;
+  industry: IndustryFieldContent;
+  entityId: TextFieldContent;
+  certifications: TextFieldContent;
+  municipality: TextFieldContent;
+  existingEmployees: TextFieldContent;
+};
+export interface UserDisplayContent extends StartingFlowContent, OwningFlowContent, ProfileContent {}
```
aloverso, line 69: the umbrella term "UserDisplayContent" isn't exactly what I'd expect... trying to think of an alternative suggestion... but since we already have a User object that's unrelated, it feels weird

web/src/lib/static/loadDisplayContent.ts
```diff
@@ -1,50 +1,57 @@
...
-  const specificHomeBased = getMarkdown(loadFile("municipality-home-based-business.md"));
-  const employerId = getMarkdown(loadFile("employer-id.md"));
-  const entityId = getMarkdown(loadFile("entity-id.md"));
-  const notes = getMarkdown(loadFile("notes.md"));
-  const taxId = getMarkdown(loadFile("tax-id.md"));
-  const certifications = getMarkdown(loadFile("certifications.md"));
-  const existingEmployees = getMarkdown(loadFile("existing-employees.md"));
+export const loadUserDisplayContent = (): LoadDisplayContent => {
+  const loadFile = (filename: string, type: UserContentType): string =>
+    fs.readFileSync(path.join(displayContentDir, "onboarding", type.toLowerCase(), filename), "utf8");
```
aloverso, line 29: I'd prefer if we had an explicit map / Record object to map from the type "STARTING" to the folder name "starting" rather than just `toLowerCase`.

It's one of those subtle things where, if it's not explicit, it's hard to know that the folder name and the type name are intrinsically connected, and you might not realize that months from now if someone wants to change it, and wouldn't know to change this line here. Whereas an explicit map will decouple that intrinsic linkage and make it visible.

Issue #977, opened 2021-11-15
Business Formation V1 (LLC)


aaronvux:
As Poppy with an LLC Business, I need the ability to submit all of my business formation information through the Navigator so I can easily complete what is required of me to form my business with the state of NJ from a single portal.

**THIS STORY IS DEPENDENT ON THE NIC API**

## Acceptance criteria

- Incorporate into "Form Your Business Entity" task.
- Use the Business Formation Test Endpoint
- Users will navigate between the steps with back and continue buttons.
- Specially designated fields will be pre-populated with available information from the Profile. 
- Editing specially designated fields will send the user to the Profile page to make any updates.
- If LLC is the Business Structure, the user will receive the new Business Formation workflow in the Form Your Business Task.
- For any email address that is required by the API, use the email address from the user's Navigator profile.
- Use existing UX patterns for inline validations for input fields and success/errors banners for API submission.

## Form Fields
------------------

- [x] - Business Structure (static display from Profile) (required)
- [x] -- Edit Link sends the user back to the Profile Page
- [x] - Business Name (static display from Profile)(required)
- [x] -- Edit Link sends the user back to the Business Name Search page.
- [x] - Business Suffix (required)
- [x] -- Dropdown
--- LLC
--- L.L.C.
--- LTD LIABILITY CO
--- LTD LIABILITY CO.
--- LTD LIABILITY COMPANY
--- LIMITED LIABILITY CO
--- LIMITED LIABILITY CO.
--- LIMITED LIABILITY COMPANY
- [x] - Business Start Date (required) 
- [x] -- Use Existing Date Picker
- [x] -- Default to Today's Date
- [x] -- Can only accept Today's Date or Future Dates

------------------
- [x] - Main Business Address (section title)
- [x] - Address Line1 (required)
- [x] - Address Line 2
- [x] - City (required)(static display from Profile)
-- Edit Link sends the user back to the Profile Page
- [x] - State
- [x] -- Static Display 
- [x] -- Default to NJ
- [x] - Zip Code (required)
- [x]  -- Validation Message:  Must be a NJ Zip Code between 07001 and 08999.

------------------

- [x] - Enter a Registered Agent Number (required)
- [x] -- Registered Agent Number must between 4-7 digits.
- [x] -User can toggle to input a Registered Agent/Office Address in substitute for Registered Agent Number
- [x] Toggle Registered Agent/Office Address
- [x] - Registered Agent Name (required)
[...]

PR #1253, opened 2021-12-27
feat: [#977] add contact info to business formation

FarazA22:


api/src/db/migrations/v33_formation_data.ts
```diff
@@ -209,7 +208,7 @@ type v33FormationSubmitError = {
 
 // ---------------- v33 factories ----------------
 
-const createEmptyFormationFormData = (): FormationFormData => {
+const createEmptyFormationFormData = (): v33FormationFormData => {
```
aloverso, line 211: good catch

aloverso: tbh surprised our fences didn't catch this :/

PR #1274, opened 2022-01-03
1065 ux enhancements

FarazA22:


web/src/components/onboarding/OnboardingCertifications.tsx
```diff
@@ -16,7 +16,7 @@ export const OnboardingCertifications = (): ReactElement => {
     });
   };
 
-  const headerLevelTwo = setHeaderRole(2, "h2-element");
+  const headerLevelTwo = setHeaderRole(2, "h3-element");
```
dsueltenfuss, line 19: Are these H2 or H3's now? 
If they are H3's, is that going to cause an accessibility issue, if there isn't an H2 before them?

FarazA22: @dsueltenfuss They have a header role of H2, but h3-element styling applied to them.

Issue #1072, opened 2021-11-29
Business Formation V2 (LLC)


aaronvux:
As Poppy, I need to have a simple and intuitive experience when submitting my business's formation information through the Navigator so that I don't have to use the State's website.

## Acceptance criteria

- Separate the single form design into 3 steps (Business, Contacts, Payment)
- Validations will fire on each page before a user will be allowed to move onto another page.
-- Display validation message below the Business Name field if it is not populated.
- Users will navigate between pages with a primary button and secondary back button.
- No change to API errors.

## Design
**Prototype**
https://www.figma.com/proto/408Krq8roicOzFVYrjwIdE/Navigator-Designs?page-id=6474%3A158222&node-id=6474%3A160334&viewport=389%2C48%2C0.02&scaling=min-zoom&starting-point-node-id=6474%3A160334

**Designs**
https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=6474%3A158222

## Additional Acceptance Criteria
- Follows Design
- Works in Mobile
- 508 Compliant

## Notes

- [x] UX Review
- [x] PO Approval

aaronvux:
@fiorella-io Business Formation MVP Design implementation story.

aaronvux:
Capture and store business formation date once the user is issued their entity ID.

aaronvux:
Create new story for API Error.

aaronvux:
Might need to repoint to include updates to payment page.

FarazA22:
@aaronvux Ready for review

aaronvux:
@FarazA22 I'm not seeing separated steps in dev.  I tried in cognito and still not seeing it.

aaronvux:
@FarazA22 the validations are not firing on the individual steps.  I can move forward with the incorrect zip codes and I don't see the error until the final step.

fiorella-io:
@FarazA22 Styling Issues I noticed:

Step 1
1. Add new content paragraph before "Business Name and Legal Structure"
2. Adjust Business Name and Legal Structure into a gray div
![image](https://user-images.githubusercontent.com/91498500/149401465-0696a198-a21b-4e59-b1e1-031ca67c3afd.png)
3. Add hr divider before Location heading and Registered Agent
4. Adjust the spacing between input fields. There's more space between "Address Line 1" and "Line 2" vs "State" and "Zip Code" (preferred spacing)
5. City, State, and Zip Code should be in one line
[...]

PR #1342, opened 2022-01-12
fix: [#1072] separated business formation into tabs

FarazA22:


web/src/components/tasks/BusinessFormation.test.tsx
```diff
@@ -502,10 +552,11 @@ describe("<BusinessFormation />", () => {
           });
         });
 
-        it("displays error message due to failed email validation", async () => {
+        it("displays error message due to failed email validation v2", async () => {
```
aloverso, line 555: v2?

FarazA22: I was using the similar title for two similar test, I'll update the test name to further clarify.

Issue #1200, opened 2021-12-19
Add Review Section to Business Formation V2


aaronvux:
As Poppy, I need the ability to review the information I am submitting to form my business so that I can ensure everything is accurate before making a payment.

## Acceptance criteria

- Add review section as the third step in the Business Formation workflow.
- Each section will have an edit link that will navigate the user to the appropriate page to update the information in that section.
- If a Registered Agent # is entered display:
-- Agent #
- If a Registered Agent is manually entered display:
-- Name, Agent Email, Address
- If Members were not added, do not show the Members section.

## Standard Acceptance Criteria
- Follows Design
- Works in Mobile
- 508 Compliant

## Design

Desktop:
https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=6693%3A148227

Mobile:
https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=7156%3A192250

## Notes

- [x] UX Review
- [ ] PO Approval
- [ ] Content Approval

FarazA22:
@fiorella-io @carye3 @aaronvux In mobile, the contextual info link is missing, but is present in the desktop. Is this by design?
<img width="336" alt="Screen Shot 2022-01-21 at 10 19 47 AM" src="https://user-images.githubusercontent.com/72581979/150552282-4756b234-5473-4982-b9b5-e331467216d2.png">
<img width="359" alt="Screen Shot 2022-01-21 at 10 21 11 AM" src="https://user-images.githubusercontent.com/72581979/150552510-f55fee41-47d7-48f1-806d-79e89e63ed21.png">


mobile - https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=7156%3A192250

desktop - https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=6693%3A148227

FarazA22:
@aaronvux Can you confirm if the 'learn more' contextual info link points to an existing markdown file (i.e. there is a contextual link that appears in the onboarding flow when the user is selecting their structure) or would we be displaying new content.


<img width="317" alt="Screen Shot 2022-01-21 at 11 03 55 AM" src="https://user-images.githubusercontent.com/72581979/150559741-4f328939-4241-4cbf-8334-cb2bd642e255.png">
.

FarazA22:
[...]

PR #1425, opened 2022-01-24
fix: [#1200] business formation review section

FarazA22:


web/src/display-defaults/roadmap/business-formation/BusinessFormationDefaults.ts
```diff
@@ -1,12 +1,12 @@
 export const BusinessFormationDefaults: Record<string, string> = {
+  llcText: "LLC",
```
FarazA22, line 2: Design requires 'LLC' to be displayed instead of the name within the legal structure object - "Limited Liability Company (LLC)". 

I decided to add the label within the defaults file. Is this reasonable? I could see us using LLC in the future, along with LLP and LP in other parts of the app. Does it make sense to update the type to include a shorthandName key? 

https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=7156%3A192250

web/src/components/tasks/BusinessFormation.test.tsx
```diff
@@ -923,49 +1017,11 @@ describe("<BusinessFormation />", () => {
       });
 
       describe("when agent number selected", () => {
-        it("agent number", () => {
+        it("agent number", async () => {
           renderWithData({ agentNumber: "", agentNumberOrManual: "NUMBER" });
+          await submitBusinessTab(false);
           expect(userDataWasNotUpdated()).toEqual(true);
         });
-        describe("when agent number selected", () => {
```
FarazA22, line 930: removing because these tests are duplicates

web/src/components/tasks/BusinessFormation.test.tsx
```diff
@@ -407,34 +407,124 @@ describe("<BusinessFormation />", () => {
...
+      fireEvent.click(subject.getByTestId("edit-legal-structure"));
+      expect(mockPush).toHaveBeenCalledWith("/profile");
+    });
+
+    describe("navigates from the review page", () => {
+      it("displays the first tab when the edit button in the main business section is clicked", async () => {
+        renderWithData({});
+        await submitBusinessTab();
+        await submitContactsTab();
+        fireEvent.click(subject.getByTestId("edit-business-name-section"));
```
aloverso, line 449: if possible (seems it is in this case) I'd prefer select by text, using the configured text (here, `BusinessFormationDefaults.editButtonText`)

aloverso: applies to several tests here, won't comment on all of them

web/src/components/tasks/BusinessFormation.test.tsx
aloverso: duplicates of what?

web/src/components/tasks/business-formation/PaymentSection.tsx
```diff
@@ -85,7 +85,7 @@ export const PaymentSection = (props: Props): ReactElement => {
     }
   };
 
-  if (!userData || state.tab !== 3) return <></>;
+  if (!userData || state.tab !== 4) return <></>;
```
aloverso, line 88: Is there a way we could make the tab numbers not hardcoded? Maybe an object that maps each thing to its page number, for example?

web/src/components/tasks/business-formation/PaymentSection.tsx
```diff
@@ -124,7 +124,7 @@ export const PaymentSection = (props: Props): ReactElement => {
           <Button
             style="secondary"
             onClick={() => {
-              setTab(2);
+              setTab(3);
```
aloverso, line 127: Could this be `state.tab + 1`?

web/src/components/tasks/business-formation/ReviewSection.tsx
```diff
@@ -0,0 +1,261 @@
...
+        </div>
+        <div>
+          {state.formationFormData.businessAddressLine1}
+          {", "}
+          {state.formationFormData.businessAddressLine2
+            ? `${state.formationFormData.businessAddressLine2}, `
+            : ""}
+          {userData?.profileData.municipality?.name}
+          {", "}
+          {state.formationFormData.businessAddressState} {state.formationFormData.businessAddressZipCode}
```
aloverso, line 73: I think that Mike's Members commit added a function to format an address into a single string. Could we reuse that here? Or in either case, turn that into a helper function?

web/src/components/tasks/business-formation/ReviewSection.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+  return (
+    <>
+      <div data-testid="review-section">
+        <BusinessNameAndLegalStructure reviewPage />
+        {getBusinessSuffixAndStartDate()}
+        {getMainBusinessLocation()}
+        {getRegisteredAgent()}
+        {getMembers()}
+        {getSignatures()}
```
aloverso, line 234: since none of these functions are ever taking any arguments, would it be cleaner to just use variables intsead?

ie, `const businessSuffixAndStartDate = <>...</>` instead of the function

web/src/components/tasks/BusinessFormation.test.tsx
FarazA22: To address a complicated conflict when rebasing, I ended up accepting both changes and deleted the tests that were duplicates. I missed a few tests.

web/src/display-defaults/roadmap/business-formation/BusinessFormationDefaults.ts
FarazA22: @aloverso do you have any thoughts on this?

web/src/components/tasks/BusinessFormation.test.tsx
FarazA22: There are 7 tests that I wrote that use "fireEvent.click(subject.getByTestId....)". They all share `BusinessFormationDefaults.editButtonText` as the content. I could use subject.getAllByText, however, I'll need to state the index within the array. I don't think this is a good idea.

web/src/components/tasks/business-formation/PaymentSection.tsx
FarazA22: will update

web/src/components/tasks/business-formation/ReviewSection.tsx
FarazA22: updated

web/src/components/tasks/BusinessFormation.test.tsx
FarazA22: I'll look into grabbing the input by the aria label

web/src/display-defaults/roadmap/business-formation/BusinessFormationDefaults.ts
aloverso: what you have here seems reasonable for now, but probably when we implement the next one, I like the idea of using a "shorthand name" property instead

web/src/components/tasks/BusinessFormation.test.tsx
aloverso: Ohh I see your point. Didn't realize there were multiple elements with the same text. In that case, I wouldn't bother changing it. Sticking with testid seems perfectly reasonable

web/src/components/tasks/business-formation/PaymentSection.tsx
aloverso: Perhaps an object that maps from sections to tab numbers? Like for example:

```
{
  MEMBER_SECTION: 2
   PAYMENT_SECTION: 3,
}
```

(or whatever the sections/values). That way the code would at least reference the sections and not the hardcoded numbers?

web/src/components/tasks/business-formation/ReviewSection.tsx
FarazA22: created helper function

web/src/components/tasks/business-formation/PaymentSection.tsx
FarazA22: updated

PR #1433, opened 2022-01-26
refactor: switched onboarding/profile to using genericfield

mikewolfd:


web/src/components/GenericNumericField.tsx
```diff
@@ -1,43 +1,12 @@
...
-      value={value.toString()}
-      additionalValidation={additionalValidation}
-      visualFilter={props.visualFilter ? props.visualFilter : regex}
-      fieldOptions={{
-        inputProps: { inputMode: "numeric" },
-      }}
-    />
-  );
+export const GenericNumericField = ({ minLength, maxLength, ...props }: Props): ReactElement => {
+  return <GenericTextField {...props} numericProps={{ minLength, maxLength }} />;
```
aloverso, line 11: what is the purpose of this component? It just seems like a passthrough. Why not just have us pass numeric props to GenericTextField when we need a numeric field?

web/src/components/onboarding/NumericField.tsx
```diff
@@ -1,53 +1,19 @@
...
-
-    props.onValidation(props.fieldName, !valid);
-  };
-
-  const handleChange = (): void => props.onValidation(props.fieldName, false);
-
-  const regex = (value: string): string => value.replace(/[^0-9]/g, "");
+interface Props extends Omit<OnboardingProps, "numericProps">, NumericFieldProps {}
 
+export const NumericField = ({ minLength, maxLength, ...props }: Props): ReactElement => {
```
aloverso, line 9: if this is specific to onboarding we should probably call it OnboardingNumericField

aloverso: but also again, now that it doesn't have any of its own logic, why does this component need to exist at all?

web/src/components/tasks/business-formation/BusinessFormationNumericField.tsx
```diff
@@ -1,41 +1,16 @@
...
-      <Content>{state.displayContent[props.fieldName].contentMd}</Content>
-      <GenericNumericField
-        value={state.formationFormData[props.fieldName]}
-        placeholder={state.displayContent[props.fieldName].placeholder}
-        error={state.errorMap[props.fieldName].invalid}
-        handleChange={handleChange}
-        onValidation={onValidation}
-        {...props}
-      />
+      <BusinessFormationTextField {...props} numericProps={{ minLength, maxLength }} />
```
aloverso, line 13: doesn't seem like this component needs to exist

web/test/pages/profile.test.tsx
```diff
@@ -184,7 +184,7 @@ describe("profile", () => {
     clickSave();
     await waitFor(() => {
       expect(
-        subject.queryByText(templateEval(OnboardingDefaults.errorTextMinimumNumericField, { length: "10" }))
+        subject.queryByText(templateEval(OnboardingDefaults.errorTextMinimumNumericField, { length: "9" }))
```
aloverso, line 187: why did this change?

mikewolfd: How it calculated the length in reference to the hyphen within the field changed, this is more accurate.

PR #1451, opened 2022-01-27
fix: fix type errors with act() on license and signup tests

namanaman:
Fixing type check errors seen here: https://app.circleci.com/pipelines/github/newjersey/navigator.business.nj.gov/2728/workflows/34efa10c-9fef-45c6-8594-3aa38b60ccd2/jobs/10098/parallel-runs/0/steps/0-112

- `act()` expects a callback of void function, but we are implicitly returning a promise
- `waitFor()` seems like a clearer way to handle this case
- can also fix by doing 
```
await act(async () => {
    await resolvedPromise
});
```


package.json
```diff
@@ -144,7 +144,7 @@
     "eslint-plugin-react": "7.28.0",
     "eslint-plugin-react-hooks": "4.3.0",
     "good-fences": "1.0.1",
-    "husky": "7.0.4",
+    "husky": "^7.0.4",
```
dsueltenfuss, line 147: Remove the ^ from this package please. We pin the version, so we are guaranteed the same version on each install

Issue #1471, opened 2022-01-28
Visually Identify Critical Steps in the Roadmap


aaronvux:
As Poppy,  I need to be able to easily identify critical tasks in the Roadmap so that I can focus on things that are required for me to start my business.

## Acceptance criteria

- Add "Required" icon to the left of the task status
- Display tool tip on hover: "Required"
- Add "Required + Copy" icon to the top of the task screen.
- Add to cms so that icon can be applied to any task 

## Additional Acceptance criteria
- Follow Design
- Works in Mobile
- 508 Compliant

## Design
https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=8198%3A193517

## Notes
- [ ] UX Review
- [ ] PO Approval
- [ ] Content Review

namanaman:
@carye3 @fiorella-io Just to confirm, do we prefer this square icon over other exclamation point options that are already in the US WDS? (from searching "exclamation point" in the Component Preview section - https://designsystem.digital.gov/components/icon/)

carye3:
The square ones @namanaman , users picked them out easily. Thx!

namanaman:
@carye3 @fiorella-io Two more questions:
1. In the design, the required icon has the color `cool-accent-darker` on the roadmap and #2E6276 on the task screen. Can we align them to the former or is the difference intentional?
2. I saw a mismatch in the heights of the tags between the design and current product, so I wanted to check which height we want to go with for all tags, so I can then align the new Required tag to have that same height. 
*In Design:*
Completed Status Tag - 18px 
Required Tag - 19px

*In Current Product:*
Completed Status Tag - 20px

fiorella-io:
@namanaman Use the cool-accent-darker and the current Completed tag height of 20px. Thanks!

namanaman:
Thanks @fiorella-io ! Is it possible to send me or update the design with a 20x20px version of the required icon's SVG. Then I can directly use that file in the code. I know it's late, so I can follow up Monday.

namanaman:
@carye3 @fiorella-io Following up about this, thanks^

aaronvux:
@Jesslax @CeciliaRam can we identify the critical tasks and add them in dev please?

Jesslax:
@aaronvux We can test it and take it off, does that work? My understanding was that we would not be expected to get this setup during this sprint.

aaronvux:
@Jesslax that's fine.

[...]

PR #1641, opened 2022-02-08
feat: [#1471] visually identify required tasks on roadmap and task screen

namanaman:


web/public/mgmt/config.yml
```diff
@@ -213,59 +214,68 @@ collections:
       - label: "Type"
         name: "type"
         widget: "select"
-        options: [ "FUNDING", "CERTIFICATION" ]
```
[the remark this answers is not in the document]
aloverso: is your prettier configured differently? strange that this wouldn't have been autoformatted already

web/src/components/TaskHeader.tsx
```diff
@@ -14,12 +15,28 @@ import {
...
 import React, { ReactElement, useState } from "react";
 import { CongratulatoryDialog } from "./CongratulatoryDialog";
+import { Tag } from "./njwds-extended/Tag";
 
 interface Props {
   task: Task;
   tooltipText?: string;
 }
 
+const RequiredTag = (): ReactElement => (
```
aloverso, line 25: if this is going to be its own independent const, I think it should probably live in its own file

web/src/display-defaults/tasks/TaskDefaults.ts
```diff
@@ -8,4 +8,6 @@ export const TaskDefaults: Record<string, string> = {
   unlockedBySingular: "Before starting this task, we recommend that you take a look at:",
   unlockedByPlural: "Before starting this task, we recommend you take a look at:",
   loadingTaskDependencies: "Loading task dependencies…",
+  requiredTagText: "Required",
+  requiredIconAltText: "Exclamation mark",
```
aloverso, line 12: Do we want/need alt text for this icon? My understanding is that alt text should convey meaning, not a description of what's on the page. So if the exclamation icon is only there to provide a visual cue, I feel like having a screenreader read out "exclamation mark" would be an unpleasant experience.

web/src/lib/utils/markdownReader.ts
```diff
@@ -62,4 +63,5 @@ export type TaskWithoutLinks = {
   callToActionText: string;
   postOnboardingQuestion: string;
   contentMd: string;
+  required?: boolean;
```
aloverso, line 66: do opportunities have this required tag also? I thought this was just for tasks

web/test/pages/[taskUrlSlug].test.tsx
```diff
@@ -204,6 +204,20 @@ describe("task page", () => {
...
   });
 
+  it("displays required tag in header if task is required", () => {
+    const subject = renderPage(generateTask({ required: true }));
+
+    expect(subject.getByAltText(TaskDefaults.requiredIconAltText)).toBeInTheDocument();
+    expect(subject.getByText(TaskDefaults.requiredTagText)).toBeInTheDocument();
+  });
+
+  it("does not displays required tag in header if task is not required", () => {
```
Comment (aloverso) [inline web/test/pages/[taskUrlSlug].test.tsx:214]: grammar - does not display

web/test/pages/roadmap.test.tsx
```diff
@@ -409,4 +410,46 @@ describe("roadmap page", () => {
...
+    const sectionPlan = subject.getByTestId("section-plan");
+    const allTasks = within(sectionPlan).getAllByRole("listitem");
+
+    expect(allTasks.length).toEqual(2);
+
+    expect(within(allTasks[0]).getByAltText(TaskDefaults.requiredIconAltText)).toBeVisible();
+    expect(within(allTasks[1]).queryByAltText(TaskDefaults.requiredIconAltText)).not.toBeVisible();
+  });
+
+  it("for a required task, hovering over icon renders tooltip", async () => {
```
aloverso, line 434: we like to keep the test names reading as sentences including the "it" (the reason is because it helps ensure that a test is describing the behavior of the component under test). So if we could rephrase this to read like a sentence where the component is the subject, that would be good. For example "it renders tooltip text when..."

web/src/lib/utils/helpers.tsx
```diff
@@ -90,7 +90,7 @@ export const setPreferencesCloseSection = (preferences: Preferences, current: Se
 export const getModifiedTaskContent = (
   roadmap: Roadmap | undefined,
   task: Task,
-  field: Exclude<keyof Task, "unlockedBy">
+  field: Exclude<keyof Task, "unlockedBy" | "required">
```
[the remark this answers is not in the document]
namanaman: Ah, since it only returned `string`, I was thinking the helper is used to get displayable content like the name and CTA text. Not sure if returning `string | boolean` will fit into the helper's design here, as all the callers are expecting a `string` back to feed into components like `<Content />` or `<TaskCTA />`.

web/src/lib/utils/markdownReader.ts
namanaman: Just for tasks I think, is the `TaskWithoutLinks` type here also being used for opportunities?

web/test/pages/roadmap.test.tsx
namanaman: Oooh that makes sense

web/src/display-defaults/tasks/TaskDefaults.ts
namanaman: I guess since there is a tooltip with a label that would get read by screen readers, there's no need for an alt text here then? I imagine a more semantic alt text would end up being the same as the tooltip label.

web/public/mgmt/config.yml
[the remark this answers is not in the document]
namanaman: It's the default, perhaps just from VSCode? Changes are changing single quotes to double, indenting long single-line arrays, adding trailing commas, and adding extra spaces around array brackets

web/src/components/TaskHeader.tsx
namanaman: I'll remove it from an independent const for now and just put it w the rest of the JSX. Easier to read and no current need for a separate component.

web/src/lib/utils/helpers.tsx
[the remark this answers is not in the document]
aloverso: Ah I see your point

web/src/lib/utils/markdownReader.ts
aloverso: oops. the way github rendered the diff made me misread this one. all good

Issue #1672, opened 2022-02-09
Task Format Improvements


carye3:
As a Navigator user, I need the ability to easily consume information on tasks screens so that I am not overwhelmed by large blocks of text.

## Acceptance criteria

- Add Issuing Agency & Form # & only display when populated
  - Add to Netlify
  - If either field is not populated, do not display that field.
- Update bullets within callout boxes (see designs)

## Designs

- [ ] [Desktop Design](https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=8711%3A200323)
- [ ] [Mobile Design](https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=8711%3A200629)

## Notes

- [ ] UX Review
- [ ] Content Review
- [ ] PO Approval

carye3:
@Jesslax

carye3:
@aaronvux this was overlooked last sprint and needs to have acceptance criteria added

aaronvux:
@namanaman does content have to be added for this to be tested?

Jesslax:
In general, this looks great. But it's not working for special license screens, for example like the designs we have for Consumer Affairs - https://dev.navigator.business.nj.gov/tasks/apply-for-shop-license... I don't actually think we need for these screens since the button says it, but would like a design take @carye3

namanaman:
Thanks @carye3 will update the callout rounding. Regarding the gray background on CTAs, currently aside from business formation, none of them have a gray background. Is this something needed in this ticket?

namanaman:
@carye3 I added the rounded corners and made the CTA backgrounds gray.

aaronvux:
@carye3 please review.

carye3:
nice! lgtm @namanaman @aaronvux

aaronvux:
Moving to deploy.

PR #1745, opened 2022-02-15
feat: [#1672] add task issuing agency and form

namanaman:


web/src/pages/tasks/[taskUrlSlug].tsx
```diff
@@ -149,12 +149,32 @@ export const TaskElement = (props: { task: Task; children?: ReactNode | ReactNod
         <Content>{beforeQuestion}</Content>
         {getPostOnboardingQuestion(props.task)}
         <Content>{afterQuestion}</Content>
+        {props.task.issuingAgency != null || props.task.formName != null ? (
```
Comment (dsueltenfuss) [inline web/src/pages/tasks/[taskUrlSlug].tsx:152]: Since these values can be strings or undefined, you can make this check simpler 
`props.task.issuingAgency || props.task.formName`

web/test/pages/[taskUrlSlug].test.tsx
```diff
@@ -214,6 +214,32 @@ describe("task page", () => {
     expect(subject.queryByText(TaskDefaults.requiredTagText)).not.toBeInTheDocument();
   });
```
Comment (dsueltenfuss) [inline web/test/pages/[taskUrlSlug].test.tsx:216]: A test that verifies neither display if they are both undefined, as well as the HR would also be good, since they are all surrounded by an if block

Issue #1695, opened 2022-02-11
Add Other Certifications and Display Dynamically on the Dashboard


aaronvux:
As Oscar, I need the ability to have Certifications dynamically displayed on the Dashboard so pertinent information to my business is easily accessible. 

## Acceptance criteria

- Onboarding

  - Business size is lower than 120 will see the SBE Certification on their Dashboard.

## Notes

- [x] PO Approval

aaronvux:
This will happen automatically:
  - Minority or Woman in the Ownership field will see the MWB Certification on their Dashboard.
  - Disabled Veteran in the Ownership field will see the DVOB Certification on their Dashboard.

FarazA22:
@Jesslax @aaronvux QQ, the veteran owned business certification markdown file is currently active. We are utilizing the  applicableOwnershipTypes key (image 1) to identify the possible dropdown values within the ownership dropdown that would display this certification. Since the SBE certification is based on the number of employees, I'd like to confirm if there is a use case that will result in populating the applicableOwnershipTypes key within the SBE markdown file (image 2).

<img width="481" alt="Screen Shot 2022-02-22 at 10 55 40 AM" src="https://user-images.githubusercontent.com/72581979/155169567-cda9290d-902f-4f2d-9400-52797596ab1b.png">

<img width="454" alt="Screen Shot 2022-02-22 at 11 02 36 AM" src="https://user-images.githubusercontent.com/72581979/155170908-03868975-6b39-4d10-90d3-ed8e05d27dd3.png">

FarazA22:
@aaronvux @Jesslax Ready for Review

Jesslax:
@FarazA22 this is not working for me. None of the other certifications are showing up

Jesslax:
@FarazA22 when I have women or minority selected DVOB license still shows. When I have disabled veteran, mwbe still shows.

FarazA22:
@Jesslax
We've added logic to show the certifications based on 1) applicableOwnershipTypes and 2) isSbe keys.

If applicableOwernshipTypes is empty (i.e. [ ]), it will show for everyone
To restrict it, you'll need to specify which ID it applies to (i.e. veteran-owned-business.md)

Note: if you wanted to show a small business certification for a specific ownershipType, you will need to fill applicableOwnershipType with that specific ID and have isSbe to equal true.

[...]

PR #1788, opened 2022-02-17
feat: [#1695] sbe certifications on dashboard

FarazA22:


web/src/lib/domain-logic/filterCertifications.ts
```diff
@@ -1,16 +1,27 @@
 import { Certification } from "@/lib/types/types";
 import { UserData } from "@businessnjgovnavigator/shared";
 
+export const SMALL_BUSINESS_MAX_EMPLOYEES = 119;
```
aloverso, line 4: If the logic in the story is "lower than 120", it feels weird to use 119 here. can we use >= 120 instead of > 119, because 120 is the magic-number called out in the story

aaronvux: @Jesslax makes sense to me if we use >=120

web/src/lib/domain-logic/filterCertifications.ts
```diff
@@ -1,16 +1,27 @@
...
+    if (it.applicableOwnershipTypes.length > 0) {
+      const ownerShipType = it.applicableOwnershipTypes.some((cert) => {
+        return userData.profileData.ownershipTypeIds.includes(cert);
       });
-    });
+      if (!ownerShipType) allowedCertification = false;
+    }
+
+    if (it.id === "sbe") {
+      allowedCertification = true;
```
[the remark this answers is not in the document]
FarazA22: The SBE cert is not based on ownership type, rather, it's based on employee count.

Issue #1557, opened 2022-02-04
Accommodate Sole Proprietorship and General Partnerships on Oscar Dashboard Calendar


aaronvux:
As an Oscar user who is an SP or GP, I need the ability to access the dashboard without providing a business formation date so I can complete onboarding.

## Acceptance criteria
- Make Business Formation Date Optional
- If Business Formation Date is not populated, no annual filing should display.
- Copy on the Business Formation question will be updated by Jess/C 

## Notes
- [ ] UX Review
- [ ] PO Approval
- [ ] Content Review

aaronvux:
@namanaman @mikewolfd Business Formation Date is showing in the calendar even though no Business Formation Date was input.
<img width="1242" alt="Screen Shot 2022-02-23 at 9 39 07 AM" src="https://user-images.githubusercontent.com/36163601/155341394-5a296af4-0049-40ba-a1c4-490275af7df4.png">

namanaman:
@aaronvux Should be fixed now

aaronvux:
Looks good.  Moving to staging.

PR #1808, opened 2022-02-18
feat: [#1557] date of formation optional for oscar

namanaman:


web/test/pages/dashboard.test.tsx
```diff
@@ -93,6 +95,14 @@ describe("dashboard", () => {
...
   });
 
+  it("displays filings calendar without annual report date if date of formation undefined", () => {
+    const userData = generateUserData({ profileData: generateProfileData({ dateOfFormation: undefined }) });
+    useMockUserData(userData);
+
+    const subject = renderPage({});
+    expect(subject.queryByText("Annual Report")).not.toBeInTheDocument();
+  });
+
```
aloverso, line 105: I don't think this test is necessary / actually testing this logic - I believe it would pass even if date of formation is required, right? Because what shows in the calendar is controlled by what is in the user's `filings` data, which is computed on the backend using their date of formation. But for a frontend test here, I can't imagine a situation in which this test would ever fail. So I don't think we need to have it.

namanaman: Ohh I see I didn't realize it connects tot the backend to provide the filings from date. Makes sense, I'll remove this

PR #1850, opened 2022-02-23
refactor: cypress industry tests / starting a new business onboarding…

FarazA22:
… flow

### DIFF
--- a/web/cypress/support/page_objects/onboardingPage.ts
+++ b/web/cypress/support/page_objects/onboardingPage.ts
@@ -1,0 +1,72 @@
+import { LookupIndustryById } from "@businessnjgovnavigator/shared/";
+
+export class OnboardingPage {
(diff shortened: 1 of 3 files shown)


web/cypress/support/page_objects/onboardingPage.ts
```diff
@@ -0,0 +1,72 @@
...
+  getHomeBased(radio?: boolean) {
+    return cy.get(`input[name="home-based-business"]${radio === undefined ? "" : `[value="${radio}"]`}`);
+  }
+
+  getLocationDropdown() {
+    return cy.get('[data-testid="municipality"]');
+  }
+
+  getNextButton() {
+    return cy.get('[data-testid="next"]');
```
aloverso, line 33: does this need to be its own function? if we're only ever using it to click the next button?

Issue #1204, opened 2021-12-19
Merge Business Name Check as New Step in Business Formation Task


aaronvux:
As Poppy, I need to understand if my Business Name is still available when I am forming my business so that I don't get frustrated during the process.

## Acceptance criteria
- Add New "Name" Step to Stepper
- Create input field for Business Name
- Clicking Check Availability or using Enter on the keyboard will call the Business Name API to check that the user's business name is available.
  - If the user changes the Business Name after a successful result is returned, do not display continue button. The user must check availability again.
- Display Success message if Business Name is available
  - Display Continue button
  - Display copy that the name is not official until formation is complete.
  - Link to Profile to update the business name.
- Display Warning message with unavailable Business Names if unsuccessful.
- Remove Business Name search task from the roadmap & dependencies for LLC.
- Move "Check Available Names and Form Your Business" task to the first task in the Start step in the roadmap and mini- roadmap.

## Analytics
- Add:  business_formation_name_step_continue_button/click/go_to_next_formation_step
- Add:  business_formation_name_step_update_name_on_roadmap/click/update_name_on_roadmap
- Copy: task_business_name_check_availability/submit/view_business_name_availability

## Design

**Desktop**
https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=9157%3A228708

**Mobile**
*Note: DO NOT use content from mobile design; use desktop copy instead*
https://www.figma.com/file/408Krq8roicOzFVYrjwIdE/Navigator-Designs?node-id=9157%3A229401

## Notes

- [ ] UX Review
- [ ] PO Approval
- [ ] Content Review

aaronvux:
Double check design to make sure Update Name on your roadmap with this one link is available.

aaronvux:
Splitting 
**UI**
- Decrease field the height of all input fields to match button height****

namanaman:
@carye3 @fiorella-io In this design compared to the existing business search task, there are differences in the error/success state styles. Do we want to also apply this new error/success state styling to the existing business search task? 

This will help plan how much code the two name checks (in business formation and separate task) should share.

aaronvux:
the designs for this ticket are specific to the LLC Business Formation and nothing should change for the other business name search functionality for the other legal structures.

namanaman:
@Jesslax @CeciliaRam Will you be adding the new content for this to the repo or should I? (not sure what the process has been in the past)

[...]

PR #1883, opened 2022-02-25
feat: [#1204] merge business name check into formation

namanaman:


web/src/lib/data-hooks/useBusinessNameSearch.ts
```diff
@@ -0,0 +1,104 @@
+import * as api from "@/lib/api-client/apiClient";
+import { useUserData } from "@/lib/data-hooks/useUserData";
+import { NameAvailability, SearchBusinessNameError } from "@/lib/types/types";
+import analytics from "@/lib/utils/analytics";
+import { ChangeEvent, FocusEvent, FormEvent, useState } from "react";
+
+export const useBusinessNameSearch = (
```
namanaman, line 7: Abstracted out the state and fetching logic into a custom hook here. This felt better than the two business name search's sharing the same React component, as there were lots of subtle differences in the display logic, it felt cleaner to have shared hook and separate UI components.

web/src/components/tasks/business-formation/BusinessNameSection.tsx
```diff
@@ -0,0 +1,173 @@
...
+    updateNameOnProfile,
+  } = useBusinessNameSearch(true);
+
+  const submitNameAndContinue = async () => {
+    if (!userData) return;
+    const newFormationFormData = {
+      ...state.formationFormData,
+      businessName: submittedName,
+    };
+    setFormationFormData(newFormationFormData);
```
namanaman, line 45: This felt a little weird to update both the local (under FormationContext) state and the user data with the same value at the same time. But I thought it's better than updating the local FormationContext state every time the name input changes, since that would trigger a re-render of the entire `<BusinessFormation />` each time. Instead, I use local non-Context state in this component to handle input changes, and only update the Context one upon submit.

PR #1912, opened 2022-03-01
chore: added structure to files folder

FarazA22:


namanaman:
Looks good to me, I feel like this is easier structure to find CSS. I may remove the numbering from the folder names, though.

PR #2047, opened 2022-03-16
1515/mdeeb/guestmode

mikewolfd:


api/src/functions/express/app.ts
```diff
@@ -118,9 +119,10 @@ const apiFormationClient = ApiFormationClient(
 );
 
 app.use(bodyParser.json({ strict: false }));
+app.use("/api", userRouterFactory(userDataClient, updateLicenseStatus));
 app.use(
-  "/api",
-  userRouterFactory(userDataClient, updateLicenseStatus, addGovDeliveryNewsletter, addToAirtableUserTesting)
+  "/api/ext",
```
aloverso, line 124: typo? incomplete word here. also, i'd prefer to follow the pattern of any sub-routes past /api being defined in the router instead of top-level app

aloverso: actually, never mind. given the change to the index.ts lambda, this makes more sense here because it is required for the auth bypass.

aloverso: i would still rather have a real word than "ext" though

content/src/fieldConfig/config.json
```diff
@@ -176,7 +176,18 @@
...
+    "navBarGuestRegistrationText": "Register now",
+    "guestAlertTitle": "### Register to Save Your Progress",
+    "guestAlertBody": "Now that you've filled out your business information, you can save your progress and come back later by [creating an account](#). ",
+    "guestSuccessTitle": "### You are Now Registered!",
+    "guestSuccessBody": "You can now save your progress, access more features, and edit your profile.",
+    "guestModalTitle": "## Create an Account to Access this Feature",
+    "guestModalBody": "To use this feature, you need to create a free account with the State of New Jersey's secure login system, MyNJ. This will ensure your privacy, and allow you to come back any time.",
+    "guestModalButtonText": "Complete my Registration",
+    "guestModalSubText": "Already have an account? [Log In](#)"
+
```
aloverso, line 190: if only there was some way to make sure tests didn't pass if these we're added to the CMS config 😉

web/src/components/SignUpDialogs.tsx
```diff
@@ -0,0 +1,225 @@
...
+            </IconButton>
+            <Content>{contentMap[registrationAlertStatus]}</Content>
+          </div>
+        </div>
+      </ToastAlert>
+    );
+  }
+  return <></>;
+};
+export const UseAuthToast = (): ReactElement => {
```
aloverso, line 87: can we separate these into different files? this makes them harder to search for and find later

aloverso: same with the tests

web/src/components/SignUpDialogs.tsx
```diff
@@ -0,0 +1,225 @@
...
+    }
+  }, [isAuthenticated, registrationAlertStatus, setRegistrationAlertStatus]);
+
+  const alertMap: Record<Exclude<RegistrationStatus, "IN_PROGRESS">, AlertVariant> = {
+    SUCCESS: "success",
+    DUPLICATE_ERROR: "error",
+    RESPONSE_ERROR: "error",
+  };
+
+  const contentMap: Record<Exclude<RegistrationStatus, "IN_PROGRESS">, string> = {
```
aloverso, line 34: we could define `Exclude<RegistrationStatus, "IN_PROGRESS">` as its own type in this file

web/src/components/SignUpDialogs.tsx
```diff
@@ -0,0 +1,225 @@
...
+            >
+              <Icon className="usa-icon--size-4">close</Icon>
+            </IconButton>
+            <Content>{contentMap[registrationAlertStatus]}</Content>
+          </div>
+        </div>
+      </ToastAlert>
+    );
+  }
+  return <></>;
```
aloverso, line 85: could we make this a guard clause instead?

web/src/components/SignUpDialogs.tsx
aloverso: also, naming a component with the `use` prefix feels weird, that should be reserved for hooks

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -119,11 +139,88 @@ describe("<NavBar />", () => {
...
+
+      userEvent.click(menuEl);
+      await waitFor(() => {
+        expect(
+          subject.queryByText(Config.navigationDefaults.navBarGuestRegistrationText)
+        ).not.toBeInTheDocument();
+      });
+    });
+
+    it("sends user to selfRegistration", async () => {
```
aloverso, line 179: feels like it should have a "...when"

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -152,30 +249,107 @@ describe("<NavBar />", () => {
...
-
-      const operateRef: Record<string, OperateReference> = {
-        "some-tax-filing-identifier-1": {
-          name: "some-filing-name-1",
-          urlSlug: "some-urlSlug-1",
-          urlPath: "some-path",
-        },
+    const renderMobileTaskNav =
+      (isAuthenticated = IsAuthenticated.TRUE) =>
+      (config?: { includeOperateRef: boolean }): RenderResult => {
```
aloverso, line 254: can we rewrite this function-that-returns-a-function as something more readable?

web/src/components/navbar/NavBarDesktop.tsx
```diff
@@ -91,9 +94,17 @@ export const NavBarLoggedInDesktop = (props: Props): ReactElement => {
...
+              className={`text-bold text-${
+                state.isAuthenticated == "TRUE" ? "primary" : "base"
+              } flex flex-align-center`}
+            >
+              <Icon className="usa-icon--size-4 margin-right-1">
+                {state.isAuthenticated == "TRUE" ? "account_circle" : "help"}
+              </Icon>
+              <div>
+                {state.isAuthenticated == "TRUE" ? userName : Config.navigationDefaults.navBarGuestText}
+              </div>
```
aloverso, line 107: with this many inline conditionals checking the same thing, might be nicer to write as a function? like

```

const textColor = state.isAuthenticated == "TRUE" ? "text-primary" : "text-base"
// ...(and same for the other variables)

<div
              className={`text-bold ${textColor} flex flex-align-center`}
            >
              <Icon className="usa-icon--size-4 margin-right-1">
                {state.isAuthenticated == "TRUE" ? "account_circle" : "help"}
              </Icon>
              <div>
                {state.isAuthenticated == "TRUE" ? userName : Config.navigationDefaults.navBarGuestText}
              </div>
```

web/src/components/navbar/NavSideBarUserSettings.tsx
```diff
@@ -28,38 +36,62 @@ export const NavSideBarUserSettings = (): ReactElement => {
...
+              className={`flex flex-align-center text-${
+                state.isAuthenticated == "TRUE" ? "primary" : "base"
+              }`}
+            >
+              <Icon className="margin-right-1 usa-icon--size-3">
+                {state.isAuthenticated == "TRUE" ? "account_circle" : "help"}
+              </Icon>
+              <span>
+                {state.isAuthenticated == "TRUE" ? userName : Config.navigationDefaults.navBarGuestText}
+              </span>
```
aloverso, line 49: same comment as other bit

aloverso: maybe this should be a component, if it's shared between here and the other navbar component?

web/src/components/navbar/NavSideBarUserSettings.tsx
```diff
@@ -28,38 +36,62 @@ export const NavSideBarUserSettings = (): ReactElement => {
...
+                      onClick={() => {
+                        analytics.event.account_menu_my_profile.click.go_to_profile_screen();
+                      }}
+                    >
+                      <span className="text-base">{Config.navigationDefaults.profileLinkText}</span>
+                    </Button>
+                  </Link>
+                </div>
+              </>
+            ) : (
```
aloverso, line 82: can we refactor this component to call functions or components? This is hard to read, would be way easier if it was like `{state.isAuthenticated == "TRUE" ? renderSomeDescriptiveName() : renderAnotherThing()}`

web/src/components/njwds-extended/Button.test.tsx
```diff
@@ -105,4 +105,12 @@ describe("Button", () => {
     ).baseElement;
     expect(subject).toMatchSnapshot();
   });
+  it("with intercomButton button displays correctly", () => {
```
aloverso, line 108: test newline sorry

web/src/components/njwds-extended/Button.tsx
```diff
@@ -33,19 +35,19 @@ export const Button = (props: Props): ReactElement => {
...
       break;
     case "secondary":
       style = "usa-button usa-button--outline padding-y-1";
-      disabledClass = "usa-button--outline-disabled";
+      disabledClass = " usa-button--outline-disabled";
       break;
     case "secondary-input-field-height":
       style = "usa-button usa-button--outline padding-y-2";
-      disabledClass = "usa-button--outline-disabled padding-y-2";
+      disabledClass = " usa-button--outline-disabled padding-y-2";
```
aloverso, line 50: why are we adding back the prefix space? we should be doing this in the template string, not in these variables

web/src/components/onboarding/MunicipalityDropdown.tsx
```diff
@@ -62,7 +63,7 @@ export const MunicipalityDropdown = (props: Props): ReactElement => {
       renderInput={(params) => (
         <TextField
           {...params}
-          id={props.fieldName}
+          // id={props.fieldName}
```
aloverso, line 66: stray comment

web/src/components/navbar/NavBar.test.tsx
mikewolfd: unfortunately no because of how displaysUserNameOrEmail uses it

web/src/lib/utils/storage-helper.test.tsx
```diff
@@ -0,0 +1,135 @@
+import { swrPrefixToIgnore, userDataPrefix, UserDataStorage, UseStorage } from "@/lib/utils/storage-helpers";
+import { generateUserData } from "@/test/factories";
+
+describe("storage-helper", () => {
```
aloverso, line 4: this file and its test should be camelCase instead of kebab case to fit our pattern

web/src/lib/utils/storage-helpers.ts
```diff
@@ -0,0 +1,146 @@
...
+  return {
+    get,
+    set,
+    keys,
+    clear,
+    delete: _delete,
+  };
+};
+
+interface UserDataStorageReturnValue {
```
aloverso, line 48: can we separate these two objects into individual files? UseStorage is generic, whileUserDataStorage is specific to user data. 

Also, same comment as before, I'd prefer to not use the `Use`... prefix for things that aren't react hooks

web/src/pages/dashboard.tsx
```diff
@@ -58,12 +71,15 @@ const DashboardPage = (props: Props): ReactElement => {
...
+                    <button
+                      role="link"
+                      style={{ border: 0, cursor: "pointer", backgroundColor: "transparent" }}
+                      className="usa-link"
+                      onClick={editOnClick}
+                      data-testid="grey-callout-link"
                     >
                       {Config.dashboardDefaults.editProfileText}
-                    </a>
+                    </button>
```
aloverso, line 82: this should probably be a component `Button`

web/src/styles/base/typography.scss
```diff
@@ -96,6 +96,16 @@ h6 {
...
 
+li,
+p,
+div,
+span,
+strong {
+  margin-block-start: 0;
+  margin-block-end: 0;
+  line-height: 150% !important;
+}
```
aloverso, line 107: did Faraz's commit get wrapped up in yours somehow?

web/test/pages/onboarding/onboarding-shared.test.tsx
```diff
@@ -88,8 +92,39 @@ describe("onboarding - shared", () => {
...
+        ...request,
+        user: {
+          ...request.user,
+          externalStatus: {
+            ...request.user.externalStatus,
+            userTesting: { status: "SUCCESS", success: true },
+          },
+        },
+      })
+    );
```
aloverso, line 127: why is this section relevant to this test?

web/test/pages/onboarding/onboarding-starting.test.tsx
```diff
@@ -198,17 +197,55 @@ describe("onboarding - starting a business", () => {
...
+        ...expectedUserData,
+        user: { ...expectedUserData.user, externalStatus: {} },
+      });
+
+      expect(mockApi.postUserTesting).toHaveBeenCalledWith({
+        ...expectedUserData,
+        user: {
+          ...expectedUserData.user,
+          externalStatus: { newsletter: { status: "SUCCESS", success: true } },
+        },
```
aloverso, line 237: again, this feels like it should be not part of this test. at the very least, it makes the test so long. If necessary, could we move the mocking step to the beforeEach or something?

Issue #1952, opened 2022-03-09
Spike: AB Testing in Production


aaronvux:
Research the ability to present two different experiences to users in Production.

aaronvux:
Create separate spike for:

**Roadmap/Mini Roadmap**
- Task Titles

**Tasks**
- Content

aloverso:
@aaronvux @skeeJay can we add some requirements to this story? This will determine how we approach this spike. For example:
- How will we determine which landing page to show a user? Should it be re-decided for every new session or kept as long as possible?
- Will the split between A & B experiences be 50/50, or some custom split that we want to be easily configurable?
- Would you prefer we optimize for static rendering? Here's what this means. Currently, we statically generate HTML for every page of our site, including the landing page. This lets a bot like google easily scrape and index our content. So far, we've defaulted to static generation, which means sometimes pages "switch" content quickly as soon as the user's data loads. A good example of this is the [business formation task](https://dev.navigator.business.nj.gov/tasks/form-business-entity) which briefly shows the static content before it realizes that you are an LLC user and then switches to show you the Formation form. This allows the paragraph content to be indexed by google and searchable. In the same way, we could do one of these: have one version of the landing page be the "default" and indexed by google. This would have the drawback that users with the "B" version would see the flicker/switch instantaneously, like on business formation. Or, otherwise, we could decide/fetch which page to show the user before showing them any content - with the downside of losing some of our landing page getting indexed by google.

aaronvux:
@aloverso Here are my thoughts:
1. I would think this could be random, but I could also see the need for this based on location or screen size.  Once, a user is shown a particular page setup, I believe we would want to keep them on that setup.  What's possible?
2. I would think 50/50 split, but could also see the need to segregate a small group with low risk to the main user population.
3.  Not sure here.  I guess it depends if we are testing with existing content for one group and some new content for the other or new content for both groups.  Probably need to talk about this one.

aloverso:
Anne notes 

Brand-new user visits landing page
-> no AB Experience data exists in browser. Landing page chooses a 50/50 experience and saves it in local storage

User clicks "Get Started", goes through onboarding
-> AB Experience is saved in user data

PATH A - User leaves site and doesn't come back for weeks
When they return to landing page, it loads AB Experience from browser and displays same experience
But if they are visiting from a different device/browser, we have no way of guaranteeing the same experience.

PATH B - User chooses to register
[...]

PR #2360, opened 2022-04-11
feat: [#1952] a/b testing for landing page

aloverso:


web/src/lib/types/types.ts
```diff
@@ -353,7 +353,7 @@ export const owningFlowDisplayFields = Object.keys(emptyOwningFlowContent) as (k
 
 export const startFlowDisplayFields = Object.keys(emptyStartingFlowContent) as (keyof StartingFlowContent)[];
 
-const businessUserDisplayFields = Object.keys(emptyBusinessUser) as (keyof BusinessUser)[];
```
mikewolfd, line 356: I think this will mess up typing in our IDE's, no?

aloverso: Will it? I'm not sure. I don't **think** it changed anything for me?

web/src/lib/types/types.ts
```diff
@@ -353,7 +353,7 @@ export const owningFlowDisplayFields = Object.keys(emptyOwningFlowContent) as (k
...
+const businessUserDisplayFields = Object.keys(createEmptyUser()) as (keyof BusinessUser)[];
```
mikewolfd, line 356: read below with this messing with the typing

aloverso: Is it actually causing issues? If the IDE knows the return type of the function then can't it do object.keys?

aloverso: I suppose I could override that, though... but yeah, I'm curious how this change disrupts the types?

PR #2480, opened 2022-04-20
feat: 2081-cannabis-priority-status

FarazA22:


web/src/pages/tasks/[taskUrlSlug].tsx
```diff
@@ -207,3 +207,14 @@ export const getStaticProps = async ({
 };
 
 export default TaskPage;
+
+export const getUnlockedBy = (
```
Comment (aloverso) [inline web/src/pages/tasks/[taskUrlSlug].tsx:211]: it's typically an anti-pattern to export helper functions from components into other components - keeping logic like this separate from React components can help with refactoring. I think this should go in utils or something if it's going to be imported into other components

Comment (aloverso) [reply web/src/pages/tasks/[taskUrlSlug].tsx:211]: however, I wonder if this functionality (filter the "unlocked by" tasks to only those that are uncompleted) should really just be logic inside of the UnlockedBy component, instead of a helper function that gets called by every component just to pass into <UnlockedBy />

web/test/helpers.tsx
```diff
@@ -128,6 +128,10 @@ export const randomIntFromInterval = (min: string, max: string) => {
   return Math.floor(Math.random() * (Number(max) - Number(min) + 1) + Number(min));
 };
 
+export const randomElementFromArray = (array: unknown[]) => {
+  return array[Math.floor(Math.random() * array.length)];
+};
+
```
aloverso, line 134: I think you could use generic types here to keep type safety. Like `export const randomElementFromArray = <T, >(array: T[]): T => {`

web/src/components/tasks/CannabisPriorityStatus.tsx
```diff
@@ -0,0 +1,342 @@
...
+  const [displayNoPriorityType, setDisplayNoPriorityType] = useState(false);
+
+  const priorityTypes = useMemo(() => {
+    return [
+      ...priorityTypesObj.minorityOrWomen,
+      ...priorityTypesObj.veteran,
+      ...priorityTypesObj.impactZone,
+      ...priorityTypesObj.socialEquity,
+    ];
+  }, []);
```
aloverso, line 62: if this is constant, why use useMemo and not just define it as a constant external to the component, like the priorityTypesObj?

web/src/components/tasks/CannabisPriorityStatus.tsx
```diff
@@ -0,0 +1,342 @@
...
+};
+
+export const CannabisPriorityStatus = (props: Props) => {
+  const { userData, update } = useUserData();
+  const taskFromRoadmap = useTaskFromRoadmap(props.task.id);
+  const unlockedByTask = getUnlockedBy(taskFromRoadmap, userData);
+  const [successToastIsOpen, setSuccessToastIsOpen] = useState(false);
+
+  const [displayFirstTab, setDisplayFirstTab] = useState(true);
+  const [displayNextTabButton, setDisplayNextTabButton] = useState(false);
```
aloverso, line 47: given the amount of if/else logic around `if !displayFirstTab` and such, could we reduce complexity by making the two tabs as two separate components?

web/src/components/tasks/CannabisPriorityStatus.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+      setDisplayMWPriorityType(minorityOrWomenPriorityTypeSelected ? true : false);
+      setDisplayVeteranPriorityType(veteranPriorityTypeSelected ? true : false);
+      setDisplayImpactZonePriorityType(impactZonePriorityTypeSelected ? true : false);
+      setDisplaySocialEquityPriorityType(socialEquityPriorityTypeSelected ? true : false);
+      setDisplayNoPriorityType(userData.taskItemChecklist[priorityTypesObj.noneOfTheAbove] ? true : false);
+    }
+  }, [userData, update, priorityTypes, displayFirstTab]);
+
+  const handleCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
aloverso, line 111: might rename this to "handleNoneOfTheAboveCheckboxChange" or something because there's a lot of other checkboxes on the page that this does not handle

web/src/components/tasks/CannabisPriorityStatus.tsx
```diff
@@ -0,0 +1,342 @@
...
+      });
+
+      const impactZonePriorityTypeSelected = priorityTypesObj.impactZone.some((key) => {
+        if (key in userData.taskItemChecklist) return userData.taskItemChecklist[key] === true;
+        else return false;
+      });
+
+      const socialEquityPriorityTypeSelected = priorityTypesObj.socialEquity.some((key) => {
+        if (key in userData.taskItemChecklist) return userData.taskItemChecklist[key] === true;
+        else return false;
```
aloverso, line 100: these could all be refactored to `return userData.taskItemChecklist[key] === true` right? because if the key isn't in `userData.taskItemChecklist` then `userData.taskItemChecklist[key]` will return undefined

web/src/components/tasks/CannabisPriorityStatus.tsx
```diff
@@ -0,0 +1,342 @@
...
+    return [
+      ...priorityTypesObj.minorityOrWomen,
+      ...priorityTypesObj.veteran,
+      ...priorityTypesObj.impactZone,
+      ...priorityTypesObj.socialEquity,
+    ];
+  }, []);
+
+  useEffect(() => {
+    if (userData && userData.taskItemChecklist && displayFirstTab) {
```
aloverso, line 65: why the check for `userData.taskItemChecklist` - that should never be undefined, right? only ever `{}`

web/src/components/tasks/CannabisPriorityStatus.tsx
```diff
@@ -0,0 +1,342 @@
...
+          <div className="margin-y-2">
+            <FormControlLabel
+              label={Config.cannabisPriorityStatus.noPriortyStatusCheckboxText}
+              control={
+                <Checkbox
+                  onChange={handleCheckboxChange}
+                  checked={
+                    userData && userData.taskItemChecklist[priorityTypesObj.noneOfTheAbove]
+                      ? userData.taskItemChecklist[priorityTypesObj.noneOfTheAbove]
+                      : false
```
aloverso, line 183: similar comment as above, i think this could be simplified to `userData?.taskItemChecklist[priorityTypesObj.noneOfTheAbove] === true`

web/src/components/tasks/CannabisPriorityStatus.tsx
FarazA22: will address with accordion ticket

PR #2637, opened 2022-04-29
poppy side bar

FarazA22:


content/src/fieldConfig/config.json
```diff
@@ -330,7 +330,15 @@
...
-    "greyBoxNotEnteredText": "Not entered"
+    "greyBoxNotEnteredText": "Not entered",
+    "sideBarHeading": "For You",
+    "sideBarWelcomeCardHeader": "Welcome!",
+    "sideBarWelcomeText": "Welcome to the Navigator, your personalized step-by-step guide to running your business in New Jersey.",
+    "sideBarSecondaryWelcomeText": "As soon as you form your business and register for taxes, you will see opportunities here including funding, certifications, and much more.",
+    "sideBarGuestModeHeadingText": "Register to Save Your Progress",
+    "sideBarGuestModeText": "**Don’t lose your progress.** [Create an account now.](#)",
+    "sideBarRegisteredHeadingText": "You’re Registered!",
+    "sideBarRegisteredText": "You can now **save your progress**, complete **forms**, add information to your **Business Profile**, and more!"
```
dsueltenfuss, line 341: I think each of the sidebar cars should be a separate markdown file, that is self contained for the card

Trying to manage them as multiple lines in the general config will get very difficult, very quickly

Additionally, each card should have an id, so we can track which ones they have hidden (such as the successful registration one)

shared/src/userData.ts
```diff
@@ -48,6 +49,7 @@ export interface Preferences {
   roadmapOpenSteps: number[];
   hiddenFundingIds: string[];
   hiddenCertificationIds: string[];
+  showSuccessfulRegistrationCard: boolean | undefined;
```
dsueltenfuss, line 52: Instead of a boolean, I think this would be better as a list of ids for hidden sidebar cards.

Issue #2502, opened 2022-04-22
Gradual Graduation (MVP): Add Sidebar to Poppy Experience with Appropriate Prompt Cards


aaronvux:
As Poppy, I need to have a better experience that allows me to see information that helps me graduate to Oscar so my business can become operational more easily.

## Acceptance Criteria
- Add Content Column to Poppy Roadmap landing page.
  - Display welcome card and content 
  - Welcome content cannot be closed.
- Display Snackbar for Registration
- Once Snackbar is closed, display registration content.
  - Clicking the link will bring the user to myNJ.
- After successful registration, display You're Registered content in Content Column.
  - Clicking close icon will remove the content from the column.

## Additional Acceptance Criteria
- 508 Compliant
- Follows Design
- Works in Mobile

## Design
- Desktop
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=12037%3A232935
- Mobile
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=12037%3A239964
**Prototype**
https://www.figma.com/proto/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?page-id=11551%3A187892&node-id=11915%3A236185&viewport=241%2C48%2C0.16&scaling=min-zoom&starting-point-node-id=11915%3A236185

## Approvals
- [ ] Content Review
- [ ] UX Review
- [ ] PO Approval

aaronvux:
Need to prioritize this with updating the registration snackbar displaying story.

CeciliaRam:
@aaronvux (@Jesslax )
"Clicking the link will bring the user to myNJ." - will that take any user info from onboarding over to MyNJ?

aaronvux:
I don't believe so.  We don't pass anything currently

Jesslax:
@aaronvux @aloverso dont we pass the name and email? Just wanted to make sure that would continue

aaronvux:
@Jesslax @CeciliaRam you're right.  We pass name and email. @aloverso cannot we not pass the other fields so we can avoid the myNJ UI all together. 

My assumption that we would continue to pass this info to myNJ from the new registration link.

aaronvux:
@kmelefsky Discuss removing links in content bar for form your business and register for taxes for 16th minute.

[...]

PR #2717, opened 2022-05-06
feat: [#2502] added sidebar to roadmap

FarazA22:


shared/src/userData.ts
```diff
@@ -29,6 +29,7 @@ export const createEmptyUserData = (user: BusinessUser): UserData => {
       roadmapOpenSteps: [],
       hiddenCertificationIds: [],
       hiddenFundingIds: [],
+      hiddenRoadmapSidebarCards: ["successful-registration", "not-registered"],
```
aloverso, line 32: Why do these start hidden? Shouldn't this be controlled by whether they are registered or not?

web/src/components/auth/SignUpToast.tsx
```diff
@@ -55,7 +55,22 @@ export const SignUpToast = (): ReactElement => {
...
+                );
+                update({
+                  ...userData,
+                  preferences: {
+                    ...userData.preferences,
+                    hiddenRoadmapSidebarCards: showRegistrationCard,
+                  },
+                });
+              }
+            }}
```
aloverso, line 73: small stylistic thing, but with this many lines of code, it might be more readable if the onClick were its own function

Issue #2247, opened 2022-04-04
Feedback: Add "Request a Feature" to Feedback Modal


aaronvux:
As a Navigator User, I need the ability to request features so that I can get information to support my business.

## Acceptance Criteria
**Feedback Modal**
- Add Request a Feature to the Feedback modal after onboarding has been completed.
  - Inline validation on text field.  "This field is required."
- Clicking the option displays Request a Feature Form with 1 question (see design)
- After clicking submit, display Thank you message.

**Storage**
- Write user submitted content to Feature Tab in Airtable

**Airtable Columns**
- Detail
- Date & Time
- Email
- Industry
- Persona
- Page Report was initiated on
- Device
- Browser
- Screen Width

**AIrtable is Down**
- Display inline error message on the form (see design)

## Additional Acceptance Criteria
- 508 Compliant
- Follows Design
- Works in Mobile

## Design
- Desktop Modal
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=11817%3A226804
- Desktop Feature Request
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=10984%3A210264
- Desktop Thank You Modal
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=10984%3A210418 
- Mobile
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=12312%3A250623

## Approvals
- [ ] Content Review
- [ ] UX Review
- [ ] PO Approval

FarazA22:
@aaronvux @fiorella-io @carye3 
- I'd like to add logic that prevents the submission of the feature request if the input is empty. Any concerns? Should the app give the user feedback if the request is not submitted due to an empty input field?
- Is there a max character size for the input field?

fiorella-io:
@FarazA22 
1. I agree to not allow users to submit an empty form
2.  I'm not sure what the character limit should be? 400? @aaronvux @carye3 
3. Yes, it should go to the main Feedback modal

FarazA22:
[...]

Issue #2825, opened 2022-05-17
Create separate AirTable tables for dev/staging and prod


aloverso:
- [ ] user testing table
- [ ] feedback feature request tabel
- [ ] feedback - user issue table env variable for dev/prod

aloverso:
verified - all good

PR #2886, opened 2022-05-18
feat: [#2247] request a feature feedback modal

FarazA22:


web/src/lib/types/types.ts
```diff
@@ -676,3 +676,5 @@ export type NaicsCodeObject = {
   TwoDigitCode?: number[];
   industryIds?: string[];
 };
+
+export type feedbackRequestDialogNames = "Select Feedback" | "Feature Request" | "Request Submitted";
```
aloverso, line 680: types should have a capital first letter

api/src/client/AirtableFeedbackClient.ts
```diff
@@ -0,0 +1,52 @@
...
+import { FeedbackClient } from "../domain/types";
+
+type AirtableConfig = {
+  apiKey: string;
+  baseId: string;
+  baseUrl: string;
+};
+
+export const AirtableFeedbackClient = (config: AirtableConfig, logWriter: LogWriterType): FeedbackClient => {
+  const table = "tblHZp9vjRIdooR4Z";
```
dsueltenfuss, line 14: This should be moved to an env variable, so that it's more easily configurable

api/src/client/AirtableFeedbackClient.ts
```diff
@@ -0,0 +1,52 @@
...
+    return new Promise((resolve, reject) => {
+      const fields = {
+        fldEnXB3SBrgib5sD: feedbackRequest.detail,
+        fldcCp3dxMI5Ntn2m: userData.user.email,
+        fld5nU9Jeldk7bW8Y: userData.profileData.industryId,
+        fldEPNY8Aub2I8WEf: userData.profileData.hasExistingBusiness ? "owning" : "starting",
+        fldvKgl7DYMSXZTsv: feedbackRequest.pageOfRequest,
+        fldMGEB5JoyVIZhuD: feedbackRequest.device,
+        fldN1mp5PrXQLt1Fg: feedbackRequest.browser,
+        fldlo19Z2Ud1p0R24: feedbackRequest.screenWidth,
```
dsueltenfuss, line 33: These should all be real field names, similar to the User Testing airtable

web/src/lib/api-client/apiClient.ts
```diff
@@ -24,22 +24,26 @@ export const getCompletedFiling = (): Promise<UserData> => {
...
   return post(`/external/newsletter`, userData, false);
 };
 
-export const postUserTesting = async (userData: UserData): Promise<UserData> => {
+export const postUserTesting = (userData: UserData): Promise<UserData> => {
   return post(`/external/userTesting`, userData, false);
 };
 
-export const postGetAnnualFilings = async (userData: UserData): Promise<UserData> => {
+export const postGetAnnualFilings = (userData: UserData): Promise<UserData> => {
```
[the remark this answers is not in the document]
aloverso: I don't think of it like that? It is accurate that the return type is Promise, and that's what indicates that it's an async function. In my opinion, we don't need the async keyword unless we're awaiting something within the function body.

api/src/client/AirtableFeedbackClient.ts
aloverso: Faraz and I talked about this when we implemented it, and discovered that we can reference fields by IDs instead of name (both here and in the table name). This seems more robust because now it will work, even if the field names get tweaked without us realizing.

We actually created a chore to switch the User Testing airtable to use the IDs instead of names also: https://github.com/newjersey/navigator.business.nj.gov/issues/2825

web/src/components/betaBar/BetaBar.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+      expect((screen.getByLabelText("Feature request") as HTMLInputElement).value).toBe("random text");
+
+      fireEvent.click(screen.getByLabelText("close"));
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalButtonText));
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText));
+      expect((screen.getByLabelText("Feature request") as HTMLInputElement).value).toBe("");
+      expect(screen.queryByText(Config.betaBar.featureRequestModalInlineErrorText)).not.toBeInTheDocument();
+    });
+
+    it("submitting feature request renders submission successful modal", async () => {
```
aloverso, line 103: test names should read like a sentence as in "it <action>"

web/src/components/betaBar/BetaBar.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+      useUndefinedUserData();
+      render(<BetaBar />);
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalButtonText));
+      expect(screen.queryByText(Config.betaBar.betaModalFeatureRequestButtonText)).not.toBeInTheDocument();
+    });
+
+    it("clears feature request input when modal is closed and is revisted", () => {
+      useMockUserData({
+        formProgress: "COMPLETED",
+        profileData: generateProfileData({ hasExistingBusiness: false }),
```
aloverso, line 85: why does it matter for this test whether they have an existing business?

api/src/client/AirtableFeedbackClient.ts
dsueltenfuss: I was thinking we need 2 tables, 1 for dev/staging and 1 for prod. That way we can test without polluting the real data
Hence why we would want to do this by name. Same for the UserTesting table.

web/src/components/betaBar/BetaBar.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText));
+      fireEvent.change(screen.getByLabelText("Feature request"), { target: { value: "random text" } });
+      expect((screen.getByLabelText("Feature request") as HTMLInputElement).value).toBe("random text");
+
+      fireEvent.click(screen.getByLabelText("close"));
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalButtonText));
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText));
+      expect((screen.getByLabelText("Feature request") as HTMLInputElement).value).toBe("");
+      expect(screen.queryByText(Config.betaBar.featureRequestModalInlineErrorText)).not.toBeInTheDocument();
+    });
```
aloverso, line 101: having expect statements scattered through this test makes me think it's actually testing several different things. I think these should be broken out into independent tests so that it is more clear and readable what logic is at play here.

- it only shows the Feature Request button when form progress is completed
- it does not show the Feature Request button when form progress is NOT completed
- it clears feature request input when modal is closed and is revisited

web/src/components/betaBar/BetaBar.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    it("submitting feature request renders submission successful modal", async () => {
+      useMockUserData({
+        formProgress: "COMPLETED",
+        profileData: generateProfileData({ hasExistingBusiness: false }),
+      });
+
+      render(<BetaBar />);
+
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalButtonText));
+      expect(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText)).toBeInTheDocument();
```
aloverso, line 112: once we've have a test for something, I think we don't need to re-test and re-expect it in every test. If we have a test for "it only shows the Feature Request button when form progress is completed" then we don't need this line in every subsequent test. This helps with readability by keeping our tests concise so that every line is necessary and nothing is extraneous.

web/src/components/betaBar/BetaBar.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+      fireEvent.click(screen.getByText(Config.betaBar.featureRequestModalSubmitButtonText));
+
+      await waitFor(() => {
+        expect(screen.getByText(Config.betaBar.successfulSubmissionModalButtonText)).toBeInTheDocument();
+      });
+      expect(screen.queryByText(Config.betaBar.featureRequestModalInlineErrorText)).not.toBeInTheDocument();
+    });
+
+    it("does not submit feature request when input is empty", () => {
```
aloverso, line 126: this name could be more specific, it's not testing JUST that it does not submit but also that it shows an error

web/src/components/betaBar/BetaBar.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+      });
+      render(<BetaBar />);
+
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalButtonText));
+      expect(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText)).toBeInTheDocument();
+
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText));
+      fireEvent.click(screen.getByText(Config.betaBar.featureRequestModalSubmitButtonText));
+      expect(screen.queryByText(Config.betaBar.successfulSubmissionModalButtonText)).not.toBeInTheDocument();
+      expect(screen.getByText(Config.betaBar.featureRequestModalInlineErrorText)).toBeInTheDocument();
```
aloverso, line 139: can we also expect that the API was not called? this seems important

web/src/components/betaBar/BetaBar.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    });
+
+    it("clicking the submit another feedback button displays the beta modal", async () => {
+      render(<BetaBar />);
+
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalButtonText));
+      expect(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText)).toBeInTheDocument();
+
+      fireEvent.click(screen.getByText(Config.betaBar.betaModalFeatureRequestButtonText));
```
aloverso, line 148: we're doing this series of 2 button clicks in nearly every one of these tests. I think it should be a helper function. Or perhaps a direct test of the Feedback component specifically so we don't have to open the BetaBar every single time?

web/src/components/betaBar/BetaBar.test.tsx
aloverso: This holds for every test

api/src/client/AirtableFeedbackClient.ts
aloverso: Hm 🤔  that's a good point

aloverso: I'm gonna switch the chore I linked to be a chore to create separate dev/prod tables and start using them. For this code, I think Dave is right, we should be using the table name and the field names so that we can switch seamlessly between tables

api/src/client/AirtableFeedbackClient.ts
FarazA22: Updating the table variable to table name and not table id. Therefore this no longer needs to be a env variable right?

FarazA22: will be address as part of the chore

Issue #3056, opened 2022-05-26
Industry Dropdown: Changes to All Other Businesses option


carye3:
We need to add a line below the "All Other Businesses" option in the industry dropdown list and make it the first option that appears. Original story: #2822 

Desktop
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=12556%3A256304

Mobile
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=12590%3A264149

carye3:
@aaronvux this was split out from #2822 to be prioritized in this sprint

FarazA22:
@carye3 @aaronvux  I spoke to the Dev team about this ticket. We are not able to introduce the horizontal line since this is an auto complete component. I'll work on updating the sort order so it displays `other businesses` first.

aaronvux:
LGTM.  @carye3 are we doing the styling in another ticket?  I only see the All Other Business Options as the first value.

PR #3067, opened 2022-05-27
fix: [#3056] update industry dropdown - all other option as first option

FarazA22:


content/src/roadmaps/industries/generic.json
```diff
@@ -35,9 +35,9 @@
       "task": "get-insurance"
     }
   ],
-  "name": "Any Other Business Type",
+  "name": "All Other Businesses",
```
dsueltenfuss, line 38: Is there a reason we are renaming this industry? I don't see that called out in the tickets.

web/src/lib/utils/helpers.tsx
```diff
@@ -310,3 +310,11 @@ export const makeButtonIcon = (svgFilename: string, size = "20px"): ReactElement
...
   />
 );
+
+export const isIndustryIdGeneric = (industry: Industry) => {
+  return industry.id === "generic";
+};
+
+export const getIndustryName = (industry: Industry) => {
+  return industry.name;
+};
```
dsueltenfuss, line 320: I don't think this helper adds any value to the code, since it is just returning something we already have access to.

web/src/lib/utils/helpers.tsx
```diff
@@ -310,3 +310,11 @@ export const makeButtonIcon = (svgFilename: string, size = "20px"): ReactElement
     alt=""
   />
 );
+
+export const isIndustryIdGeneric = (industry: Industry) => {
+  return industry.id === "generic";
+};
```
dsueltenfuss, line 316: I think this should be moved to the industry file in shared, so it's available in any project.

web/src/lib/utils/helpers.tsx
FarazA22: I removed the function and passed the `name` key as an argument to the orderBy function

web/src/lib/utils/helpers.tsx
FarazA22: I'll make this update

PR #3083, opened 2022-06-02
wip

FarazA22:


content/src/fieldConfig/config.json
```diff
@@ -26,28 +26,36 @@
...
     "featureRequestModalHeadingText": "Request a Feature",
-    "betaModalButtonText": "Share your feedback!"
+    "reportBugModalHeadingText": "Report an Issue",
+    "betaModalFeedbackButtonText": "Share Feedback",
+    "reportBugModalIssueSummaryBodyText": "**What were you trying to do?**",
+    "reportBugModalIssueSummaryAdditionalBodyText": "What is the issue you experienced? (i.e. I was not able to log in.)",
+    "reportBugModalIssueDetailBodyText": "**What was the issue you encountered?**",
+    "reportBugModalIssueDetailAdditionalBodyText": "Walk us through, step-by-step what happened. (i.e. This is my first time trying to create an account. When I was on the landing page I clicked “Get Started” but then nothing happened.)"
+  },
+  "bugBar": { "bugBarText": "Report an issue." },
```
aloverso, line 55: nitpick, but this is a button, not a Bar, right?

web/src/components/BetaBar.test.tsx
```diff
@@ -13,8 +13,8 @@ describe("<BetaBar />", () => {
 
   it("opens feedback modal on button click", () => {
     render(<BetaBar />);
-    expect(screen.queryByText(Config.betaBar.betaModalTitle)).not.toBeInTheDocument();
+    expect(screen.queryByText(Config.feedbackModal.betaModalTitle)).not.toBeInTheDocument();
     fireEvent.click(screen.getByText(Config.betaBar.betaModalButtonText));
-    expect(screen.getByText(Config.betaBar.betaModalTitle)).toBeInTheDocument();
+    expect(screen.getByText(Config.feedbackModal.betaModalTitle)).toBeInTheDocument();
```
aloverso, line 18: I like this renaming. Would it make sense to extend to the fields also (ie, `feedbackModalTitle` instead of `betaModalTitle`)?

web/src/components/FeedbackModal/FeedbackModal.test.tsx
```diff
@@ -0,0 +1,294 @@
...
+      fireEvent.click(screen.getByText(Config.feedbackModal.betaModalBugReportButtonText));
+      fireEvent.change(screen.getByLabelText("Issue details"), { target: { value: "random details text" } });
+      fireEvent.click(screen.getByText(Config.feedbackModal.feedbackModalSubmitButtonText));
+
+      expect(screen.getByText(Config.feedbackModal.feedbackModalInlineErrorText)).toBeInTheDocument();
+      expect(mockApi.postIssue).not.toHaveBeenCalled();
+    });
+
+    it("opens issues dialog when isBug prop is used and then returns to feedback selection dialog after issue submission", async () => {
+      renderFeedbackModal(true);
```
aloverso, line 227: for readability, I'd recommend using an object for the parameters here. `renderFeedbackModal({isBugBar: true})` is much easier to understand what the param is doing than just the naked boolean here

web/src/components/FeedbackModal/FeedbackModal.test.tsx
```diff
@@ -0,0 +1,294 @@
...
+      useMockUserData(userData);
+
+      const issue = generateIssueRequest({
+        context: "random summary text",
+        browser: "Firefox v.6.5",
+        device: "Mac OS 10 Google Pixel Mobile",
+        screenWidth: "500 px",
+        detail: "random details text",
+        pageOfRequest: "roadmap/test",
+      });
```
aloverso, line 262: I wouldn't recommend using a generator for an object that you're using as an expectation. This is the `expectedIssue` that you expect something to have been called with. The thing about a generator is that it will fill in random values for anything you don't provide.

If we add a new field to the issue object, this code would still compile and it would have a random value floating around in there that may or may not be correct. Whereas if you build the expected object explicitly here like:
```
const expectedIssue: IssueRequest = {
... all the stuff here
}
```
then it would fail to compile if we add a new field, which is good, because it would force us to update the test and be clear about what we expect to happen to this field in the situation.

content/src/fieldConfig/config.json
aloverso: reading more, I think I understand that sometimes the "beta bar" is actually just "report an issue" an that's why it's Bar here, right?

web/src/components/FeedbackModal/FeedbackModal.tsx
```diff
@@ -1,22 +1,28 @@
...
+import { ReportIssueDialog } from "@/components/FeedbackModal/ReportIssueDialog";
+import { RequestFeatureDialog } from "@/components/FeedbackModal/RequestFeatureDialog";
+import { SelectFeedbackDialog } from "@/components/FeedbackModal/SelectFeedbackDialog";
 import { FeedbackRequestDialogNames } from "@/lib/types/types";
 import { ReactElement, useEffect, useState } from "react";
 
 type Props = {
   handleClose: () => void;
   isOpen: boolean;
+  isBugBar?: boolean;
```
aloverso, line 11: it might be nice to have consistent domain language throughout. Everywhere else in the code & content we're using the term "issue" - is there a reason we switch to "bug" when referencing this behavior?

aloverso: we even call the component ReportAnIssueBar already

web/src/components/GenericTextField.tsx
```diff
@@ -101,9 +101,9 @@ export const GenericTextField = (props: GenericTextFieldProps): ReactElement =>
     : "";
   return (
     <div
-      className={`${props.formInputWide && "form-input-wide"} ${
-        !props.formInputFull && !props.formInputWide && "form-input"
-      }  margin-top-2`}
+      className={`${props.formInputWide ? "form-input-wide " : ""}${
+        !props.formInputFull && !props.formInputWide ? "form-input " : ""
+      }margin-top-2`}
```
aloverso, line 106: why this change? I prefer the first version more, where the class strings are just the string, instead of relying on something to have a trailing space 

To me, I prefer:
```
const x = "some-class"
className=`{$x} margin-top-2`
```

over:

```
const x = "some-class "
className=`{$x}margin-top-2`
```

web/src/components/PageSkeleton.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <PageSkeleton>
+        <></>
+      </PageSkeleton>
+    );
+
+    expect(screen.queryByTestId("bugBar")).not.toBeInTheDocument();
+  });
+
+  it("renders report an issues bar after onboarding is completed", () => {
+    useMockUserData({});
```
aloverso, line 25: I didn't realize useMockUserData wasn't correctly randomizing values, so ideally this shouldn't work. But as a rule, if we're relying on some property in a test, we should be explicit about setting that property and not relying on what happens to come from the generator.

This should explicitly set `formProgress: "COMPLETED"` if that's what the test is relying on

Issue #2818, opened 2022-05-17
Update Profile Box on Left Side of Poppy Roadmap


aaronvux:
As a Poppy, I need a more streamlined Profile section on the Roadmap so I'm not overwhelmed with content.

## Acceptance Criteria
- Display business name as Profile link if the user is registered and has a business name saved to their profile, current date (Local NJ Date) on the right with ET designation
  - If user is in Guest mode, display "Guest Account" link & date 
  - If no Business Name exists and the user is registered, display "Business Profile" link & date 
- Remove Profile block
- Remove Copy block

## Additional Acceptance Criteria
- 508 Compliant
- Follows Design
- Works in Mobile

## Design
**Poppy**
- Desktop - Registered
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=12623%3A262580

- Desktop - Unregistered
https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=13498%3A177017

## Analytics
roadmap_profile_edit_button/click/go_to_profile_screen

## Approvals
- [ ] UX Review
- [ ] PO Approval

aaronvux:
Spoke to EJ, split Oscar update to separate ticket.

carye3:
We need to have slightly different copy for Guest Mode:
<img width="967" alt="image" src="https://user-images.githubusercontent.com/62653834/170735468-6a9c53cf-269d-4264-9571-eb0a607cf190.png">

@CeciliaRam does this look good? Figma: [desktop](https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=13498%3A177017) and [mobile](https://www.figma.com/file/IKcrTIyvaS593LuJbAVFR3/Navigator-Designs-(Sprint-33%2B)?node-id=13502%3A176917)

aaronvux:
@FarazA22
1. I'll update the AC to include the unregistered user.
2. I don't think we need to include the designator as we will only know if they go through formation.  
3. Reuse the roadmap_profile_edit_button analytics on the link.  Not sure about the return_to_onboarding.  I'll check it out.

FarazA22:
@aaronvux Are we now using the 'ET' designation?

aaronvux:
@FarazA22 ET.  Updated AC.

[...]

PR #3118, opened 2022-06-07
feat: [#2818] remove profile box from roadmap

FarazA22:


web/src/components/Header.tsx
```diff
@@ -0,0 +1,66 @@
...
+    if (
+      (userData?.profileData.businessName === undefined || userData?.profileData.businessName === "") &&
+      state.isAuthenticated === "TRUE"
+    )
+      buttonText = Config.roadmapDefaults.genericToProfileButtonText;
+
+    if (userData?.profileData.businessName && state.isAuthenticated === "TRUE")
+      buttonText = userData.profileData.businessName;
+
+    return buttonText;
```
aloverso, line 45: I think this would be cleaner if we just returned the value in the if-block instead of setting a variable and returning at the end? It would also prevent the possibility of accidentally changing a value if multiple if-statements apply to a situation

web/src/components/Header.tsx
```diff
@@ -0,0 +1,66 @@
...
+      buttonText = Config.roadmapDefaults.guestModeToProfileButtonText;
+
+    if (
+      (userData?.profileData.businessName === undefined || userData?.profileData.businessName === "") &&
+      state.isAuthenticated === "TRUE"
+    )
+      buttonText = Config.roadmapDefaults.genericToProfileButtonText;
+
+    if (userData?.profileData.businessName && state.isAuthenticated === "TRUE")
+      buttonText = userData.profileData.businessName;
```
aloverso, line 43: would strongly prefer that we use curly braces for an if-statement in any case where it's not a one-line guard-clause (like the `if (x) return` lines)

web/test/pages/roadmap.test.tsx
```diff
@@ -175,247 +174,65 @@ describe("roadmap page", () => {
...
-        </ThemeProvider>,
-        IsAuthenticated.FALSE,
-        { modalIsVisible: false, setModalIsVisible }
-      )
-    );
-    fireEvent.click(screen.getByTestId("grey-callout-link"));
-    expect(mockPush).toHaveBeenCalled();
-    expect(setModalIsVisible).not.toHaveBeenCalled();
-  });
+    renderRoadmapPage({});
```
aloverso, line 180: this comment applies to most of the changes in this file - I think this would be a great opportunity to do a component test for <Header/> instead of having all these tests in the roadmap page

web/test/pages/roadmap.test.tsx
```diff
@@ -108,7 +102,7 @@ describe("roadmap page", () => {
     return { page };
   };
 
-  const renderPageWithAuth = ({
+  const renderPageWithAuthAlert = ({
```
aloverso, line 105: why this rename?

FarazA22: This function was using the `withAuthAlert` helper function and rendered the Roadmap page with SignUp Toast component. I didn't need the SignUp Toast component to render so I created a separate function called `renderPageWithAuth` and used the `withAuth` helper function. This function was renamed because it shared the same name.

web/src/components/Header.tsx
FarazA22: I've switched `buttonText =` with a `return` statement. This if statement has a one-line guard clause, but it wraps to the second line. I've added the curly braces, wondering if you think it's necessary in this scenario?

web/test/pages/roadmap.test.tsx
FarazA22: will do

PR #3235, opened 2022-06-15
chore: add cms preview for cannabis license task

FarazA22:


web/src/contexts/configContext.ts
```diff
@@ -1,3 +1,5 @@
+import * as cannabisLicenseTab1 from "@businessnjgovnavigator/content/fieldConfig/cannabis-license-tab1.json";
+import * as cannabisLicenseTab2 from "@businessnjgovnavigator/content/fieldConfig/cannabis-license-tab2.json";
```
aloverso, line 2: small nitpick but I think these should be capitalized first letter to match the others

web/public/mgmt/config.yml
```diff
@@ -135,95 +135,104 @@ collections:
...
+        file: content/src/fieldConfig/config.json
         editor:
-          preview: true
+          preview: false
         fields:
-          - { label: "Content", name: "body", widget: "markdown" }
-      - label: "Impact-Zone Requirements"
-        name: "impact-zone-requirements"
-        file: "content/src/display-content/cannabis-license/impact-zone-requirements.md"
+          - label: Cannabis Eligibility Modal
```
aloverso, line 150: we could make a preview for this, right? just another preview component that just shows the modal?

aloverso: wait no... I remember trying to do this for profile... Material really doesn't let us show the modal in-page instead of opening it across the whole page like a real modal. 

maybe we could fix that at some point but I retract my comment for the purpose of merging this branch

web/public/mgmt/config.yml
```diff
@@ -135,95 +135,104 @@ collections:
...
             name: cannabisApplyForLicense
             collapsed: false
             widget: object
             summary: "{{cannabisApplyForLicense}}"
             fields:
+              - {
+                  label: Priority Status None Selected Text,
+                  name: priorityStatusNoneSelectedText,
+                  widget: string,
+                  hint: This field supports Markdown,
```
aloverso, line 234: i think we should make this one a markdown widget, because it has a link in it, so it'll be easier for jess and ce to edit if they can use the rich text editor

aloverso: actually I think this comment holds for pretty much every field here in the first tab, because they all have contextual info or links, and it's much easier for them to edit those with the rich text markdown editor

web/src/lib/cms/previews/cannabis-license/cannabis-license.tsx
```diff
@@ -0,0 +1,81 @@
...
+        )}
+        {tab === "2" && (
+          <>
+            <h2>Conditional General Requirements</h2>
+            <Content>{config.cannabisApplyForLicense.conditionalGeneralRequirements}</Content>
+            <hr className="margin-y-5" />
+            <h2>Conditional Bottom of Task</h2>
+            <Content>{config.cannabisApplyForLicense.conditionalBottomOfTask}</Content>
+            <hr className="margin-y-5" />
+            <h2 className="margin-bottom-5">Annaul License</h2>
```
aloverso, line 68: typo

web/src/lib/cms/previews/cannabis-license/cannabis-license.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <h2 className="margin-bottom-5">Priority Status Selected</h2>
+          </>
+        )}
+        {tab === "2" && (
+          <>
+            <h2>Conditional General Requirements</h2>
+            <Content>{config.cannabisApplyForLicense.conditionalGeneralRequirements}</Content>
+            <hr className="margin-y-5" />
+            <h2>Conditional Bottom of Task</h2>
+            <Content>{config.cannabisApplyForLicense.conditionalBottomOfTask}</Content>
```
aloverso, line 66: there's nothing wrong with this preview, to be clear, but i was wondering if it might be easier to read if we did one of:
- use the <CannabisApplyForLicenseTask /> component twice, instead of only using it for annual and breaking conditional down into its sub-parts
- make Tab 2 Conditional and Tab 2 Annual as two different files in the collection, so that there's not quite as much text / content on this tab

PR #3314, opened 2022-06-28
nexus WIP

aloverso:


api/src/db/migrations/v65_add_nexus_location_to_profile.ts
```diff
@@ -0,0 +1,292 @@
+import {v65UserData} from "./v65_add_task_progess_card";
```
dsueltenfuss, line 1: This should be updated to v66, v65 is already in use

api/src/db/migrations/v65_add_nexus_location_to_profile.ts
```diff
@@ -0,0 +1,292 @@
...
+
+export const migrate_v65_to_v66 = (v65Data: v65UserData): v66UserData => {
+  return {
+    ...v65Data,
+    profileData: {
+      ...v65Data.profileData,
+      nexusLocationInNewJersey: undefined
+    },
+    version: 66,
+  };
```
dsueltenfuss, line 24: Do you need to add a field for the dba name?

content/src/roadmaps/industries/home-contractor.json
```diff
@@ -40,7 +40,7 @@
     {
       "step": 3,
       "weight": 15,
-      "task": "get-insurance-home-contractor"
+      "task": "get-insurance-  home-contractor"
```
dsueltenfuss, line 43: Is this space intentional?

web/src/components/tasks/search-business-name/AvailableProps.ts
```diff
@@ -0,0 +1,5 @@
+export interface AvailableProps {
```
dsueltenfuss, line 1: Do we want to start splitting these props out to separate files?

api/src/db/migrations/v65_add_nexus_location_to_profile.ts
dsueltenfuss: (I'm referring to the file name)

api/src/db/migrations/v65_add_nexus_location_to_profile.ts
aloverso: yep! wip :)

web/src/components/tasks/search-business-name/AvailableProps.ts
dsueltenfuss: We have been leaving the prop types in the same file as the class. 
I was asking if separating them was a pattern we want to start following

PR #3315, opened 2022-06-28
feat: [#182374966] the dakota navigator unsupported page

somabadri:


web/src/components/InformationCards.tsx
```diff
@@ -0,0 +1,81 @@
...
+        "",
+        cardConfig.card3Button,
+        "info",
+        true
+      )}
+    </div>
+  );
+};
+
+export default InformationCards;
```
aloverso, line 81: As a pattern, we don't do `export default` on any components except for Next page components (and there, we only do it because Next requires it)

For everything else, we do `export const ComponentName ...` inline when we declare it

web/test/pages/onboarding/onboarding-foreign.test.tsx
```diff
@@ -2,8 +2,13 @@ import { getMergedConfig } from "@/contexts/configContext";
 import { templateEval } from "@/lib/utils/helpers";
 import { generateProfileData, generateUserData } from "@/test/factories";
 import { expectContent } from "@/test/helpers";
+import * as mockRouter from "@/test/mock/mockRouter";
 import { useMockRouter } from "@/test/mock/mockRouter";
```
aloverso, line 6: you have two styles of imports for the same thing being imported here

web/test/pages/onboarding/onboarding-foreign.test.tsx
```diff
@@ -177,6 +182,92 @@ describe("onboarding - foreign business", () => {
         screen.queryByText(Config.profileDefaults.FOREIGN.foreignBusinessType.errorTextRequired)
       ).not.toBeInTheDocument();
     });
+
+    it("deselects every other option if none is selected", async () => {
+      // GIVEN the none of the above option exists in the config file
```
aloverso, line 187: remove comments before merging

web/test/pages/unsupported.test.tsx
```diff
@@ -0,0 +1,11 @@
+import UnsupportedPage from "@/pages/unsupported";
+import { render, screen } from "@testing-library/react";
+import Config from "../../../content/src/fieldConfig/config.json";
```
aloverso, line 3: for a test file, we never want to import Config directly anymore, with our new config pattern. we should do `const Config = getMergedConfig()` at the top of the test file.

web/test/pages/onboarding/onboarding-foreign.test.tsx
```diff
@@ -177,6 +182,92 @@ describe("onboarding - foreign business", () => {
...
+          screen.getByLabelText(
+            Config.profileDefaults.FOREIGN.foreignBusinessType.optionContent.revenueInNJ
+          ) as HTMLInputElement
+        ).checked
+      ).toEqual(false);
+
+      act(() => page.clickNext());
+      await waitFor(() => {
+        expect(userDataWasNotUpdated).toBeTruthy();
+      });
```
aloverso, line 228: what are these last two lines testing? It doesn't seem to relate to the stated title of the test

web/test/pages/onboarding/onboarding-foreign.test.tsx
```diff
@@ -177,6 +182,92 @@ describe("onboarding - foreign business", () => {
...
+          ) as HTMLInputElement
+        ).checked
+      ).toEqual(true);
+      expect(
+        (
+          screen.getByLabelText(
+            Config.profileDefaults.FOREIGN.foreignBusinessType.optionContent.employeesInNJ
+          ) as HTMLInputElement
+        ).checked
+      ).toEqual(false);
```
aloverso, line 253: Having an assertion about `employeesInNJ` feels like a non-sequitur in this test. In your "when" action section, you check the `none` checkbox and the `transactionsInNJ` checkbox, so it makes sense to see assertions regarding their state. Is there a reason for asserting on a seemingly unrelated checkbox state?

web/test/pages/onboarding/onboarding-foreign.test.tsx
aloverso: we typically avoid `import * as X` imports (there is a single exception to this, which is for when we do module mocking, like of the `api` - happy to talk further on that if need be)

It looks like this was added because you want to use `mockPush` but I think you could just import that directly by adding it to the existing destructuring import line, instead of this new line being necessary

web/src/components/onboarding/OnboardingForeignBusinessType.tsx
```diff
@@ -74,13 +85,14 @@ export const OnboardingForeignBusinessType = ({ headerAriaLevel = 2 }: Props): R
...
 
-      {state.profileData.foreignBusinessType !== undefined && (
-        <Alert variant="info">
-          <Content>
-            {Config.profileDefaults.FOREIGN.foreignBusinessType[state.profileData.foreignBusinessType]}
-          </Content>
-        </Alert>
-      )}
+      {state.profileData.foreignBusinessType !== undefined &&
+        state.profileData.foreignBusinessType !== "NONE" && (
```
aloverso, line 89: this is a stylistic preference, but personally, when I start to have more than a single item in an inline JSX boolean check (like this), I like to extract it out into a function to make it more readable. Something like `shouldShowForeignBusinessTypeAlert` or something. Prevents the JSX from getting too wordy with inline javascript logic checks

web/public/mgmt/config.yml
```diff
@@ -2021,6 +2021,14 @@ collections:
               - { label: Optional Tasks Remaining (plural), name: optionalTasksPlural, widget: string }
               - { label: Required Tasks Remaining (singular), name: requiredTasksSingular, widget: string }
               - { label: Required Tasks Remaining (plural), name: requiredTasksPlural, widget: string }
+          - label: Unsupported Page
```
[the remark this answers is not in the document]
somabadri: Done, changed it to "Navigator Unsupported User Page"

web/test/pages/onboarding/onboarding-foreign.test.tsx
AnJuHyppolite: Done

web/test/pages/onboarding/onboarding-foreign.test.tsx
AnJuHyppolite: updated

PR #3357, opened 2022-06-30
feat: [#182476305] navbar displays register and log in for guest

somabadri:


web/src/components/navbar/NavBarDesktop.tsx
```diff
@@ -147,7 +126,34 @@ export const NavBarDesktop = (): ReactElement => {
               <img className="height-4" src="/img/Navigator-logo.svg" alt="Business.NJ.Gov Navigator" />
             </a>
           </Link>
-          <div className="z-100">
+          <div className="flex z-100">
+            {!isAuthenticated && (
```
dsueltenfuss, line 130: This is showing up on onboarding when it shouldn't be.

PR #3363, opened 2022-07-01
feat: #182476312 updated navbar for onboarding

somabadri:


web/src/components/navbar/NavBar.test.tsx
```diff
@@ -69,6 +69,68 @@ describe("<NavBar />", () => {
...
+        renderDesktopNav();
+        expect(screen.getByText(Config.navigationDefaults.logInButton)).toBeInTheDocument();
+      });
+
+      it("doesn't render register button when in onboarding", () => {
+        renderDesktopNav();
+        expect(screen.queryByTestId("registration-button")).not.toBeInTheDocument();
+      });
+
+      it("doesn't render a menu to be opened when in onboarding", async () => {
```
aloverso, line 99: in this test name and the one before it, the "when in onboarding" piece is redundant because it already says that in the `describe`

web/src/components/navbar/NavBarDesktop.tsx
```diff
@@ -26,12 +26,19 @@ export const NavBarDesktop = (): ReactElement => {
...
+      return;
+    }
     if (!open) {
       analytics.event.account_name.click.expand_account_menu();
     }
     setOpen((prevOpen) => !prevOpen);
   };
 
+  const currentlyOnboarding = (): boolean => {
+    return router.pathname === "/onboarding" ? true : false;
```
aloverso, line 39: this ternary can be simplified, the true/false is redundant, you can just say `return router.pathname === "/onboarding"`

web/src/components/navbar/NavBarDesktop.tsx
```diff
@@ -26,12 +26,19 @@ export const NavBarDesktop = (): ReactElement => {
   const anchorRef = useRef<HTMLButtonElement | null>(null);
 
   const toggleDropdown = (): void => {
+    if (currentlyOnboarding()) {
+      return;
+    }
```
aloverso, line 31: this is unnecessary, right? Because the dropdown will never even show up when in onboarding?

PR #3428, opened 2022-07-07
feat: [#182374939] add priority status feedback selection

AnJuHyppolite:


web/src/components/tasks/cannabis/CannabisPriorityTypes.tsx
```diff
@@ -53,6 +55,71 @@ export const CannabisPriorityTypes = (props: Props): ReactElement => {
     }
   }, [userData, update]);
 
+  useEffect(() => {
+    if (!userData) return;
+    const anyMinorityOrWomenCheckboxesSelected: boolean = priorityTypesObj.minorityOrWomen.some(
```
dsueltenfuss, line 60: This logic all works great, but it could be consolidated. 
Instead of doing the check, storing it as a variable and then using that variable in the if block, you could have the if block be the check. 

This would make it more clear and concise. 

That said, I don't think that needs to be changed before this ticket is merged to main. It could be done as a refactoring later.

web/src/components/tasks/cannabis/CannabisPriorityStatusTask.test.tsx
```diff
@@ -362,4 +363,167 @@ describe("<CannabisPriorityStatusTask />", () => {
...
+    fireEvent.click(screen.getByTestId(randomMinorityOrWomenPriorityType));
+    fireEvent.click(screen.getByTestId(randomSocialEquityPriorityType));
+    expect(screen.getByText(eligibilityPhrase)).toBeInTheDocument();
+
+    fireEvent.click(screen.getByTestId(randomMinorityOrWomenPriorityType));
+    fireEvent.click(screen.getByTestId(randomSocialEquityPriorityType));
+    expect(screen.queryByText(eligibilityPhrase)).not.toBeInTheDocument();
+  });
+
+  it("displays minority/women, impact zone, and social equity eligibilities when minority/women, impact zone, and social equity priority types are selected", () => {
```
aloverso, line 498: this is a nitpick, (I'm sorry) but one thing to think about is that tests are documentation for future developers. The names of tests should succinctly communicate the behavior, so that if someone is looking to understand or make changes, they should be able to look at test names and understand and make changes quickly.

These test names are descriptive, but they're fairly wordy and repetitive - different tests here have very similar names. It might be worth considering ways to simplify the name of the test to be clearer about what they're doing. When we were developing together, we were referring to them as "the test for the two-case" and the "test for the three-case". Because that was the Thing they were testing for. A future developer would have to work hard reading the test names & body to see that some are testing the two-case vs three-case.

A few examples that might be more readable. I'm giving more than one to show that this is totally a stylistic nitpick! Your code is just fine, and you're welcome to take any of these suggestions, or none of them. Just communicating that thinking about that "future developer" trying to read & understand the tests is an audience I think about when writing my test names.
- it "displays phrase with 2 eligibilities when minority/women AND impact zone checked"
- it "inserts minority/women and impact zone into 2-part eligibility phrase"
- it "displays 2-part eligibility phrase for minority/women and impact zone"

web/src/components/tasks/cannabis/CannabisPriorityTypes.tsx
aloverso: Same as my comment :) your idea for making it more concise is actually different from mine. it might be fun to do a refactoring party of sorts. to play with different possible refactorings and teach/show that there's never exactly one "correct" answer or "best" solution

PR #3493, opened 2022-07-11
wip

FarazA22:


web/src/components/FilingsCalendar.test.tsx
```diff
@@ -111,4 +149,38 @@ describe("<FilingsCalendar />", () => {
       )
     ).toBeInTheDocument();
   });
+
+  it("does not displays filings calendar as list when isRoadmap prop is true and formation date is undefined", () => {
```
aloverso, line 153: two thoughts:
- if we're introducing new tests for what happens when `isRoadmap` prop is true, then we should have a matching test for when it's false that describes that behavior
- I think it might be better if we renamed this something other than `isRoadmap`. It doesn't communicate to me what will change in the component when I switch the prop between true and false.

web/src/components/FilingsCalendar.tsx
```diff
@@ -138,6 +140,23 @@ export const FilingsCalendar = (props: Props): ReactElement => {
     </div>
   );
 
+  if (props.isRoadmap) {
+    return taxFilings.length > 0 && userData?.profileData.dateOfFormation ? (
```
aloverso, line 144: why do we need the new `isRoadmap` prop?

Looking further, it seems the difference is the styling surrounding the list view? It makes me wonder if this `renderList` function should be its own component. It feels weird to have a `FilingsCalendar` component that renders two completely different things based on the roadmap prop. It's like they don't share anything except a concept of being "related to the calendar" and calling the renderList function

web/src/components/TaskHeader.test.tsx
```diff
@@ -249,7 +249,10 @@ describe("<TaskHeader />", () => {
         expect(currentUserData().profileData.dateOfFormation).toEqual(date.format("YYYY-MM-DD"))
       );
       expect(currentUserData().taskProgress[id]).toEqual("COMPLETED");
-      expect(mockPush).toHaveBeenCalledWith("/roadmap");
+      expect(mockPush).toHaveBeenCalledWith({
+        pathname: userData.profileData.businessPersona === "OWNING" ? "/dashboard" : "/roadmap",
```
aloverso, line 253: if you're specifying "STARTING" up above then this switch should not necessary right?

aloverso: why are we testing the query param logic in 2 places?

web/src/styles/base/background-colors.scss
```diff
@@ -98,6 +98,10 @@
   background-color: rgba(255, 255, 255, 0) !important;
 }
 
+.bg-base-lighter {
+  background-color: $base-lighter !important;
+}
```
aloverso, line 103: doesn't this exist in NJWDS already? Or am I confused at the purpose here?

web/src/styles/base/border-colors.scss
```diff
@@ -30,3 +30,7 @@
 .border-right-base {
   border-right-color: $base !important;
 }
+
+.border-base-lightest {
+  border-right-color: $base-lightest !important;
```
aloverso, line 35: `border-base-lightest` is already a class that exists from NJWDS, right? Should we be overriding it to just be border-right? Why not make a new class name specific to setting the right-border?

web/src/pages/roadmap.tsx
```diff
@@ -100,20 +103,33 @@ const RoadmapPage = (props: Props): ReactElement => {
...
+              setProfileUpdatedAlert(false);
               router.replace({ pathname: "/roadmap" }, undefined, { shallow: true });
             }}
             heading={Config.profileDefaults.successTextHeader}
-            dataTestid="toast-alert-SUCCESS"
           >
             {Config.profileDefaults.successTextBody}
           </ToastAlert>
         )}
+        {calendarAlert && (
```
mikewolfd, line 119: These alerts look pretty duplicative, you could probably refactor them, but not necessary to complete this PR

web/src/styles/base/background-colors.scss
FarazA22: Deleting the css.

web/src/styles/base/border-colors.scss
FarazA22: We don't need this styling, deleting.

web/src/pages/roadmap.tsx
FarazA22: Added a story in the refactoring section of the backlog. We can look at it during refactor fridays.

PR #3510, opened 2022-07-13
refactor: roadmap data structure

somabadri:


web/src/components/roadmap/MiniRoadmapStep.tsx
```diff
@@ -16,14 +17,14 @@ interface Props {
...
   const [isOpen, setIsOpen] = useState<boolean>(props.isOpen ?? false);
   const stepNumber = props.step.step_number;
   const { isOpen: sectionIsOpen } = useContext(SectionAccordionContext);
   const isActive = useMemo(() => {
-    if (!props.activeTaskId) return undefined;
-    return !!props.step.tasks.some((task) => task.id === props.activeTaskId);
-  }, [props.activeTaskId, props.step]);
-
+    if (!props.activeTaskId || !roadmap?.tasks) return undefined;
+    return !!roadmap?.tasks.some((task) => task.id === props.activeTaskId && task.stepNumber == stepNumber);
```
aloverso, line 26: This `!!` is unnecessary, `.some` already returns a boolean

web/src/lib/roadmap/fixtures/tasks/blocked-task.md
```diff
@@ -2,6 +2,7 @@
 id: "blocked-id"
 urlSlug: "blocked-url-slug"
 name: "Blocked Task"
+stepNumber: 1
```
aloverso, line 5: this comment applies to all the task files changed in `/fixtures/tasks` - I thought that the step number was coming from the add-ons, right? Not from the task file itself. I think these are probably just leftover unused lines of code that can be removed

web/src/lib/roadmap/roadmapBuilder.test.ts
```diff
@@ -84,9 +82,9 @@ describe("roadmapBuilder", () => {
       ])
     );
 
-    expect(roadmap.steps[1].tasks.map((it) => it.id)).toEqual(expect.arrayContaining(["coffee-task-3-id"]));
+    expect(roadmap.tasks.map((it) => it.id)).toEqual(expect.arrayContaining(["coffee-task-3-id"]));
 
-    expect(roadmap.steps[2].tasks.map((it) => it.id)).toEqual(expect.arrayContaining(["coffee-task-4-id"]));
+    expect(roadmap.tasks.map((it) => it.id)).toEqual(expect.arrayContaining(["coffee-task-4-id"]));
```
aloverso, line 87: so with these changes here, we're actually losing a piece of confidence that we used to have while testing.

In the previous version, the tests were asserting _not just_ that the tasks were showing up in the roadmap but _also_ that the step numbers were being interpreted correctly. 

We were previously testing that step 1 contained certain tasks, and step 2 contained certain tasks, and so on. I think in order to keep the same degree of confidence now, we need to test that the step numbers are being used correctly in the new structure.

I have a suggestion on how to do this, let me know if you want me to share, but otherwise wanted to give you a chance to think through it on your own

aloverso: this is true for pretty much all the test changes in this file

web/src/lib/roadmap/roadmapBuilder.ts
```diff
@@ -128,37 +131,32 @@ const findTaskInRoadmapByFilename = (
   roadmapBuilder: RoadmapBuilder,
   taskFilename: string
 ): TaskBuilder | undefined => {
-  for (const step of roadmapBuilder.steps) {
-    const found = step.tasks.find((task) => task.filename === taskFilename);
-    if (found) return found;
-  }
+  const found = roadmapBuilder.tasks.find((task) => task.filename === taskFilename);
+  if (found) return found;
```
aloverso, line 135: this could be combined into a single line now that it's not a for loop

web/src/lib/roadmap/roadmapBuilder.ts
```diff
@@ -128,37 +131,32 @@ const findTaskInRoadmapByFilename = (
...
 const convertToRoadmap = async (roadmapBuilder: RoadmapBuilder): Promise<Roadmap> => {
   const roadmap = {
-    steps: await Promise.all(
-      roadmapBuilder.steps.map(async (step) => ({
-        ...step,
-        tasks: await Promise.all(
-          step.tasks.sort(orderByWeight).map((task) => fetchTaskByFilename(task.filename))
-        ),
-      }))
+    steps: await Promise.all(roadmapBuilder.steps),
```
aloverso, line 140: this doesn't need to be an `await Promise.all` anymore, for steps here, because it's no long an asynchronous function

web/src/lib/roadmap/roadmapBuilder.ts
```diff
@@ -128,37 +131,32 @@ const findTaskInRoadmapByFilename = (
...
   };
 
-  const allFilenames = roadmap.steps.reduce(
-    (acc: string[], currStep: Step) => [...acc, ...currStep.tasks.map((task) => task.filename)],
-    []
+  roadmap.tasks.map(
+    (task: Task) =>
+      (task.stepNumber = (
+        findTaskInRoadmapByFilename(roadmapBuilder, task.filename) as TaskBuilder
+      ).step_number)
```
aloverso, line 150: what is this line doing? we're not saving the return value of this map...

somabadri: yeah, I'm not sure. It's not doing anything, I don't where it even came from, I will get rid of it.

web/src/lib/roadmap/roadmapBuilder.test.ts
somabadri: I will try to take a stab at it right now.

web/src/lib/roadmap/roadmapBuilder.ts
somabadri: so after some digging, this line is updating the step numbers within in roadmap but I'm not sure why it is because as you said we are not saving the values?

web/src/lib/roadmap/roadmapBuilder.ts
```diff
@@ -128,37 +131,32 @@ const findTaskInRoadmapByFilename = (
...
   );
 
+  const allFilenames = new Set(roadmap.tasks.map((task) => task.filename));
```
aloverso, line 153: why a set? they should already be all unique right?

web/src/lib/roadmap/roadmapBuilder.ts
```diff
@@ -128,37 +131,32 @@ const findTaskInRoadmapByFilename = (
...
     ...roadmap,
-    steps: roadmap.steps.map((step) => ({
-      ...step,
-      tasks: step.tasks.map((task) => ({
-        ...task,
-        unlockedBy: task.unlockedBy.filter((it) => allFilenames.includes(it.filename)),
-      })),
+    tasks: roadmap.tasks.map((task) => ({
+      ...task,
+      unlockedBy: task.unlockedBy.filter((it) => allFilenames.has(it.filename)),
```
aloverso, line 159: got super confused by this for a second, I'm all "js doesn't have an array.has method" but realized it's because of the set. is there a reason we're using a set?

web/src/lib/roadmap/roadmapBuilder.ts
```diff
@@ -188,12 +187,12 @@ interface StepBuilder {
   timeEstimate: string;
   section: SectionType;
   description: string;
-  tasks: TaskBuilder[];
 }
 
 interface TaskBuilder {
   filename: string;
   weight: number;
+  step_number: number;
```
aloverso, line 195: i know you're just following what we did for Step, but I have no idea why we did snake case instead of camel case for this property. Ideally we'd refactor both to be camel case. Maybe that could be a separate commit though, to separate out the complexity

web/src/lib/utils/helpers.tsx
```diff
@@ -239,13 +247,16 @@ export const camelCaseToSentence = (text: string): string => {
...
+  roadmap?.steps.reduce((accumulator, currentStep: Step) => {
+    const currentStepTasks = roadmap.tasks.filter((task) => task.stepNumber === currentStep.step_number);
+    accumulator[currentStep.section] = [...(accumulator[currentStep.section] || []), ...currentStepTasks];
     return accumulator;
   }, {} as Record<SectionType, Task[]>);
 
 const stepInRoadmap = (roadmap: Roadmap | undefined, taskId: string): Step | undefined =>
-  roadmap?.steps.find((step) => step.tasks.find((task) => task.id === taskId));
+  roadmap?.steps.find(
+    (step) => step.step_number === roadmap?.tasks.find((task) => task.id === taskId)?.stepNumber
```
aloverso, line 258: I think it might be nice (more readable, for one) to break out the found task `roadmap?.tasks.find((task) => task.id === taskId)` as its own variable and then do a guard clause in case we don't find it.

In addition to the readability of not having nested finds, it also protects us from a weird situation. I know that `step.step_number` should never be undefined. But imagine if it was somehow. The right side of the `===` has a bunch of `?` in it, which means it could evaluate to undefined. Which means that if `step.step_number` ever was undefined somehow, then undefined === undefined, and it would evaluate as true, and we would return that step despite it not actually being a match. 

Imo, better to guard clause if `roadmap?.tasks.find((task) => task.id === taskId)` is undefined before we do this find.

web/src/pages/tasks/[taskUrlSlug].tsx
```diff
@@ -50,7 +50,6 @@ const TaskPage = (props: Props): ReactElement => {
     };
   }, [props.task.urlSlug, roadmap]);
   const isTabletAndUp = useMediaQuery(MediaQueries.tabletAndUp);
-
```
Comment (aloverso) [inline web/src/pages/tasks/[taskUrlSlug].tsx:53]: but I like newlines :/ 

:P

web/src/lib/roadmap/roadmapBuilder.ts
somabadri: I figured it out, I was using the map as for loop to traverse and assign step numbers so I just made it a for of loop so we don't waste memory.

web/src/lib/roadmap/roadmapBuilder.ts
somabadri: es-lint kept complaining and suggested the refactor to a `Set` and using `has`.

PR #3527, opened 2022-07-14
fix: [#182737527] nexus business profile

somabadri:


content/src/fieldConfig/profile-info-oscar.json
```diff
@@ -50,7 +50,7 @@
       },
       "nexusDbaName": {
         "placeholder": "",
-        "header": "",
+        "header": "### Doing Business As (DBA) New Jersey Business Name",
```
aloverso, line 53: why are these not empty string? Shouldn't Poppy & Oscar never have the DBA profile field? They should never be used, right?

content/src/fieldConfig/profile-info-poppy.json
```diff
@@ -45,7 +45,7 @@
       },
       "nexusDbaName": {
         "placeholder": "",
-        "header": "",
+        "header": "### Doing Business As (DBA) New Jersey Business Name",
```
aloverso, line 48: Same comment as above, why not empty?

web/src/components/onboarding/ProfileNexusBusinessNameField.tsx
```diff
@@ -5,36 +5,44 @@ export const ProfileNexusBusinessNameField = (): ReactElement => {
...
-              {Config.profileDefaults.nexusBusinessName.outOfStateNameHeader}
-            </h3>
-            <a href="/tasks/search-business-name-nexus">
-              {Config.profileDefaults.nexusBusinessName.editButton}
-            </a>
-          </div>
-          <div>{userData.userData?.profileData.businessName}</div>
-        </div>
-      )}{" "}
+      userData.userData?.profileData.businessName === undefined
```
aloverso, line 43: This could be shortened to `!userData.userData?.profileData.businessName`

content/src/fieldConfig/profile-info-oscar.json
somabadri: I thought they had to look the same in all three profile info files, i will update that to be empty.

web/src/lib/cms/previews/profile.tsx
```diff
@@ -54,6 +55,7 @@ const ProfilePreview = (props: Props) => {
           municipalities={[]}
           CMS_ONLY_tab={tab as ProfileTabs}
           CMS_ONLY_businessPersona={businessPersona}
+          CMS_ONLY_foreignBusinessType={foreignBusinessType}
```
aloverso, line 58: See comment above. Instead of returning `foreignBusinessType` from getMetadataFromSlug, why not just set this line as `persona == "dakota" ? "NEXUS" : "NONE"`?

This avoids changing the function signature on a widely-used function in support of solving a narrow one-time use case

PR #3572, opened 2022-07-19
feat: [#182641410] add new search func industry dropdown

AnJuHyppolite:


web/public/mgmt/config.yml
```diff
@@ -1732,6 +1732,13 @@ collections:
             summary: "{{fundingDefaults}}"
             fields:
               - { label: Issuing Agency Label, name: issuingAgencyText, widget: string }
+          - label: Industry Dropdown
+            name: industryDropdown
+            collapsed: true
+            widget: object
+            summary: "{{industryDropdown}}"
+            fields:
+              - { label: Mapped Industry Copy, name: mappedIndustryCopy, widget: string }
```
aloverso, line 1741: see my above comment that this should be moved

aloverso: also, I think we could revisit the naming of this field. Remember, the intention of the name and label is to communicate clearly to Jess and Ce what the purpose of the field is and where it gets used. I think we could name this field something that makes it more clear what it means

web/src/components/onboarding/IndustryDropdown.test.tsx
```diff
@@ -10,4 +11,40 @@ describe("Industry Dropdown", () => {
...
+  it("updates the value of the industry dropdown input when the no-matching industries button is clicked", () => {
+    render(<IndustryDropdown />);
+
+    const inputElement = screen.getByLabelText("Industry");
+    fireEvent.click(inputElement);
+    fireEvent.change(inputElement, { target: { value: "ac5" } });
+    fireEvent.click(screen.getByTestId("generic"));
+
+    expect((screen.getByLabelText("Industry") as HTMLInputElement).value).toBe("All Other Businesses");
+  });
```
aloverso, line 39: I'm not sure this test is adding value, because it's testing the intrinisic behavior of the dropdown. Part of our testing philosophy is avoiding testing the "internals" of a component that we use from a third-party library like Material UI. We can test their interactions with our code, and we test any configuration we do to them. 

But (if I'm reading this test correctly), this will always pass because it's expecting the value of the input field to change to the option we click, when we click an option. I'm pretty sure that's the default behavior for this MUI component, and there's no world in which this test would fail?

Again, I could be missing something, let me know if there's something else you were intending here. But if it's just testing the MUI component behavior, I think we can delete this test altogether.

web/src/components/onboarding/IndustryDropdown.test.tsx
```diff
@@ -10,4 +11,40 @@ describe("Industry Dropdown", () => {
...
+
+  it("displays the user input and a message that the input falls under the industries below", () => {
```
aloverso, line 41: two small notes here about naming tests:
- You say "displays the user input" but I don't see anything in the test that appears to be asserting about the user input.
- You describe the message it displays as "a message that the input falls under the industries below". This is fine for now, because it is what that string says. But that is configurable content, which means it could change at any time. 

We want to write robust tests that will accurately describe behavior so that future developers can read the tests to learn about the intended behavior of a component. So good naming is important. 

Because of that, I think we should remove clauses from the test name that are not actually being tested.

Similarly, we should avoid referencing the specific value of the message (like how it says "falls under industries below") because that's liable to change. In fact, for this case it's very likely to change almost immediately because I know Jess & Ce wanted to make it more generic, because sometimes the options are displayed above, not below. Choosing a good, descriptive name for what _we_ as developers call this message, and then being consistent about it, is important to keeping our code cohesive and understandable. 

Right now we have three disparate ways of referring to this message:
- In config, it is `mappedIndustryCopy`
- In the data-testid, it is `styled-search-term`
- In the test name, it is "a message that the input falls under the industries below"

Try to think of a descriptive, concise name that communicates what this message represents. And then we should use that same name to refer to it everywhere in the code.

web/public/mgmt/config.yml
aloverso: I have a comment below that addresses the naming more broadly btw

AnJuHyppolite:
- [ ] fix & clean up PR comments
- [ ] commit all the code into 1 nice commit
- [ ] checkout main branch, pull, checkout our branch, `git rebase main`
- [ ] force-push our code to this branch

PR #3623, opened 2022-07-22
feat: [#182682790] certifications-nudge-tax-registration

AnJuHyppolite:


api/src/domain/updateRoadmapSidebarCards.test.ts
```diff
@@ -1,4 +1,5 @@
 import { formationTaskId } from "@shared/gradualGraduationStages";
+
```
dsueltenfuss, line 2: This extra line should be removed

api/src/domain/updateRoadmapSidebarCards.ts
```diff
@@ -32,6 +32,19 @@ export const updateRoadmapSidebarCards: UpdateRoadmapSidebarCards = (userData: U
...
+    hideCard("access-certifications-nudge");
+  }
+
+  const hasCompletedTaxes = userData.taskProgress["register-for-taxes"] === "COMPLETED";
+  if (hasCompletedTaxes) {
+    hideCard("access-certifications-nudge");
+  } else {
+    showCard("access-certifications-nudge");
+  }
+
```
dsueltenfuss, line 47: I think we should turn this into 1 if block, instead of two
Having 2 separate sets of logic for displaying the card could lead to confusion

The first if block could be  (pseudo code)
`if (hasCompletedFormation and !hasCompletedTaxes) 
  showCard(-cardname-) 
else
  hideCard(-cardname-)`

dsueltenfuss (approved):

PR #3650, opened 2022-07-27
chore: add operatingPhase field to profileData

dsueltenfuss:


api/src/domain/updateRoadmapSidebarCards.test.ts
```diff
@@ -98,6 +108,9 @@ describe("updateRoadmapSidebarCards", () => {
     it("adds tax-registration-nudge if formation task is complete", () => {
       const taskId = formationTaskId;
       const userData = generateUserData({
+        profileData: generateProfileData({
+          operatingPhase: "NEEDS_TO_REGISTER_FOR_TAXES",
+        }),
         taskProgress: { [taskId]: "COMPLETED" },
```
aloverso, line 114: i think we should delete lines like this and change the test name because now the taskProgress isn't the field that's determining what happens anymore, right?

api/src/domain/user/updateOperatingPhase.ts
```diff
@@ -0,0 +1,38 @@
...
+    updatedPhase = "NEEDS_TO_REGISTER_FOR_TAXES";
+  } else if (currentPhase === "NEEDS_TO_REGISTER_FOR_TAXES" && isPublicFiling && !hasCompletedFormation) {
+    updatedPhase = "NEEDS_TO_FORM";
+  }
+
+  if (currentPhase === "NEEDS_TO_REGISTER_FOR_TAXES" && hasCompletedTaxes) {
+    updatedPhase = "FORMED_AND_REGISTERED";
+  } else if (currentPhase === "FORMED_AND_REGISTERED" && !hasCompletedTaxes) {
+    updatedPhase = "NEEDS_TO_REGISTER_FOR_TAXES";
+  }
```
aloverso, line 29: I understand that the else-ifs here are in case the user un-completes a task, right? Does it cover the case where what if they un-complete them in an unexpected order?

api/src/domain/user/updateOperatingPhase.test.ts
```diff
@@ -0,0 +1,93 @@
...
+          legalStructureId: "general-partnership",
+        }),
+        taskProgress: { [taskId]: "IN_PROGRESS" },
+      });
+      expect(updateOperatingPhase(userData).profileData.operatingPhase).toBe("NEEDS_TO_REGISTER_FOR_TAXES");
+    });
+  });
+
+  describe("FORMED_AND_REGISTERED", () => {
+    it("removes tax-registration-nudge when register for taxes is complete", () => {
```
aloverso, line 80: copy-paste error?

api/src/db/migrations/v73_add_operating_status_field.ts
```diff
@@ -0,0 +1,324 @@
...
+  let newOperatingPhase: v73OperatingPhase;
+  if (
+    v72Data.profileData.initialOnboardingFlow === "OWNING" ||
+    v72Data.profileData.businessPersona === "OWNING"
+  ) {
+    newOperatingPhase = "UP_AND_RUNNING";
+  } else if (v72Data.taskProgress["register-for-taxes"] === "COMPLETED") {
+    newOperatingPhase = "FORMED_AND_REGISTERED";
+  } else if (v72Data.taskProgress[formationTaskId] === "COMPLETED") {
+    newOperatingPhase = "NEEDS_TO_REGISTER_FOR_TAXES";
```
aloverso, line 29: is this correct for the distinction based on legal structures? An SP/GP who has not completed taxes should be NEEDS_TO_REGISTER right?

PR #3693, opened 2022-08-02
feat: [#182470455] tax registration modal, certifications snackbar, certifications

somabadri:


shared/src/gradualGraduationStages.ts
```diff
@@ -0,0 +1,5 @@
+export const formationTaskId = "form-business-entity";
+
+export const isFormationTask = (id: string): boolean => {
+  return id === formationTaskId;
+};
```
aloverso, line 5: duplicated code? I dont think this file should exist

shared/src/misc.test.ts
```diff
@@ -0,0 +1,12 @@
...
+describe("createEmptyNameAndAddress", () => {
+  it("creates an empty address object", () => {
+    expect(createEmptyNameAndAddress()).toEqual({
+      name: "",
+      addressLine1: "",
+      addressLine2: "",
+      zipCode: "",
+    });
+  });
+});
```
aloverso, line 12: I think this happened from a rebase issue, this file should no longer exist

shared/src/misc.ts
```diff
@@ -0,0 +1,13 @@
...
+  addressLine2: string;
+  zipCode: string;
+};
+
+export const createEmptyNameAndAddress = (): NameAndAddress => ({
+  name: "",
+  addressLine1: "",
+  addressLine2: "",
+  zipCode: "",
+});
```
aloverso, line 13: I think this happened from a rebase issue, this file should no longer exist

shared/src/operatingPhase.ts
```diff
@@ -24,21 +26,26 @@ export const OperatingPhases: OperatingPhase[] = [
...
   },
   {
     id: "NEEDS_TO_REGISTER_FOR_TAXES",
     hasCompletedTaxRegistration: false,
+    displayCertifications: false,
   },
   {
     id: "UP_AND_RUNNING",
     hasCompletedTaxRegistration: true,
+    displayCertifications: true,
```
aloverso, line 49: why should we track this separately from `hasCompletedTaxRegistration` if they are always the same?

web/public/mgmt/config.yml
```diff
@@ -1848,6 +1848,20 @@ collections:
...
+            collapsed: true
+            widget: object
+            summary: "{{taxRegistrationModal}}"
+            fields:
+              - { label: Title, name: title, widget: string }
+              - { label: Subtitle, name: subtitle, widget: markdown }
+              - { label: Save button text, name: saveButtonText, widget: string }
+              - { label: Cancel button text, name: cancelButtonText, widget: string }
+              - { label: Are-You-Sure Tax Title, name: areYouSureTaxTitle, widget: string }
+              - { label: Are-You-Sure Tax Body, name: areYouSureTaxBody, widget: string }
```
aloverso, line 1862: typically the body should be a markdown widget

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+        taskProgress: {
+          [taxTaskId]: "NOT_STARTED",
+        },
+        preferences: generatePreferences({ roadmapOpenSections: ["START"] }),
+      });
+
+      useMockRoadmap({
+        steps: [generateStep({ stepNumber: 1, section: "START" })],
+        tasks: [taxTask],
+      });
```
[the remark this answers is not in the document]
aloverso: All of these new tests should be edited to remove all the unnecessary setup. Having extra setup that's not relevant to the outcome makes the tests hard to read and interpret for future users, and can be actively confusing because people will be trying to figure out the answers to those questions, like "wait why do we need a particular section here???" when the answer is, we don't.

shared/src/operatingPhase.ts
dsueltenfuss: If we want to keep them separate, we should rename 'hasCompletedTaxRegistration' to 'dispalyTaxProfileFields'
My thinking was to have flags for specific fields/control to make it easier to change/separate if we need to

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+
+      useMockRoadmap({
+        steps: [generateStep({ stepNumber: 1, section: "PLAN" })],
+        tasks: [taxTask],
+      });
+
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      fireEvent.click(screen.getByTestId("modal-button-secondary"));
```
aloverso, line 236: it's better to select by text (in this case, `taxRegistrationModal.cancelButtonText`) when possible

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+        tasks: [taxTask],
+      });
+
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      fireEvent.click(screen.getByTestId("modal-button-primary"));
+      await waitFor(() => {
+        const dropdown = screen.getByTestId("taskProgress");
+        expect(within(dropdown).getByText(Config.taskProgress.COMPLETED)).toBeInTheDocument();
```
aloverso, line 264: the `within` technique is a good idea here, and it'd be nice if we refactored the rest of this test file to use it, instead of how the other tests use `[0]` or `getAllByText` to handle this

aloverso: not as part of this story but as a follow-up refactor

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+        tasks: [taxTask],
+      });
+
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      fireEvent.click(screen.getByTestId("modal-button-secondary"));
+      await waitFor(() => {
+        const dropdown = screen.getByTestId("taskProgress");
+        expect(within(dropdown).getByText(Config.taskProgress.NOT_STARTED)).toBeInTheDocument();
```
aloverso, line 239: Maybe we should also expect userData not to have been updated

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      fireEvent.click(screen.getByTestId("modal-button-primary"));
+      await waitFor(() => {
+        const dropdown = screen.getByTestId("taskProgress");
+        expect(within(dropdown).getByText(Config.taskProgress.COMPLETED)).toBeInTheDocument();
+      });
+      expect(mockPush).toHaveBeenCalledWith({
+        pathname: ROUTES.roadmap,
```
aloverso, line 292: might be more robust to use `routeForPersona` here

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+      });
+
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      fireEvent.click(screen.getByTestId("modal-button-primary"));
+      await waitFor(() => {
+        const dropdown = screen.getByTestId("taskProgress");
+        expect(within(dropdown).getByText(Config.taskProgress.COMPLETED)).toBeInTheDocument();
+      });
```
aloverso, line 290: we already asserted this in another test, we shouldn't have to do it again

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+      useMockRoadmap({
+        steps: [generateStep({ stepNumber: 1, section: "PLAN" })],
+        tasks: [taxTask],
+      });
+
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      expect(
+        screen.getByText(markdownToText(Config.profileDefaults.STARTING.existingEmployees.header))
```
aloverso, line 320: the test name says we're looking at the business name field... why are we asserting about existing employees?

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+
+      useMockRoadmap({
+        steps: [generateStep({ stepNumber: 1, section: "PLAN" })],
+        tasks: [taxTask],
+      });
+
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      expect(screen.getByText("New Jersey Tax ID")).toBeInTheDocument();
```
aloverso, line 421: should be config

aloverso: (in all three tests)

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+      });
+
+      renderTaskHeader(taxTask, userData);
+      selectCompleted();
+
+      expect(screen.getByText(Config.profileDefaults.STARTING.ownership.header)).toBeInTheDocument();
+      expect(
+        screen.getByText(markdownToText(Config.profileDefaults.STARTING.existingEmployees.header))
+      ).toBeInTheDocument();
+    });
```
aloverso, line 556: okay, so we have a *lot* of repetition in these last few tests. I think that for the goal of "test should be readable documentation of the thing's behavior" then we could male some refactoring here to make these more readable. I think making them more concise would be the path here.

here's the tests you have that are all very closely related:
- shows the business name field if it's not populated for non sp/gp
- doesn't show the business name field if it's populated for non sp/gp
- doesn't show the business name field for sp
- doesn't show the business name field for gp
- shows the tax id field for non sp/gp
- doesn't show the tax id field for sp
- doesn't show the tax id field for gp
- shows ownership and employee count for non sp/gp
- shows ownership and employee count for sp
- shows ownership and employee count for gp

For one, I think you could make use of random test generators to avoid duplicating tests for sp and for gp. Imagine if you had a function for `randomSpOrGpLegalStructure` and `randomNonSpGpLegalStructure`. Then you avoid duplication while still covering both the SP and GP cases.

If you had those two functions, it makes it clear that our tests break down into 2 behavior categories: SP/GP and non-SP/GP. We could make that explicit in the tests using describe blocks

Imagine something like:
```
describe('when SP/GP', () => {
 it "shows all fields except business name"
    expect(not business-name-field)
    expect(tax-id-field) 
    expect(ownership-field)
    expect(employee-count-field)

}

describe('when not SP/GP', () => {
  it "shows all fields except tax ID when business name not populated"
    expect(business-name-field)
    expect(not tax-id-field) 
    expect(ownership-field)
    expect(employee-count-field)

  it "does not show business name field when already populated"
    expect(not business-name-field)
}
```

obviously the names are not perfect but this is just one example breakdown that, at a high level, is readable so that people could actually reasonably look through the tests and be expected to understand what's actually going on

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+
+    it("shows the warning modal if the user tries to change status from completed", async () => {
```
aloverso, line 558: I don't think this needs to be async?

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+      });
+
+      useMockRoadmap({
+        steps: [generateStep({ stepNumber: 1, section: "PLAN" })],
+        tasks: [taxTask],
+      });
+
+      renderTaskHeader(taxTask, userData);
+      fireEvent.click(screen.getByTestId("taskProgress"));
+      fireEvent.click(screen.getByTestId("IN_PROGRESS"));
```
aloverso, line 605: the other tests in this file use getAllByText instead to select these. Is it intentional to switch to a testid model? What do you prefer, or think the tradeoffs are?

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +193,456 @@ describe("<TaskHeader />", () => {
...
+
+      expect(screen.getByText(Config.taxRegistrationModal.areYouSureTaxBody)).toBeInTheDocument();
+
+      fireEvent.click(screen.getByText(Config.taxRegistrationModal.areYouSureTaxContinueButton));
+      const dropdown = screen.getByTestId("taskProgress");
+      expect(within(dropdown).getByText(Config.taskProgress.IN_PROGRESS)).toBeInTheDocument();
```
aloverso, line 611: maybe also we should expect user data was not updated

aloverso: same in below test also

web/src/components/TaxRegistrationModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+      userData?.profileData.businessName === "" &&
+      userData?.profileData.legalStructureId !== "sole-proprietorship" &&
+      userData?.profileData.legalStructureId !== "general-partnership"
+    );
+  };
+
+  const showTaxIdField = () => {
+    return (
+      userData?.profileData.legalStructureId !== "sole-proprietorship" &&
+      userData?.profileData.legalStructureId !== "general-partnership"
```
aloverso, line 67: for this and the function above, we should be using our `LookupByLegalStructure` function because it already tracks which legal structure IDs are tradeName versus publicFiling. We shouldn't have the ids hardcoded here

web/src/components/TaskHeader.test.tsx
aloverso: Also, I matched your existing terminology here, but as I noted below, we should be using the domain language for these terms. It's not SP/GP and non-SP/GP, its Public Filing legal structure or Trade Name legal structure

web/src/components/roadmap/SidebarCardsList.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        sidebarDisplayContent={createDisplayContent().sidebarDisplayContent}
+        certifications={certifications}
+        fundings={[]}
+      />
+    );
+    expect(screen.getByTestId("cert1")).toBeInTheDocument();
+    expect(screen.getByTestId("cert2")).toBeInTheDocument();
+  });
+
+  it("doesn't show the certification cards if the user is formed and registered", () => {
```
aloverso, line 54: this is the opposite of what the last test says... typo?

web/src/components/roadmap/SidebarCardsList.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        profileData: generateProfileData({
+          operatingPhase: "NEEDS_TO_REGISTER_FOR_TAXES",
+        }),
+      })
+    );
+
+    const certifications = [
+      generateCertification({ name: "Cert 1", applicableOwnershipTypes: ["disabled-veteran"] }),
+      generateCertification({ name: "Cert 2", applicableOwnershipTypes: [] }),
+      generateCertification({ name: "Cert 3", applicableOwnershipTypes: ["minority-owned"] }), // Filtered out
```
aloverso, line 66: we don't care about this "filtered out" certification for this test, right? why have it here?

web/src/components/roadmap/SidebarCardsList.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+    render(
+      <SidebarCardsList
+        sidebarDisplayContent={createDisplayContent().sidebarDisplayContent}
+        certifications={certifications}
+        fundings={[]}
+      />
+    );
+    expect(screen.queryByTestId("cert1")).not.toBeInTheDocument();
+    expect(screen.queryByTestId("cert2")).not.toBeInTheDocument();
```
aloverso, line 77: is it possible to look by text instead?

web/src/components/roadmap/SidebarCardsList.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const { hideCard } = useRoadmapSidebarCards();
+  const [hiddenAccordionIsOpen, setHiddenAccordionIsOpen] = useState<boolean>(false);
+
+  //   const filteredSortedFundings = userData ? sortFundings(filterFundings(props.fundings, userData)) : [];
+  const filteredSortedCertifications = userData
+    ? sortCertifications(filterCertifications(props.certifications, userData))
+    : [];
+  //   const nonHiddenSortedFundings = filteredSortedFundings.filter(
+  //     (it) => !userData?.preferences.hiddenFundingIds.includes(it.id)
+  //   );
```
aloverso, line 32: cleanup comments (below as well). I recommend looking through added files when making your commit

web/src/components/roadmap/SidebarCardsList.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const nonHiddenSortedCertifications = filteredSortedCertifications.filter(
```
aloverso, line 33: instead of "nonHidden" it might be clearer to use the positive case instead of the negative case. "shown" or "visible" maybe?

web/src/components/roadmap/SidebarCardsList.tsx
```diff
@@ -0,0 +1,154 @@
...
+          {/* {nonHiddenSortedFundings.map((funding) => (
+            <OpportunityCard key={funding.id} opportunity={funding} urlPath="funding" />
+          ))} */}
+          {/* {nonHiddenSortedFundings.length + nonHiddenSortedCertifications.length === 0 && (
+            <div className="fdc fac margin-y-3">
+              <h3 className="text-normal">{Config.dashboardDefaults.emptyOpportunitiesHeader}</h3>
+              <img src={`/img/signpost.svg`} className="" alt="" />
+              <p className="text-center">{Config.dashboardDefaults.emptyOpportunitiesText}</p>
+            </div>
+          )} */}
```
aloverso, line 145: there's a lot of comments in this file about the fundings. What's going on? Will this code work for our existing Oscar experience for fundings?

web/test/pages/roadmap.test.tsx
```diff
@@ -68,15 +68,21 @@ describe("roadmap page", () => {
   const renderRoadmapPage = ({
     sidebarDisplayContent,
     operateReferences,
+    fundings,
+    certifications,
   }: {
     sidebarDisplayContent?: Record<string, SidebarCardContent>;
     operateReferences?: Record<string, OperateReference>;
+    fundings?: Funding[];
+    certifications?: Certification[];
```
aloverso, line 77: these don't need to be part of the arguments if they're never being used by a test. we can just pass `[]` always on line 84/85. same for the function below as well.

web/src/components/TaskHeader.test.tsx
```diff
@@ -37,6 +46,17 @@ const renderTaskHeader = (task: Task, initialUserData?: UserData) =>
...
+const randomTradeNameLegalStructure = () => {
+  return randomInt() % 2 ? "sole-proprietorship" : "general-partnership";
+};
+
+const randomPublicFilingLegalStructure = () => {
+  const nonTradeNameLegalStructures = LegalStructures.filter(
+    (x) => x.id !== "sole-proprietorship" && x.id !== "general-partnership"
+  );
+  return nonTradeNameLegalStructures[randomInt() % nonTradeNameLegalStructures.length].id;
+};
```
aloverso, line 58: part of my rationale for this was to avoid hardcoding the logic into our tests of _which_ legal structures are public filing or trade name, since we already have that logic elsewhere

you don't have to recreate this logic, and instead could do a filter on the `LegalStructures` const that's available based on the `requiresPublicFiling` or `hasTradeName` booleans

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +212,237 @@ describe("<TaskHeader />", () => {
...
+      });
+    });
+
+    it("shows the warning modal if the user tries to change status from completed", () => {
+      const taxTask = generateTask({ id: taxTaskId });
+      const userData = generateUserData({
+        profileData: generateProfileData({
+          businessName: "test-business",
+          legalStructureId:
+            randomInt() % 2 ? randomTradeNameLegalStructure() : randomPublicFilingLegalStructure(),
```
aloverso, line 290: in the spirit of simplifying tests, if this could be either legal structure, can we just get rid of this line and let the generator pick one randomly? And same for businessName, can this be deleted here?

aloverso: same comment in test below

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +212,237 @@ describe("<TaskHeader />", () => {
...
+      });
+      expect(userDataWasNotUpdated).toBeTruthy();
+    });
+
+    describe("when trade name legal structure", () => {
+      it("shows all fields except business name and tax field", () => {
+        const taxTask = generateTask({ id: taxTaskId });
+        const userData = generateUserData({
+          profileData: generateProfileData({
+            businessName: "",
```
aloverso, line 359: i'd delete this, because it doesn't impact the outcome of the tests right? we don't show businessName field regardless of this value

web/src/components/TaskHeader.test.tsx
```diff
@@ -192,6 +212,237 @@ describe("<TaskHeader />", () => {
...
+        ).not.toBeInTheDocument();
+        expect(
+          screen.getByText(markdownToText(Config.profileDefaults.STARTING.taxId.header))
+        ).toBeInTheDocument();
+        expect(
+          screen.getByText(markdownToText(Config.profileDefaults.STARTING.existingEmployees.header))
+        ).toBeInTheDocument();
+        expect(
+          screen.getByText(markdownToText(Config.profileDefaults.STARTING.ownership.header))
+        ).toBeInTheDocument();
```
aloverso, line 441: I'd say these last 3 assertions are unncessary because we've already asserted them in the previous test

web/src/components/TaxRegistrationModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <div className="margin-bottom-3">
+          <Content>{Config.taxRegistrationModal.subtitle}</Content>
+        </div>
+        <hr />
+        {showBusinessField() ? (
+          <OnboardingBusinessName onValidation={onValidation} fieldStates={fieldStates} />
+        ) : (
+          <></>
+        )}
+        {showTaxIdField() ? <OnboardingTaxId onValidation={onValidation} fieldStates={fieldStates} /> : <></>}
```
aloverso, line 100: these could both be just `{showTaxIdField() && <OnboardingTaxId onValidation={onValidation} fieldStates={fieldStates} />}`, we don't need a ternary with an empty case

PR #3792, opened 2022-08-12
feat: [#182490248] add toggle calendar button

AnJuHyppolite:


web/.env-template
```diff
@@ -5,7 +5,7 @@ CYPRESS_COGNITO_USER_POOL_ID=
 CYPRESS_COGNITO_WEB_CLIENT_ID=
 API_BASE_URL=
 AUTH_DOMAIN=
-REDIRECT_URL=
+REDIRECT_URL=http://localhost:3000/loading
```
mikewolfd, line 8: This is the template, no need for putting values in here

aloverso: same comment for the api env template

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -111,4 +135,119 @@ describe("<FilingsCalendar />", () => {
...
+    const dueDate = getCurrentDate().add(2, "months");
+    const annualReport = generateTaxFiling({
+      identifier: "annual-report",
+      dueDate: dueDate.format("YYYY-MM-DD"),
+    });
+
+    const userData = generateUserData({
+      taxFilingData: generateTaxFilingData({ filings: [annualReport] }),
+      preferences: generatePreferences({ isCalendarFullView: true }),
+    });
```
aloverso, line 151: All of this setup (due date, annual report, user data, operate references) are repeated and reused exactly across all these tests. I think it could be pulled out into a beforeEach to clean up and be less repetitive

web/src/components/FilingsCalendar.tsx
```diff
@@ -84,6 +85,17 @@ export const FilingsCalendar = (props: Props): ReactElement => {
     );
   };
 
+  const handleCalendarOnclick = () => {
```
aloverso, line 88: small typo - the C in click should be capitalized for proper camelCase

web/src/components/FilingsCalendar.tsx
```diff
@@ -84,6 +85,17 @@ export const FilingsCalendar = (props: Props): ReactElement => {
...
+    if (userData)
```
aloverso, line 89: we prefer guard clauses (as in, `if (!userData) return` instead of nesting inside of an if-statement. The reason is that it's easier to read code (especially the core code of the function) that has as little nesting as possible, and that is as close to left-aligned with the function margin as possible.

web/src/components/FilingsCalendarAsList.tsx
```diff
@@ -28,18 +33,30 @@ export const FilingsCalendarAsList = (props: Props): ReactElement => {
...
+              {isDesktopAndUp && (
+                <Button
+                  style="light"
+                  className="font-body-2xs padding-y-05 padding-x-1"
+                  onClick={props.onToggle}
+                >
+                  <Icon className="usa-icon--size-3 margin-right-05">menu</Icon>
+                  {Config.dashboardDefaults.calendarFullViewButton}
+                </Button>
+              )}
```
aloverso, line 59: I think there's a bug here - I don't think this button should be visible on the list view when the onToggle function doesn't exist. Currently, it shows up in the list-view calendar after a user enters their formation date 
image

web/src/components/FilingsCalendar.tsx
```diff
@@ -93,18 +105,28 @@ export const FilingsCalendar = (props: Props): ReactElement => {
...
+            </div>
           </div>
+          <Button
+            style="light"
+            className="font-body-2xs padding-y-05 padding-x-1"
+            onClick={handleCalendarOnclick}
+          >
+            <Icon className="usa-icon--size-3 margin-right-05">list</Icon>
+            {Config.dashboardDefaults.calendarListViewButton}
+          </Button>
```
aloverso, line 129: I'm not seeing this button styled and left-aligned the way it is in the designs 
image

web/src/components/FilingsCalendarAsList.tsx
dsueltenfuss: The double header is also back... The first Upcoming header should be removed

aloverso: I figured these were related...but yeah it looks like the rendering is off somehow

web/src/components/FilingsCalendar.tsx
AnJuHyppolite: I could not duplicate this behavior.

web/src/components/FilingsCalendarAsList.tsx
AnJuHyppolite: 1. I am not able to duplicate the double header behavior. 
2. I updated the button logic.
Thanks!

aloverso (approved):

PR #3889, opened 2022-08-23
feat: [#182955641] display hideable roadmap

FarazA22:


api/src/db/migrations/v82_add_up_and_running_owning_operating_phase.ts
```diff
@@ -0,0 +1,318 @@
...
+    v81Data.profileData.businessPersona === "OWNING" &&
+    v81Data.profileData.operatingPhase === "UP_AND_RUNNING"
+      ? "UP_AND_RUNNING_OWNING"
+      : v81Data.profileData.operatingPhase;
+
+  return {
+    ...v81Data,
+    preferences: {
+      ...v81Data.preferences,
+      isHideableRoadmapVisible: false,
```
aloverso, line 27: I know it feels annoying but I think we should have a separate migration for adding this. Or at least change the file name to include it as well maybe?

api/src/db/migrations/v82_add_up_and_running_owning_operating_phase.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  v81UserData,
+} from "./v81_add_completed_filing_payment";
+import { migrate_v81_to_v82 } from "./v82_add_up_and_running_owning_operating_phase";
+
+describe("migrate_v81_to_v82", () => {
+  it("updates operating phase to up_and_running_owning from up_and_running for owning businessPersona", () => {
+    const v81 = makeUserData({ businessPersona: "OWNING", operatingPhase: "UP_AND_RUNNING" });
+    const v82 = migrate_v81_to_v82(v81);
+    expect(v82.profileData.operatingPhase).toEqual("UP_AND_RUNNING_OWNING");
+  });
```
aloverso, line 15: maybe not strictly necessary, but it feels nice to me to cover the additional cases:
- when OWNING but not up-and-running, operating phase should remain the same
- when UP_AND_RUNNING and starting, operating phase should remain the same

api/src/domain/user/updateOperatingPhase.test.ts
```diff
@@ -12,7 +12,7 @@ describe("updateOperatingPhase", () => {
           legalStructureId: "limited-partnership",
         }),
       });
-      expect(updateOperatingPhase(userData).profileData.operatingPhase).toBe("UP_AND_RUNNING");
+      expect(updateOperatingPhase(userData).profileData.operatingPhase).toBe("UP_AND_RUNNING_OWNING");
```
aloverso, line 15: this change should change the test name as well

shared/src/userData.ts
```diff
@@ -55,6 +56,7 @@ export interface Preferences {
   visibleSidebarCards: string[];
   returnToLink: string;
   isCalendarFullView: boolean;
+  isHideableRoadmapVisible: boolean;
```
aloverso, line 59: I think this name could be potentially confusing, because there's both the concept of it being "visible" ie "on the page" in some cases (for the up-and-running phase) as well as "visible" ie "open" when on the page. I think by this you mean the second one, so maybe we could avoid that confusion by naming it something like `isHideableRoadmapOpen` or something instead?

web/src/components/dashboard/HideableTasks.tsx
```diff
@@ -0,0 +1,71 @@
...
+                  .filter((step) => step.section === section)
+                  .map((step, index, array) => (
+                    <Step key={step.stepNumber} step={step} last={index === array.length - 1} />
+                  ))}
+            </SectionAccordion>
+          ))}
+        </>
+      ) : (
+        <div className="text-base">
+          {roadmap?.tasks.length} {Config.dashboardDefaults.hiddenTasksText}
```
aloverso, line 66: typically what we do for phrases that have a dynamic component like a count is to use `${}` notation in the config string itself. For example, take a look at `hiddenOpportunitiesHeader` in the config. 

We want to write the whole phrase including the part that will be substituted for the number. That way, Jess and Ce can rearrange the phrase as needed without a code change. If we change the value of `hiddenTasksText` to `"${count} Hidden Tasks"` then they could, for example, change the content later on to read as `"You have ${count} Hidden Tasks"` without needing us to change any code. And we use `templateEval` to insert the value for `count` into the string.

web/src/pages/dashboard.tsx
```diff
@@ -121,6 +122,7 @@ const DashboardPage = (props: Props): ReactElement => {
             {LookupOperatingPhaseById(userData?.profileData.operatingPhase).displayFullCalendar && (
               <FilingsCalendar operateReferences={props.operateReferences} />
             )}
+            <HideableTasks />
```
aloverso, line 125: It feels like we could move the logic of checking `LookupOperatingPhaseById(...).displayHideableRoadmapTasks` from within the component to here, to follow the pattern of all the components above. Was there a reason to choose one over the other?

web/src/components/dashboard/HideableTasks.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <>
+          {getSectionNames(roadmap).map((section) => (
+            <SectionAccordion key={section} sectionType={section}>
+              {roadmap &&
+                roadmap.steps
+                  .filter((step) => step.section === section)
+                  .map((step, index, array) => (
+                    <Step key={step.stepNumber} step={step} last={index === array.length - 1} />
+                  ))}
+            </SectionAccordion>
```
aloverso, line 61: given that this a copy-pasted of the Roadmap from `dashboard.tsx` it'd be cool if we extracted that duplication out into its own component

web/src/components/dashboard/HideableTasks.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <div className="fdr margin-bottom-105">
+        <h2>{Config.dashboardDefaults.upAndRunningTaskHeader}</h2>
+        <div className="mla">
+          <Button style="narrow-light" onClick={handleToggleClick}>
+            <div className="fdr fac">
+              <Icon>{userData?.preferences.isHideableRoadmapVisible ? "visibility_off" : "visibility"}</Icon>
+              <span className="margin-left-05 line-height-sans-2">
+                {userData?.preferences.isHideableRoadmapVisible
+                  ? Config.dashboardDefaults.hideTaskText
+                  : Config.dashboardDefaults.showTaskText}
```
aloverso, line 43: should we use a MUI Accordion for this? That way we get the nice animation on the show/hide movement like we do for sections in the roadmap

web/src/components/dashboard/HideableTasks.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    const userData = generateUserData({
+      profileData: generateProfileData({ operatingPhase: randomOperatingPhase.id }),
+      preferences: generatePreferences({ isHideableRoadmapVisible: true }),
+    });
+    renderHideableTask(userData);
+
+    fireEvent.click(screen.getByText(Config.dashboardDefaults.hideTaskText));
+
+    expect(screen.getByText("2 Hidden Tasks")).toBeInTheDocument();
```
aloverso, line 67: we should reference config here instead of hardcoding the string in case the text gets changed by the content team

web/src/components/dashboard/HideableTasks.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    const userData = generateUserData({
+      profileData: generateProfileData({ operatingPhase: randomOperatingPhase.id }),
+      preferences: generatePreferences({ isHideableRoadmapVisible: false }),
+    });
+    renderHideableTask(userData);
+
+    fireEvent.click(screen.getByText(Config.dashboardDefaults.showTaskText));
+
+    expect(screen.getByTestId("section-plan")).toBeInTheDocument();
```
aloverso, line 83: i think this test could also expect that the hidden text is not there. And the test previous could expect that this text is not in the document

web/src/components/dashboard/HideableTasks.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { UserData } from "@businessnjgovnavigator/shared/userData";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { HideableTasks } from "./HideableTasks";
+
+jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
+jest.mock("@/lib/data-hooks/useRoadmap", () => ({ useRoadmap: jest.fn() }));
+
+const Config = getMergedConfig();
+
+describe("HideableTasks", () => {
```
aloverso, line 17: typically we use `<HideableTasks />` as the test name for component unit tests

web/src/lib/utils/analytics-helpers.ts
```diff
@@ -73,6 +73,8 @@ const getPhaseDimension = (phase: OperatingPhaseId) => {
       return "Formed and Registered";
     case "UP_AND_RUNNING":
       return "Up and Running";
+    case "UP_AND_RUNNING_OWNING":
+      return "Up and Running";
```
aloverso, line 77: I asked EJ and I did get confirmation that this is correct

web/src/components/dashboard/HideableTasks.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+      <WithStatefulUserData initialUserData={initialUserData}>
+        <HideableTasks />
+      </WithStatefulUserData>
+    );
+
+  it("does not render HideableTasks for operating phases that don't display HideableRoadmapTasks", () => {
+    const randomOperatingPhase = randomElementFromArray(
+      OperatingPhases.filter((obj) => obj.displayHideableRoadmapTasks !== true)
+    );
+    useMockUserData({ profileData: generateProfileData({ operatingPhase: randomOperatingPhase.id }) });
```
aloverso, line 35: As a heads up, you're using 2 different mocking patterns for userData in the same test here. It works, but it's a little confusing to read.

Your render function is expecting initialUserData, and using the <WithStatefulUserData> wrapper to set up a mocked _dynamic_ userData hook. We use this anytime we want to test that changing the state of the userData changes something on the page. You're right that you need this stateful wrapped, because the open/close is controlled by the userData preferences, and without this, then preferences wouldn't actually change and the component wouldn't work.

Whereas `useMockUserData` is for _static_ mocking. It only mocks what happens when we ask for userData. We typically use this when we just need to start with some value but don't ever change it.

Using both of them in the same test is a little confusing and hard-to-follow because you have to think harder about how the two different mocking patterns interact to ensure it actually reflects reality.

I'd recommend removing `useMockUserData` altogether, because we do need the dynamic mocking here. And you can pass the information you have defined here into your render function as initialUserData

api/src/db/migrations/v82_add_up_and_running_owning_operating_phase.ts
FarazA22: I'll create another migration file

web/src/components/dashboard/HideableTasks.tsx
FarazA22: keeping it as is for now, we may create a design qa ticket if Cary wants the accordion animation

PR #4008, opened 2022-09-08
feat: [#182963868] add formation nudge modal

AnJuHyppolite:


web/src/components/dashboard/SidebarCardFormationNudge.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </WithStatefulUserData>
+    );
+  };
+
+  describe("when clicking funding button", () => {
+    jest.resetAllMocks();
+    useMockRouter({});
+    setupStatefulUserDataContext();
+    card = generateSidebarCardContent({ id: "formation-nudge" });
+
```
aloverso, line 39: these couple lines should be within a `beforeEach` block. Because you have them floating at the top of a `describe`, they will only get run once. This *happens* to work right now because you only have one test in this describe block, if you or someone else were to add another one, having these free-floating at the top of the describe would cause some unexpected behavior.

web/src/components/dashboard/SidebarCardFormationNudge.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+      renderWithUserData({
+        profileData: generateProfileData({
+          businessPersona: "STARTING",
+        }),
+        preferences: generatePreferences({ visibleSidebarCards: ["formation-nudge"] }),
+      });
+
+      fireEvent.click(screen.getByTestId("cta-formation-nudge"));
+      expect(currentUserData().profileData.operatingPhase).toEqual("NEEDS_TO_FORM");
+    });
```
aloverso, line 50: A couple related items here that I want to address:
1. This test will always pass and will never fail. The operating phase here *starts* as NEEDS_TO_FORM before you ever even click that button, and it never changes.
2. Because this test never fails, that's how I can tell it was written *after* the code was written, not before. Part of the reason I encourage using TDD in the code is so that we can never wind up in a situation where the test is not testing the behavior of the code, because you never write any code until you have a failing test. And then when you write your code and see the test go from red to green, then you know that your test is accurately testing that specific behavior.
3. Because the test never fails, it hides the fact that it does not actually describe the behavior under test, in a few ways. For one, the test name says "updates operating phase to NEEDS_TO_FORM" and that is what the expectation here on line 49 says as well. But when I look at your code in the component, on line 32-40, it appears to me that when that cta button is clicked, then it opens a modal. Because of that, there is also a whole missing step that is completely untested. What happens after the modal is opened, when the save button on the modal is clicked? That flow should be tested as well.

web/src/components/dashboard/SidebarCardFormationNudge.tsx
```diff
@@ -0,0 +1,52 @@
...
+type Props = {
+  card: SidebarCardContent;
+};
+
+export const SidebarCardFormationNudge = (props: Props): ReactElement => {
+  const [modalOpen, setModalOpen] = useState<boolean>(false);
+  const router = useRouter();
+  const { userData, update } = useUserData();
+
+  const needsToForm = async () => {
```
aloverso, line 19: I'm a little confused by the name of this function. I recommend naming functions in a way that is action-oriented and accurately describes the action that this function will perform when it is invoked.

web/src/components/dashboard/SidebarCardFormationNudge.tsx
```diff
@@ -0,0 +1,52 @@
...
+    router.push({ query: { fromForming: "true" } }, undefined, { shallow: true });
+  };
+
+  const onClick = async () => {
+    if (!userData) return;
+    if (userData.profileData.industryId === "generic" || !userData.profileData.industryId) {
+      setModalOpen(true);
+    } else {
+      await needsToForm();
+    }
```
aloverso, line 38: In addition to my comment in the test, I see this if/else structure that says that two different things happen if the industry ID is generic or not.
1. Is this correct? I do not see anything in the acceptance criteria of the story that indicates that the modal should only open if they are a generic industry. Why does this exist?
2. If it is correct, the tests should describe and test this behavior.

PR #4087, opened 2022-09-16
182962504 tax calendar mvp

mikewolfd:


content/src/fieldConfig/config.json
```diff
@@ -367,7 +367,27 @@
...
+    "taxCalendarModalTaxErrorName": "New Jersey Tax ID",
+    "taxCalendarModalTaxIdHeader": "`#### New Jersey Tax ID|tax-id`",
+    "taxCalendarModalTaxIdMarkdown": "Enter the 12-digit number below (xxx-xxx-xxx/xxx). \n\n (You can find your Tax ID listed as your 'Filer ID' in your email confirmation subject 'Confirmation of New Jersey Business Registration').",
+    "taxCalendarFailedErrorMessageMarkdown": "Review your Tax ID and/or business name for a match with the information in your Business Registration Certificate. If you need assistance, you can chat with one of our business advocates.",
+    "taxCalendarFailedBusinessFieldHelper": "Enter your correct business name",
+    "taxCalendarFailedTaxIdHelper": "Enter a correct 12-digit NJ Tax ID",
+    "taxCalendarFailedUnknownMarkdown": "There was a problem with your verification process, please try again later.",
+    "taxCalendarPendingCopyMarkdown": "#### Your Tax Calendar is pending. \n\n Your tax due dates are being added, check back in **7 days**.",
+    "taxCalendarSnackbarSuccessHeader": "Tax Calendar Added!",
+    "taxCalendarSnackbarSuccessBody": "Your businesses tax deadlines are now in your calendar."
```
aloverso, line 390: just from a content team perspective, might be nice if all `the taxCalendar` configs were their own key, which would make it easier to use in the CMS and also make all the keys shorter

api/src/api/taxFilingRouter.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const userData = generateUserData({});
+    apiTaxFilingClient.onboarding.mockResolvedValue(userData);
+    const taxIdAndBusinessName = generateTaxIdAndBusinessName({});
+    const response = await request(app).post(`/onboarding`).send(taxIdAndBusinessName);
+    expect(apiTaxFilingClient.onboarding).toHaveBeenCalledWith({
+      userId: "some-id",
+      ...taxIdAndBusinessName,
+    });
+    expect(response.status).toEqual(200);
+  });
```
aloverso, line 65: a couple things here:
- these tests aren't testing that the userDataClient.put is being called with the returned results from the taxFilingClient. If you deleted that line of code then the tests would still pass
- I think we need to test the error cases here. In the code, you have the userDataClient call in a try/catch but not the call to the taxFilingClient. What happens when the taxFilingClient call fails? We should test that case

api/src/client/ApiTaxFilingClient.test.ts
```diff
@@ -0,0 +1,237 @@
...
+            {
+              Error: `some-error-content-${randomInt()}`,
+              Field: randomInt() % 2 === 0 ? "Taxpayer ID" : "Business Name",
+            },
+          ]
+        : [],
+    Notice: statusCode == 200 ? "You are signed up for the Tax Calendar Reminder Service." : null,
+    StatusCode: statusCode as 200 | 400 | 500,
+    ...overrides,
+  };
```
aloverso, line 70: there's a lot of logic in these two generators. It might be worth considering simplifying this to something like `generate400OnboardingResponse` and `generate200OnboardingResponse` to make it clearer what happens when. Might make it easier for someone to edit in the future. 🤷

api/src/client/ApiTaxFilingClient.test.ts
```diff
@@ -0,0 +1,237 @@
...
+          },
+        ],
+      };
+
+      await client.lookup(taxIdAndBusinessNameAndEmail.taxId, taxIdAndBusinessNameAndEmail.businessName);
+      expect(mockAxios.post).toHaveBeenCalledWith(`${config.baseUrl}/lookup`, postBody);
+    });
+
+    it("returns taxFiling data on success", async () => {
+      const Results = generateTaxFilingResults(["test-id"], 1) ?? [];
```
aloverso, line 111: I don't understand this line. Is there a chance that your generator will return undefined? Why do you have the `?? []` here?

aloverso: In general, I prefer using overrides in the generators over passing arguments like this, so that the generators are dumb. When they start to have all this logic in them then it starts to get to a place where you might want to test your test helpers, which is a potentially long recursive loop haha. 

So I'd ask that:
- Why pass a number and list of IDs here instead of using overrides to set the sub-properties of this response like we do when using other generators?
- If you do think this is the best way to do it, at the very least, I think these things would be much more readable if they were nested in an object. Having a magic number (or boolean, or string) in the parameters of a function doesn't give you any insight into what it *means* - I have to into the definition of the function to understand why there is a `1` here and how it is even used. This could be really easily remedied by having the parameters of the generator be an object, so that you can label the parameters as you call the function. Like: `generateTaxFilingResults({ids: ["test-id"], numberOfX: 1 })`

** I want note here that I wrote `numberOfX` because I went into the definition of the generator and tried for 20 seconds to parse out what the number parameter was being used for. Number of dates? Number of values? The code there with all that logic is not very clear and that is sort of my point when it comes to adding complex logic to the generator functions, which are typically a more "dumb" helper

api/src/client/ApiTaxFilingClient.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    it("returns only taxFiling state on failure", async () => {
+      const stubResponse = generateApiTaxFilingLookupResponse({}, "FAILED");
+      mockAxios.post.mockRejectedValue({ response: { data: stubResponse, status: 400 } });
+      const response = await client.lookup(
+        taxIdAndBusinessNameAndEmail.taxId,
+        taxIdAndBusinessNameAndEmail.businessName
+      );
+      expect(response.state).toEqual("FAILED");
+      expect(response.filings).toEqual([]);
```
aloverso, line 132: is there a reason why not to expect the whole object? It's nice, when looking at tests as documentation, to see exactly what the response would look like on failure, instead of just asserting on individual elements. As in, what is the `identifier` on failure? Is it skipped here because it's undefined? Or does it have a value that's not shown here? These are the questions that my brain asks and that could be clear in the test by asserting against the whole object

api/src/domain/tax-filings/taxFilingsFactor.test.ts
```diff
@@ -0,0 +1,240 @@
+import { randomInt } from "@shared/intHelpers";
```
aloverso, line 1: this filename has a typo in it `taxFilingsFactor`

api/src/client/ApiTaxFilingClient.ts
```diff
@@ -0,0 +1,167 @@
...
+  TaxFilingResults,
+} from "../domain/types";
+import { LogWriterType } from "../libs/logWriter";
+
+type ApiConfig = {
+  apiKey: string;
+  baseUrl: string;
+};
+
+export const apiTaxFilingClient = (config: ApiConfig, logger: LogWriterType): ApiTaxFilingClient => {
```
aloverso, line 16: I almost thought that this was a violation of dependency inversion, but it's not quite. It's just written/named in a way that makes it look like it.

The idea behind dependency inversion is that nothing in our domain is specific to any particular implementation. Instead, it is an abstract descriptor of what occurs, in the interface type, and then when we implement it here, we provide a specific way of doing that abstract thing.

So in this case, it's incorrect to name something in our domain with the type name `ApiTaxFilingClient` because using the prefix `Api` is a specificity to _how_ it is happening - that is, that we're invoking an Api. instead, it should simply be `TaxFilingClient` and _this_ is the ApiTaxFilingClient - a _specific type_ of TaxFilingClient that _happens_ to invoke the NICUSA api.

It's just like how we would never say `BoomiClient` in our domain, because boomi is a specific way of doing it, but the abstract domain type is a `LicenseSearchClient`

api/src/domain/tax-filings/taxFilingsFactory.ts
```diff
@@ -0,0 +1,43 @@
...
+export const taxFilingsFactory = (
+  userDataClient: UserDataClient,
+  taxFilingClient: ApiTaxFilingClient
+): TaxFilingClient => {
+  const lookup = async (props: {
+    userId: string;
+    taxId: string;
+    businessName: string;
+  }): Promise<UserData> => {
+    let userData = await userDataClient.get(props.userId);
```
aloverso, line 13: I think we talked about this a month or so ago when we were adding operating phase to the backend, but I believe we agreed that we wanted to avoid having domain functions like this do the get/put userDataClient operations. Instead, we wanted these to be more "pure" functions where they get passed a userData, do stuff with it, make some changes, and return the updated userData, and then the caller is responsible for both the get and the put. 

This makes things easier because:
- we can chain update functions like this if need be, because they just operate on what they're passed instead of what's in the DB
- it makes the test simpler because we just pass in a userData object instead of having to mock the userDataClient in this test
- it makes this function more reusable

api/src/domain/tax-filings/taxFilingsFactory.ts
```diff
@@ -0,0 +1,43 @@
...
+    const response = await taxFilingClient.lookup(props.taxId, props.businessName);
+    userData =
+      response.state == "SUCCESS"
+        ? { ...userData, taxFilingData: { ...userData.taxFilingData, ...response } }
+        : { ...userData, taxFilingData: { ...userData.taxFilingData, state: response.state } };
+    return userData;
+  };
+
+  const onboarding = async (props: {
```
aloverso, line 22: props is a frontend word, it does not belong here :(

api/src/domain/tax-filings/taxFilingsFactor.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      beforeEach(() => {
+        userData = generateUserData({
+          taxFilingData: generateTaxFilingData({
+            state: undefined,
+            filings: [generateTaxFiling({})],
+          }),
+        });
+      });
+
+      it("when lookup response is API_ERROR", async () => {
```
aloverso, line 117: `it` tests should read as a sentence. Enforcing this is what allows us to enforce behavior-driven testing, because using the `it` sentence as a descriptor of what "it" (the thing under test) _does_ is making sure that you describe the _behavior_ of it.

So all these tests should describe not the SETUP of the test (ie, "it when the response is error") but instead the OUTCOME of the test (ie, "it does not call onboarding when...") or whatever the outcome may be

api/src/domain/tax-filings/taxFilingsFactor.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        apiTaxFilingClient.onboarding.mockResolvedValue({ state: "SUCCESS" } as TaxFilingOnboardingResponse);
+        stubUserDataClient.get.mockResolvedValue(userData);
+        expect(await taxFilingClient.onboarding(userIdTaxIdBusinessName)).toEqual({
+          ...userData,
+          taxFilingData: { ...userData.taxFilingData, state: "PENDING" },
+        });
+      });
+
+      describe("returns onboarding state when", () => {
+        it("responds with api error", async () => {
```
aloverso, line 198: similar comment as above, this is very confusing naming scheming because "it" (the object under test) is NOT the thing that responds with api error. one of its dependencies responds with an api error and _it_ handles / does something with that

api/src/domain/tax-filings/taxFilingsFactory.ts
```diff
@@ -0,0 +1,43 @@
...
+    userId: string;
+    taxId: string;
+    businessName: string;
+  }): Promise<UserData> => {
+    let userData = await userDataClient.get(props.userId);
+    const response = await taxFilingClient.lookup(props.taxId, props.businessName);
+    userData =
+      response.state == "SUCCESS"
+        ? { ...userData, taxFilingData: { ...userData.taxFilingData, ...response } }
+        : { ...userData, taxFilingData: { ...userData.taxFilingData, state: response.state } };
```
aloverso, line 18: don't these two cases from the ternary result in the same output? either way, it's using the state from the response?

api/src/domain/tax-filings/taxIdHelper.ts
```diff
@@ -0,0 +1,34 @@
+import { TaxFiling } from "@shared/taxFiling";
+import { TaxFilingResults } from "src/domain/types";
+
+export const reformatTaxId = (id: string) => id.trim().replaceAll(" ", "").replaceAll("/", "_").toLowerCase();
```
aloverso, line 4: would be nice if this was named in a way that we know what "reformat" means. Reformat to what? For what?

api/src/domain/tax-filings/taxIdHelper.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const getTaxIds = (id: string, map: Record<string, string[]> = {}): string[] => {
```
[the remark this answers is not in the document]
aloverso: same comment for function below

api/src/domain/tax-filings/taxIdHelper.ts
```diff
@@ -0,0 +1,34 @@
...
+  const val = reformatTaxId(id);
+  return map[val] ?? [val];
+};
+
+export const flattenDeDupAndMapTaxFilings = (
+  results: TaxFilingResults,
+  map?: Record<string, string[]>
+): TaxFiling[] =>
```
aloverso, line 14: curly braces / parens around function bodies

api/src/domain/tax-filings/taxIdHelper.ts
```diff
@@ -0,0 +1,34 @@
...
+  Object.values(
+    results?.reduce(
+      (accumulator: Record<string, TaxFiling[]>, result) => ({
+        ...accumulator,
+        ...getTaxIds(result.Id, map).reduce(
```
aloverso, line 19: a reduce within a reduce? 😳 

could this be rewritten with more helper functions and intermediate variables that have descriptive names, so that someone else would be able to look at this and understand how it works? There's no reason for this to be a one-liner, and it's truly impossible to parse. 

A good heuristic is to write something imagining the most junior member of our team coming in and trying to read it / make a change, and writing in a way that will enable them to do that without help or without talking to you to explain it.

api/src/domain/types.ts
```diff
@@ -50,6 +61,16 @@ export interface SelfRegClient {
   resume: (authID: string) => Promise<SelfRegResponse>;
 }
 
+export type TaxFilingResults = { Content: string; Id: string; Values: string[] }[] | null;
```
aloverso, line 64: I think this type does not belong in the domain? It is an api-specific type, and we should never have types directly referencing external APIs in our domain, because changes on their end should never cause changes in our domain (it should only cause changes in our api adapters)

This is another dependency inversion principle. I know you missed that deep-dive when I did it with anju, soma, and faraz a few weeks ago, but I'd be happy to go through that with you to see the why and how of what the DI should look like in our api.

In any case, this should be an easy fix because you don't appear to be actually USING this type anywhere in the domain, just defining it here. It's only needed in the api client anyways, so it should just live there.

api/test/factories.ts
```diff
@@ -191,6 +210,26 @@ export const generateLicenseStatusResult = (overrides: Partial<LicenseStatusResu
...
+  overrides: Partial<{ businessName: string; taxId: string }>
+): { businessName: string; taxId: string } => {
+  return {
+    businessName: `some-name-${randomInt()}`,
+    taxId: `${randomInt(12)}`,
+    ...overrides,
+  };
+};
+
+export const generateTaxIdAndBusinessNameAndEmail = (
```
aloverso, line 223: if this is only used in the API, it doesn't really belong here.

api/test/factories.ts
```diff
@@ -191,6 +210,26 @@ export const generateLicenseStatusResult = (overrides: Partial<LicenseStatusResu
   };
 };
 
+export const generateTaxIdAndBusinessName = (
+  overrides: Partial<{ businessName: string; taxId: string }>
+): { businessName: string; taxId: string } => {
```
aloverso, line 215: can we give this type a real name and type definition?

web/src/components/FilingsCalendar.tsx
```diff
@@ -24,7 +25,12 @@ export const FilingsCalendar = (props: Props): ReactElement => {
   const Config = getMergedConfig();
   const { userData, update } = useUserData();
   const isLargeScreen = useMediaQuery(MediaQueries.tabletAndUp);
-  const taxFilings = userData?.taxFilingData.filings || [];
+  const taxFilings =
+    userData?.taxFilingData.filings.sort((a, b) =>
+      parseDateWithFormat(a.dueDate, "YYYY-MM-DD").isBefore(parseDateWithFormat(b.dueDate, "YYYY-MM-DD"))
```
aloverso, line 30: a common comment I'm leaving, but I think this one-liner would be more readable if you extracted some of the logic into a well-named function. For example:

```
const sortFilingsEarliestToLatest = () => {...} // is this even right? custom sorts are hard to read. hence my desire for descriptive names so that the code can be understood
const orderedTaxFilings = userData?.taxFilingData.filings ? sortFilingsEarliestToLatest(userData.taxFilingData.filings) : []
```

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -140,6 +142,99 @@ describe("<FilingsCalendar />", () => {
...
+          urlPath: "annual_report-url-path",
+        },
+      };
+
+      renderFilingsCalendar(operateReferences, userData);
+
+      expect(screen.queryByTestId("get-tax-access")).not.toBeInTheDocument();
+    });
+  });
+
```
aloverso, line 237: at first, I was confused by these tests here because this is the FilingsCalendar test and there is no logic in the FilingsCalendar component that handles whether to show/hide this button the way these tests describe. 

Looking further, I discovered the logic that actually controls this lives in the TaxAccess component.  Remembering our techtro conversation with Faraz a few weeks ago, I think that logic doesn't belong at the component level. I believe we agreed that child components should not control whether or not they exist. Instead, parent components should decide whether or not to render that child on the page, and the child should just exist, just do its thing without any logic where it just doesn't exist in certain conditions.

All that to say, these tests are correctly placed but the logic should be moved into the FilingsCalendar component

web/src/components/FilingsCalendarTaxAccess.tsx
```diff
@@ -0,0 +1,74 @@
+import { Content } from "@/components/Content";
+import { SnackbarAlert } from "@/components/njwds-extended/SnackbarAlert";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { useUserData } from "@/lib/data-hooks/useUserData";
+import { LookupLegalStructureById, LookupOperatingPhaseById } from "@businessnjgovnavigator/shared/";
+
+import { ReactElement, useState } from "react";
+import { Gov2GoModal } from "./Gov2GoModal";
```
aloverso, line 8: same comment as elsewhere about dependency inversion. Especially on the frontend, we should never be naming a component after a specific backend API implementation that we don't have control over. Instead, this should be named in a way that describes its purpose in the context of the user and the frontend application experience

web/src/components/FilingsCalendarTaxAccess.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <div className="tax-calendar-button-container grid-row" data-testid="button-container">
+          <div className="margin-bottom-2 tablet:margin-bottom-0 margin-right-2 mobile-lg:grid-col-6 grid-col-12">
+            <Content>{Config.dashboardDefaults.taxCalendarAccessBody}</Content>
+          </div>
+          <Button
+            dataTestid="get-tax-access"
+            style={"secondary"}
+            noRightMargin
+            onClick={() => {
+              setShowModal(!showModal);
```
aloverso, line 49: this should be the callback pattern for a state change, like we were discussing

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+    );
+    userData = generateUserData({
+      profileData: generateProfileData({
+        legalStructureId: legalStructure.id,
+        operatingPhase: randomElementFromArray(
+          OperatingPhases.filter((obj) => obj.displayTaxAccessButton === true)
+        ).id,
+      }),
+      formationData: generateFormationData(
+        { completedFilingPayment: false },
```
aloverso, line 53: why are we specifying this? it seems inaccurate, right? By the time a user is accessing tax filings, they have completed and paid for formation. Although, why does this even matter for this test?

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+  render(
+    <WithStatefulUserData initialUserData={initialUserData}>
+      <FilingsCalendarTaxAccess />
+    </WithStatefulUserData>
+  );
+};
+
+describe("<FilingsCalendarTaxAccess />", () => {
+  let userData: UserData;
+  let flow: Exclude<BusinessPersona, undefined>;
```
aloverso, line 37: why exclude here?

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+        operatingPhase: randomElementFromArray(
+          OperatingPhases.filter((obj) => obj.displayTaxAccessButton === true)
+        ).id,
+      }),
+      formationData: generateFormationData(
+        { completedFilingPayment: false },
+        legalStructure.id as FormationLegalType
+      ),
+    });
+    flow = userData.profileData.businessPersona ?? flow;
```
aloverso, line 57: what is the `?? flow` at the end of this line doing? `userData.profileData.businessPersona` should never be undefined

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+    setupStatefulUserDataContext();
+  });
+
+  it("opens Gov2Go modal when on button click", () => {
+    renderFilingsCalendarTaxAccess(userData);
+    fireEvent.click(screen.getByTestId("get-tax-access"));
+    expect(screen.getByTestId("modal-content")).toBeInTheDocument();
+  });
+
+  it("Pre-populates fields with userData values", async () => {
```
aloverso, line 67: lowercase test names. Also, I don't think this test is async

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+    renderFilingsCalendarTaxAccess({
+      ...userData,
+      profileData: { ...userData.profileData, taxId: "123456789123", businessName: "1234" },
+    });
+    fireEvent.click(screen.getByTestId("get-tax-access"));
+    expect((screen.queryByLabelText("Business name") as HTMLInputElement)?.value).toEqual("1234");
+    expect((screen.queryByLabelText("Tax id") as HTMLInputElement)?.value).toEqual("123-456-789/123");
+  });
+
+  it("makes businessName un-editable if they have completed formation with us", async () => {
```
aloverso, line 77: no need for async

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+  it("opens Gov2Go modal when on button click", () => {
+    renderFilingsCalendarTaxAccess(userData);
+    fireEvent.click(screen.getByTestId("get-tax-access"));
+    expect(screen.getByTestId("modal-content")).toBeInTheDocument();
+  });
+
+  it("Pre-populates fields with userData values", async () => {
+    renderFilingsCalendarTaxAccess({
+      ...userData,
+      profileData: { ...userData.profileData, taxId: "123456789123", businessName: "1234" },
```
aloverso, line 70: not to nitpick, but often, it can make tests more readable if the test values we use are in line with what people expect them to be. Using "1234" as a business name is a little unexpected because it's a number, not a name. It may cause a double take. Especially when used next to "real" numeric field like taxId. Using anything else that's actually a word, not a number, might increase readability

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -0,0 +1,338 @@
...
+
+  it("displays in-line error and alert when businessName field is empty and save button is clicked", () => {
+    renderFilingsCalendarTaxAccess({
+      ...userData,
+      profileData: { ...userData.profileData, taxId: randomInt(12).toString(), businessName: "" },
+    });
+    fireEvent.click(screen.getByTestId("get-tax-access"));
+    fireEvent.click(screen.getByTestId("modal-button-primary"));
+    expect(screen.getByRole("alert")).toHaveTextContent(Config.dashboardDefaults.taxCalendarModalErrorHeader);
+    expect(screen.getByText(Config.profileDefaults[flow].businessName.errorTextRequired)).toBeInTheDocument();
```
aloverso, line 129: maybe an additional expectation here that the API was not called?

web/src/components/Gov2GoModal.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const [fieldStates, setFieldStates] = useState<ProfileFieldErrorMap>(createProfileFieldErrorMap());
+
+  const [apiFailed, setOnAPIfailed] = useState<undefined | "FAILED" | "UNKNOWN">(undefined);
+  const [onSubmitClicked, setOnSubmitClicked] = useState<boolean>(false);
+
+  const getErrors = () => ({
+    businessName: { invalid: profileData.businessName.trim().length === 0 },
+    taxId: { invalid: profileData.taxId?.trim().length != 12 ?? true },
```
aloverso, line 42: magic number, should be extracted as TAX_ID_LENGTH or something

web/src/components/Gov2GoModal.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    if (userDataToSet.taxFilingData.state == "PENDING") {
+      setIsLoading(false);
+      props.close();
+    }
+
+    if (userDataToSet.taxFilingData.state == "FAILED") {
+      setFieldStates((prev) => ({
+        ...fields.reduce(
+          (p, c) => {
```
aloverso, line 89: variable naming

web/src/components/Gov2GoModal.tsx
```diff
@@ -0,0 +1,199 @@
...
+      </Backdrop>
+      <ModalTwoButton
+        isOpen={props.isOpen}
+        close={onClose}
+        title={Config.dashboardDefaults.taxCalendarModalHeader}
+        primaryButtonText={Config.dashboardDefaults.taxCalendarModalNextButton}
+        primaryButtonOnClick={onSubmit}
+        secondaryButtonText={Config.dashboardDefaults.taxCalendarModalPreviousButton}
+      >
+        {hasErrors() && onSubmitClicked && !apiFailed ? (
```
aloverso, line 144: we prefer to use `&&` syntax in the jsx instead of a ternary with an unnecessary `: ( <> </> )}` at the end, which just takes up space and makes things harder to read

aloverso: same comment below

web/src/components/onboarding/OnboardingTaxId.tsx
```diff
@@ -1,25 +1,142 @@
...
+
+interface Props extends Omit<OnboardingProps, "fieldName" | "handleChange"> {
   fieldStates: ProfileFieldErrorMap;
-  headerAriaLevel?: number;
+  splitField?: boolean;
   handleChangeOverride?: (value: string) => void;
 }
-
-export const OnboardingTaxId = (props: Props): ReactElement => {
+export const OnboardingTaxId = ({
```
aloverso, line 17: we have an outstanding refactoring ticket to rename all of the `OnboardingFieldName` fields to `ProfileFieldName` instead, so it'd be good to integrate that new pattern with brand-new fields at least

web/src/contexts/profileDataContext.ts
```diff
@@ -17,7 +17,7 @@ interface ProfileDataState {
   municipalities: Municipality[];
 }
 
-interface ProfileDataContextType {
+export interface ProfileDataContextType {
```
aloverso, line 20: does this need to be exported? I don't see where it's being used

web/src/lib/api-client/apiClient.ts
```diff
@@ -37,6 +37,13 @@ export const getCompletedFiling = (): Promise<UserData> => {
   return get(`/completed-filing`);
 };
 
+export const postTaxRegistrationOnboarding = (props: {
+  businessName: string;
+  taxId: string;
+}): Promise<UserData> => {
+  return post(`/gov2go/onboarding`, props);
+};
```
aloverso, line 45: I'm noticing there is no apiClient function that calls the `/lookup` endpoint. Why does that endpoint need to exist?

web/test/pages/profile.test.tsx
```diff
@@ -468,7 +468,7 @@ describe("profile", () => {
       chooseTab("numbers");
       expect(getEmployerIdValue()).toEqual("12-3456789");
       expect(getEntityIdValue()).toEqual("1234567890");
-      expect(getTaxIdValue()).toEqual("123456790");
+      expect(getTaxIdValue()).toEqual("123-456-790");
```
aloverso, line 471: We're not storing the number with hyphens in the user data, right? This is just the html display element?

mikewolfd: correct

web/src/components/FilingsCalendarTaxAccess.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return LookupOperatingPhaseById(userData?.profileData.operatingPhase).displayTaxAccessButton &&
+      LookupLegalStructureById(userData?.profileData.legalStructureId).requiresPublicFiling &&
+      userData?.taxFilingData.state !== "SUCCESS" ? (
+      <>
+        <Gov2GoModal
+          isOpen={showModal}
+          close={() => setShowModal(false)}
+          onSuccess={async () => {
+            setShowModal(false);
+            await new Promise((resolve) => setTimeout(resolve, 500));
```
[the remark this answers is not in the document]
mikewolfd: I included it for a smoother user experience, it felt jarring when it was just all happening at once.

web/src/components/FilingsCalendar.test.tsx
mikewolfd: I would usually agree, but this component is being rendered in multiple locations with identical logic, I think we need to refactor all the calendar stuff, but I didn't really want to do that at this moment. Maybe I should just duplicate it in both locations thoug

web/src/components/FilingsCalendarTaxAccess.test.tsx
mikewolfd: Part of the AC is that we don't allow users to alter their businessName if they had completed formation with us

web/src/components/FilingsCalendarTaxAccess.test.tsx
mikewolfd: Because otherwise it's annoying to re-use flow in all the config lookups below

web/src/components/FilingsCalendarTaxAccess.test.tsx
mikewolfd: BusinessPersona can be undefined in the type

PR #4089, opened 2022-09-16
feat: [#182963868] add formation nudge modal

AnJuHyppolite:


web/src/components/dashboard/SidebarCardFormationNudge.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  describe("when clicking funding button", () => {
+    beforeEach(() => {
+      jest.resetAllMocks();
+      useMockRouter({});
+      setupStatefulUserDataContext();
+      card = generateSidebarCardContent({ id: "formation-nudge" });
+    });
+
+    it("allows user in NEEDS_TO_FORM Phase to see modal", () => {
```
aloverso, line 47: this test name isn't quite accurate because there's nothing in the content of the test that indicates anything about "NEEDS_TO_FORM"

a better test name might indicate that the modal becomes visible when the user clicks the CTA button

web/src/components/dashboard/SidebarCardFormationNudge.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      expect(currentUserData()).toEqual(updatedUserData);
+      expect(screen.queryByTestId(Config.formationDateModal.header)).not.toBeInTheDocument();
+      await waitFor(() =>
+        expect(mockPush).toHaveBeenCalledWith({ query: { fromForming: "true" } }, undefined, {
+          shallow: true,
+        })
+      );
+    });
+
+    it("allows user to click the cancel button", () => {
```
aloverso, line 93: I'm being a little nitpicky but I really want to help you practice writing descriptive test names, because it is a skill that will help guide you to writing tests more easily.

"allows" here is a little vague because to me it implies that there is a converse - is the user sometimes not "allowed" to click the button?

The outcome of this test, as written in the code, is not whether or not the user is "allowed" to click the button, but rather, what actually happens when they do so.

A more descriptive test name might read something like "it [] when the user clicks the cancel button" - where you fill in the brackets with the _outcome_ that we expect to happen.

PR #4388, opened 2022-10-04
feat: [#183117206] updated 404 page

somabadri:


web/src/pages/404.tsx
```diff
@@ -0,0 +1,34 @@
...
+const PageNotFound = (): ReactElement => {
+  const { Config } = useConfig();
+
+  return (
+    <PageSkeleton>
+      <main className="usa-section padding-top-0 desktop:padding-top-8" id="main">
+        <SingleColumnContainer>
+          <div className="flex flex-column flex-align-center">
+            <img src="./img/page-not-found-icon.svg" alt="404 icon" />
+            <h2 className="margin-top-4 text-accent-cool-darker">{Config.pageNotFoundError.errorHeader}</h2>
```
aloverso, line 17: For accessibility purposes, every page needs to have an `h1` element ([more info](https://fae.disability.illinois.edu/rulesets/HEADING_1/))

You should use an h1 html tag here, but style it to look like an h2 if that's the desired font size/style/etc. Take a look at the `Header` component, for example, to see an h1 styled as an h2

PR #4407, opened 2022-10-07
feat: [#182478361] mobile view dashboard has tabs

FarazA22:


FarazA22:
@aloverso Can I get your eyes on this PR

web/src/components/dashboard/TwoTabDashboardLayout.tsx
```diff
@@ -0,0 +1,105 @@
...
+        userData as UserData
+      ).length;
+    }
+
+    if (LookupOperatingPhaseById(userData?.profileData.operatingPhase).displayFundings) {
+      count += getVisibleFundings(
+        filterFundings(fundings, userData as UserData),
+        userData as UserData
+      ).length;
+    }
```
aloverso, line 70: this is a really clean solution, using the `displayFundings` and `displayCertifications` indicators on the phase as the if-condition. Well done

web/src/components/dashboard/TwoTabDashboardLayout.tsx
```diff
@@ -0,0 +1,105 @@
...
+  firstTab: ReactElement;
+  secondTab: ReactElement;
+  certifications: Certification[];
+  fundings: Funding[];
+}
+
+export default function TwoTabDashboardLayout({ firstTab, secondTab, certifications, fundings }: Props) {
+  const { userData } = useUserData();
+  const [tabIndex, setTabIndex] = React.useState(0);
+  const Config = getMergedConfig();
```
aloverso, line 25: inside of components, we want to use the `useConfig` hook instead. The `getMergedConfig` helper is just for tests or helpers that can't use a hook because they're not a component

web/src/lib/utils/helpers.tsx
```diff
@@ -377,3 +379,13 @@ export const flattenObject = (obj: any) => {
...
+
+export const getVisibleCertifications = (certifications: Certification[], userData: UserData) => {
+  if (!certifications) return [];
+  return certifications.filter((it) => !userData?.preferences.hiddenCertificationIds.includes(it.id));
+};
+
+export const getVisibleFundings = (fundings: Funding[], userData: UserData) => {
+  if (!fundings) return [];
+  return fundings.filter((it) => !userData?.preferences.hiddenFundingIds.includes(it.id));
+};
```
aloverso, line 391: maybe these could be in their own file? similar to how the filter functions are.

web/src/components/dashboard/TwoTabDashboardLayout.tsx
FarazA22: Yay!!!

PR #4515, opened 2022-10-18
feat: [#183014824] see taxes banner for up and running non sp/gp guests

somabadri:


web/src/components/FilingsCalendar.test.tsx
```diff
@@ -409,6 +409,62 @@ describe("<FilingsCalendar />", () => {
       expect(screen.queryByTestId("button-container")).not.toBeInTheDocument();
       process.env.FEATURE_TAX_CALENDAR = "true";
     });
+
+    it("displays button on filings calendar in up and running guest mode for non sp/gp", () => {
+      const dueDate = getCurrentDate().add(2, "months");
+      const annualReport = generateTaxFiling({
+        identifier: "annual-report",
+        dueDate: dueDate.format("YYYY-MM-DD"),
+      });
```
aloverso, line 418: just in terms of the simplest possible setup for the test, this could be removed. The existence (and due date) of the annual report doesn't impact whether the Get Access button shows or hides, so this setup is extraneous

aloverso: (and removing it will make your test much simpler :) )

aloverso: same comment for the test below

web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -80,6 +99,51 @@ describe("<FilingsCalendarTaxAccess />", () => {
     expect(screen.getByTestId("modal-content")).toBeInTheDocument();
   });
 
+  it("open the sign up modal when button is clicked in up and running guest mode for non sp/gp instead of tax modal", async () => {
```
aloverso, line 102: grammar (sorry) - it "opens"

web/src/components/auth/SignUpModal.tsx
```diff
@@ -61,7 +62,14 @@ export const SignUpModal = (): ReactElement => {
...
             onClick={() => {
-              analytics.event.guest_modal.click.go_to_myNJ_registration();
+              if (
+                userData?.preferences.returnToLink ===
+                `${ROUTES.dashboard}?${QUERIES.openTaxFilingsModal}=true`
+              ) {
+                analytics.event.myNJ_prompt_modal_complete_button.click.go_to_myNJ_registration();
+              } else {
+                analytics.event.guest_modal.click.go_to_myNJ_registration();
+              }
```
aloverso, line 72: typically when there's a significant amount of logic in an onClick (more than 2ish lines, or anything it has additional nesting like and if/else her) I recommend extracting and making it a named function, for readability

web/src/lib/utils/analytics.js
```diff
@@ -686,6 +686,8 @@ export default {
       click: {
         show_tax_calendar_modal: () =>
           sendEvent("tax_calendar_banner_button", "click", "show_tax_calendar_modal"),
+        show_myNJ_registration_prompt_modal: () =>
+          sendEvent("tax_calendar_banner_button", "click", "how_myNJ_registration_prompt_modal"),
```
aloverso, line 690: type here - this is important, the event won't send as EJ expects if this is wrong

web/src/components/FilingsCalendar.test.tsx
somabadri: I changed it to only contain specific userData as that's the only thing affecting the behavior.

PR #4609, opened 2022-10-28
feat: [#183316444] add functionality to skip industry if no essential…

AnJuHyppolite:
… question


web/src/pages/profile.tsx
```diff
@@ -36,7 +36,7 @@ import { IsAuthenticated } from "@/lib/auth/AuthContext";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { useRoadmap } from "@/lib/data-hooks/useRoadmap";
 import { useUserData } from "@/lib/data-hooks/useUserData";
-import { isHomeBasedBusinessApplicable } from "@/lib/domain-logic/isHomeBasedBusinessApplicable";
+import { isHomeBasedBusinessApplicable } from "@/lib/domain-logic/essentialQuestions/isHomeBasedBusinessApplicable";
```
aloverso, line 39: I think this file should probably be moved out of this folder given that homeBasedBusiness isn't really an essential industry-based question

PR #4634, opened 2022-10-31
feat: [#183559302] tax id encryption mvp

somabadri:


api/src/api/taxFilingRouter.test.ts
```diff
@@ -19,6 +19,7 @@ describe("taxFilingRouter", () => {
 
   let stubUserDataClient: jest.Mocked<UserDataClient>;
   let apiTaxFilingClient: jest.Mocked<TaxFilingInterface>;
+  let stubTaxEncryptionDecryptionClient: jest.Mocked<EncryptionDecryptionClient>;
```
aloverso, line 22: A nitpick, but technically this is just a `stubEncryptionDecryptionClient` (no "Tax") because there's nothing in the abstract interface (nor the implementation, even) that makes it specific to tax. It can encrypt and decrypt anything.

api/src/api/taxFilingRouter.test.ts
```diff
@@ -102,6 +164,25 @@ describe("taxFilingRouter", () => {
...
+      apiTaxFilingClient.onboarding.mockResolvedValue(responseUserData);
+      const response = await request(app).post(`/onboarding`).send(taxIdAndBusinessName);
+      expect(response.body).toEqual(responseUserData);
+      expect(apiTaxFilingClient.onboarding).toHaveBeenCalledWith({
+        userData,
+        taxId: "123456789000",
+        businessName: "my-cool-business",
+      });
+      expect(stubUserDataClient.put).toHaveBeenCalledWith(responseUserData);
+      expect(stubUserDataClient.get).toHaveBeenCalledWith("some-id");
```
aloverso, line 182: there's an expectation missing that, once you add it, will reveal that this test isn't exactly set up to mirror the way it would look in the real world.

We should have an expectation to assert what the stubEncryptionDecrytionClient.decrypt was called with

api/src/api/taxFilingRouter.test.ts
```diff
@@ -52,11 +60,51 @@ describe("taxFilingRouter", () => {
...
+      expect(response.status).toEqual(200);
+    });
+
+    it("decrypts the taxId field using the encryptedTaxId field if it is masked", async () => {
```
aloverso, line 94: is there a duplicate here? Is this the same test as line 167?

api/src/api/taxFilingRouter.test.ts
aloverso: (same comment for the other mirrored test in lookup)

api/src/api/taxFilingRouter.test.ts
```diff
@@ -52,11 +60,51 @@ describe("taxFilingRouter", () => {
...
     it("returns userData", async () => {
-      const taxIdAndBusinessName = generateTaxIdAndBusinessName({});
+      const taxIdAndBusinessName = generateTaxIdAndBusinessName({
+        businessName: "my-cool-business",
+        taxId: "123456789000",
+      });
+      apiTaxFilingClient.lookup.mockResolvedValue(responseUserData);
+      const response = await request(app).post(`/lookup`).send(taxIdAndBusinessName);
+      expect(response.body).toEqual(responseUserData);
+      expect(stubUserDataClient.put).toHaveBeenCalledWith(responseUserData);
```
aloverso, line 70: should we expect this `put` to have put in the data with the encyrpted value?

aloverso: same comment for all the tests

api/src/api/taxFilingRouter.ts
```diff
@@ -1,19 +1,32 @@
...
+  return encryptedTaxId === undefined && taxId.includes(maskingCharacter) === false
+    ? taxId
+    : await taxEncryptionDecryption.decryptValue(taxId);
+};
+
 export const taxFilingRouterFactory = (
   userDataClient: UserDataClient,
-  taxFilingInterface: TaxFilingInterface
+  taxFilingInterface: TaxFilingInterface,
+  taxEncryptionDecryption: EncryptionDecryptionClient
```
aloverso, line 19: our pattern is to name the variable the same as the type but lowercase. So `encryptionDecrytionClient`

This is so that the name clearly communicates what type it is. But you know it's a variable not a type because of the lowercase

api/src/api/userRouter.test.ts
```diff
@@ -50,6 +50,7 @@ describe("userRouter", () => {
   let stubUpdateLicenseStatus: jest.Mock;
   let stubUpdateRoadmapSidebarCards: jest.Mock;
   let stubUpdateOperatingPhase: jest.Mock;
+  let stubTaxEncryptionDecryptionClient: jest.Mocked<EncryptionDecryptionClient>;
```
aloverso, line 53: same comment going to be all over the place - no need for the word "Tax" in this variable name (and it is technically not accurate)

api/src/api/userRouter.test.ts
```diff
@@ -391,5 +396,92 @@ describe("userRouter", () => {
         });
       });
     });
+
+    describe("Tax ID changes", () => {
```
aloverso, line 400: can we make this something a little more descriptive like "when user changes tax id" or something

api/src/api/userRouter.test.ts
```diff
@@ -391,5 +396,92 @@ describe("userRouter", () => {
...
+        stubTaxEncryptionDecryptionClient.encryptValue.mockResolvedValue("some-encrypted-value");
+
+        await request(app).post(`/users`).send(updatedUserData).set("Authorization", "Bearer user-123-token");
+
+        expect(stubUserDataClient.put).toHaveBeenCalledWith({
+          ...updatedUserData,
+          profileData: {
+            ...updatedUserData.profileData,
+            taxId: "*******89123",
+          },
```
aloverso, line 459: what is the point of this test? isn't it testing the same thing as the one above it?

api/src/api/userRouter.ts
```diff
@@ -76,11 +79,31 @@ const businessHasFormed = (userData: UserData): boolean => {
   return userData.formationData.getFilingResponse?.success ?? false;
 };
 
+const handleTaxId = async (
+  userData: UserData,
+  taxEncryptionDecryptionFactory: EncryptionDecryptionClient
+): Promise<UserData> => {
```
aloverso, line 85: there's no reason for this function to exist outside the context of the userRouter. if it was contained within the router, then you would not have to pass the encryption client in as an argument

aloverso: also, let's be careful about naming here. is that variable really a taxEncryptionDecryption_Factory_?

remember, a factory is a function that _returns_ an object

api/src/db/migrations/v98_added_encrypted_tax_id.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    // expect(v98).toEqual({
+    //   ...v97,
+    //   profileData: {
+    //     ...v97.profileData,
+    //     taxId: "*****89000",
+    //     encryptedId: "my encrypted value"
+    //   },
+    //   version: 98,
+    // });
+    expect(true).toBe(true);
```
aloverso, line 28: commented code should be removed if not relevant - looks like this file needs a but of cleanup?

api/src/db/migrations/v98_added_encrypted_tax_id.ts
```diff
@@ -0,0 +1,469 @@
...
+
+const taxEncryptionDecryptionClient: EncryptionDecryptionClient = TaxEncryptionDecryptionFactory(
+  process.env.AWS_CRYPTO_KEY ?? "",
+  [],
+  {
+    stage: process.env.AWS_CRYPTO_CONTEXT_STAGE ?? "",
+    purpose: process.env.AWS_CRYPTO_CONTEXT_PURPOSE ?? "",
+    origin: process.env.AWS_CRYPTO_CONTEXT_ORIGIN ?? "",
+  }
+);
```
aloverso, line 27: cleanup this file too

api/src/domain/user/maskTaxId.ts
```diff
@@ -0,0 +1,5 @@
+export const maskTaxId = (taxId: string): string => {
+  return taxId.length == 12
+    ? `*******${taxId.slice(7, taxId.length)}`
+    : `*****${taxId.slice(5, taxId.length)}`;
```
aloverso, line 4: I believe the whole reason that we talked about extracting `*` into a shared variable was so that anytime it was referenced in code (the tests are fine) that we should be referencing it by variable name

api/src/domain/user/taxEncryptionDecryption.test.ts
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import * as AWSCrypto from "@aws-crypto/client-node";
+import { TextEncoder } from "node:util";
+import { EncryptionDecryptionClient } from "../types";
+import { TaxEncryptionDecryptionFactory } from "../user/taxEncryptionDecryptionFactory";
+
+type context = {
```
aloverso, line 7: types should always start with a capital letter

aloverso: and if it's aws-mock specific i might call it like `AWSContext` or something

api/src/domain/user/taxEncryptionDecryption.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      },
+    });
+    expect(mockAWSCrypto.optionsCalledWith).toEqual("REQUIRE_ENCRYPT_REQUIRE_DECRYPT");
+  });
+
+  it("calls decrypt with the correct value", async () => {
+    await client.decryptValue("2323232");
+    expect(mockAWSCrypto.decryptValueCalledWith).toEqual("2323232");
+  });
+});
```
aloverso, line 64: there is a missing expectation in both tests that doesn't test what either of these functions _returns_

api/src/domain/user/taxEncryptionDecryption.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      const encrypt = (cmm: AWSCrypto.KeyringNode, plaintext: string, op?: context) => {
+        this.encryptValueCalledWith = plaintext;
+        this.contextCalledWith = { ...op };
+        return Promise.resolve(plaintext).then(() => {
+          return { result: plaintext, messeageHeader: "message-header" };
+        });
+      };
+      // eslint-disable-next-line unicorn/consistent-function-scoping
+      const decrypt = (keyring: AWSCrypto.KeyringNode, value: string) => {
+        const encoder = new TextEncoder();
```
aloverso, line 86: what is this encoder doing? why do we need it in a test object?

api/src/domain/user/taxEncryptionDecryptionFactory.ts
```diff
@@ -0,0 +1,50 @@
...
+import { TextDecoder } from "node:util";
+import { EncryptionDecryptionClient } from "../types";
+
+type context = {
+  stage: string;
+  purpose: string;
+  origin: string;
+};
+
+export const TaxEncryptionDecryptionFactory = (
```
aloverso, line 13: again, no need for `Tax` to be here. Nothing here is Tax specific.

This implementation is a specific type of an Encryption client - but it's not tax-specific.

What makes this implementation unique? It's that it uses AWS. We _could_ write any encryption client that uses any random encrytion algorithm. This one is unique because it's an AWS Encryption Client - and that's how it should be named

api/src/domain/user/taxEncryptionDecryptionFactory.ts
```diff
@@ -0,0 +1,50 @@
...
+  generatorKeyId: string,
+  keyIds: string[],
```
aloverso, line 15: why does this parameter exist? It looks like we're never using it in the app.ts file?

web/src/components/FilingsCalendarTaxAccess.tsx
```diff
@@ -53,6 +53,7 @@ export const FilingsCalendarTaxAccess = (): ReactElement => {
         const updatedUserData = await postTaxRegistrationLookup({
           businessName: userData.taxFilingData.businessName as string,
           taxId: userData.profileData.taxId as string,
+          encryptedTaxId: userData.profileData.encryptedTaxId as string,
```
aloverso, line 56: there should be a test for this?

web/src/components/TaxRegistrationModal.test.tsx
```diff
@@ -450,6 +450,66 @@ describe("<TaxRegistrationModal>", () => {
...
+    renderComponent(userData);
+    fireEvent.click(screen.getByLabelText("Tax id"));
+    fireEvent.change(screen.getByLabelText("Tax id"), { target: { value: "123456" } });
+    fireEvent.blur(screen.getByLabelText("Tax id"));
+    fireEvent.click(screen.getByTestId("modal-button-primary"));
+    expect(modalOnClose).not.toHaveBeenCalled();
+    expect(
+      screen.getByText(Config.profileDefaults[getFlow(userData.profileData)].taxId.errorTextRequired)
+    ).toBeInTheDocument();
+  });
```
aloverso, line 512: did you write all these tests? for that oneline change? or are you backfilling missing testst?

aloverso: it looks like these tests are duplicated above... is there a difference here that i'm missing?

yarn.lock
```diff
@@ -60,6 +60,46 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@aws-crypto/cache-material@npm:^3.1.0":
```
aloverso, line 63: @mikewolfd i can't remember but I feel like dave said something at one point about not committing the yarn-lock file? Or was that package-lock?

web/src/components/TaxRegistrationModal.test.tsx
somabadri: I had to move the test around because of a rebase, they were pre-existing.

PR #4646, opened 2022-11-01
fix: [#183679981] refactoring essential questions with more comprehen…

mikewolfd:
…sive testing


web/src/lib/types/types.ts
```diff
@@ -146,6 +146,17 @@ export type ProfileFields =
       keyof typeof profileFieldsFromConfig)
   | keyof BusinessUser;
 
+export type ProfileLabelField = Exclude<ProfileFields, keyof BusinessUser | "businessPersona">;
```
aloverso, line 149: can we name this type something else? We already have a React component named `FieldLabelProfile` and this is very similar.

web/src/components/onboarding/OnboardingIndustry.test.tsx
```diff
@@ -40,131 +54,99 @@ describe("<OnboardingIndustry />", () => {
...
+                `radioButton${capitalizeFirstLetter(
+                  kabobSnakeSentenceToCamelCase(choices[0].toString())
+                )}Text`
+              ]
+            )
+          );
+          expect(currentProfileData()[el.fieldName]).toEqual(choices[0]);
+          selectIndustry("generic");
+          expect(currentProfileData()[el.fieldName]).toEqual(emptyIndustrySpecificData[el.fieldName]);
+        });
```
aloverso, line 149: this is hard to read at points, and maybe we could help that by adding named helper functions like `chooseFirstRadioButton` or `chooseRadioWithContent` but assuming this test works and tests everything the way it appears to, this seems like a huge win

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
...
+  ProfileData,
+} from "@businessnjgovnavigator/shared";
+
+type EssentialQuestionFunction = (industry: Industry) => boolean;
+interface EssentialQuestion
+  extends Omit<EssentialQuestionsProps<ProfileData[keyof ProfileData]>, "fieldName" | "choices"> {
+  fieldName: keyof IndustrySpecificData;
+  validationFunc: EssentialQuestionFunction;
+  reset: boolean;
+  convenienceFunc: (industryId: string | undefined) => boolean;
```
aloverso, line 16: I'm not sure that the name `convenienceFunc` describes this function well... I guess I wouldn't be sure what that means at all. Maybe we could rename that to something more descriptive? idk, like `isQuestionApplicable` or `isQuestionApplicableToIndustry` or something

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
+import { EssentialQuestionsProps, ProfileLabelField } from "@/lib/types/types";
+import {
+  emptyIndustrySpecificData,
+  Industry,
+  IndustrySpecificData,
+  LookupIndustryById,
+  ProfileData,
+} from "@businessnjgovnavigator/shared";
+
+type EssentialQuestionFunction = (industry: Industry) => boolean;
```
aloverso, line 10: this type name is super generic and I think it should be renamed if we're only using it as a validation function type?

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
...
+interface EssentialQuestion
+  extends Omit<EssentialQuestionsProps<ProfileData[keyof ProfileData]>, "fieldName" | "choices"> {
+  fieldName: keyof IndustrySpecificData;
+  validationFunc: EssentialQuestionFunction;
```
aloverso, line 14: "validationFunc" that returns a boolean is a poor name because it doesn't tell me anything about what it does. Does it return true if it's valid? True if there are validation errors and it is invalid? A clearer name could help with readability

aloverso: Changed my mind... I looked below and it doesn't seem like this validation has anything to do with the user data? It's testing if the essential question is applicable to the industry? This seems the same as "conveniencFunc"? Can you explain the difference here, I think I am missing it

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
...
+  reset: boolean;
```
aloverso, line 15: "reset" here reads as an action, like a function, but it's actually a boolean and I think it'd be helpful for us to understand if it was named more clearly. I'm trying to guess what this field represents based on the code below and I'm guessing something like `shouldBeResetWhenIndustryChanges`?

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
...
+export const getResetIndustrySpecificData = (industryId: string | undefined) => {
+  const industry = LookupIndustryById(industryId);
+  return EssentialQuestions.filter((eQ) => {
+    return eQ.reset && !eQ.validationFunc(industry);
+  }).reduce((reducer, eQ) => {
+    return { ...reducer, [eQ.fieldName]: emptyIndustrySpecificData[eQ.fieldName] };
+  }, {} as Partial<IndustrySpecificData>);
+};
+
+export const getValidationFuncByFieldName = (fieldName: ProfileLabelField) => {
```
aloverso, line 28: I didn't realize that we didn't have the eslint rule that enforces return types on functions, but we should add it.

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
...
+  return EssentialQuestions.find((eQ) => {
+    eQ.contentFieldName == fieldName || eQ.fieldName == fieldName;
+  })?.validationFunc;
+};
+
+export const validateByIndustryIdAndFieldName = (fieldName: ProfileLabelField) => {
```
aloverso, line 34: I have the same comment on the use of the word "validate" here - validate what?

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
...
+export const isCertifiedInteriorDesignerApplicable =
+  validateByIndustryIdAndFieldName("certifiedInteriorDesigner");
+export const isChildcareForSixOrMoreApplicable = validateByIndustryIdAndFieldName("isChildcareForSixOrMore");
+export const isCpaRequiredApplicable = validateByIndustryIdAndFieldName("requiresCpa");
+export const isCannabisLicenseApplicable = validateByIndustryIdAndFieldName("cannabisLicenseType");
+export const isLiquorLicenseApplicable = validateByIndustryIdAndFieldName("liquorLicense");
+export const isProvidesStaffingServicesApplicable =
+  validateByIndustryIdAndFieldName("providesStaffingService");
+export const isRealEstateAppraisalManagementApplicable = validateByIndustryIdAndFieldName(
+  "realEstateAppraisalManagement"
```
aloverso, line 63: going through these, it seems like `validationFunc` and `convenienceFunc` are the same? Why two fields?

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -0,0 +1,155 @@
...
+);
+
+export const EssentialQuestions: EssentialQuestion[] = [
+  {
+    reset: true,
+    validationFunc: (industry) => {
+      return (
+        !!industry.industryOnboardingQuestions.isInterstateTransportApplicable && industry.id === "logistics"
```
aloverso, line 71: I don't think we need `!!` on all of these because they are already booleans, right? You only need that if you're going to type-coerce something like a string into a boolean type. But things like `industry.industryOnboardingQuestions.isInterstateTransportApplicable` should already be a boolean

web/test/factories.ts
```diff
@@ -722,26 +722,37 @@ export const randomOwnershipType = (): OwnershipType => {
   return ownershipTypes[randomIndex];
 };
 
-export const randomIndustry = (canHavePermanentLocation = false): Industry => {
+export const randomFilteredIndustry = (func: (industry: Industry) => boolean, enabled = true): Industry => {
```
aloverso, line 725: This is feedback I gave to Soma last week and I want to start enforcing as much as possible - we should try to never have raw booleans in function parameters. It's pretty confusing to see something like `randomFilteredIndustry(func, false)` and have no idea what that false represents. Anytime we have a boolean parameter it should be inside of an object so that it can be a named parameter like `randomFilteredIndustry(func, {isEnabled: false})` which is much clearer

web/test/factories.ts
```diff
@@ -722,26 +722,37 @@ export const randomOwnershipType = (): OwnershipType => {
...
-  return homeBasedIndustries[randomIndex].id;
 };
 
-export const randomNonHomeBasedIndustry = () => {
-  const nonHomeBasedIndustries = Industries.filter((it) => {
-    return !it.industryOnboardingQuestions.canBeHomeBased && it.canHavePermanentLocation && it.isEnabled;
-  });
-  const randomIndex = Math.floor(Math.random() * nonHomeBasedIndustries.length);
-  return nonHomeBasedIndustries[randomIndex].id;
+export const randomIndustry = (canHavePermanentLocation = false): Industry => {
```
aloverso, line 739: same boolean comment as above

web/src/components/onboarding/OnboardingIndustry.test.tsx
```diff
@@ -40,131 +54,99 @@ describe("<OnboardingIndustry />", () => {
...
+              Content({
+                children: fieldContent.description,
+              })
+            )
+          );
+          choices.map((choice) => {
+            fireEvent.click(
+              screen.getByText(
+                fieldContent[
+                  `radioButton${capitalizeFirstLetter(kabobSnakeSentenceToCamelCase(choice.toString()))}Text`
```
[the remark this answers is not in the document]
aloverso: also, I was going to make a comment about how "kebab" is spelled "kebab" and not "kabob" but then I googled it to check myself and it turns out it's either! how crazy is that. But I still think "kebab" is considered the standard, at least according to other JS developers 
image

web/setupTests.js
```diff
@@ -1,12 +1,16 @@
...
+import { TextDecoder, TextEncoder } from "util";
 
 process.env.API_BASE_URL = "";
 process.env.REDIRECT_URL = "http://www.example.com";
 
 global.scrollTo = jest.fn();
 global.gtag = jest.fn();
 
+global.TextEncoder = TextEncoder;
+global.TextDecoder = TextDecoder;
```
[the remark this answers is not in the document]
mikewolfd: For the ReactDomServer.renderToMarkup function that I was exploring as a way to properly render and test content using the `<content>` widget

web/src/lib/domain-logic/essentialQuestions.ts
mikewolfd: a few of them can be resolved as undefined, so the double bang is needed to avoid a typescript error

PR #4746, opened 2022-11-17
Landing page redesign get started #183443994

AnJuHyppolite:


web/src/components/LandingPageTiles.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { ReactElement } from "react";
+import { LandingPageActionTile } from "./LandingPageActionTile";
+
+export const LandingPageTiles = (): ReactElement => {
+  const router = useRouter();
+  const isMobile = useMediaQuery(MediaQueries.isMobile);
+  const { Config } = useConfig();
+
+  const autoSelectPoppyAndRouteToIndustry = () => {
```
aloverso, line 14: I think this function should be renamed; it is not fully accurate to the action that the function is performing

web/src/components/njwds/Hero.tsx
```diff
@@ -57,24 +52,13 @@ export const Hero = (): ReactElement => {
...
-                  onClick={primaryCTAOnClick}
-                  dataTestid="hero-login-button"
-                  widthAutoOnMobile
-                  noRightMargin
-                >
-                  {landingPageConfig.heroCallToActionText}
-                </Button>
-              </div>
-
-              <AuthButton position="HERO" landing />
```
aloverso, line 72: Same thing here regarding code cleanup. If we are removing AuthButton from the hero component, then the AuthButton component doesn't need to support "HERO" as a prop position anymore, since it's not getting used anywhere else.

These kinds of changes are often known as the "scout rule" aka "leaving the campground cleaner than when you found it." Even though your feature does not require changes to the AuthButton component, your change allows the AuthButton component to be changed to have a good deal less complexity, which is always desirable in the codebase. We don't want to support props like "HERO" that we don't use anymore. So part of being a good steward of the codebase is to notice the effect that your changes has on other components you might not otherwise have changed.

This technique doesn't require you to know every single component and what it does, but just be observant of what's being changed. I didn't previously know/remember that AuthButton took a position prop. But seeing this line of code being removed, with a prop labeled "HERO", I'm thinking "huh, if we're removing this from the Hero component, I bet that component doesn't need to take a prop called 'HERO' anymore". Then you can go to the component, check your assumptions, and clean up the code accordingly.

web/src/lib/domain-logic/routes.ts
```diff
@@ -41,6 +41,7 @@ export enum QUERIES {
   industry = "industry",
   code = "code",
   openTaxFilingsModal = "openTaxFilingsModal",
+  flow = "flow",
```
aloverso, line 44: Anytime we add a query value here, we also need to add the type definition for its allowed values in the interface QUERY_PARAMS_VALUES. This allows us to do static type checking when we route to query parameters in the code

web/src/pages/onboarding.tsx
```diff
@@ -171,19 +179,28 @@ const OnboardingPage = (props: Props): ReactElement => {
...
           const queryPage = Number(router.query[QUERIES.page]);
           const queryIndustryId = router.query[QUERIES.industry] as string | undefined;
+          const queryFlow = (router.query[QUERIES.flow] as string | undefined)?.toUpperCase();
+
+          // if (flowQueryParamIsValid(queryFlow)) {
+          //   currentUserData = {
+          //     ...currentUserData,
+          //     profileData: { ...currentUserData.profileData, businessPersona: queryFlow as FlowType },
+          //   };
+          // }
```
aloverso, line 196: commented code should be removed

web/src/pages/onboarding.tsx
```diff
@@ -143,6 +143,15 @@ const OnboardingPage = (props: Props): ReactElement => {
...
   };
 
+  const flowQueryParamIsValid = (flow: string | undefined): boolean => {
+    return (
+      //eslint-disable-next-line unicorn/prefer-array-some
+      businessPersonas.find((el) => {
+        return el === flow;
+      }) !== undefined
+    );
+  };
```
aloverso, line 153: why are we disabling this linting rule? It seems to be giving you helpful advice about how this code could be improved and simplified.

web/src/pages/onboarding.tsx
```diff
@@ -221,6 +238,20 @@ const OnboardingPage = (props: Props): ReactElement => {
     }
   };
 
+  const setBusinessPersonaAndRouteToPage = async (flow: FlowType): Promise<void> => {
+    if (flow === undefined) {
+      return;
+    }
```
aloverso, line 244: why is this line necessary? According to your type definition in the `types` file, FlowType can never be undefined

web/src/styles/components/buttons.scss
```diff
@@ -114,3 +114,46 @@
...
+
+  @media (min-width: 643px) {
+    flex-direction: column;
+    padding: 20px 20px 24px 20px;
+    gap: 18px;
+    width: 200px;
+    height: 155px;
+    border-radius: 20px;
+  }
+}
```
aloverso, line 159: A couple things here: 
1. We always use rem, not px. This is for accessibility purposes; if you want to learn more you can take a look at [articles like this](https://uxdesign.cc/why-designers-should-move-from-px-to-rem-and-how-to-do-that-in-figma-c0ea23e07a15)
2. We always use variables in our sass files, which ensures that our styleguides are always being followed. In your case, that means replacing any hardcoded hex colors with the appropriate variable name, and also replacing the hardcoded pixel widths with their variable names. These can be found in `variables.scss`
3. There's a ton of repeated / duplicated styles between these two classes. It seems like a better-architected approach would be to have a single class called "landing-page-tile" that defined the styles, and then have classes for "primary" and "secondary" (or even, maybe just "secondary"... "primary" is _primary_ after all...), which would define only the styles that are unique to being "secondary"

web/test/pages/onboarding/onboarding-starting.test.tsx
```diff
@@ -274,4 +274,13 @@ describe("onboarding - starting a business", () => {
   describe("validates self-reg step", () => {
     runSelfRegPageTests({ businessPersona: "STARTING", selfRegPage: "4" });
   });
+
+  describe("when onboarding flow is starting", () => {
```
aloverso, line 278: this test name is a little confusing because it's not really testing when the "onboarding flow" is starting. Instead, it's testing when the _query parameter_ flow is starting. These are different things, because the query parameter flow actually _sets_ the onboarding flow. So saying that the description of the test is "_when_ the onboarding flow is starting" is kind of backwards

web/test/pages/onboarding/onboarding-starting.test.tsx
```diff
@@ -274,4 +274,13 @@ describe("onboarding - starting a business", () => {
...
+    it("routes user to step 2 when query parameter exists and value is starting", () => {
+      useMockRouter({ isReady: true, query: { flow: "starting" } });
+      renderPage({});
+
+      expect(screen.getByTestId("step-2")).toBeInTheDocument();
+    });
+  });
```
aloverso, line 285: I think it might be more accurate to put this new test in `onboarding-shared` because as soon as we add more flows, it will no longer apply just to "starting". `onboarding-shared` is where we test all the other query parameters that can apply to the onboarding flow.

aloverso:
I don't think the styling here matches the designs. Here's what I'm seeing when I run your code:
image


Here's what I see in the designs:
image


I don't think the tiles should not be spaced out all the way across the page. It looks like the designers want them lined up next to one another?

web/src/components/LandingPageTiles.tsx
AnJuHyppolite: updated to setFlowToStartingAndRouteToIndustry

web/src/components/AuthButton.tsx
```diff
@@ -9,7 +9,7 @@ import { useRouter } from "next/router";
 import { ReactElement, useContext } from "react";
 
 interface Props {
-  position: "HERO" | "NAVBAR";
+  position: "NAVBAR";
```
aloverso, line 12: this prop does not need to exist at all if it only ever has one possible value. It should just BE the value, it doesn't need to come from a prop anymore

web/src/styles/components/buttons.scss
```diff
@@ -114,3 +114,32 @@
...
+    padding: 1.25rem 1.25rem 1.5rem 1.25rem;
+    gap: 1.125rem;
+    width: 12.5rem;
+    height: 9.6875rem;
+    border-radius: 1.25rem;
+  }
+}
+
+.landing-page-secondary-tile {
+  @extend .landing-page-tile;
```
aloverso, line 141: this isn't what I was expecting, but it accomplishes what I was raising. Well done with using @extend here, that's a neat idea

web/test/pages/onboarding/onboarding-shared.test.tsx
```diff
@@ -411,4 +413,25 @@ describe("onboarding - shared", () => {
...
     });
   });
+
+  describe("when query parameter sets onboarding flow to starting", () => {
+    beforeEach(() => {
+      jest.resetAllMocks();
+      useMockRouter({});
+      useMockUserData({});
+      useMockRoadmap({});
+      setupStatefulUserDataContext();
```
aloverso, line 423: I believe all of these functions are already being called at the top-level `beforeEach` in this file, so this `beforeEach` is redundant and could be removed

aloverso (approved):

content/src/fieldConfig/config.json
```diff
@@ -395,14 +395,16 @@
...
     "heroSupportingText": "Create your free, personalized guide to register and find funding for your New Jersey or out-of-state business.",
     "section5FirstHeaderText": "Step-by-Step Guide",
     "heroCallOutFirstLineText": "Your Business,",
     "section4ThirdIconText": "NJ Government Services",
     "section5SecondHeaderText": "Funding & Certifications",
-    "section6Header": "Looking for More Support?"
+    "section6Header": "Looking for More Support?",
+    "LandingPageTile1Text": "Get Started",
+    "LandingPageTile2Text": "I'm starting a business",
+    "LandingPageTile5Text": "How do I register my business?"
```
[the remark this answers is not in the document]
AnJuHyppolite: Thanks! I fixed the YAML file too.

web/src/styles/components/buttons.scss
AnJuHyppolite: 😄 Thank you!

PR #4749, opened 2022-11-18
feat: refactor to allow for dakota formations

mikewolfd:


api/src/api/userRouter.test.ts
```diff
@@ -390,83 +390,6 @@ describe("userRouter", () => {
           },
         });
       });
-
-      it("does not add a blank signer if the legal structure is a corp", async () => {
```
aloverso, line 394: why are we deleting these tests?

aloverso: or i guess, why are we deleting this functionality?

api/src/db/migrations/v97_updating_formation.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  v96UserDataGenerator,
+} from "./v96_added_date_field_to_tax_filing_data";
+import { migrate_v96_to_v97 } from "./v97_updating_formation_types";
+
+describe("migrate_v96_to_v97", () => {
+  beforeEach(async () => {
+    jest.resetAllMocks();
+  });
+
+  it("updates formationFormData for a limited-liability-company", () => {
```
aloverso, line 16: you have HUGE code diffs here that are honestly really hard to read through. This files is called "updating formation" and the test name is called "updates formationFormData" and is like 50 lines long.

Part of the value of tests is "tests-as-documentation" so that someone can read them and understand the behavior of the thing under test.

The way this is written, I have NO IDEA what changes have been made to formationFormData. 

One way to rewrite this in readable way would be to make use of a describe and beforeEach, for example. Something like:

describe: updating formationFormData for LLC
beforeEach: do the whole thing generate v96 and convert it to v97
it "changes x field in y way": make one assertion about one new behavior
it "changes q field in z way": same thing

Just have an it block with an assertion in it for every change. That way the tests would actually be describing what pieces of this massive object have changes to them

api/test/factories.ts
```diff
@@ -352,7 +355,7 @@ export const generateFormationData = (
   legalStructureId?: FormationLegalType
 ): FormationData => {
   return {
-    formationFormData: generateFormationFormData({}, legalStructureId),
+    formationFormData: generateFormationFormData({}, legalStructureId, {}),
```
aloverso, line 358: this doesn't look good to read. As I read this line, my brain is thinking: "What is that empty object doing in this parameter list, what does it represent?" It's a lot like having a plain true/false in a parameter list. I think that if it's gonna be there, we should have named parameters so that this says `options: {}` or maybe this options thing should just optional, actually? even better. that way it's easy to read when it IS being used, but it's not causing questions when it is empty

shared/src/formationData.ts
```diff
@@ -1,16 +1,56 @@
...
   "limited-liability-company",
   "limited-partnership",
   "c-corporation",
   "s-corporation",
 ] as const;
 
-export type FormationLegalType = typeof FormationLegalTypes[number];
+export type PublicFilingLegalType = typeof publicFilingLegalTypes[number];
+
+export const formationLegalTypes = [
```
aloverso, line 27: just a style thing, but when I have a list like this, I'd rather name it something like "allFormationLegalTypes" to make it clear it's a list of all the elements.

Otherwise you're expecting developers to know the difference between `formationLegalTypes` and `FormationLegalType` which is a very subtle naming distinction for two things that are used very differently

web/cypress/support/helpers.ts
```diff
@@ -868,22 +874,86 @@ export const generateMunicipality = (overrides: Partial<Municipality>): Municipa
...
+        return state.shortCode != "NJ";
+      })
+    ),
+    addressCountry: "US",
+    addressZipCode: randomInt(5).toString(),
+    addressMunicipality: undefined,
+    addressProvince: undefined,
+    ...overrides,
+  };
+};
```
aloverso, line 893: where/how are these being used?

web/src/components/FilingsCalendarTaxAccess.tsx
```diff
@@ -25,7 +25,7 @@ export const shouldRenderFilingsCalendarTaxAccess = (userData?: UserData) => {
     userData &&
     LookupOperatingPhaseById(userData.profileData.operatingPhase).displayTaxAccessButton &&
     LookupLegalStructureById(userData.profileData.legalStructureId).requiresPublicFiling &&
-    process.env.FEATURE_TAX_CALENDAR == "true"
+    process.env.FEATURE_BUSINESS_FLC == "true"
```
aloverso, line 28: is this intentional?

web/test/helpers-formation.tsx
```diff
@@ -76,11 +76,18 @@ export const preparePage = (
...
+  const internalMunicipalities = [
+    profileData?.municipality ?? generateMunicipality({ displayName: "GenericTown" }),
+    ...(municipalities ?? []),
+  ];
+  initialUserData.formationData.formationFormData.addressMunicipality &&
+    internalMunicipalities.push(initialUserData.formationData.formationFormData.addressMunicipality);
+  initialUserData.formationData.formationFormData.agentOfficeAddressMunicipality &&
+    internalMunicipalities.push(
+      initialUserData.formationData.formationFormData.agentOfficeAddressMunicipality
+    );
```
aloverso, line 88: what is going on with this logic?

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -686,7 +681,7 @@ describe("<BusinessFormationPaginator />", () => {
             ...initialUserData,
             formationData: {
               ...initialUserData.formationData,
-              formationFormData: generateFormationFormData({}, "limited-liability-company"),
+              formationFormData: generateFormationFormData({}, {}, "limited-liability-company"),
```
aloverso, line 684: :/

aloverso: i dunno what it means

web/src/components/tasks/business-formation/contacts/AddressModal.tsx
```diff
@@ -13,21 +13,24 @@ interface DisplayContent {
...
+  addressData: T[];
   displayContent: DisplayContent;
-  defaultAddress?: Partial<FormationAddress>;
+  defaultAddress?: Partial<T>;
   fieldName: string;
-  setData: (addressData: FormationAddress[]) => void;
+  setData: (addressData: T[]) => void;
   index?: number;
   handleClose: () => void;
+  generatorFunction: () => T;
```
aloverso, line 25: I think this should be renamed as something like "createEmptyAddress"

web/src/components/tasks/business-formation/contacts/Addresses.tsx
```diff
@@ -29,22 +23,31 @@ interface DisplayContent {
...
-  defaultAddress?: Partial<FormationAddress>;
+interface Props<T> {
+  defaultAddress?: Partial<T>;
   fieldName: FormationFields;
-  addressData: FormationAddress[];
-  setData: (addressData: FormationAddress[]) => void;
+  addressData: T[];
+  setData: (addressData: T[]) => void;
   needSignature?: boolean;
+  generatorFunction: () => T;
```
aloverso, line 32: same comment as AddressModal

api/src/api/userRouter.test.ts
mikewolfd: moved it to a simple useEffect hook in the signer component,

web/test/helpers-formation.tsx
mikewolfd: depending on the generators, there might be 3 different generatedMunicipality objects in the data, and if they're not all passed into the MunicipalitiesContext, then the tests will throw errors as the fields are being populated with invalid data

aloverso: where do the 3 different ones come from? When i refactored the municipality context, i was explicitly passing the municipality into the array, I noticed you refactored that out?

web/src/components/tasks/business-formation/contacts/Signatures.tsx
```diff
@@ -7,58 +8,75 @@ import { BusinessFormationContext } from "@/contexts/businessFormationContext";
...
 import Config from "@businessnjgovnavigator/content/fieldConfig/config.json";
-import {
-  createEmptyFormationAddress,
-  FormationAddress,
-  FormationFields,
-} from "@businessnjgovnavigator/shared";
+import { createEmptyFormationSigner, FormationFields, FormationSigner } from "@businessnjgovnavigator/shared";
 import { useMediaQuery } from "@mui/material";
 import React, { ChangeEvent, ReactElement, useContext } from "react";
+import { useMountEffect } from "../../../../lib/utils/helpers";
```
aloverso, line 14: should use @ notation

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -396,42 +396,6 @@ describe("<FilingsCalendar />", () => {
       expect(screen.queryByTestId("get-tax-access")).not.toBeInTheDocument();
     });
 
-    it("hides button on filings calendar when feature flag is disabled", () => {
```
aloverso, line 399: i agree this should be removed but in the future i think we should avoid putting disparate things in one refactor commit

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
```diff
@@ -24,10 +24,10 @@ export const ReviewSignatures = (): ReactElement => {
   return (
     <>
       <ReviewSectionHeader header={getHeader()} stepName="Contacts" testId="signature" />
-      {state.formationFormData.signers.length === 0 && (
+      {state.formationFormData.signers?.length === 0 && (
         <i>{Config.businessFormationDefaults.reviewStepNotEnteredText}</i>
       )}
-      {state.formationFormData.signers.map((signer, index) => {
+      {state.formationFormData.signers?.map((signer, index) => {
```
aloverso, line 30: why are we changing so many fields to be able to be undefined? especially a list - why not just empty list?

web/src/lib/domain-logic/allowFormation.ts
```diff
@@ -0,0 +1,34 @@
...
+    "foreign-c-corporation": process.env.FEATURE_BUSINESS_FCORP == "true",
+  };
+  if (publicFilingLegalTypes.includes(legalStructureId as PublicFilingLegalType)) {
+    return (
+      featureFlagMap[
+        castPublicFilingLegalTypeToFormationType(legalStructureId as PublicFilingLegalType, persona)
+      ] ?? true
+    );
+  }
+  return false;
```
aloverso, line 33: I think this file is a little hard to read... do you think the logic here could be simplified a bit?

If I were to describe this aloud to someone, I'd say something like:
- if Trade Name, return false always
- otherwise, if not foreign, return true always
- otherwise, return based on whether the feature flag is enabled

it seems like that could be translated more directly into code without a lot of the layers of abstraction and re-structuring and re-direction here that makes it harder for someone to look at the code and understand the logic driving it

web/test/mock/mockUseUserData.ts
```diff
@@ -27,6 +26,18 @@ export const useMockProfileData = (profileData: Partial<ProfileData>): void => {
...
+    userData: generateUserData({}),
+    update: jest.fn().mockResolvedValue({}),
+    error: undefined,
+    isLoading: false,
+    refresh: jest.fn().mockResolvedValue({}),
+    updateQueue: undefined,
+    ...overrides,
+  };
+};
+
```
aloverso, line 40: why is this moved?

mikewolfd: circular import issues

PR #4801, opened 2022-11-29
feat: [#182906423] added info alert widget to cms

FarazA22:


web/.dependency-cruiser.js
```diff
@@ -206,7 +206,18 @@ module.exports = {
...
+      from: {},
+      to: { path: "unist-util-visit" },
+    },
+    {
+      from: {},
+      to: { path: "remark-directive" },
+    },
+    {
+      from: {},
+      to: { path: "@types" },
```
aloverso, line 219: I wouldn't expect this to be necessary because you're never importing directly from the @types package... does it work without this one?

web/src/components/Content.test.tsx
```diff
@@ -49,4 +49,15 @@ describe("<Content />", () => {
...
   });
+
+  describe("infoAlert", () => {
+    it("renders info alert with header", () => {
+      const mdString =
+        ":::infoAlert{header='Header Text'}\n" + 'id: "some-id-1"\n' + 'body test"\n' + ':::"\n';
+
+      render(<Content>{mdString}</Content>);
+      expect(screen.getByText("Header Text")).toBeInTheDocument();
+      expect(screen.getByTestId("contentInfoAlert")).toBeInTheDocument();
```
aloverso, line 60: I think it might be nice to have an assertion here that expects that someone specific about the Alert component itself is being rendered.

Maybe we could set a default testId in the <Alert> component, and expect that to exist? I see in the component that you can pass one in, but maybe we could have there be a default if you don't pass one, and then look for it here

aloverso, line 60: I think it'd be good to expect the body text to exist as well

web/.dependency-cruiser.js
FarazA22: The customRemarkPlugin function within PureMarkdownContent has `  | void
  | import("unified").Transformer<import("mdast").Root, import("mdast").Root>` as the return type. When this is removed, I get the following error from the dependency check `error not-in-allowed: src/components/PureMarkdownContent.tsx → ../node_modules/@types/mdast/index.d.ts`

PR #4803, opened 2022-11-30
feat: [#183444050] complete landing page out-of-state redesign

AnJuHyppolite:


web/src/components/LandingPageTiles.tsx
```diff
@@ -20,6 +20,15 @@ export const LandingPageTiles = (): ReactElement => {
...
 
+  const setFlowToOutOfStateAndRouteToForeignBusinessTypeIds = () => {
+    routeWithQuery(router, {
+      path: ROUTES.onboarding,
+      queries: {
+        [QUERIES.flow]: "out-of-state",
+      },
+    });
+  };
+
```
aloverso, line 31: After you merge this commit, I recommend coming back here and doing some refactoring of these two functions. They're almost entirely duplicated code. I think it's a good opportunity to refactor them into a single function that can handle either scenario.

aloverso (approved):

PR #4842, opened 2022-12-05
feat: [#183855154] lambda function for database tax id encryption

somabadri:


api/src/domain/types.ts
```diff
@@ -46,6 +47,7 @@ export interface EncryptionDecryptionClient {
   decryptValue: (valueToBeDecrypted: string) => Promise<string>;
 }
 
+export type EncryptTaxId = (UserData: UserData) => Promise<UserData>;
```
aloverso, line 50: variable name should be lowercase u

api/src/domain/user/encryptTaxIdBatch.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    stubUserDataQlClient = {
+      getNeedNewsletterUsers: jest.fn(),
+      getNeedToAddToUserTestingUsers: jest.fn(),
+      getNeedTaxIdEncryptionUsers: jest.fn(),
+      search: jest.fn(),
+    };
+    encryptTaxId = encryptTaxIdFactory(stubEncryptionDecryptionClient);
+  });
+
+  it("adds all users and returns success, failed, and total count when all succeed", async () => {
```
aloverso, line 30: "adds" is not right here, it "encrypts tax id" right? and also not for all users, just those who need it

api/src/domain/user/encryptTaxIdBatch.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    stubUserDataClient.put.mockImplementation((userData: UserData): Promise<UserData> => {
+      return Promise.resolve(userData);
+    });
```
aloverso, line 33: this works but could be significantly simplified if you used `mockResolvedValue`

api/src/domain/user/encryptTaxIdFactory.ts
```diff
@@ -0,0 +1,22 @@
...
+    const user: UserData = {
+      ...userData,
+      profileData: {
+        ...userData.profileData,
+        taxId: maskedTaxId,
+        encryptedTaxId: encryptedTaxId,
+      },
+    };
+    return user;
+  };
```
aloverso, line 21: could we make use of this code in userRouter also? so that we're not duplicating code

api/src/domain/user/encryptTaxIdFactory.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  let encryptTaxId: EncryptTaxId;
+
+  beforeEach(async () => {
+    jest.resetAllMocks();
+    stubEncryptionDecryptionClient = { encryptValue: jest.fn(), decryptValue: jest.fn() };
+    encryptTaxId = encryptTaxIdFactory(stubEncryptionDecryptionClient);
+    stubEncryptionDecryptionClient.encryptValue.mockResolvedValue("some-encrypted-value");
+  });
+
+  it("updates the users database entry", async () => {
```
aloverso, line 16: not sure about this test name.... i don't see anything database related here

api/src/domain/types.ts
somabadri: It's a type right, should it still be lowercase? I was using the same notation as the `AddNewsLetter` and `AddToUserTesting` below it.

PR #4900, opened 2022-12-15
feat: [#183444029] add operating and carousel and landing-page-redesign

AnJuHyppolite:


web/src/components/LandingPageActionTile.tsx
```diff
@@ -1,13 +1,16 @@
 import { MediaQueries } from "@/lib/PageSizes";
 import { useMediaQuery } from "@mui/material";
-import { ReactElement } from "react";
+import { MutableRefObject, ReactElement } from "react";
 
 interface Props {
+  className: string;
+  aTag: string;
```
aloverso, line 7: what is this prop? it seems it's never being used

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+import analytics from "@/lib/utils/analytics";
+import { useRouter } from "next/router";
+import { ReactElement, useRef, useState } from "react";
+import Slider from "react-slick";
+import "slick-carousel/slick/slick-theme.css";
+import "slick-carousel/slick/slick.css";
+import { useMountEffect } from "../lib/utils/helpers";
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+function SampleNextArrow(props: any) {
```
aloverso, line 13: we should not be using "any" unless there really is no other choice. In this case, there's a lot of options for providing types, but the easiest is to just use the types provided by the library.

I went ahead and inspected the definition of the "Slider" element, which showed me its type definition, and the type definitions in the Settings and in the library. From there, it was easy to see that the library exports a type called "CustomArrowProps" which is exactly what you should be using here

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const { onClick } = props;
```
aloverso, line 14: seems like an extra line to destructure when we could just reference via props.onClick. Is there a reason for the destructuring?

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+  return (
+    <button onClick={onClick} className="slider-button next">
+      {">"}
```
aloverso, line 17: This should be an icon, not the angle bracket character. You can see it doesn't match the designs as-is

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+    </button>
+  );
+}
+function SamplePrevArrow(props: any) {
```
aloverso, line 21: these components are still incorrectly named with the copy-pasted "Sample"

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+    <button onClick={onClick} className="slider-button prev">
+      {"<"}
+    </button>
+  );
+}
+
+export const LandingPageSlider = (): ReactElement => {
+  const router = useRouter();
+  const { Config } = useConfig();
+  const [activeSlide, setActiveSlider] = useState(0);
```
aloverso, line 33: typo here? is this state an "active slide" or an "active slider"?

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+    arrows: true,
+    infinite: false,
+    speed: 500,
+    nextArrow:
+      activeSlide < slideBreakPoint ? (
+        <SampleNextArrow />
+      ) : (
+        <>
+          <span></span>
+        </>
```
aloverso, line 84: 1) there's no need to wrap an empty span in an empty react fragment
2) there's no need for any of these empty html tags. If you don't want the arrow to show, you can just set it to `null` or `undefined`

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const cardRef = useRef<null | HTMLButtonElement>(null);
+  const sliderRef = useRef<null | HTMLDivElement>(null);
+
+  const getNumberOfTilesToShow = (): number => {
+    const sliderWidth = sliderRef.current?.clientWidth;
+    const cardWidth = cardRef.current?.parentElement?.clientWidth;
+    if (!sliderWidth || !cardWidth) return 7;
+    const numberOfTilesToFit = Math.ceil(sliderWidth / (cardWidth + 100));
+    return 7 - numberOfTilesToFit;
+  };
```
aloverso, line 44: Do you know what a magic number is? Here's a reference: https://www.codebyamir.com/blog/software-anti-patterns-magic-numbers

You are using the number 7 in three different places here. This is a great example of a magic number. I have no idea what "7" means, why you picked 7 and not 6 or 8, what it represents, when or if it should change, or anything else about it. It should be captured in a well-named variable.

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const router = useRouter();
+  const { Config } = useConfig();
+  const [activeSlide, setActiveSlider] = useState(0);
+  const [slideBreakPoint, setSlideBreakPoint] = useState(7);
+  const cardRef = useRef<null | HTMLButtonElement>(null);
+  const sliderRef = useRef<null | HTMLDivElement>(null);
+
+  const getNumberOfTilesToShow = (): number => {
+    const sliderWidth = sliderRef.current?.clientWidth;
+    const cardWidth = cardRef.current?.parentElement?.clientWidth;
```
aloverso, line 40: I'm confused by this line - if you want the width of the parent element, why not just put the ref on the parent element instead of selecting the parent here? This is potentially risky code because if someone say, adds another wrapping div around that ref, the "parent element" would change and then this code would break

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+    if (!sliderWidth || !cardWidth) return 7;
+    const numberOfTilesToFit = Math.ceil(sliderWidth / (cardWidth + 100));
```
aloverso, line 42: 100 is another magic number - what does it represent? why are we adding it?

aloverso: I'm confused by the naming here: 

this function is called "getNumberOfTilesToShow" and this variable is called "numberOfTilesToFit". Unless "fit" does not mean the same thing as "show" (which it seems to?) then I would expect "numberOfTilesToFit" to be the value returned from this function, given these names. But it's not - you return seven minus the "numberOfTilesToFit". Does that mean that this isn't actually the number of tiles to fit? If so, what *is* it?

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const setFlowAndRouteUser = (flow: "starting" | "out-of-state" | "up-and-running") => {
+    routeWithQuery(router, {
+      path: ROUTES.onboarding,
+      queries: { [QUERIES.flow]: flow },
+    });
+  };
+
+  const settings = {
+    className: "slider variable-width",
+    slidesToShow: 1,
```
aloverso, line 71: Can you explain this to me? It's working, but how is it working when the slidesToShow is always set to 1?

web/src/components/LandingPageActionTile.tsx
```diff
@@ -33,15 +36,18 @@ export const LandingPageActionTile = (props: Props): ReactElement => {
...
-      onClick={props.onClick}
-    >
-      <img src={props.imgPath} alt="" />
-      {setTileText(props, isMobile)}
-    </button>
+    <div className={props.className}>
+      <button
+        ref={props.reference}
+        className={`${props.isPrimary ? "landing-page-tile" : "landing-page-secondary-tile"} margin-1 ${
+          isMobile ? "tal" : ""
```
aloverso, line 43: I must have missed this in previous stories:
1) You added this "tal" class to the "flex.scss" file but it is unrelated to flexbox and should have been added to something relevant like "typography.scss" because it has to do with text
2) The couple css classes we have that are three-letter abbreviations are an old pattern we're moving away from. This should have been named something descriptive like "text-align-left"
3) Regardless, neither of those matter because there is in fact already a class named "text-align-left" in the typography.scss file, so that should have been used instead of creating something new

aloverso: also - your landing-page-tile class already has a css media query on it. Why is this styling in javascript at all instead of just added into the media query in the css file?

web/src/styles/components/buttons.scss
```diff
@@ -115,9 +115,40 @@
...
+  border: none;
+  // background-color: transparent;
+
+  &.next {
+    margin-right: 10px;
+    right: 0;
+    z-index: 20;
+    opacity: 0.8;
+    // border: 2px solid red;
+  }
```
aloverso, line 138: commented code should be removed

web/src/components/LandingPageSlider.tsx
```diff
@@ -0,0 +1,154 @@
...
+      setActiveTile(next);
+    },
+    prevArrow: activeTile > 0 ? <PrevArrow /> : undefined,
+  };
+
+  return (
+    <div className="slider" ref={sliderRef} style={{ marginLeft: "25px", marginRight: "25px" }}>
+      <Slider {...settings}>
+        <LandingPageActionTile
+          // reference={cardRef}
```
aloverso, line 88: commented code should be removed... but why is this commented? don't you need this ref?

PR #4914, opened 2022-12-18
feat: [#183677756] tax calendar modal error and tax id differentiation

somabadri:


web/src/components/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -337,16 +337,34 @@ describe("<FilingsCalendarTaxAccess />", () => {
...
+      expect(screen.getByRole("alert")).toHaveTextContent(Config.taxCalendar.failedErrorMessageHeader);
+      expect(screen.getByRole("alert")).toHaveTextContent(Config.taxCalendar.modalBusinessFieldErrorName);
       expect(screen.getByText(Config.taxCalendar.failedTaxIdHelper)).toBeInTheDocument();
       expect(screen.getByText(Config.taxCalendar.failedBusinessFieldHelper)).toBeInTheDocument();
       expect(
         screen.queryByText(Config.taxCalendar.failedResponsibleOwnerFieldHelper)
       ).not.toBeInTheDocument();
     });
 
+    it("states business name specific field error in alert if api fails", async () => {
```
aloverso, line 349: I see in your code that you have logic that changes the error message shown depending on whether it needs to show responsible owner name or business name. That part isn't tested for the Responsible Owner Name because this test only exists for the Public Filing case. You should have a corresponding test to cover the rest of the logic you coded.

web/src/components/TaxFilingLookupModal.tsx
```diff
@@ -73,6 +73,37 @@ export const TaxFilingLookupModal = (props: Props): ReactElement => {
...
   };
 
+  const responsibleOwnerOrBusinessName = () => {
+    return LookupLegalStructureById(profileData.legalStructureId).requiresPublicFiling
+      ? Config.taxCalendar.modalBusinessFieldErrorName
+      : Config.taxCalendar.modalResponsibleOwnerFieldErrorName;
+  };
+
+  const errorHeader = (): ReactElement => {
+    if (userData?.taxFilingData.errorField == "Business Name" && apiFailed == "FAILED") {
```
aloverso, line 83: Maybe this could be a refactor after this story is merged, because I didn't realize this behavior already exists, but this code should be refactored. 

The field value "Business Name" is completely coupled to the specific implementation that the Tax Filing API uses to refer to this field. If their code changes, then our frontend React code would have to be changed to account for the error. This very much breaks dependency inversion principle, which is that only the adapter component (in our case, the ApiTaxFilingClient) should know anything or depend on any logic that is specific to the dependency.

Think about it like this: at any point, NICUSA could change the value of this field from "Business Name" to "Tax Business Name" or to "business name" or to anything else.

As soon as they do, our code breaks. And to fix it, you'd have to go into our _frontend react code_ and change not only a Modal component that handles display logic, but also a test. That's very, very brittle code. The more that *their* direct domain model of what specific string they use to represent an error proliferates *our* frontend code, the more broken we would wind up if _anything_ ever changes.

The TaxFilingClient should be responsible for transforming the API-specific error representation into something *we* control and define. It's okay for the frontend to be dependent on _our_ representation of an error. And then, if anything ever changes, then the only code we need to change is to modify the TaxFilingClient to be able to interpret the new value into our domain model. And our frontend code would stay the same.

Not only is this good coding practice for us, it's good coding practice in general. Just imagine we had, like, an iOS frontend app and an Android frontend app as well. Imagine how many things you'd have to change if all three of those frontends depended on this API-specific logic. It's a good heuristic to keep in mind when thinking about dependencies between backend and frontend.

aloverso: cc @mikewolfd something to flag for your attention also

web/src/components/FilingsCalendarTaxAccess.test.tsx
somabadri: yeah that makes sense, I forgot to replicate the test for the parallel scenario. I will fix that.

PR #4921, opened 2022-12-21
feat: nexus formation flow

mikewolfd:


content/src/display-content/business-formation/nexus/dba-resolution-foreign.md
```diff
@@ -0,0 +1,18 @@
+---
+urlSlug: DBA-resolution
+filename: dba-resolution-foreign
+name: Get a Resolution Adopting Your DBA Name
+id: DBA-resolution-foreign
+---
```
aloverso, line 6: first thought - why does this have all this grey matter here? I tested the app and the url doesn't change across each step of the DBA flow. Why do we defined a url slug and other stuff here?

second thought, more for longer-term thought, but I thought we were trying to move away from "display-content" markdowns and storing this stuff in just a string in the config JSONs anyways. It makes editing in the CMS easier and also makes our lives way easier in coding because we don't have to static import the displayContent, instead we can just import the config directly. I know we still have a few left, but it's worth thinking about not increasing the stack of what is essentially debt that needs to be turned into a config string,

content/src/fieldConfig/nexus-dba-formation.json
```diff
@@ -1,5 +1,5 @@
 {
-  "nexusFormationTask": {
+  "DbaFormationTask": {
```
aloverso, line 2: we should start with lowercase letters for config keys

content/src/roadmaps/add-ons/public-record-filing.json
```diff
@@ -1,14 +1,9 @@
...
-    {
-      "step": 2,
-      "weight": 19,
-      "task": "search-business-name"
-    },
     {
       "step": 2,
       "weight": 21,
-      "task": "form-business-entity"
+      "task": "form-business-entity-nj"
```
aloverso, line 6: I feel myself getting very confused with the quantity of "form business entity" tasks and what they all do... it feels like it needs a high-level assessment at some point

web/src/components/tasks/business-formation/BusinessFormation.tsx
```diff
@@ -215,6 +216,7 @@ export const BusinessFormation = (props: Props): ReactElement => {
       </div>
     );
   }
+  const isForeign = userData?.profileData.businessPersona == "FOREIGN";
```
aloverso, line 219: we could use this variable in the if-statement above to simplify the code and not do the same check twice

web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+  const onPreviousButtonClick = (): void => {
+    setStepperState((stepperState) => {
+      stepperState[1].isComplete = state.stepIndex != 2;
+      return stepperState;
+    });
+    moveToStep(state.stepIndex - 1);
+  };
+
+  const isNotDba =
```
aloverso, line 60: it took me a solid moment to grasp why we would be checking if it "isNotDba" in the "DbaFormationPaginator"... I think the component should be renamed to "Nexus" or something, right?

web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const onMoveToStep = async (
+    stepIndex: number,
+    config: { moveType: "NEXT_BUTTON" | "STEPPER" }
+  ): Promise<void> => {
+    onStepChangeAnalytics(userData?.formationData.formationFormData, stepIndex, config.moveType);
+    if (
+      config.moveType == "NEXT_BUTTON" &&
+      stepIndex == 1 &&
+      isNotDba &&
+      isAuthenticated === IsAuthenticated.FALSE
```
aloverso, line 72: why do we need all these checks? Shouldn't it just be, if they're not authenticated, we need to show the modal?

web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
```diff
@@ -0,0 +1,215 @@
...
+      return (
+        <TaskCTA
+          link={state.dbaContent.Authorize.callToActionLink}
+          text={state.dbaContent.Authorize.callToActionText}
+          onClick={() => {
+            return setShowCtaModal(true);
+          }}
+        >
+          <BackButton />
+        </TaskCTA>
```
aloverso, line 173: I think it feels weird to use the TaskCTA here... it has a specific purpose as a general Task component and it feels like putting it in a "special" component like this... would it be hard to just make a new button?

web/src/components/tasks/business-formation/NexusFormationStepsConfiguration.ts
```diff
@@ -0,0 +1,23 @@
+import { DbaStepNames } from "@/lib/types/types";
+
+export const NexusFormationStepsConfiguration: {
+  name: DbaStepNames;
+  stepIndex: number;
+  disableStepper?: boolean;
```
aloverso, line 6: why is this false or undefined? shouldn't the other steps just have disableStepper = false?

web/src/components/tasks/business-formation/billing/BillingStep.test.tsx
```diff
@@ -15,6 +16,7 @@ import {
 } from "@businessnjgovnavigator/shared";
 import * as materialUi from "@mui/material";
 import { fireEvent, screen } from "@testing-library/react";
+import { generateFormationDbaContent } from "../../../../../test/factories";
```
aloverso, line 19: relative import 👎

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -47,11 +50,19 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
     searchBusinessName,
     updateNameOnProfile,
     resetSearch,
-  } = useBusinessNameSearch({ isBusinessFormation: false, isDba: props.isDba || false });
+  } = useBusinessNameSearch({
+    isBusinessFormation: !!props.isBusinessFormation,
+    isDba: props.isDba || false,
```
aloverso, line 55: two things: 
1) why take `isBusinessFormation` - is now ever a time when we are doing a business name search that is NOT formation?
2) I think this is an example of the double-bool thingy I was mentioning after the tech talk. We use 2 bools here (4 states) to represent a 3-state situation. There's no way for isDba to be true when isBusinessFormation is false now, right? Our type model should reflect that

web/src/lib/static/loadTasks.ts
```diff
@@ -29,7 +29,7 @@ export const loadTaskByUrlSlug = (urlSlug: string): Task => {
   return loadTaskByFileName(matchingFileName);
 };
 
-const loadTaskByFileName = (fileName: string): Task => {
+export const loadTaskByFileName = (fileName: string): Task => {
```
aloverso, line 32: why are we exporting this? I don't see it being used anywhere? i could be missing it?

web/src/lib/types/types.ts
```diff
@@ -312,8 +336,14 @@ export type FormationDisplayContentMap = Record<FormationLegalType, FormationDis
 
 export interface FormationSignedAddress extends FormationMember, Partial<Omit<FormationSigner, "name">> {}
 
+export type FormationDbaContent = {
+  DbaResolution: TaskWithoutLinks;
+  Authorize: TaskWithoutLinks;
+  Formation: TaskWithoutLinks;
```
aloverso, line 342: typically variables are lowercase

web/src/pages/tasks/[taskUrlSlug].tsx
```diff
@@ -121,17 +118,7 @@ const TaskPage = (props: Props): ReactElement => {
...
-      return <NexusFormationTask task={taskInRoadmap} />;
-    } else {
-      return (
-        <BusinessFormation
-          task={taskInRoadmap}
-          displayContent={props.displayContent.formationDisplayContent}
-        />
-      );
-    }
+    return <BusinessFormation task={taskInRoadmap} displayContent={props.displayContent} />;
```
Comment (aloverso) [inline web/src/pages/tasks/[taskUrlSlug].tsx:121]: Could we just keep these completely separate? Why combine them all into one Business Formation task that splits off into 2 different flows?

web/test/helpers/helpers-formation.tsx
```diff
@@ -201,7 +221,7 @@ export const createFormationPageHelpers = (): FormationPageHelpers => {
     });
   };
 
-  const getStepStateInStepper = (index: number): string => {
+  const getStepStateInStepper = (index: number | undefined): string => {
```
aloverso, line 224: why this change? Why would index be undefined, and if index *is* undefined, why would we be requesting its step state?

PR #4922, opened 2022-12-21
Essential question required

FarazA22:


web/test/factories.ts
```diff
@@ -129,28 +129,34 @@ export const generateTaxFiling = (overrides: Partial<TaxFiling>): TaxFiling => {
...
+    homeBasedBusiness: randomElementFromArray([true, false]),
+    cannabisLicenseType: randomElementFromArray(["CONDITIONAL", "ANNUAL"]),
+    cannabisMicrobusiness: randomElementFromArray([true, false]),
+    constructionRenovationPlan: randomElementFromArray([true, false]),
+    providesStaffingService: randomElementFromArray([true, false]),
+    certifiedInteriorDesigner: randomElementFromArray([true, false]),
+    realEstateAppraisalManagement: randomElementFromArray([true, false]),
+    carService: randomElementFromArray(["STANDARD", "HIGH_CAPACITY", "BOTH"]),
+    interstateTransport: randomElementFromArray([true, false]),
+    isChildcareForSixOrMore: randomElementFromArray([true, false]),
```
aloverso, line 146: I think that for booleans, something like `!(randomInt() % 2)` like we do in other generators might be a little less verbose

PR #4969, opened 2023-01-03
feat: [#183901632] update calendar past and current months highlight

AnJuHyppolite:


web/src/components/FilingsCalendar.tsx
```diff
@@ -125,7 +126,13 @@ export const FilingsCalendar = (props: Props): ReactElement => {
             return (
               <tr key={rowIndex}>
                 {monthIndicesForRow.map((month) => {
-                  return (
+                  const date = getJanOfCurrentYear().add(month, "months");
+
+                  return date.month() < getCurrentDate().month() ? (
+                    <td key={month} className="td-gray-border bg-base-extra-light">
+                      {getMonth(month)}
+                    </td>
```
aloverso, line 134: I think that instead of rewriting the entire `td` in this ternary just to add one class name, it would be cleaner to refactor this to use a dynamic className. Then you don't need to duplicate the `td` code and you don't need a ternary 

I went searching for an example, and there's a lot of places. For example, line 67 of `PaymentTypeTable`: 
```
<div className={`${hasError ? "error" : ""} input-error-bar`}>
```

web/src/components/njwds-extended/Tag.tsx
```diff
@@ -54,6 +55,9 @@ export const Tag = (props: Props): ReactElement => {
     case "accent-cooler-lightest":
       styling = "bg-accent-cooler-lightest text-base-darkest border border-accent-cooler-light";
       break;
+    case "warning-light":
+      styling = "bg-warning-light";
+      break;
```
aloverso, line 60: I think you need to double-check the colors? I'm seeing a bright yellow with white text: 
image

web/src/components/FilingsCalendar.tsx
```diff
@@ -75,34 +75,35 @@ export const FilingsCalendar = (props: Props): ReactElement => {
...
         <div>
-          {thisMonthFilings
-            .filter((filing) => {
-              return props.operateReferences[filing.identifier];
-            })
-            .map((filing) => {
-              return (
-                <div key={filing.identifier} className="line-height-1 margin-bottom-1" data-testid="filing">
-                  <Tag backgroundColor="warning-extra-light" isHover isRadiusMd isWrappingText>
+          {date.month() >= getCurrentDate().month() &&
```
aloverso, line 78: I'm not sure I understand why this change was made? What does it do?

web/src/components/FilingsCalendar.tsx
```diff
@@ -125,7 +126,13 @@ export const FilingsCalendar = (props: Props): ReactElement => {
             return (
               <tr key={rowIndex}>
                 {monthIndicesForRow.map((month) => {
-                  return (
+                  const date = getJanOfCurrentYear().add(month, "months");
```
aloverso, line 129: I don't think this line of code does anything. You're using the variable `month` to create a new variable `date`, so that you can make use of `date.month()` on the next line.

Right after this line, do a console.log of `month` and of `date.month()` and you should see what I mean

shared/src/dateHelpers.ts
```diff
@@ -11,7 +11,7 @@ export const getCurrentDate = (): Dayjs => {
 };
 
 export const getJanOfCurrentYear = (): Dayjs => {
-  return dayjs().month(0);
+  return dayjs().month(0).date(1);
```
aloverso, line 14: Why is this addition necessary?

web/src/components/FilingsCalendar.tsx
AnJuHyppolite: It will prevent past events from showing up. I have now created a function, isCalendarMonthLessThanCurrentMonth, and calling !isCalendarMonthLessThanCurrentMonth() here and using isCalendarMonthLessThanCurrentMonth for the conditional <td>.

web/src/components/FilingsCalendar.tsx
AnJuHyppolite: I have now created a function, isCalendarMonthLessThanCurrentMonth, and set the className to  that function with a ternary to conditionally render the gray background.

shared/src/dateHelpers.ts
AnJuHyppolite: We added it b/c Faraz was wondering about specifically-dated events. I have updated it to return dayjs().startOf('year') so that it resets to January 1st of the specific year.

**Source:** https://day.js.org/docs/en/manipulate/start-of#docsNav

aloverso (approved):

PR #5005, opened 2023-01-05
feat: [#183982666] add tax id disclaimer for trade name legal structure

AnJuHyppolite:


web/src/components/TaxRegistrationModal.tsx
```diff
@@ -90,9 +90,9 @@ export const TaxRegistrationModal = (props: Props): ReactElement => {
     );
   };
 
-  const showTaxIdField = (): boolean => {
-    return LookupLegalStructureById(userData?.profileData.legalStructureId).requiresPublicFiling;
-  };
+  // const showTaxIdField = (): boolean => {
+  //   return LookupLegalStructureById(userData?.profileData.legalStructureId).requiresPublicFiling;
+  // };
```
aloverso, line 95: commented code

web/src/components/TaxRegistrationModal.tsx
```diff
@@ -131,12 +131,12 @@ export const TaxRegistrationModal = (props: Props): ReactElement => {
...
         )}
-        {showTaxIdField() && (
-          <>
-            <FieldLabelModal fieldName="taxId" />
-            <OnboardingTaxId onValidation={onValidation} fieldStates={fieldStates} required />
-          </>
-        )}
+        <>
+          <FieldLabelModal fieldName="taxId" />
+          <br />
```
aloverso, line 136: In this place (and the others as well), we should use margins to control spacing rather than inserting an empty line

web/src/components/TaxRegistrationModal.tsx
```diff
@@ -56,7 +56,7 @@ export const TaxRegistrationModal = (props: Props): ReactElement => {
         invalid: !profileData.existingEmployees,
       },
       taxId: {
-        invalid: profileData.taxId?.length !== 12 && showTaxIdField(),
+        invalid: profileData.taxId?.length !== 12,
```
aloverso, line 59: Given that we're changing this logic, I think the test should change to reflect this new behavior

web/src/pages/profile.tsx
```diff
@@ -535,6 +536,8 @@ const ProfilePage = (props: Props): ReactElement => {
         />
         <hr className="margin-top-4 margin-bottom-2" aria-hidden={true} />
         <FieldLabelProfile fieldName="taxId" />
+        <br />
+        <Content>{Config.profileDefaults.fields.taxId.default.disclaimerMd}</Content>
```
aloverso, line 540: In this and all the places where it changed, there is missing logic here. The disclaimer should only show for Trade Name businesses

content/src/fieldConfig/profile.json
```diff
@@ -164,7 +164,8 @@
           "header": "New Jersey Tax ID",
           "placeholder": "Enter New Jersey Tax ID number",
           "description": "What is your `New Jersey Tax ID?|tax-id`",
-          "errorTextRequired": "Enter a 12-digit NJ Tax ID"
+          "errorTextRequired": "Enter the 12-digit Tax ID shown in your tax confirmation email.",
```
aloverso, line 167: I think this is a rebase issue. The old text "Enter a 12-digit NJ Tax ID" is what should be here

web/src/components/tasks/TaxTask.test.tsx
```diff
@@ -204,4 +205,54 @@ describe("<TaxTask />", () => {
...
+  return tradeNameLegalStructures[randomIndex].id;
+};
+
+const randomPublicFilingLegalStructure = () => {
+  const nonTradeNameLegalStructures = LegalStructures.filter((x) => {
+    return x.requiresPublicFiling;
+  });
+  const randomIndex = Math.floor(Math.random() * nonTradeNameLegalStructures.length);
+  return nonTradeNameLegalStructures[randomIndex].id;
+};
```
aloverso, line 258: I think these functions already exist in helpers/support, we shouldn't need to recreate them

web/test/pages/profile.test.tsx
```diff
@@ -2026,4 +2056,20 @@ describe("profile", () => {
...
+    return tradeNameLegalStructures[randomIndex].id;
+  };
+
+  const randomPublicFilingLegalStructure = () => {
+    const nonTradeNameLegalStructures = LegalStructures.filter((x) => {
+      return x.requiresPublicFiling;
+    });
+    const randomIndex = Math.floor(Math.random() * nonTradeNameLegalStructures.length);
+    return nonTradeNameLegalStructures[randomIndex].id;
+  };
```
aloverso, line 2074: Same comment as above - we should be able to import the existing helper functions here instead of recreating

PR #5011, opened 2023-01-06
feat: [#183901632] update calendar past months logic in list view

AnJuHyppolite:


web/src/components/FilingsCalendar.test.tsx
```diff
@@ -476,6 +476,51 @@ describe("<FilingsCalendar />", () => {
...
+          ).id,
+        }),
+        taxFilingData: generateTaxFilingData({ filings: [pastReport, futureReport] }),
+      });
+
+      const operateReferences: Record<string, OperateReference> = {
+        past: {
+          name: "Past Report",
+          urlSlug: "past-report-url",
+          urlPath: "past_report-url-path",
```
aloverso, line 508: if these values aren't being used in the test, we should use a generator instead of writing them out explicitly. That way, it's clear to a reader of the test that it's not important/relevant

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -476,6 +476,51 @@ describe("<FilingsCalendar />", () => {
...
     });
+
+    it("doesn't display past months in list view", () => {
+      const pastDueDate = getCurrentDate().subtract(1, "months");
+      const pastReport = generateTaxFiling({
+        identifier: "past",
+        dueDate: pastDueDate.format(defaultDateFormat),
+      });
+
+      const futureDueDate = getCurrentDate().add(2, "months");
```
aloverso, line 487: This works, and knowing what we know about upcoming stories, I don't think it needs to be changed. That said, I just want to point out that this is a dangerous test because if it was November or December, this test would suddenly start failing out of nowhere. Because our current logic is to only show events in the future but still in the calendar year. If the current date is November, then 2 months out is January - a different year, so it wouldn't show up. 

I wanted to point this out so that you have it in your mind in general when writing date tests. Again, doesn't need to be changed since we know this behavior is changing so soon. But if I were to change this, I would probably explicitly pick a date I know to be in the future but in this year: `futureDueDate = dayjs(new Date(getCurrentDate.year(), 12, 31))`

web/src/lib/domain-logic/filterFilings.ts
```diff
@@ -15,12 +15,16 @@ export const sortFilingsEarliestToLatest = (filings: TaxFiling[]): TaxFiling[] =
   });
 };
 
-export const deadlinesWithinAYear = (filings: TaxFiling[]): TaxFiling[] => {
+export const currentAndFutureDeadlines = (filings: TaxFiling[]): TaxFiling[] => {
```
aloverso, line 18: it's still within a year, though, I might add that to the function name

aloverso: also, "Upcoming" might be a good word to use instead of "currentAndFuture" to make this function name a little shorter and easier to read (as well as the other renamed variables like "sortedFilteredFilingsCurrentAndFutureMonths")

PR #5073, opened 2023-01-14
feat: [#183874527] split tax id fields on register for taxes task

somabadri:


web/src/styles/base/global.scss
```diff
@@ -357,3 +365,7 @@ ol {
 .no-wrap {
   white-space: nowrap;
 }
+
+.width-38rem {
```
mikewolfd, line 369: the class name seems unintuitive, maybe rename it to include max

mikewolfd (approved):

PR #5105, opened 2023-01-20
feat: [#183934860] add calendar list view as default

AnJuHyppolite:


web/src/components/FilingsCalendar.test.tsx
```diff
@@ -51,13 +52,19 @@ jest.mock("@/lib/data-hooks/useRoadmap", () => {
 jest.mock("next/router", () => {
   return { useRouter: jest.fn() };
 });
-
+jest.mock("@/lib/api-client/apiClient", () => ({
+  postTaxFilingsOnboarding: jest.fn(),
+  postTaxFilingsLookup: jest.fn(),
+}));
```
aloverso, line 58: why are we mocking these here? It doesn't appear to be used anywhere

web/src/components/FilingsCalendar.tsx
```diff
@@ -196,7 +196,12 @@ export const FilingsCalendar = (props: Props): ReactElement => {
     const type = LookupOperatingPhaseById(userData?.profileData.operatingPhase).displayCalendarType;
     if (type === "LIST") return renderCalendarAsList();
     if (type === "FULL") {
-      if (isLargeScreen && userData?.preferences.isCalendarFullView) return renderCalendarAsGrid();
+      if (
+        isLargeScreen &&
+        userData?.preferences.isCalendarFullView &&
+        userData.taxFilingData.state === "SUCCESS"
+      )
+        return renderCalendarAsGrid();
```
aloverso, line 204: @AnJuHyppolite the logic added here is not correct, I think you might need to start this one over. The story is asking us to display List View *by default* if they do not yet have a success on the tax access, and Grid View *by default* if they have a successfully accessed the tax calendar.

"By default" means the initial state the user sees. The way this is written right now, it is showing ONLY List View pre-tax-access without allowing the user to switch to List View. This story is about changing the default view ... but the user is still allowed to change their view as desired.

web/src/components/FilingsCalendar.test.tsx
AnJuHyppolite: We need it to fix the **No User Pool in the configuration** error.

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -51,13 +52,19 @@ jest.mock("@/lib/data-hooks/useRoadmap", () => {
...
 const Config = getMergedConfig();
+const mockApi = api as jest.Mocked<typeof api>;
 
 const renderFilingsCalendar = (
   operateReferences: Record<string, OperateReference>,
   initialUserData?: UserData
 ) => {
+  mockApi.postTaxFilingsLookup.mockResolvedValue(initialUserData ?? generateUserData({}));
+  mockApi.postTaxFilingsOnboarding.mockResolvedValue(initialUserData ?? generateUserData({}));
```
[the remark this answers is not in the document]
AnJuHyppolite: We need it to fix the **No User Pool in the configuration** error.

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -648,5 +666,29 @@ describe("<FilingsCalendar />", () => {
...
+        ...userData,
+        preferences: { ...userData.preferences, isCalendarFullView: false },
+        taxFilingData: { ...userData.taxFilingData, state: "PENDING" },
+      };
+
+      renderFilingsCalendar(operateReferences, userData);
+
+      expect(screen.getByText(Config.dashboardDefaults.calendarGridViewButton)).toBeInTheDocument();
+    });
   });
```
aloverso, line 693: Why has this test file changed if none of the code for the FilingsCalendar is changed? I think that all these changes can be removed, right?

api/src/domain/tax-filings/taxFilingsInterfaceFactory.test.ts
```diff
@@ -202,6 +211,10 @@ describe("TaxFilingsInterfaceFactory", () => {
         } as TaxFilingLookupResponse);
         expect(await taxFilingInterface.onboarding({ userData, ...taxIdBusinessName })).toEqual({
           ...userData,
+          preferences: generatePreferences({
+            ...userData.preferences,
+            isCalendarFullView: true,
+          }),
```
aloverso, line 217: We've discussed tests-as-documentation before. When we add new behavior like this, there should be test(s) explicitly describing the new behavior so that someone can read the plain english and understand what the intention is. In this case, there are 3 new behaviors, so I'd expect 3 new tests that cover each case:
 a. when the lookup returns success, the isCalendarFullView gets set to true 
b. when the lookup is anything but success, and the previous value of isCalendarFullView was false, it should stay false
c. when the lookup is anything but success, and the previous value of isCalendarFullView was true, it should stay true

All these are currently tested *implicitly* in the tests you've changed, but it should **also** be described *explicitly* in its own test

PR #5122, opened 2023-01-23
fix: [#184258348] business name specific error on tax filing lookup m…

somabadri:


api/src/client/ApiTaxFilingClient.test.ts
```diff
@@ -96,15 +96,15 @@ const generateSuccessfulApiTaxFilingOnboardingResponse = (
...
+) => {
   return {
     ApiKey: `some-ApiKey-${randomInt()}`,
     Errors:
       statusCode == 400
         ? [
             {
               Error: `some-error-content-${randomInt()}`,
-              Field: randomInt() % 2 === 0 ? "formFailure" : "businessName",
+              Field: randomInt() % 2 === 0 ? "Taxpayer ID" : "Business Name",
```
aloverso, line 107: What's the purpose of this change? I see you've removed the return type on this function now. It appears the type no longer matches "ApiTaxFilingOnboardingResponse" - is that correct? If so, what IS its type? Should this function really be called "generateErroredApiTaxFilingOnboardingResponse" if that's not its return type?

api/src/client/ApiTaxFilingClient.test.ts
```diff
@@ -277,7 +277,14 @@ describe("ApiTaxFilingClient", () => {
...
         taxIdAndBusinessNameAndEmail.businessName
       );
-      expect(response).toEqual({ state: "FAILED", errorField: stubResponse.Errors[0].Field });
+      let errorField;
+      if (stubResponse.Errors[0].Field == "Taxpayer ID") {
+        errorField = "formFailure";
+      }
+      if (stubResponse.Errors[0].Field == "Business Name") {
+        errorField = "businessName";
+      }
```
aloverso, line 286: I think that what's going on here, if I am reading it correctly, is that we're generating a stub response with a random error field that has one of two values. And then here, we're setting up our expectation based on which of those two random values it is - if it's X, then we expect A, if it's Y, then we expect B.

If that's right - this kind of logic shouldn't really exist in a test. We're kind of just duplicating the logic of the code we're testing. If you want to test both of these cases (which I assume you do) then we should have 2 tests to cover both outcomes. In each test, set one value explicitly in the stub response, and then explicitly expect what you should receive back in the response based on that input.

api/src/domain/tax-filings/taxFilingsInterfaceFactory.test.ts
```diff
@@ -140,83 +140,81 @@ describe("TaxFilingsInterfaceFactory", () => {
...
+        const filingData = generateTaxFiling({});
+        taxFilingClient.onboarding.mockResolvedValue({
+          state: "SUCCESS",
+          filings: [filingData],
+        } as TaxFilingOnboardingResponse);
         taxFilingClient.lookup.mockResolvedValue({
-          state: "API_ERROR",
-          filings: [],
+          state: "SUCCESS",
+          filings: [filingData],
```
aloverso, line 151: I think that because these two calls have the same mocked response value, it doesn't feel like the test is actually testing that it "returns the lookup response"

Like, if Lookup returns ObjectA, and Onboarding also returns the same ObjectA - if the expectation is that we get back ObjectA, it's not actually validating if that's the thing that was returned from Lookup or from Onboarding. If your code was written such that it just passed through the response from the onboarding success without ever calling lookup, I think this test would still pass. It's not validating that the thing that's getting returned is specifically the return from the lookup call.

api/src/domain/tax-filings/taxFilingsInterfaceFactory.test.ts
```diff
@@ -140,83 +140,81 @@ describe("TaxFilingsInterfaceFactory", () => {
...
-      });
-
-      it("returns lookup response when lookup returns PENDING", async () => {
-        taxFilingClient.lookup.mockResolvedValue({
-          state: "PENDING",
-          filings: [],
-        } as TaxFilingLookupResponse);
+        taxFilingClient.onboarding.mockResolvedValue({
+          state: "API_ERROR",
+          filings: [filingData],
```
aloverso, line 179: is this correct? Does an API_ERROR response also come with tax filing data?

api/src/domain/tax-filings/taxFilingsInterfaceFactory.test.ts
```diff
@@ -140,83 +140,81 @@ describe("TaxFilingsInterfaceFactory", () => {
...
+        const filingData = generateTaxFiling({});
+        const userData = generateUserData({
+          taxFilingData: generateTaxFilingData({
+            state: undefined,
+            lastUpdatedISO: undefined,
             registeredISO: undefined,
-            businessName: taxIdBusinessName.businessName,
-            lastUpdatedISO: currentDate.toISOString(),
-          },
+            filings: [filingData],
```
aloverso, line 174: why are we starting with filingData for the user before the onboarding/lookup happens? Would this state ever actually exist?

aloverso: same comment on all these

api/src/domain/tax-filings/taxFilingsInterfaceFactory.ts
```diff
@@ -23,36 +23,54 @@ export const taxFilingsInterfaceFactory = (apiTaxFilingClient: TaxFilingClient):
...
-          registeredISO:
-            userData.taxFilingData.registeredISO ??
-            (registered ? new Date(Date.now()).toISOString() : undefined),
-        },
-      };
+    const response = await apiTaxFilingClient.onboarding(
+      request.taxId,
+      request.userData.user.email,
+      request.businessName
+    );
```
aloverso, line 30: Not something that you need to do for this story, but this function taking 3 string types as a parameter list is a code smell. If any of them wind up out of order, the whole thing won't work and we wouldn't even get a type failure. This is a great candidate for a Named Parameters refactoring

api/src/domain/tax-filings/taxFilingsInterfaceFactory.ts
```diff
@@ -23,36 +23,54 @@ export const taxFilingsInterfaceFactory = (apiTaxFilingClient: TaxFilingClient):
...
+            state: response.state,
+            businessName: request.businessName,
+          },
+        };
+      }
+      case "SUCCESS": {
+        return await lookup(request);
+      }
+
+      default: {
```
aloverso, line 62: I notice this is the same as the "FAILED" case, right? and also that API_ERROR case doesn't have its own specific case, so I assume it's meant to be covered by the default?

A cleaner way to write this could be to use the fall-through feature of switch statements to group shared behavior:

```
case "API_ERROR":
case "FAILED":
default:
  return ... 
```

This a) groups shared behavior to reduce duplication and b) makes it clear all the possible cases we know about, so that someone reading it doesn't think that API_ERROR has just been forgotten. Instead, it acknowledges that its a known case, and explicitly states its outcome

PR #5129, opened 2023-01-25
feat: [#183901788] added view more button to calendar list view

adidner:


web/src/components/FilingsCalendar.tsx
```diff
@@ -96,6 +107,20 @@ export const FilingsCalendar = (props: Props): ReactElement => {
...
+            onClick={() => setNumberOfVisibleCalendarEntries((previous) => previous + listViewMoreIncrement)}
+            onKeyDown={() =>
+              setNumberOfVisibleCalendarEntries((previous) => previous + listViewMoreIncrement)
+            }
+            role="button"
+            tabIndex={0}
+            aria-label={`View ${listViewMoreIncrement} More Calendar Entries`}
+          >
+            View More
+          </div>
```
aloverso, line 122: You don't need to recreate all this from scratch (Same with your `view-more-button` css). For one, for semantic html we want to use <button> components instead of div role=button. 

But we keep all the button types in our `Button` component. You can achieve the styling you're looking for with `<Button style="tertiary">

web/src/components/FilingsCalendar.tsx
```diff
@@ -96,6 +107,20 @@ export const FilingsCalendar = (props: Props): ReactElement => {
...
+            className={"view-more-button"}
+            onClick={() => setNumberOfVisibleCalendarEntries((previous) => previous + listViewMoreIncrement)}
+            onKeyDown={() =>
+              setNumberOfVisibleCalendarEntries((previous) => previous + listViewMoreIncrement)
+            }
+            role="button"
+            tabIndex={0}
+            aria-label={`View ${listViewMoreIncrement} More Calendar Entries`}
+          >
+            View More
```
aloverso, line 121: All plaintext in the application is configured so that it can be externally edited in the CMS. We can pair and I can show you how to add this to the config for CMS. but as a rule of thumb, there should never be plaintext words in the JSX

web/src/components/FilingsCalendar.tsx
```diff
@@ -59,16 +62,24 @@ export const FilingsCalendar = (props: Props): ReactElement => {
       return <></>;
     }
 
+    const sortedFilteredOperatingFilings = groupBy(
```
aloverso, line 65: I wonder if we could have a more descriptive variable name here. I think the groupby is confusing (I'm glad you pulled it out and named it so that we have the opportunity to clear up that confusion and name it something descriptive)

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -649,4 +649,58 @@ describe("<FilingsCalendar />", () => {
...
+        }),
+        taxFilingData: generateTaxFilingData({ filings: filings }),
+        preferences: generatePreferences({ isCalendarFullView: false }),
+      });
+
+      renderFilingsCalendar(operateReferences, userData);
+    }
+
+    it("does not show the view more button if 5 or less events are in the calendar", () => {
+      renderCalendarWithEntries(3);
```
aloverso, line 686: with a test like this, i'd usually test the edge of the edge case (ie, 5 filings here instead of 3)

same with the one below on the other side of the edge - I'd use 6 filings instead of 7. This forces the tests to confirm that yes, it is in fact 5 and nothing else. I think these tests would still pass if the numberToShow was set to 4 or 6 in the code, and we don't want that

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -649,4 +649,58 @@ describe("<FilingsCalendar />", () => {
...
+      expect(screen.getByText("View More")).toBeInTheDocument();
+    });
+
+    it("shows 5 more events when the view more button is clicked", () => {
+      renderCalendarWithEntries(12);
+      expect(screen.getAllByTestId("calendar-list-entry")).toHaveLength(5);
+      fireEvent.click(screen.getByText("View More"));
+      expect(screen.getAllByTestId("calendar-list-entry")).toHaveLength(10);
+      fireEvent.click(screen.getByText("View More"));
+      expect(screen.getAllByTestId("calendar-list-entry")).toHaveLength(12);
```
aloverso, line 703: I think there's a missing final expectation where the View More button goes away when there are no more events to show

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -649,4 +649,58 @@ describe("<FilingsCalendar />", () => {
...
+        preferences: generatePreferences({ isCalendarFullView: false }),
+      });
+
+      renderFilingsCalendar(operateReferences, userData);
+    }
+
+    it("does not show the view more button if 5 or less events are in the calendar", () => {
+      renderCalendarWithEntries(3);
+      expect(screen.getAllByTestId("calendar-list-entry")).toHaveLength(3);
+      expect(screen.queryByText("View More")).not.toBeInTheDocument();
```
aloverso, line 688: same comment as in the code - with externalized content, we don't want to hardcode any labels into the tests, because if content changes the label, we wouldn't want the tests to start failing. So the tests should reference the externalized content also, same as the code

web/src/components/FilingsCalendar.test.tsx
```diff
@@ -649,4 +649,58 @@ describe("<FilingsCalendar />", () => {
       expect(screen.queryByText(Config.dashboardDefaults.calendarGridViewButton)).not.toBeInTheDocument();
     });
   });
+
+  describe("calendar list view, view more functionality", () => {
+    function renderCalendarWithEntries(numberOfCalendarEntries: number) {
```
aloverso, line 654: love this helper :) we usually use arrow functions pretty much everywhere, for no particular reason, but it might be nice here also just to keep style consistency

web/src/components/FilingsCalendar.test.tsx
adidner: Yeah good catch

PR #5143, opened 2023-01-26
feat: [#184125138] pass operating phase to intercom

AnJuHyppolite:


web/src/lib/data-hooks/useUserData.ts
```diff
@@ -24,6 +26,11 @@ export const useUserData = (): UseUserDataResponse => {
     },
   });
   const dataExists = !!data;
+  const { setOperatingPhaseId } = useContext(IntercomContext);
+
+  useEffect(() => {
+    setOperatingPhaseId(data?.profileData.operatingPhase);
+  }, [data?.profileData.operatingPhase]);
```
aloverso, line 33: This works, but another option could be to make use the of `onProfileDataChange` function down below, which already runs every time something changes in profileData. It would run more frequently, but would also group related behavior (things that update contexts when something in profile data is changed)

web/src/lib/data-hooks/useUserData.ts
```diff
@@ -12,6 +13,7 @@ import { setAnalyticsDimensions } from "@/lib/utils/analytics-helpers";
 import { UserData } from "@businessnjgovnavigator/shared/";
 import { useContext, useEffect } from "react";
 import useSWR from "swr";
+// import { useMountEffect } from "../utils/helpers";
```
aloverso, line 16: commented code

web/src/pages/_app.tsx
```diff
@@ -29,13 +29,43 @@ import { UserDataErrorContext } from "@/contexts/userDataErrorContext";
...
+//         data-user-hash={user?.intercomHash}
+//         data-user-name={user?.name}
+//         data-user-email={user?.email}
+//         data-user-operating-phase={operatingPhaseId}
+//       />
+//     );
+//   }
+//   return <></>;
+// };
+
```
aloverso, line 68: commented code

web/src/components/tasks/IntercomScript.tsx
```diff
@@ -0,0 +1,32 @@
...
+  operatingPhaseId,
+}: {
+  user?: BusinessUser;
+  operatingPhaseId?: OperatingPhaseId;
+}): ReactElement => {
+  useEffect(() => {
+    if (!operatingPhaseId) return;
+    const item = document.querySelector("#intercom-identity");
+    item?.setAttribute("data-user-operating-phase", operatingPhaseId);
+  }, [operatingPhaseId]);
```
aloverso, line 17: What is the purpose of this useEffect?

web/src/contexts/intercomContext.ts
```diff
@@ -0,0 +1,12 @@
+import { OperatingPhaseId } from "@businessnjgovnavigator/shared/operatingPhase";
+import { createContext } from "react";
+
+export interface IntercomContextType {
+  operatingPhaseId: OperatingPhaseId | undefined;
```
aloverso, line 5: I think that this context doesn't actually need to include the operatingPhaseId. Consumers of this context are only ever using the setter.

web/src/components/tasks/IntercomScript.tsx
mikewolfd: Because <Script> is a next component, it wasn't re-rendering after initial render, so we have to manually edit/modify it's dom element attribute

web/src/lib/data-hooks/useUserData.ts
AnJuHyppolite: There seems to be a lag in the operatingPhase when I use `onProfileDataChange`.

aloverso (approved):

PR #5173, opened 2023-01-30
feat: [#183448263] add max length validation for simple text fields

aloverso:
Iterative commit for fields:
- address line 1 (NJ only)
- contact first name
- contact last name
- registered agent email
- registered agent name
- registered agent address line 1
- registered agent city


web/src/components/tasks/business-formation/getErrorStateForField.ts
```diff
@@ -52,7 +67,6 @@ export const getErrorStateForField = (
     "canMakeDistribution",
     "addressState",
     "addressCountry",
-    "addressMunicipality",
```
mikewolfd, line 55: Why are you removing this? addressMunicipality is an object field, not a text field

PR #5224, opened 2023-02-07
feat: [#183677932] tax encryption v2

somabadri:


web/src/components/onboarding/OnboardingTaxId.tsx
```diff
@@ -46,6 +71,8 @@ export const OnboardingTaxId = ({
   const [locationValue, setLocationValue] = useState(state.profileData[fieldName]?.trim().slice(9, 12) ?? "");
   const [taxIdValue, setTaxIdValue] = useState(state.profileData[fieldName]?.trim().slice(0, 9) ?? "");
   const [errorMap, setErrorMap] = useState<ErrorMapType>({ taxId: false, taxIdLocation: false });
+  const initialDisplayType = useRef<"MASKED" | "PLAIN">(getInitialDisplayType());
```
aloverso, line 74: why are we tracking this value separately from the `displayTaxId` using a `useRef`? It seems like the only place this is used is to set the displayTaxId state?

web/src/components/onboarding/OnboardingTaxId.tsx
```diff
@@ -100,65 +125,136 @@ export const OnboardingTaxId = ({
...
-        formInputFull={props.formInputFull ? true : false}
-        {...props}
-      />
+      <UnStyledButton style="tertiary" align="start" onClick={toggleTaxIdDisplay}>
+        <ButtonIcon svgFilename={displayTaxId ? "hide" : "show"} />
+        <span className="underline">{displayTaxId ? hideButtonText() : showButtonText()}</span>
+      </UnStyledButton>
     );
-  }
+  };
```
aloverso, line 173: stuff like this show/hide toggle could be extracted into its own component, along with extracting its own unit tests. It would make this component AND this component's tests less long and more focused on *its* behavior

web/src/components/onboarding/OnboardingTaxId.tsx
```diff
@@ -100,65 +125,136 @@ export const OnboardingTaxId = ({
...
+      const decryptedTaxId = await decryptTaxId({ encryptedTaxId: state.profileData.encryptedTaxId });
+      setProfileData({ ...state.profileData, taxId: decryptedTaxId });
+      if (initialType.current == "SPLIT") {
+        setSplitTaxId(decryptedTaxId);
+      }
+      setDisplayTaxId(true);
+    } else {
+      setDisplayTaxId(false);
+    }
+  };
```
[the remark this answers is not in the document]
somabadri: I agree, I will figure out a way to make it more readable tomorrow.

web/src/components/onboarding/OnboardingTaxId.tsx
```diff
@@ -46,6 +71,7 @@ export const OnboardingTaxId = ({
   const [locationValue, setLocationValue] = useState(state.profileData[fieldName]?.trim().slice(9, 12) ?? "");
   const [taxIdValue, setTaxIdValue] = useState(state.profileData[fieldName]?.trim().slice(0, 9) ?? "");
   const [errorMap, setErrorMap] = useState<ErrorMapType>({ taxId: false, taxIdLocation: false });
+  const [displayTaxId, setDisplayTaxId] = useState<boolean>(getInitialDisplayType());
```
aloverso, line 74: this variable seems a little confusing named to me - when I see a boolean controlling "displayX" I'd expect that when it is false, then the thing is not displayed on the page. But that's not the case here. It's always displaying tax ID - this is just controlling whether it's input-view or password-view, right? I think the variable could be renamed and even you could use a specific descriptive type with two descriptive values (like you even did something similar in the useRef above that's not used!) instead of a boolean ("primitive obsession" code smell)

aloverso: I see you updated it - I now doubly recommend this maneuver because seeing a function called "getInitialDisplayType" that returns a true/false value is super baffling to come across. I ask myself, 'the two display _types_ are true and false? what does that mean?'

web/src/components/onboarding/OnboardingTaxId.tsx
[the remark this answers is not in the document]
aloverso: consider our refactoring code smell "Repeated Switches" and its associated solution "replace conditional with polymorphism"

At least to me, looking at the return values of this component, it seems to be begging for two components for split/full instead of having one thing handle all these split/full conditional logic

web/src/components/onboarding/OnboardingTaxId.test.tsx
```diff
@@ -139,18 +162,109 @@ describe("<OnboardingTaxId />", () => {
...
+          ...profileData,
+          taxId: "********9000",
+          encryptedTaxId: "some-encrypted-value",
+        },
+        errorProps
+      );
+      expect(screen.getByLabelText("Tax id")).toBeDisabled();
+    });
+
+    it("decrypts the tax id if needed when show is clicked and button toggles text to hide", async () => {
```
aloverso, line 188: what does the "if needed" here mean? that sounds like it should be two tests describing two cases?

web/src/components/onboarding/OnboardingTaxId.test.tsx
```diff
@@ -139,18 +162,109 @@ describe("<OnboardingTaxId />", () => {
...
+        {
+          ...profileData,
+          taxId: "********9000",
+          encryptedTaxId: "some-encrypted-value",
+        },
+        errorProps
+      );
+      expect(screen.queryByText(Config.tax.hideButtonText)).not.toBeInTheDocument();
+      fireEvent.click(screen.getByText(Config.tax.showButtonText));
+      expect(mockApi.decryptTaxId).toHaveBeenCalled();
```
aloverso, line 200: i would pretty much always test `toHaveBeenCalledWith` to be specific here

web/src/components/onboarding/OnboardingTaxId.test.tsx
```diff
@@ -139,18 +162,109 @@ describe("<OnboardingTaxId />", () => {
...
+        expect((screen.getByLabelText("Tax id") as HTMLInputElement).value).toEqual("123-456-789/000");
+      });
+      fireEvent.click(screen.getByText(Config.tax.hideButtonText));
+      await waitFor(() => {
+        expect(currentProfileData().taxId).toEqual("123456789000");
+      });
+      fireEvent.click(screen.getByText(Config.tax.showButtonText));
+      await waitFor(() => {
+        expect(mockApi.decryptTaxId).toHaveBeenCalledTimes(1);
+      });
```
aloverso, line 228: does this need a waitFor if it's not async? I'd that only the first button click needs a waitFor since it's the only one making an API call

web/src/components/onboarding/OnboardingTaxId.test.tsx
```diff
@@ -139,18 +162,109 @@ describe("<OnboardingTaxId />", () => {
...
+        },
+        errorProps
+      );
+      expect((screen.getByLabelText("Tax id") as HTMLInputElement).type).toEqual("password");
+      fireEvent.click(screen.getByText(Config.tax.showButtonText));
+      await waitFor(() => {
+        expect((screen.getByLabelText("Tax id") as HTMLInputElement).type).toEqual("text");
+      });
+      fireEvent.click(screen.getByText(Config.tax.hideButtonText));
+      await waitFor(() => {
```
aloverso, line 247: same as above here - do we need async if the hide isn't making an API call?

web/src/components/onboarding/OnboardingTaxId.test.tsx
```diff
@@ -139,18 +162,109 @@ describe("<OnboardingTaxId />", () => {
...
+          ...profileData,
+          taxId: "********9000",
+          encryptedTaxId: "some-encrypted-value",
+        },
+        errorProps
+      );
+      expect(screen.queryByText(Config.tax.hideButtonTextMobile)).not.toBeInTheDocument();
+      fireEvent.click(screen.getByText(Config.tax.showButtonTextMobile));
+      await waitFor(() => {
+        expect(screen.getByText(Config.tax.hideButtonTextMobile)).toBeInTheDocument();
```
aloverso, line 266: for this and the expect above, I think each one should have a partner expect of what should NOT be in the document. Otherwise, if i just had 2 buttons on the page, one for "show" and one for "hide" and they were just always present, then this test would still pass

PR #5260, opened 2023-02-13
feat: [#183901987] add add'l dakota business types for onboarding

AnJuHyppolite:


web/src/lib/domain-logic/determineForeignBusinessType.test.ts
```diff
@@ -5,14 +5,28 @@ describe("determineForeignBusinessType", () => {
     expect(determineForeignBusinessType([])).toBeUndefined();
   });
 
-  it("returns NEXUS for operationsInNJ as highest priority", () => {
-    expect(determineForeignBusinessType(["operationsInNJ"])).toEqual("NEXUS");
+  it("returns NEXUS for employeeOrContractorInNJ as highest priority", () => {
```
aloverso, line 8: I think this test will need to be expanded. It is testing that when a user selects multiple options that correspond to multiple foreign business types, the NEXUS type takes priority. But right now, that priority is only being tested for employeeOrContractorInNJ and not for the other nexus types.

mikewolfd (approved):

PR #5309, opened 2023-02-23
fix: [#184506447] location is not asked for dakota nexus with no nj l…

somabadri:
…ocation

### DIFF
--- a/web/src/components/FormationDateModal.tsx
+++ b/web/src/components/FormationDateModal.tsx
@@ -35,6 +35,23 @@ export const FormationDateModal = (props: Props): ReactElement => {
     setFieldStates({ ...fieldStates, [field]: { invalid } });
   };
 
+  const shouldShowMunicipalityQuestion = (): boolean => {
+    if (!userData) {
+      return false;
+    }
+    if (userData.profileData.nexusLocationInNewJersey === false) {
+      return false;
+    }
+    return userData.profileData.municipality === undefined;
+  };
+
+  const shouldValidateMunicipalityQuestion = (): boolean => {
+    if (!shouldShowMunicipalityQuestion()) {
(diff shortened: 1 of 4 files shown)


web/src/components/FormationDateModal.tsx
```diff
@@ -43,7 +50,7 @@ export const FormationDateModal = (props: Props): ReactElement => {
       onValidation("dateOfFormation", true);
       return;
     }
-    if (!profileData.municipality || fieldStates.municipality.invalid) {
+    if (shouldShowMunicipalityQuestion() && (!profileData.municipality || fieldStates.municipality.invalid)) {
```
aloverso, line 53: I think this logic should be combined. If `shouldShowMunicipalityQuestion` determines (as the name states) whether the municipality question is shown, I'd expect all that logic to live in the function. Here, it's only handling *some* of the logic around when-to-show, and the rest is in the if-clause. That logic should be moved into the function too.

web/src/components/RegisteredForTaxesModal.tsx
```diff
@@ -97,7 +97,7 @@ export const RegisteredForTaxesModal = (props: Props): ReactElement => {
   };
 
   const showMunicipalityField = (): boolean => {
-    if (!userData) {
+    if (!userData || userData.profileData.nexusLocationInNewJersey === false) {
```
aloverso, line 100: I understand the temptation to combine these, since it already returns false there. But this is a guard clause, it's not intended to contain logic - just to check that userData actually exists before we perform the REAL logic on it. Due to typescript, it's necessary, but we know userData will pretty much never be undefined so you can think of just skipping past it. This logic should be added further down in the body of the function, rather than combined with the guard clause.

PR #5316, opened 2023-02-24
fix: [#184506340] display LLC name NOT id in locked profile field

AnJuHyppolite:


web/test/pages/profile.test.tsx
```diff
@@ -1727,6 +1727,29 @@ describe("profile", () => {
...
   });
 
+  describe("Information Section", () => {
+    describe("business structure field", () => {
+      it("displays Limited Liability Company (LLC) when legal structure id is limited-liability-company", () => {
+        const legalStructure = "limited-liability-company";
+
+        renderPage({
+          userData: generateUserData({
+            formationData: generateFormationData({
```
somabadri, line 1737: Is there a reason to why you had to explicitly set the getFilingResponse to true because the test still passes if when I deleted it.

somabadri (commented):
Everything looks good, just had one question regarding the generated userData within the test.

web/test/pages/profile.test.tsx
AnJuHyppolite: Yes, b/c it will always pass otherwise. I implemented the test first and with red-green factoring, the test was passing on implementation. When I added the formationData, the test failed. Once I implemented my code, then the test passed.

web/test/pages/profile.test.tsx
```diff
@@ -1727,6 +1727,29 @@ describe("profile", () => {
     });
   });
 
+  describe("Information Section", () => {
+    describe("business structure field", () => {
+      it("displays Limited Liability Company (LLC) when legal structure id is limited-liability-company", () => {
```
aloverso, line 1732: I think it'd be helpful to rename this test to be more descriptive of what we care about:
- LLC is just one example of a case, the test isn't specific to LLC. we care about displaying the _name_ of the legal structure instead of the _ID_
- This test is specifically about when the field is locked, which happens after the user has completed formation. The test should specify then the scenario under test here is regarding after formation has completed. I think if the test name had that, it would have clarified soma's question without him having to ask. The test name should make that obvious so that when an engineer comes across this test in 6 months, they ALSO understand why you had to set getFilingResponse

web/test/pages/profile.test.tsx
```diff
@@ -1727,6 +1727,29 @@ describe("profile", () => {
     });
   });
 
+  describe("Information Section", () => {
+    describe("business structure field", () => {
```
aloverso, line 1731: we still call this legal structure in the code, and we should continue to use only one terminology for it

PR #5319, opened 2023-02-24
feat: [#184306179] display error in stepper for unavailable name

somabadri:


web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -364,20 +386,20 @@ describe("<BusinessFormationPaginator />", () => {
...
         switchStepFunction();
-        expect(page.getStepStateInStepper(LookupStepIndexByName("Name"))).toEqual("INCOMPLETE");
+        expect(page.getStepStateInStepper(LookupStepIndexByName("Name"))).toEqual("ERROR");
       });
 
-      it("marks step one as incomplete if business name search is error", async () => {
+      it("marks step one as error if business name search is error", async () => {
         const page = preparePage(initialUserData, displayContent);
         page.fillText("Search business name", "Pizza Joint LLC");
         await page.searchBusinessName({ status: "DESIGNATOR" });
```
AnJuHyppolite, line 400: Would it be worthwhile to rename the DESIGNATOR status (or the test) so that it clearly reflects an error state? Both Soma and I had to trace the status back to `APIBusinessNameClient.ts`. We're thinking about self-contained code so thought this would be an example of code block we could change. Thoughts?

aloverso: Sure, what are you thinking?

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -186,6 +186,14 @@ describe("<BusinessFormationPaginator />", () => {
       });
     });
 
+    it("persists the error state on step one even after switching steps", async () => {
+      const page = preparePage(initialUserData, displayContent);
+      page.fillText("Search business name", "Pizza Joint");
+      await page.searchBusinessName({ status: "UNAVAILABLE" });
+      await page.stepperClickToBusinessStep();
+      expect(page.getStepStateInStepper(LookupStepIndexByName("Name"))).toEqual("ERROR");
+    });
```
aloverso, line 195: This test is labelled "even after" switching steps but it is ONLY testing what happens after switching steps - not what happens before switching steps.
- if it's valuable to test BOTH before AND after switching steps, this needs another assertion before the switch
- if its only testing AFTER then then "even" should be dropped from the label
(i suspect it's the first one)

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
somabadri: I was thinking just having status suffixed with the keyword error, so like `DESIGNATOR_ERROR` so if anyone were to read the test or anywhere else in the code it would be obvious it's a type of error.

AnJuHyppolite: How about `it ("marks step one as error if business name search returns an error response status")`?

PR #5336, opened 2023-03-01
feat: [#183314453] welcome card for up and running and oscar

somabadri:


web/src/pages/onboarding.tsx
```diff
@@ -364,6 +364,18 @@ const OnboardingPage = (props: Props): ReactElement => {
...
 
+      if (newProfileData.operatingPhase === "GUEST_MODE_OWNING") {
+        newPreferencesData.visibleSidebarCards = newPreferencesData.visibleSidebarCards.filter(
+          (cardId: string) => {
+            return cardId !== "welcome";
+          }
+        );
+        newPreferencesData.visibleSidebarCards = [
+          ...newPreferencesData.visibleSidebarCards,
+          "welcome-up-and-running",
```
AnJuHyppolite, line 375: Shouldn't we include this new card for the up-and-running persona (owning)?

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -125,6 +125,86 @@ describe("updateRoadmapSidebarCards", () => {
...
+          industryId: "generic",
+        }),
+        preferences: generatePreferences({ visibleSidebarCards: ["welcome"] }),
+      });
+
+      const updatedUserData = updateSidebarCards(userData);
+      expect(updatedUserData.preferences.visibleSidebarCards).not.toContain("welcome");
+    });
+
+    it("doesn't remove the generic welcome card when operating phase is not UP_AND_RUNNING_OWNING", () => {
```
aloverso, line 181: this is a copy-paste of the exact same test on line 142?

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -125,6 +125,86 @@ describe("updateRoadmapSidebarCards", () => {
...
+        profileData: generateProfileData({
+          operatingPhase: "UP_AND_RUNNING",
+          industryId: "generic",
+        }),
+        preferences: generatePreferences({ visibleSidebarCards: ["welcome"] }),
+      });
+
+      const updatedUserData = updateSidebarCards(userData);
+      expect(updatedUserData.preferences.visibleSidebarCards).toContain("welcome-up-and-running");
+    });
```
aloverso, line 166: this test could be combined with "removes the generic welcome card when operating phase is UP_AND_RUNNING" - they have the exact same setup and are asserting the same scenario

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -125,6 +125,86 @@ describe("updateRoadmapSidebarCards", () => {
...
+      const userData = generateUserData({
+        profileData: generateProfileData({
+          operatingPhase: "UP_AND_RUNNING_OWNING",
+          industryId: "generic",
+        }),
+        preferences: generatePreferences({ visibleSidebarCards: ["welcome"] }),
+      });
+
+      const updatedUserData = updateSidebarCards(userData);
+      expect(updatedUserData.preferences.visibleSidebarCards).toContain("welcome-up-and-running");
```
aloverso, line 204: this test could be combined with "removes the generic welcome card when operating phase is UP_AND_RUNNING_OWNING" - they have the exact same setup and are asserting the same scenario. just one extra assertion statement

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -125,6 +125,86 @@ describe("updateRoadmapSidebarCards", () => {
...
+    });
+  });
+
```
aloverso, line 207: I think there is a missing test (and behavior) here. What happens when a user "un-graduates" and reverts their phase backwards FROM up-and-running TO formed-and-registered or something else? I assume the welcome card should no longer display the special up-and-running card?

web/src/components/dashboard/SidebarCardsContainer.test.tsx
```diff
@@ -91,6 +91,23 @@ describe("<SidebarCardsContainer />", () => {
...
+      };
+
+      useMockUserData({
+        preferences: generatePreferences({ visibleSidebarCards: ["welcome-up-and-running"] }),
+      });
+
+      renderPage({ sidebarCards });
+      const topCardSection = screen.getByTestId("top-cards");
+      expect(within(topCardSection).getByTestId("welcome-up-and-running")).toBeInTheDocument();
+    });
```
aloverso, line 109: I don't think I would add this test. If I'm understanding it correctly, the card weight determines the ordering, right? And the weight is controlled by content, and could be changed by the content team at any point. We don't want to write a test that would fail if they change that "weight: 1" line to "weight: 10", which I believe this would do?

web/src/pages/onboarding.tsx
aloverso: No, that's happening on the backend in the updateSidebarCard changes

web/test/pages/onboarding/onboarding-owning.test.tsx
```diff
@@ -267,6 +268,32 @@ describe("onboarding - owning a business", () => {
...
+    expect(page.getRadioButton("Business Status - Owning")).toBeChecked();
+    await page.visitStep(2);
+    page.clickNext();
+
+    await waitFor(() => {
+      expect(currentUserData()).toEqual({
+        ...initialUserData,
+        preferences: {
+          ...initialUserData.preferences,
+          visibleSidebarCards: ["welcome-up-and-running"],
```
aloverso, line 291: If this change is happening between when the user moves from step 1 to step 2, then I think you'll have to write another test / more code for this situation: a user selects OWNING on step 1, moves to step 2, then goes Back to step 1, and goes through Poppy flow instead. In that word, I think they would get the up-and-running card (but they SHOULD get the regular welcome card).

You could either add logic to "undo" the change if they are Poppy. Or you could just have the card-switch logic happen after submitting from the final step, instead of happening between step 1 and 2. That way there's no way to go back.

somabadri: If I'm not wrong, currently the code to add and remove of welcome cards happens in the else clause of the onSubmit function which fires on the final step.

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -125,6 +126,85 @@ describe("updateRoadmapSidebarCards", () => {
...
+      expect(updatedUserData.preferences.visibleSidebarCards).toContain("welcome-up-and-running");
+      expect(updatedUserData.preferences.visibleSidebarCards).not.toContain("welcome");
+
+      const revertedUserData = {
+        ...updatedUserData,
+        profileData: {
+          ...updatedUserData.profileData,
+          operatingPhase: "FORMED_AND_REGISTERED" as OperatingPhaseId,
+        },
+      };
```
aloverso, line 177: I don't think this test sufficiently describes the case here... the case would be if the phase is FORMED_AND_REGISTERED *and* the user has the welcome-up-and-running card visible. That sets up the "difference" from that to the expectations where it is no longer visible

web/test/pages/onboarding/onboarding-owning.test.tsx
aloverso: Got it, missed that. looks good

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -125,6 +126,85 @@ describe("updateRoadmapSidebarCards", () => {
...
+
+    const operatingPhases = [
+      "GUEST_MODE",
+      "GUEST_MODE_OWNING",
+      "NEEDS_TO_FORM",
+      "NEEDS_TO_REGISTER_FOR_TAXES",
+      "FORMED_AND_REGISTERED",
+    ];
+    for (const operatingPhase of operatingPhases) {
+      it(`doesn't remove the generic welcome card when operating phase is not ${operatingPhase}`, () => {
```
aloverso, line 192: is / is not

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -125,6 +126,85 @@ describe("updateRoadmapSidebarCards", () => {
...
+          operatingPhase: "FORMED_AND_REGISTERED" as OperatingPhaseId,
+        },
+      };
+
+      const updatedRevertedUserData = updateSidebarCards(revertedUserData);
+      expect(updatedRevertedUserData.preferences.visibleSidebarCards).not.toContain("welcome-up-and-running");
+      expect(updatedRevertedUserData.preferences.visibleSidebarCards).toContain("welcome");
+    });
+
+    const operatingPhases = [
```
aloverso, line 184: if you type this array using `const operatingPhases: Operating Phase =`
then you don't need `as` later on

PR #5349, opened 2023-03-06
feat: [#183444690] update profile formation date field so user can de…

AnJuHyppolite:
…lete the date

### DIFF
--- a/web/src/components/onboarding/OnboardingDateOfFormation.tsx
+++ b/web/src/components/onboarding/OnboardingDateOfFormation.tsx
@@ -26,6 +26,7 @@ interface Props {
   required?: boolean;
   disabled?: boolean;
   errorTextOverride?: string;
+  valueFormatter?: ((value: string) => string) | undefined;
 }
 
 export const OnboardingDateOfFormation = (props: Props): ReactElement => {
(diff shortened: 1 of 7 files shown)


web/src/components/onboarding/OnboardingDateOfFormation.tsx
```diff
@@ -26,6 +26,7 @@ interface Props {
   required?: boolean;
   disabled?: boolean;
   errorTextOverride?: string;
+  valueFormatter?: ((value: string) => string) | undefined;
```
mikewolfd, line 29: Why is this being passed in, where is being used?

web/src/pages/profile.tsx
```diff
@@ -281,7 +331,7 @@ const ProfilePage = (props: Props): ReactElement => {
     }
     return isHomeBasedBusinessApplicable(profileData.industryId);
   };
-
+  isFormValidationError;
```
mikewolfd, line 334: remove

web/src/pages/profile.tsx
```diff
@@ -519,18 +569,15 @@ const ProfilePage = (props: Props): ReactElement => {
...
+              <LockedProfileField fieldName="dateOfFormation" valueFormatter={formatDate} />
             ) : (
               <>
                 <FieldLabelProfile fieldName="dateOfFormation" />
                 <OnboardingDateOfFormation
                   onValidation={onValidation}
                   fieldStates={fieldStates}
                   futureAllowed={true}
-                  required
+                  valueFormatter={formatDate}
```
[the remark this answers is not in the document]
AnJuHyppolite: This is old code when I was going to use a different approach. I'll delete it.

PR #5353, opened 2023-03-07
feat: [#184595589] update annual filings logic to return next 3 years

jphechter:
## Purpose

Calculate the next 3 annual tax filing events to be displayed on a user's calendar.

## Ticket

[#184595589](https://www.pivotaltracker.com/story/show/184595589)

## Approach
- Updated the `calculateNextAnnualFIlingDate` function to return the next 3 years as an array.
- Modified test helper functions to reference this new logic to reduce redundancy.

### DIFF
--- a/api/src/domain/annual-filings/calculateNextAnnualFilingDates.test.ts
+++ b/api/src/domain/annual-filings/calculateNextAnnualFilingDates.test.ts
@@ -19,7 +19,7 @@ describe("calculateNextAnnualFilingDates", () => {
       currentDateMock.mockReturnValue(dayjs("2021-10-05", defaultDateFormat));
     });
 
-    it("finds the next-year date for an earlier-this year date", () => {
+    it("finds the next 3 filing dates for an earlier-this year date", () => {
       expect(calculateNextAnnualFilingDates("2021-03-01")).toEqual([
         "2022-03-31",
         "2023-03-31",
@@ -27,7 +27,7 @@ describe("calculateNextAnnualFilingDates", () => {
       ]);
     });
 
-    it("finds the next-year date for a date from last year", () => {
+    it("finds the next 3 filing dates for a date from last year", () => {
       expect(calculateNextAnnualFilingDates("2020-06-01")).toEqual([
         "2022-06-30",
         "2023-06-30",
@@ -35,7 +35,7 @@ describe("calculateNextAnnualFilingDates", () => {
       ]);
     });
 
-    it("finds the next-year date for a much older formation date", () => {
+    it("finds the next 3 filing dates for a much older formation date", () => {
       expect(calculateNextAnnualFilingDates("2001-04-01")).toEqual([
         "2022-04-30",
         "2023-04-30",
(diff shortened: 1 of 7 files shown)


api/test/helpers.ts
```diff
@@ -1,16 +1,12 @@
...
-  if (dateOfFormationDate.month() < currentDate.month()) {
-    year = year + 1;
-  }
-  const nextMonth = dateOfFormationDate.month() + 2;
-  return parseDate(`${year}-${nextMonth}-01`).add(-1, "day").format(defaultDateFormat);
+export const generateAnnualFilings = (dateOfFormation: string): TaxFiling[] => {
+  return calculateNextAnnualFilingDates(dateOfFormation).map((dueDate: string) => ({
+    identifier: "ANNUAL_FILING",
+    dueDate,
+  }));
```
aloverso, line 9: I would tend against having a test-helper that calls the real function, but it winds up being a sort of self-fulfilling prophecy of a test.

I wonder if a simpler implementation of a helper like this that we could use in all the places you need it, would be a function that takes an array of 3 dates and just transforms them into the proper taxFiling object. But doesn't do any calculation logic.

That way, it's the responsibility of the caller to define what the expected dates are, which is part of making the tests more specific and deterministic

api/src/domain/annual-filings/calculateNextAnnualFilingDates.test.ts
```diff
@@ -13,46 +13,82 @@ jest.mock("@shared/dateHelpers", () => {
...
-describe("calculateNextAnnualFilingDate", () => {
+describe("calculateNextAnnualFilingDates", () => {
   describe("when today is October 5, 2021", () => {
     beforeEach(() => {
       currentDateMock.mockReturnValue(dayjs("2021-10-05", defaultDateFormat));
     });
 
     it("finds the next-year date for an earlier-this year date", () => {
-      expect(calculateNextAnnualFilingDate("2021-03-01")).toEqual("2022-03-31");
+      expect(calculateNextAnnualFilingDates("2021-03-01")).toEqual([
```
aloverso, line 23: I think this (and all of the test names in here) should probably be changed from "date" to "dates" or "next 3 years dates" or something like that

api/test/helpers.ts
mikewolfd: Yeah we talked about it for a while, the old helper function was just duplicating the logic of the the real function and given how this helper is being used, it seems appropriate to just use the real logic here. To do what you're talking about we'd have to mock the calculateNextAnnualFilingDates function itself, right?

aloverso: Yeah, I can see that the old implementation was also duplicating the logic. I don't think that's good, and this seems like an opportunity to improve it ("clean up the campground")

I'm not suggesting mocking the function, I'm imagining something simple like:
```
export const generateAnnualFilings = (dueDates: string[]): TaxFiling[] => {
  return dueDates.map((dueDate: string) => ({
    identifier: "ANNUAL_FILING",
    dueDate,
  }));
```

and it gets called in stuff like the `getAnnualFilings` test like:
```
taxFilingData: {
    ...postedUserData.taxFilingData,
    filings: generateAnnualFilings(["2021-03-01","2022-03-01","2023-03-01"]),
  },
```

api/src/domain/annual-filings/calculateNextAnnualFilingDates.test.ts
jphechter: Makes sense. Included!

api/test/helpers.ts
jphechter: I think that makes sense. I'm not super torn one way or another. Included.

PR #5362, opened 2023-03-10
refactor: [#184606606] daycare essential question to facilitate addin…

AnJuHyppolite:
…g it to any industry

### DIFF
--- a/web/src/lib/domain-logic/essentialQuestions.ts
+++ b/web/src/lib/domain-logic/essentialQuestions.ts
@@ -108,7 +108,7 @@ export const EssentialQuestions: EssentialQuestion[] = [
   new EssentialQuestion({
     shouldBeResetWhenIndustryChanges: true,
     isQuestionApplicableToIndustry: (industry) => {
-      return !!industry.industryOnboardingQuestions.isChildcareForSixOrMore && industry.id === "daycare";
+      return !!industry.industryOnboardingQuestions.isChildcareForSixOrMore;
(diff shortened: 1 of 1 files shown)


web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -108,7 +108,7 @@ export const EssentialQuestions: EssentialQuestion[] = [
   new EssentialQuestion({
     shouldBeResetWhenIndustryChanges: true,
     isQuestionApplicableToIndustry: (industry) => {
-      return !!industry.industryOnboardingQuestions.isChildcareForSixOrMore && industry.id === "daycare";
+      return !!industry.industryOnboardingQuestions.isChildcareForSixOrMore;
     },
     fieldName: "isChildcareForSixOrMore",
   }),
```
jphechter, line 114: Beyond the "scope" of this chore, but given that all of these `EssentialQuestions` are generated in a _very_ similar, maybe there's another refactor opportunity here.

jphechter (approved):
This looks good to me! The updated logic seems to follow the existing patterns in this file.

The ticket # referenced in the PR name is incorrect, but it is correct in the branch name. I'm updating this to reflect.

mikewolfd (approved):

aloverso:
@AnJuHyppolite the commit message should be edited to fix the story number, per JP's comment. otherwise looks good

PR #5367, opened 2023-03-13
chore: [#184185244] consolidate legal structure logic

somabadri:


shared/src/legalStructure.ts
```diff
@@ -4,6 +4,14 @@ export interface LegalStructure {
...
   readonly hasTradeName: boolean;
   readonly onboardingOrder: number;
+  readonly displayTaxDisclaimer: boolean;
+  readonly displayResponsibleOwnerName: boolean;
+  readonly displayBusinessName: boolean;
+  readonly displayNexusBusinessElements: boolean;
+  readonly displayStartingBusinessDocuments: boolean;
+  readonly displayEntityId: boolean;
+  readonly displayDbaName: boolean;
+  readonly displayOwningDateOfFormation: boolean;
```
aloverso, line 14: here's some thoughts on these fields. Just my personal suggestions / ideas:

- `displayTaxDisclaimer` could be a little more specific as `displayTaxIdDisclaimer`
- `displayResponsibleOwnerName` and `displayBusinessName` are always an either/or.  It's like the strong-typing technique we've talked about. It's never a valid case to have false/false or true/true with these two booleans - they must always be opposites and so the type should probably reflect that. Same with `displayDbaName` where a DBA name is just another type of business name, so if businessName is true then DBA name will be too. I wonder if all these three could be simplified into something like `typeOfBusinessName` and the values would be like `"NAME_AND_DBA_NAME" | "OWNER_NAME_AND_TRADE_NAME"`. Idk about variable names exactly but I'm also thinking ahead to the story we just pointed where Trade Name will soon be introduced in profile also.
- Seems like `displayNexusBusinessElements` would also be covered by the comment above because it's the same logic as displaying the nexus/business/dba name - they all go together
- `displayStartingBusinessDocuments` - I might rename this as `formationDocumentsInProfile` or maybe `hasFormationDocuments`
- `displayOwningDateOfFormation` could be more like `hasFormationDate`... or maybe that's similar enough to "hasFormationDocuments" to be combined somehow?

adidner: I totally agree with this, in that I think this style gives us more potential to be specific which is always nice.

PR #5377, opened 2023-03-14
feat: [#184306179] display and persist error in stepper for unavailable name

somabadri:
## Description

The business name search component now saves the state of it's response for non-dakota in a businessNameSearch object within formationData. Both the stepper and the inline error fire on search and the state of it is saved using the autosave that runs every second within formation.

### Ticket

[184306179](https://www.pivotaltracker.com/story/show/184306179)

### Steps to Test
Use the Business Name Search as poppy and then wait a second for the auto save to run and the response state will persist even after reloading the page or switching steps.

### Notes

Anne and I talked about restructuring formation because portions of it are overly abstracted and not as relevant anymore. (e.g. SearchBusinessNameForm)

## Code author checklist


api/src/db/migrations/v108_add_business_name_search_to_formation_data.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  it("adds the businessNameSearch field to formation data", () => {
+    const v107 = generateV107UserData({});
+    const v108 = migrate_v107_to_v108(v107);
+    expect(v108).toEqual({
+      ...v107,
+      formationData: {
+        ...v107.formationData,
+        businessNameSearch: undefined,
+      },
+      version: 108,
```
aloverso, line 14: There's no issue here, but as a note for the future, if there's no logic in your migration, then it doesn't actually need a test. Since this test is just a copy-paste of what the actual code does, it doesn't add any additional value.

It's not an issue here, of course. But yeah, we don't need to test if there's no conditional or no logic at all

shared/src/formationData.ts
```diff
@@ -216,6 +225,13 @@ export const createEmptyFormationIncorporator = (): FormationIncorporator => {
   };
 };
 
+export const createEmptyFormationNameAvailability = (): NameAvailability => {
```
aloverso, line 228: where is this used?

shared/src/test/formationFactories.ts
```diff
@@ -129,6 +130,17 @@ export const generateFormationIncorporator = (
   };
 };
 
+export const generateBusinessNameSearch = (overrides: Partial<BusinessNameSearch>): BusinessNameSearch => {
+  return {
+    businessNameAvailability: {
+      similarNames: [],
+      status: undefined,
+    },
```
aloverso, line 138: same comment as before - generators should fill random values

shared/src/formationData.ts
```diff
@@ -69,8 +70,14 @@ export interface FormationData {
   readonly formationResponse: FormationSubmitResponse | undefined;
   readonly getFilingResponse: GetFilingResponse | undefined;
   readonly completedFilingPayment: boolean;
+  readonly businessNameSearch: BusinessNameSearch | undefined;
 }
 
+export type BusinessNameSearch = {
+  businessNameAvailability: NameAvailability | undefined;
+  hasBusinessNameBeenSearched: boolean | undefined;
```
aloverso, line 78: how do we use this `hasBusinessNameBeenSearched` variable? 

1. Why does `hasBusinessNameBeenSearched` have type `boolean | undefined`? If it is not yet searched, I imagine it would be false. Why is undefined a possible value?
2. Although, do we even need it at all? Isn't it always the same as the `businessNameAvailability` value? ie, if `businessNameAvailability` is undefined, it has not been searched. If `businessNameAvailability` is not undefined, it has been searched

PR #5428, opened 2023-03-20
fix: [#184718036] welcome card now closes when the close button is hit

adidner:


api/src/domain/updateSidebarCards.test.ts
```diff
@@ -127,6 +127,45 @@ describe("updateRoadmapSidebarCards", () => {
...
+            industryId: "generic",
+          }),
+          preferences: generatePreferences({ visibleSidebarCards: [] }),
+        });
+        const updatedUserData = updateSidebarCards(userData);
+        expect(updatedUserData.preferences.visibleSidebarCards).not.toContain("welcome");
+      }
+    );
+
+    it.each<OperatingPhaseId | undefined>(["UP_AND_RUNNING", "UP_AND_RUNNING_OWNING"])(
```
aloverso, line 153: why `| undefined` here? (and in the test above)

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -127,6 +127,45 @@ describe("updateRoadmapSidebarCards", () => {
...
+      "removes welcome-up-and-running card when the updated visibleSidebarCards do NOT have welcome-up-and-running card",
+      (operatingPhase) => {
+        console.log(operatingPhase);
```
aloverso, line 156: remove console log

api/src/domain/updateSidebarCards.test.ts
```diff
@@ -127,6 +127,45 @@ describe("updateRoadmapSidebarCards", () => {
...
+    const operatingPhases: OperatingPhaseId[] = [
+      "GUEST_MODE",
+      "GUEST_MODE_OWNING",
+      "NEEDS_TO_FORM",
+      "NEEDS_TO_REGISTER_FOR_TAXES",
+      "FORMED_AND_REGISTERED",
+    ];
+
+    it.each<OperatingPhaseId | undefined>(operatingPhases)(
+      "removes welcome card when the updated visibleSidebarCards do NOT have welcome card",
```
aloverso, line 139: I'm not sure "removes" is the right word here because in the setup here, `visibleSidebarCards` is an empty array. So nothing is being removed. Same in test below

api/src/domain/updateSidebarCards.test.ts
adidner: Oh, that was just the type my IDE said it should match (but you're totally right that it should be narrower)

api/src/domain/updateSidebarCards.test.ts
adidner: I thought I got all of them and I was wrong. Nice catch

api/src/domain/updateSidebarCards.test.ts
adidner: Sure, we're more testing that it's allowed to be empty in this state, rather than constantly re-filled with the initial value like in the bug case.

PR #5431, opened 2023-03-20
refactor: [#184675379] interstateTransport essential questions to fac…

AnJuHyppolite:
…ilitate adding to any industry


web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -71,23 +71,18 @@ export const EssentialQuestions: EssentialQuestion[] = [
   new EssentialQuestion({
     shouldBeResetWhenIndustryChanges: true,
     isQuestionApplicableToIndustry: (industry) => {
-      return (
-        !!industry.industryOnboardingQuestions.isInterstateTransportApplicable && industry.id === "logistics"
-      );
+      return !!industry.industryOnboardingQuestions.isInterstateLogisticsApplicable;
     },
-    fieldName: "interstateTransport",
+    fieldName: "interstateLogistics",
```
AnJuHyppolite, line 76: **Refactor Context:** These were two separate questions that were covered by the same industry onboarding criteria. In deciding how to best approach the refactor, JP and I thought it best to separate out the logistics and moving company industries.

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.ts
```diff
@@ -0,0 +1,382 @@
...
+  version: number;
+}
+
+export const migrate_v106_to_v107 = (v106Data: v106UserData): v107UserData => {
+  return {
+    ...v106Data,
+    profileData: {
+      ...v106Data.profileData,
+      isInterstateLogisticsApplicable: v106Data.profileData.industryId == "logistics" ? true : undefined,
+      isInterstateMovingApplicable: v106Data.profileData.industryId === "moving-company" ? true : undefined,
```
aloverso, line 23: I don't think this logic is correct... if a user has industry=logistics, they are presented with a radio question for the interstate transport, and some of them may have selected "no".

We need to make use of their previous answer to the question. We can't just use set the answer to "yes" for every logistics industry business

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.ts
```diff
@@ -0,0 +1,414 @@
...
+  formationData: v107FormationData;
+  lastUpdatedISO: string | undefined;
+  version: number;
+}
+
+export const migrate_v106_to_v107 = (v106Data: v106UserData): v107UserData => {
+  const getLogisticsResponse = () => {
+    if (
+      v106Data.profileData.industryId === "logistics" &&
+      v106Data.profileData.interstateTransport === true
```
aloverso, line 21: Brought this up in techtro last week, we don't want to be checking `=== true` or `=== false` and instead we should simplify the expression

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.ts
```diff
@@ -0,0 +1,414 @@
...
+    ) {
+      return true;
+    } else if (
+      v106Data.profileData.industryId === "logistics" &&
+      v106Data.profileData.interstateTransport === false
+    ) {
+      return false;
```
aloverso, line 28: This if/else-if chain of logic looks correct, but could be simplified quite a lot to make it more readable.

```
if (v106Data.profileData.industryId === "logistics") {
 return ...
} else {
  return ...
}
```

Same with the one below as well

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.ts
```diff
@@ -0,0 +1,414 @@
...
+    } else {
+      return;
```
aloverso, line 30: it's strange to have a return here without a value. If we want to return undefined, we should say so explicitly. I also would encourage giving an explicit type to the return value in the function declaration. Usually when I see `return;`  with no value, it means an early exit from a logical flow, not that it's explicitly returning a value that we're going to use later on.

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      const v107 = migrate_v106_to_v107(v106);
+      expect(v107).toEqual({
+        ...v106,
+        profileData: {
+          ...v106.profileData,
+          isInterstateLogisticsApplicable: true,
+        },
+        version: 107,
+      });
+    });
```
aloverso, line 22: given that our code change is adding 2 new fields, I'd expect the test to be testing explicitly for what it sets both fields to in the output. `isInterstateLogisticsApplicable` becomes true for this user, but what should `isInterstateMovingApplicable` be? 

Since the function purpose is to start with one field and determines the value of 2 fields, then in tests, I'd suggest that the tests have assertions that make it clear what we expect from both fields. This comment would hold true for pretty much all the tests here - it'd be nice for them to be explicit about the end-result value of both new fields that are being calculated

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.test.ts
```diff
@@ -0,0 +1,134 @@
...
+          isInterstateMovingApplicable: false,
+        },
+        version: 107,
+      });
+    });
+
+    it("sets interstateMovingApplicable onboarding question to undefined for any non-moving company existing users", () => {
+      const v106 = generateV106UserData({
+        profileData: generateV106ProfileData({
+          industryId: "moving-company",
```
aloverso, line 104: The test title says "non-moving-company users" but the industry here **is** moving-company?

api/src/db/migrations/v106_add_pet_care_housing_essential_question.ts
```diff
@@ -377,3 +378,223 @@ type v106GetFilingResponse = {
   standingDoc: string;
   certifiedDoc: string;
 };
+
```
jphechter, line 381: This is interesting: we're adding generator functions to the previous migration as a part of the work for `v107`. This feels like something that should be the responsibility of the author of `v106`. I guess this is mostly copypaste, but if this piece blocks the next migration it feels somewhat unfinished to me if we leave it to the next person to generate.

I'm feeling more strongly about generating a migration generator script.

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.ts
```diff
@@ -0,0 +1,414 @@
...
+
+type v107GetFilingResponse = {
+  success: boolean;
+  entityId: string;
+  transactionDate: string;
+  confirmationNumber: string;
+  formationDoc: string;
+  standingDoc: string;
+  certifiedDoc: string;
+};
```
jphechter, line 414: I would advocate for adding `v107` generator functions to this file to make life easier for the author of `v108`.

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.ts
```diff
@@ -0,0 +1,414 @@
...
+      return true;
+    } else if (
+      v106Data.profileData.industryId === "moving-company" &&
+      v106Data.profileData.interstateTransport === false
+    ) {
+      return false;
+    } else {
+      return;
+    }
+  };
```
jphechter, line 48:
```suggestion
  const getMovingResponse = () => v106Data.profileData.industryId === "moving-company" && v106Data.profileData.interstateTransport;
```
Same as above.

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -23,7 +23,7 @@ export const getIsApplicableToFunctionByFieldName = (
   return (industryId: string | undefined): boolean => {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     return EssentialQuestions.find((eQ) => {
-      return eQ.fieldName == fieldName || eQ.contentFieldName == fieldName;
+      return eQ.fieldName == fieldName;
```
jphechter, line 26:
```suggestion
      return eQ.fieldName === fieldName;
```
Another one of those techtro items from last week.

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.test.ts
```diff
@@ -0,0 +1,134 @@
...
+          ...v106.profileData,
+          isInterstateMovingApplicable: undefined,
+        },
+        version: 107,
+      });
+    });
+
+    it("sets the interstateMovingApplicable question to undefined for any new user", () => {
+      const v106 = generateV106UserData({
+        profileData: generateV106ProfileData({}),
```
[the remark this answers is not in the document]
AnJuHyppolite: Your feedback makes sense. So that would mean the similar pet care essential question test in the v106 migration doesn't make sense either.

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.test.ts
AnJuHyppolite: I wanted to account for existing users who are not in the moving-company industry.

api/src/db/migrations/v106_add_pet_care_housing_essential_question.ts
AnJuHyppolite: I agree, JP, and was confused as to why that would be the "work" of the person implementing the next migration. I'd love additional feedback.

A migration generator script would be cool.

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.ts
AnJuHyppolite: Waiting feedback from previous commentary...

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.test.ts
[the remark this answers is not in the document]
aloverso: Yes, you're right I think. I didn't catch that one

api/src/db/migrations/v107_refactor_interstate_transport_essential_question.test.ts
aloverso: Gotcha, shouldn't this line that sets industryId set it to something else? Right now it is testing a user whose industry is moving-company

api/src/db/migrations/v106_add_pet_care_housing_essential_question.ts
aloverso: The rationale was that not every migration needs generators - it only needs them if the following migration has a test written. 

We write migrations sometimes even if they don't have logic, like adding a new field that will always have a static value. Those migrations don't typically get tests added.

So if v107 was a migration that didn't need a test, then we'd have no use for generators in v106 - they would be unused functions. Which is generally, something to avoid, in my opinion.

That's the context why (so far) we don't add the generators at the same time as the migration file. Happy to discuss what we prefer in the tradeoff of having-unused-functions-laying-around versus having-to-edit-a-previous-migration-file-when-testing

PR #5446, opened 2023-03-21
Find and replace all instances of "==" with "==="

jphechter:
## Description

We've talked enough about == that I figured it would be worth checking how many instances were in the code. Did a global search for " == " and " != "  and found `195 occurrences across 108 files`. Did a global replace of " == " and " != "with " === " and " !== ". In theory there could be some == and != that are jammed up against a variable and wouldn't match, but I'm not a regex wizard and I didn't feel like looking.

If tests pass, then this should be good to go. Hopefully no behaviors are changed.

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/name/BusinessNameStep.tsx
+++ b/web/src/components/tasks/business-formation/name/BusinessNameStep.tsx
@@ -127,7 +127,7 @@ export const BusinessNameStep = (): ReactElement => {
           </div>
         </WithErrorBar>
       </form>
-      {error != null && (
+      {error && (
         <Alert variant="error" dataTestid={`error-alert-${error}`}>
           {SearchBusinessNameErrorLookup[error]}
         </Alert>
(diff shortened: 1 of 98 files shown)


somabadri (approved):
Looks good to me! I just skimmed over the changes. It passed CI so I think it's good.

web/src/components/tasks/business-formation/name/BusinessNameStep.tsx
```diff
@@ -127,7 +127,7 @@ export const BusinessNameStep = (): ReactElement => {
           </div>
         </WithErrorBar>
       </form>
-      {error != null && (
+      {error && (
```
jphechter, line 130: Screen Shot 2023-03-21 at 11 02 31 PM

`SearchBusinessNameError` is a string. Dropped this comparison because the strict comparison causes the build to fail and strings are truthy in JS.

adidner (approved):
LGTM

PR #5453, opened 2023-03-22
refactor: [#184432315] extract shared component for shared formatting in profile fields

aloverso:
## Description

Pre-factor that will make it easier to add an anchor tag ID to every profile field element, which is necessary for the linked story.

This refactor makes the styling (headers, spacing, margin, horizontal lines, etc) consistent across all the personas in profile, and across all the fields. Previously every single one was managing their spacing independently and it was all over the place. Now, this wrapper component manages it all.

### Ticket

[#184432315](https://www.pivotaltracker.com/story/show/184432315)

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/profile.json
+++ b/content/src/fieldConfig/profile.json
@@ -12,7 +12,7 @@
       },
       "notes": {
         "default": {
-          "header": "Notes",
+          "header": "",
           "description": "Use this space to add any information for quick reference."
         }
       },
@@ -266,7 +266,7 @@
       },
       "documents": {
         "default": {
-          "header": "Reference Documents",
+          "header": "",
           "placeholder": "If you are an LLC, Corporation, or LLP, we will automatically add your formation documents once you [Form Your Business](/tasks/form-business-entity) on our site."
         }
       },
--- a/web/src/lib/types/types.ts
+++ b/web/src/lib/types/types.ts
@@ -77,11 +77,8 @@ export type FormationFieldErrorState = {
 
 export const profileFieldsFromConfig = getMergedConfig().profileDefaults.fields;
 
-export type IndustrySpecificDataAddOnFields = "interstateLogistics" | "interstateMoving";
-
 export type ProfileContentField = Exclude<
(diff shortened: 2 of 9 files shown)


content/src/fieldConfig/profile.json
```diff
@@ -266,7 +266,7 @@
       },
       "documents": {
         "default": {
-          "header": "Reference Documents",
+          "header": "",
```
aloverso, line 269: These are no longer rendered since there's a page header that says the same thing, and that better fits accessibility hierarchy

web/src/lib/types/types.ts
```diff
@@ -77,11 +77,8 @@ export type FormationFieldErrorState = {
 
 export const profileFieldsFromConfig = getMergedConfig().profileDefaults.fields;
 
-export type IndustrySpecificDataAddOnFields = "interstateLogistics" | "interstateMoving";
-
 export type ProfileContentField = Exclude<
-  (keyof ProfileData | keyof IndustrySpecificData | IndustrySpecificDataAddOnFields) &
-    keyof typeof profileFieldsFromConfig,
+  (keyof ProfileData | keyof IndustrySpecificData) & keyof typeof profileFieldsFromConfig,
```
aloverso, line 81: Technically this is cleanup from the logistics/transport refactor, but I was changing similar code so I thought I'd do it as well. Should have been a separate commit, apologies

web/test/pages/profile.test.tsx
```diff
@@ -1900,130 +1900,112 @@ describe("profile", () => {
...
-        }),
+    describe("if Oscar", () => {
+      it("has no document section", () => {
+        const userData = generateUserData({
+          profileData: generateProfileData({
+            businessPersona: "OWNING",
+          }),
+        });
+        renderPage({ userData });
+        expect(screen.queryByTestId("documents")).not.toBeInTheDocument();
```
aloverso, line 2007: Changed test logic. I am pretty sure this is the correct logic, what we had before was old. Oscar can never have formation documents

PR #5457, opened 2023-03-22
feat: [#184432315] add opportunities alert to profile

aloverso:
### Ticket

[#184432315](https://www.pivotaltracker.com/story/show/184432315)

## Code author checklist

### DIFF
--- a/web/src/lib/utils/helpers.ts
+++ b/web/src/lib/utils/helpers.ts
@@ -41,6 +41,19 @@ export const useMountEffectWhenDefined = (fun: () => void, thingToBeDefined: unk
   }, [thingToBeDefined, fun]);
 };
 
+export const useScrollToPathAnchor = () => {
+  useEffect(() => {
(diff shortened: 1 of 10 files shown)


web/src/lib/utils/helpers.ts
```diff
@@ -41,6 +41,19 @@ export const useMountEffectWhenDefined = (fun: () => void, thingToBeDefined: unk
...
+    if (path && path.includes("#")) {
+      const id = path.replace("#", "");
+      const element = document.querySelector(`#${id}`);
+      if (id && element) {
+        element.scrollIntoView({ behavior: "smooth" });
+      }
+    }
+  });
+};
+
```
aloverso, line 56: This is, for some unknown reason, necessary because of [this nextjs issue](https://github.com/vercel/next.js/issues/11109)

Without it, anchor links don't work from a URL on page load. they work if you're already on the page, but not on reload or when navigating from another page

PR #5482, opened 2023-03-24
[#184783187] Consolidate ReviewSection Components

jphechter:
## Description

While working on [#183443674](https://www.pivotaltracker.com/story/show/183443674) I noticed some redundancies between components in the review step. This is an attempt to streamline some of that logic.

### Ticket

[#184783187](https://www.pivotaltracker.com/story/show/184783187)

### Approach

This is largely centered around `ReviewText`, `ReviewSectionHeader`, and `ReviewLineItem` all of which had redundancies. I created a wrapper component `ReviewSection` that has the shared elements and logic of these, and can take children. I then replaced instances of `ReviewSectionHeader` with `ReviewSection` and provided the other components as children where applicable. Ideally, this should have no effect on the UI.

### Steps to Test

Hopefully, test coverage will be adequate to confirm this is a seamless swap. Otherwise, the Review step should behave as expected prior to the change.

### Notes

In doing this, I only encountered one place that wasn't a clean drop-in replacement for this change. The `BusinessNameAndLegalStructure` component has a slightly different pattern, where the same component is used in both the review and form entry, but takes a boolean to determine which version of the component to render. This feels like it could be cleaned up, but I didn't want to start digging into that just yet.

## Code author checklist


adidner (approved):
Seems good to me from looking at code. 

Well thought out refactor of logic

PR #5524, opened 2023-03-28
feat: [#183913408] create header styling to support contextual links …

adidner:
…in the profiles

## Description

We wanted to make the Headers in profile support contextual links

### Ticket

[183913408](https://www.pivotaltracker.com/story/show/#183913408)

### Approach

As discussed during a standup we wanted to render the contextual link as a separate field instead of a markdown field for reusability of the header text across the application.

I also did some of the copy but not all of it as some of it doesn't have contextual link files ready yet (Specifically for Dakota DBA). 

I also made some fields, required optional as the new designs don't include these fields and they have been required up to this point

### Steps to Test

Go to profile as the various users and see that contextual links are rendered and that the tabs accurately represent the design in the pivotal ticket / figma (except for Dakota)

Also make sure you have different onboarding styling for the legal structure question (the onboarding one should have a description while the other should just be a title and the profile one should just be a contextual link)

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/profile.json
+++ b/content/src/fieldConfig/profile.json
@@ -27,8 +27,18 @@
       "sectorId": {
         "default": {
           "header": "Industry Sector",
-          "description": "Which industry sector best describes your business?",
+          "description": "",
           "errorTextRequired": "Please select an industry sector."
+        },
+        "onboarding": {
+          "default": {
+            "description": ""
+          },
+          "overrides": {
+            "OWNING": {
+              "description": "Which industry sector best describes your business?"
+            }
+          }
         }
       },
       "certifiedInteriorDesigner": {
@@ -50,7 +60,8 @@
       "taxId": {
         "default": {
           "header": "New Jersey Tax ID",
-          "description": "What is your `New Jersey Tax ID?|tax-id`",
+          "headerContextualInfo": "tax-id",
+          "description": "",
           "errorTextRequired": "Enter a 12-digit Tax ID",
           "disclaimerMd": "**Disclaimer:** Your Tax ID will be used to identify the taxes you owe and is required information if you'd like to access your tax calendar."
         }
@@ -101,6 +112,7 @@
       "taxPin": {
         "default": {
           "header": "Tax PIN",
+          "headerContextualInfo": "",
           "description": "What is your `Tax PIN|tax-pin`?",
           "errorTextRequired": "Tax PIN must be 4 digits."
         }
@@ -108,19 +120,17 @@
       "businessName": {
         "default": {
           "header": "Business Name",
+          "headerContextualInfo": "",
           "errorTextRequired": "Please enter your business name.",
-          "description": "What will you name your business?"
-        },
-        "overrides": {
-          "OWNING": {
-            "description": "What is your business' name?"
-          }
+          "description": ""
         }
       },
       "dateOfFormation": {
         "default": {
-          "header": "Effective Date",
-          "description": "When did you form your business? Find your `effective date|effective-date` on your `Certificate of Formation|certificate-formation`.",
+          "header": "Business Effective Date",
+          "headerContextualInfo": "effective-date",
+          "headerNotBolded": "(mm/yyyy)",
+          "description": "Find your on your `Certificate of Formation|certificate-formation`.",
           "errorTextRequired": "Your formation should be formatted as MM/YYYY. (This information will help us inform your annual reporting due date.)"
         }
       },
@@ -158,6 +168,7 @@
       "nexusDbaName": {
         "default": {
           "header": "Doing Business As (DBA) New Jersey Business Name",
+          "headerContextualInfo": "dba-resolution-foreign",
           "description": "What is your DBA name?",
           "errorTextRequired": "Please enter your DBA name."
         }
@@ -179,6 +190,7 @@
       "naicsCode": {
         "default": {
           "header": "NAICS Code",
+          "headerContextualInfo": "naics",
           "editText": "Edit",
           "addText": "Add",
           "notEnteredText": "*Not entered*"
@@ -216,8 +228,9 @@
       },
       "legalStructureId": {
         "default": {
-          "header": "Which Business Structure Will You Choose?",
-          "description": "Your business’ `business structure|business-structure-learn-more` can have a large impact on taxes, financing, liability, and closure of your business. We strongly suggest doing research on the best business structure for your business.  \nSelect a business structure for your Navigator profile below. Making a selection will not form your business, it will help us create your personalized guide. You can change your intended business structure anytime before formally filing it with the NJ Department of Treasury.",
+          "header": "Business Structure",
+          "headerContextualInfo": "business-structure-learn-more",
+          "description": "",
           "errorTextRequired": "Select a business structure.",
           "optionContent": {
             "general-partnership": "In a general partnership two or more people share ownership of a single business and are jointly responsible for its assets and liabilities. [Learn more about General Partnerships.](https://business.nj.gov/pages/general-partnership) ",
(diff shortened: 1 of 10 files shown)


web/src/components/onboarding/FieldLabelProfile.tsx
```diff
@@ -36,7 +37,14 @@ export const FieldLabelProfile = (props: Props) => {
       <div className="flex flex-row fac">
         {isHeaderInConfig && (
           <div role="heading" aria-level={3} className="h3-styling margin-bottom-2">
-            {contentFromConfig.header}
+            {contentFromConfig.headerContextualInfo ? (
```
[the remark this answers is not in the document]
adidner: Futhermore I could test in the profile page itself but it still feel like I'm not testing functionality as much as I'm testing content choices which seems bad as they could break the tests with a content change

aloverso:
I think you need to talk to the designers about the change to onboarding. I doubt that they want this story to affect that, and if so, that will need to be addressed before merging

content/src/fieldConfig/profile.json
```diff
@@ -101,13 +102,14 @@
       "taxPin": {
         "default": {
           "header": "Tax PIN",
-          "description": "What is your `Tax PIN|tax-pin`?",
+          "description": "What is your `Tax PIN|tax-pin` ?",
```
aloverso, line 105: It seems like there's a lot of lines of code changes that are adding a space in between the text and the question mark? This doesn't seem desirable - should be removed?

content/src/fieldConfig/profile.json
```diff
@@ -101,13 +102,14 @@
...
           "errorTextRequired": "Tax PIN must be 4 digits."
         }
       },
       "businessName": {
         "default": {
           "header": "Business Name",
+          "headerContextualInfo": "",
```
aloverso, line 112: So is the intent of the story to add the *ability* for any profile label to have a contextual info link? If so, I imagine we want to add this line to pretty much every field, right? Like directly above for example, wouldn't we want to add this same empty-string line (and the corresponding dropdown in config.yml) - otherwise, how will the content team add a contextual info to Tax PIN if they decide that's what they want to do?

web/test/pages/profile.test.tsx
```diff
@@ -284,9 +284,9 @@ describe("profile", () => {
         renderPage({ userData });
         chooseTab("numbers");
         expect(screen.getByText(Config.profileDefaults.fields.entityId.default.header)).toBeInTheDocument();
-        expect(
-          screen.queryByText(Config.profileDefaults.fields.entityId.default.description)
-        ).not.toBeInTheDocument();
+        // expect(
+        //   screen.queryByText(Config.profileDefaults.fields.entityId.default.description)
+        // ).not.toBeInTheDocument();
```
aloverso, line 289: no commented code

web/test/pages/profile.test.tsx
```diff
@@ -296,9 +296,9 @@ describe("profile", () => {
         expect(
           screen.getByText(Config.profileDefaults.fields.legalStructureId.default.header)
         ).toBeInTheDocument();
-        expect(
-          screen.queryByText(Config.profileDefaults.fields.legalStructureId.default.description)
-        ).not.toBeInTheDocument();
+        // expect(
+        //   screen.queryByText(Config.profileDefaults.fields.legalStructureId.default.description)
+        // ).not.toBeInTheDocument();
```
aloverso, line 301: no commented code

content/src/fieldConfig/profile.json
adidner: Agreed, I will fix that, not sure how that got in there.

content/src/fieldConfig/profile.json
adidner: Yes, we do I thought I got that one but I clearly have missed it

content/src/fieldConfig/profile.json
```diff
@@ -229,6 +238,11 @@
             "sole-proprietorship": "In a sole proprietorship you alone own the company and are responsible for its assets and liabilities. [Learn more about Sole Proprietorships.](https://business.nj.gov/pages/sole-proprietorship) "
           }
         },
+        "onboarding": {
+          "header": "Which Business Structure Will You Choose?",
+          "headerContextualInfo": "",
+          "description": "Your business’ `business structure|business-structure-learn-more` can have a large impact on taxes, financing, liability, and closure of your business. We strongly suggest doing research on the best business structure for your business.  \nSelect a business structure for your Navigator profile below. Making a selection will not form your business, it will help us create your personalized guide. You can change your intended business structure anytime before formally filing it with the NJ Department of Treasury."
+        },
```
aloverso, line 245: I think the overrides should the overrides be nested inside of onboarding? Otherwise they apply to Dakota businesses on the profile as well? 

(it would be true for oscar as well except we purposefully don't show legal structure on Oscar profile. But my point is that we are overriding the *onboarding* description so it feels like it should go inside the onboarding object?)

Screenshot 2023-03-30 at 3 59 16 PM

content/src/fieldConfig/profile.json
AnJuHyppolite, line 1: How true to the copy in the design should this be?

content/src/fieldConfig/profile.json
```diff
@@ -179,6 +185,7 @@
       "naicsCode": {
         "default": {
           "header": "NAICS Code",
+          "headerContextualInfo": "naics",
           "editText": "Edit",
           "addText": "Add",
```
AnJuHyppolite, line 190: There is some styling issue with `Add`. It's not aligned with `NAICS Code`.

adidner: So their are actually a lot of minor inconsistencies from design to actual at the moment, and I'm planning on a follow up with the design to team for a story to address all of the minor changes (figured it would be easier and not have this story balloon into fixing a bunch of styling)

content/src/fieldConfig/profile.json
adidner: Where more precisely do you mean? 

Mostly the copy that I went and changed was with regards to just getting the links up and running at all in the profile (except for Dakota stuff). The copy on the contextual menu's I haven't touched 

Also make sure you're running the proxy server to get all the changes

content/src/fieldConfig/profile.json
```diff
@@ -208,16 +215,18 @@
       "homeBasedBusiness": {
         "default": {
           "header": "Home-based business",
+          "headerContextualInfo": "home-based-business",
           "radioButtonTrueText": "Yes",
           "radioButtonFalseText": "No",
-          "description": "Are you a `home-based business|home-based-business`?",
-          "altDescription": "Do you plan to be a `home-based business|home-based-business`?"
+          "description": "Are you a `home-based business|home-based-business` ?",
+          "altDescription": "Do you plan to be a `home-based business|home-based-business` ?"
```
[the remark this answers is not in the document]
adidner: God, how do these keep getting in my code. Whats going on

content/src/fieldConfig/profile.json
adidner: Sure that makes sense

adidner: So all 3 onboarding flows should be correct now. And all 3 profiles should have all the contextual headers from design (with a few minor inconsistencies)

content/src/fieldConfig/profile.json
adidner: So the contextual headers from the design that are in all 3 profile pages or just 1 profile page should appear (some of the specific to one profile but not others the designers / content people are still refining)

web/src/lib/domain-logic/getProfileConfig.test.ts
```diff
@@ -5,41 +5,131 @@ import { merge } from "lodash";
...
+            description: "default description",
             errorTextRequired: "error",
+            headerContextualInfo: "default header contextual info",
+          },
+        },
+      },
+    },
+  };
+
+  const testConfig_default_with_override = {
```
aloverso, line 23: We don't use snake case in variable names (here and further above)

content/src/fieldConfig/profile.json
```diff
@@ -229,10 +248,32 @@
...
+                "limited-liability-company": "",
+                "limited-liability-partnership": "",
+                "limited-partnership": "",
+                "s-corporation": "",
+                "sole-proprietorship": ""
+              }
+            }
+          }
+        },
+        "overrides": {
```
aloverso, line 276: so I can see this working now, in that the Dakota profile does not have the contextual link on the header. But I'm looking at the code and logic here, and I'm confused about _why_ it's working, can you explain the logic I'm missing?

It seems like for Dakota profile, it would take the default and then the overrides.

The default includes the `headerContextualInfo` field and the overrides for FOREIGN doesn't have headerContextualInfo set to empty string. So why doesn't the contextual info show up on the header for Dakota in profile?

aloverso: Like, if default is
```
"header": "Business Structure",
"headerContextualInfo": "business-structure-learn-more",
"description": "",
```

and the override is 
```
"overrides": {
  "FOREIGN": {
    "header": "Business Structure of Your Existing Out-of-State Business",
    "description": "Please select...",
```

Then wouldn't the merge result in an object like:
```
"header": "Business Structure of Your Existing Out-of-State Business",
"headerContextualInfo": "business-structure-learn-more",
"description": "Please select...",
```

What am I missing?

adidner: So a few things. First I think it actually does work exactly how you described and your cache is invalid or something (like if you try this again and refresh the page you will probably get the correct behavior). 

This is my local Dakota Profile: 
image




Whereas the design for Dakota Profile looks like this 

image


The issue is, that some of the copy and design that I have for this is wrong (or at least I've questioned something and they've been amended). 

And I don't think the intent of this ticket was for me to question the design and copy and play ping pong with design and content and fix it all to be same as design which is what it kind feels like its turning into. 

So I would like to ask design and content to write a ticket for me once they've aligned and bring up my various questions during a meeting they have together.

content/src/fieldConfig/profile.json
```diff
@@ -101,26 +112,30 @@
...
           "errorTextRequired": "Please enter your business name.",
-          "description": "What will you name your business?"
+          "description": ""
         },
-        "overrides": {
-          "OWNING": {
-            "description": "What is your business' name?"
+        "onboarding" : {
+          "default": {
+            "description": "What will you name your business?"
```
aloverso, line 129: Why this change? I didn't think business name ever got asked in onboarding?

adidner: That is correct and I will fix it

PR #5530, opened 2023-03-29
fix: [#184306179] confirm availability in-line error on blur

somabadri:
## Description
Fix to address the faulty behavior with the wrong in-line error being fired on blur in business name search.

### Ticket
[184306179](https://www.pivotaltracker.com/story/show/184306179)

## Code author checklist

### DIFF
--- a/web/src/lib/data-hooks/useBusinessNameSearch.ts
+++ b/web/src/lib/data-hooks/useBusinessNameSearch.ts
@@ -34,15 +34,20 @@ export const useBusinessNameSearch = ({
   const [error, setError] = useState<SearchBusinessNameError | undefined>(undefined);
   const [updateButtonClicked, setUpdateButtonClicked] = useState<boolean>(false);
 
+  const businessNameHasBeenSearched = (): boolean => {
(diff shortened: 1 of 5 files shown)


web/src/lib/data-hooks/useBusinessNameSearch.ts
```diff
@@ -41,7 +41,7 @@ export const useBusinessNameSearch = ({
     setCurrentName(isDba ? userData.profileData.nexusDbaName || "" : userData.profileData.businessName);
     if (
       userData.formationData.businessNameAvailability?.status !== "AVAILABLE" &&
-      userData.formationData.businessNameAvailability?.status !== undefined
+      userData.formationData.businessNameAvailability !== undefined
```
aloverso, line 44: It took me a bit of reading to figure out why we were checking these two values. Maybe saving as a variable would make it more readable? Something like:

```
if (businessNameHasBeenSearched && isNotAvailable)
```

somabadri: I think that's a good idea, making the change.

PR #5538, opened 2023-03-30
feat: [184533765] add legal structure and industry to intercom

AnJuHyppolite:
## Description

Added the legal structure and industry to the intercom in an effort to phase out the Airtable model we are currently using.

### Ticket

[184533765](https://www.pivotaltracker.com/n/projects/2573455/stories/184533765)

### Approach

N/A

### Steps to Test

Once you go through a persona and select an industry and legal structure, use the dev inspector tools to check that the data is present in the `src` attribute of the `script` tag that points to **intercom/identity.js**.

### Notes

I need some guidance/context as to how to check for for user data in Airtable to ensure we add all of the user data we want present in the intercom to ensure a successful migration from the Airtable model. If you can provide this guidance/context, please let me know. Thanks!

## Code author checklist

### DIFF
--- a/web/public/intercom/identity.js
+++ b/web/public/intercom/identity.js
@@ -1,31 +1,46 @@
-var userId = document.currentScript.getAttribute("data-user-id");
+let userId = document.currentScript.getAttribute("data-user-id");
 if (typeof userId === "undefined") {
   userId = undefined;
 }
 
-var userHash = document.currentScript.getAttribute("data-user-hash");
+let userHash = document.currentScript.getAttribute("data-user-hash");
 if (typeof userHash === "undefined") {
   userHash = undefined;
 }
 
-var userName = document.currentScript.getAttribute("data-user-name");
+let userName = document.currentScript.getAttribute("data-user-name");
 if (typeof userName === "undefined") {
   userName = undefined;
 }
 
-var userEmail = document.currentScript.getAttribute("data-user-email");
+let userEmail = document.currentScript.getAttribute("data-user-email");
 if (typeof userEmail === "undefined") {
   userEmail = undefined;
 }
 
-var userOperatingPhase = document.currentScript.getAttribute("data-user-operating-phase");
+let userOperatingPhase = document.currentScript.getAttribute("data-user-operating-phase");
 if (typeof userOperatingPhase === "undefined") {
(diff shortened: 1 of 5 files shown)


aloverso:
@AnJuHyppolite the `AirtableFeedbackClient` has the full list of fields that it is sending.

web/public/intercom/identity.js
```diff
@@ -28,6 +28,21 @@ if (typeof userUuid === "undefined") {
   userUuid = undefined;
 }
 
+var userLegalStructure = document.currentScript.getAttribute("legal-structure");
```
adidner, line 31: I think let tends to be better than var because its scope is narrower which leads to fewer potential weird scoping override errors.

adidner (commented):
Generally the logic is good and makes sense. I didn't really understand what you meant by testing. I signed in as a poppy and looked at the scripts via dev tools but didn't see a script with the src or attributes you were mentioning
image

web/public/intercom/identity.js
aloverso: That's fair, but I think we should make a separate refactoring chore for this. Here AnJu is just following the pattern of the whole rest of the file. If we change it, we should change it in all places here.

web/src/components/tasks/IntercomScript.tsx
```diff
@@ -26,6 +43,9 @@ export const IntercomScript = ({
         data-user-name={user?.name}
         data-user-email={user?.email}
         data-user-operating-phase={operatingPhaseId}
+        data-user-legal-structure={legalStructureId}
+        data-user-industry-id={industryId}
+        data-user-business-persona={businessPersona}
```
aloverso, line 48: I know you created this file for the last intercom story, and I didn't notice this then, but I see now that this file lives in `components/tasks` directory. Is there a reason for that? It seems misleading since this component is used on every single page of the application, not just on tasks. I'd recommend moving this file

web/public/intercom/identity.js
aloverso: Or idk, if AnJu addresses it now, that's also fine but we should change all the instances of `var` in this whole file

aloverso (approved):

PR #5545, opened 2023-03-30
feat: [#184306201] poppy auto-name check

somabadri:
### Ticket

## Code author checklist


somabadri:
I have separated the `NameAvailability` type into two versions, `NameAvailability` and `NameAvailabilityResponse` the response doesn't contain a timestamp and contains only the fields returned from a search call to the `ApiBusinessNameClient`. The `NameAvailability` type is only used in regards to the response of the `search` call to `ApiBusinessNameClient`. I originally had only the `NameAvailability` type and the return from the search call would return an object with the timestamp set to undefined which then would overwritten in `searchAndAddTimeStampToBusinessSearch` but it felt incorrect as the return type didn't reflect the information relayed back from ApiBusinessNameClient, so I opted out to creating two types. Thoughts?

api/src/domain/user/searchAndAddTimeStampToBusinessSearch.ts
```diff
@@ -0,0 +1,33 @@
+import { NameAvailabilityResponse, NameAvailabilityStatus } from "@shared/businessNameSearch";
+import { getCurrentDateISOString } from "@shared/dateHelpers";
+import { BusinessNameClient } from "../types";
+
+export const searchAndAddTimeStampToBusinessSearch = async (
+  businessName: string,
+  businessNameClient: BusinessNameClient
+): Promise<BusinessNameSearchResponse> => {
```
aloverso, line 8: I would encourage you to follow the pattern we have for other domain functions that they are a function that returns a function.

The function signature here is a little strange to use because it requires a business name (something that changes on every single query) and a businessNameClient (something that is an object that is always the same).

The reason we have a function-that-returns-a-function is allow the top-level (the app.ts file) create an inject all the necessary dependencies. Here, we're requiring the caller of this function (the routers) to inject the dependency instead of just setting them up ahead of time.

This will simplify the tests you wrote for the routers, which are right now intrinsically testing the behavior of this function as well. Instead, you would pass an instance of a searchAndAddTimeStampToBusinessSearch function as a dependency to the router (instead of a business name client). Think of it this way. The router is never calling a function on the businessNameClient - and it never should. it should ALWAYS use this searchAndAddTimeStampToBusinessSearch to search a name. So why are we giving it access to the whole client? (fun fact - in SOLID principles, this this the "I")

aloverso, line 8: I would encourage you to follow the pattern we have for other domain functions that they are a function that returns a function.

The function signature here is a little strange to use because it requires a business name (something that changes on every single query) and a businessNameClient (something that is an object that is always the same).

The reason we have a function-that-returns-a-function is allow the top-level (the app.ts file) create an inject all the necessary dependencies. Here, we're requiring the caller of this function (the routers) to inject the dependency instead of just setting them up ahead of time.

This will simplify the tests you wrote for the routers, which are right now intrinsically testing the behavior of this function as well. Instead, you would pass an instance of a searchAndAddTimeStampToBusinessSearch function as a dependency to the router (instead of a business name client). Think of it this way. The router is never calling a function on the businessNameClient - and it never should. it should ALWAYS use this searchAndAddTimeStampToBusinessSearch to search a name. So why are we giving it access to the whole client? (fun fact - in SOLID principles, this this the "I")

shared/src/businessNameSearch.ts
```diff
@@ -6,8 +6,15 @@ export type NameAvailabilityStatus =
...
+  status: NameAvailabilityStatus;
+  similarNames: string[];
+  invalidWord?: string;
+};
+
 export type NameAvailability = {
   status: NameAvailabilityStatus;
   similarNames: string[];
   invalidWord?: string;
+  lastUpdatedTimeStamp: string;
```
aloverso, line 19: could make use of `extends` here if you wanted

shared/src/test/formationFactories.ts
```diff
@@ -142,7 +142,26 @@ export const generateBusinessNameAvailability = (overrides: Partial<NameAvailabi
     undefined,
   ];
   return {
-    similarNames: [`random-name-${randomInt()}`],
+    similarNames: [`some-name-${randomInt()}`],
+    status: statusValues[randomIntFromInterval("0", "5")],
```
aloverso, line 146: I don't think this is an ideal usage of `randomIntFromInterval` which was designed for zip codes. 

we have a `randomElementFromArray` fn that might suit better. You could filter out the undefined before calling it

shared/src/businessNameSearch.ts
```diff
@@ -6,8 +6,20 @@ export type NameAvailabilityStatus =
...
   status: NameAvailabilityStatus;
   similarNames: string[];
   invalidWord?: string;
+}
+
+export interface NameAvailability extends NameAvailabilityResponse {
+  lastUpdatedTimeStamp: string;
+}
+
+export type BusinessNameSearchResponse = {
```
aloverso, line 19: Why the introduction of the third type? I understood your distinction between `NameAvailability` and `NameAvailabilityResponse` that you described but not sure I understand the purpose of `BusinessNameSearchResponse` also

api/src/api/userRouter.ts
```diff
@@ -101,6 +114,21 @@ export const userRouterFactory = (
...
+          const response = await timeStampBusinessSearch.search(userData.profileData.businessName);
+          userData = {
+            ...userData,
+            formationData: {
+              ...userData.formationData,
+              businessNameAvailability: {
+                status: response.status,
+                similarNames: response.similarNames ?? [],
+                lastUpdatedTimeStamp: response.lastUpdatedTimeStamp,
+                invalidWord: response.invalidWord ?? undefined,
```
aloverso, line 127: `?? undefined` is redundant

api/src/api/userRouter.ts
```diff
@@ -215,6 +229,29 @@ export const userRouterFactory = (
       });
   };
 
+  const updateBusinessNameSearchIfNeeded = async (userData: UserData): Promise<UserData> => {
+    if (shouldUpdateBusinessNameSearch(userData)) {
```
aloverso, line 233: Might be more readable if this was a guard clause? Idk, tradeoff between fewer layers of nesting, which is more readable, against having to switch the logic to `if (!shouldUpdateBusinessNameSearch)` which is slightly less readable because of the "not". Up to you

shared/src/test/formationFactories.ts
```diff
@@ -139,11 +139,28 @@ export const generateBusinessNameAvailability = (overrides: Partial<NameAvailabi
...
     "RESTRICTED_ERROR",
-    undefined,
   ];
   return {
-    similarNames: [`random-name-${randomInt()}`],
-    status: statusValues[randomIntFromInterval("0", "5")],
+    similarNames: [`some-name-${randomInt()}`],
+    status: randomElementFromArray(statusValues),
+    lastUpdatedTimeStamp: getCurrentDateISOString(),
+    ...overrides,
```
aloverso, line 147: This could reduce duplication by using `...generateBusinessNameAvailabilityResponse` if you want

PR #5602, opened 2023-04-07
chore: [#184814236] add ESLint rule to require use of === and !== (over == and !=)

AnJuHyppolite:
## Description

Add ESLint rule to require the use of === and !== (over == and !=)

### Ticket

[#184814236](https://www.pivotaltracker.com/story/show/#184814236)

### Approach

N/A

### Steps to Test

Run `yarn eslint` to determine whether == and != is being used in the codebase.

### Notes

N/A

## Code author checklist

### DIFF
--- a/api/.eslintrc.json
+++ b/api/.eslintrc.json
@@ -19,6 +19,7 @@
   "rules": {
     "@typescript-eslint/no-empty-function": "off",
     "@typescript-eslint/explicit-function-return-type": "warn",
+    "eqeqeq": "warn",
     "prefer-template": "error",
     "unicorn/filename-case": [
       "error",
--- a/api/src/domain/updateSidebarCards.ts
+++ b/api/src/domain/updateSidebarCards.ts
@@ -48,7 +48,7 @@ export const updateSidebarCards: UpdateSidebarCards = (userData: UserData): User
     hideCard("task-progress");
   }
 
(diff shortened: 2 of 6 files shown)


jphechter (approved):
LGTM! 🚀 🌔

jphechter:
Looks like its working!

"I never trust a test I've never seen fail." - @aloverso 

Screen Shot 2023-04-07 at 4 10 18 PM

api/.eslintrc.json
```diff
@@ -19,6 +19,7 @@
   "rules": {
     "@typescript-eslint/no-empty-function": "off",
     "@typescript-eslint/explicit-function-return-type": "warn",
+    "eqeqeq": "error",
```
aloverso, line 22: Something to consider might be to set this to "warn" instead of "error"

Either one will fail the `yarn lint` command and thus will fail in CircleCI (which is what we want).

The difference is that depending on your personal IDE settings, an eslint "error" will highlight red as you're writing the code, whereas an eslint "warn" will not "scream at you" in the IDE. It'll likely show up as a yellow suggestion instead of a red error. 

Either way, everyone would have to fix before pushing their code, which is good. But the warn/error would just change how it displays in the IDE before it's fixed. Personal preference.

aloverso (approved):

PR #5603, opened 2023-04-07
feat: [#183443674] update c-corp formation

jphechter:
NOTE: Waiting on feedback from the PrepareFiling API team. Currently, cannot successfully submit. Getting the following error:

```
[{"Valid":false,"ErrorMessage":"Required","Name":"Formation.ForeignGoodStandingFile"}]
```

---

## Description

Updates the flow for foreign c-corp formation.

### Ticket

[183443674](https://www.pivotaltracker.com/story/show/183443674)

### Approach

- Create new components associated with new formation questions
  - `PracticesLaw`
  - `ForeignCertificate`
- Create a new `FileInput` component using the USWDS component and functionality as a launchpad
- Established new types and updated config files as necessary
- Updated existing tests to be more flexible and to accommodate for the new `foreignGoodStandingFile` that will be passed around alongside formation data

We are not storing the `foreignGoodStandingFile` but it needs to be maintained in state while the user is filling out the form prior to submission. To accomplish this, we created a new state object. Passing this around required updating a number of functions to accept a props object (easier to handle optional props) and modifying some types.

### Steps to Test

You can experiment with this work by creating a foreign c-corp. The UI changes should be on the formation Business step and Review step. Clicking submit should hit the NICUSA API. Check for errors and successful submission to NICUSA.

### Notes

n/a

## Code author checklist

  - This does in fact use a relative import to get the USWDS library to play nicely. 

### DIFF
--- a/api/src/client/ApiFormationClient.test.ts
+++ b/api/src/client/ApiFormationClient.test.ts
@@ -16,6 +16,7 @@ import axios from "axios";
 import {
   generateFormationData,
   generateFormationUserData,
+  generateInputFile,
   generateProfileData,
   generateUserData,
 } from "../../test/factories";
@@ -567,13 +568,19 @@ describe("ApiFormationClient", () => {
           formationFormData
         );
 
-        await client.form(userData, "hostname.com/form-business");
+        const foreignGoodStandingFile = generateInputFile({});
+
+        await client.form(userData, "hostname.com/form-business", foreignGoodStandingFile);
 
         expect(mockAxios.post).toHaveBeenCalledWith("example.com/formation/PrepareFiling", {
           Account: "12345",
           Key: "abcdef",
           ReturnUrl: "hostname.com/form-business?completeFiling=true",
           FailureReturnUrl: "hostname.com/form-business?completeFiling=false",
+          ForeignGoodStandingFile: {
+            Extension: foreignGoodStandingFile.fileType,
+            Content: foreignGoodStandingFile.base64Contents,
+          },
           Payer: {
             CompanyName: formationFormData.businessName,
             Address1: "",
@@ -593,7 +600,6 @@ describe("ApiFormationClient", () => {
 
             BusinessInformation: {
               CompanyOrigin: "Foreign",
-              PracticesLaw: false,
               Business: "ForeignForProfitCorporation",
               BusinessName: formationFormData.businessName,
               BusinessDesignator: formationFormData.businessSuffix,
@@ -663,6 +669,92 @@ describe("ApiFormationClient", () => {
             ContactLastName: formationFormData.contactLastName,
             ContactPhoneNumber: formationFormData.contactPhoneNumber,
           },
+        });
+      });
+
+      describe("getPracticesLaw", () => {
+        let stubResponse: ApiResponse;
+
+        beforeEach(() => {
+          jest.resetAllMocks();
+          stubResponse = generateApiResponse({});
+          mockAxios.post.mockResolvedValue({ data: stubResponse });
+        });
+
+        it("PracticesLaw is undefined if not provided", async () => {
+          const legalStructureId = "c-corporation";
+          const formationFormData = generateFormationFormData(
+            {},
+            { legalStructureId: `foreign-${legalStructureId}` }
+          );
+
+          const userData = generateFormationUserData(
+            { legalStructureId, businessPersona: "FOREIGN" },
+            {},
+            formationFormData
+          );
+
+          await client.form(userData, "hostname.com/form-business");
+
+          const payload = mockAxios.post.mock.calls[0][1] as ApiSubmission;
+          expect(payload.Formation.BusinessInformation.PracticesLaw).toBe(undefined);
+        });
+
+        it("PracticesLaw is 'Yes' if willPracticeLaw is true", async () => {
+          const legalStructureId = "c-corporation";
+          const formationFormData = generateFormationFormData(
+            { willPracticeLaw: true },
+            { legalStructureId: `foreign-${legalStructureId}` }
+          );
+
+          const userData = generateFormationUserData(
+            { legalStructureId, businessPersona: "FOREIGN" },
+            {},
+            formationFormData
+          );
+
+          await client.form(userData, "hostname.com/form-business");
+
+          const payload = mockAxios.post.mock.calls[0][1] as ApiSubmission;
+          expect(payload.Formation.BusinessInformation.PracticesLaw).toBe("Yes");
+        });
+
+        it("PracticesLaw is 'No' if willPracticeLaw is false", async () => {
+          const legalStructureId = "c-corporation";
+          const formationFormData = generateFormationFormData(
+            { willPracticeLaw: false },
+            { legalStructureId: `foreign-${legalStructureId}` }
+          );
+
+          const userData = generateFormationUserData(
+            { legalStructureId, businessPersona: "FOREIGN" },
+            {},
+            formationFormData
+          );
+
+          await client.form(userData, "hostname.com/form-business");
+
+          const payload = mockAxios.post.mock.calls[0][1] as ApiSubmission;
+          expect(payload.Formation.BusinessInformation.PracticesLaw).toBe("No");
+        });
+
+        it("PracticesLaw is undefined if not foreign c-corp", async () => {
+          const legalStructureId = "s-corporation";
+          const formationFormData = generateFormationFormData(
+            { willPracticeLaw: true },
+            { legalStructureId: `${legalStructureId}` }
+          );
+
+          const userData = generateFormationUserData(
+            { legalStructureId, businessPersona: "OWNING" },
+            {},
+            formationFormData
+          );
+
+          await client.form(userData, "hostname.com/form-business");
+
+          const payload = mockAxios.post.mock.calls[0][1] as ApiSubmission;
+          expect(payload.Formation.BusinessInformation.PracticesLaw).toBe(undefined);
         });
       });
     });
--- a/api/src/client/ApiFormationClient.ts
+++ b/api/src/client/ApiFormationClient.ts
@@ -3,12 +3,12 @@ import { getCurrentDateISOString, parseDateWithFormat } from "@shared/dateHelper
 import { defaultDateFormat } from "@shared/defaultConstants";
 import {
   BusinessSuffix,
-  ForeignGoodStandingFileObject,
   formationApiDateFormat,
   FormationLegalType,
   FormationSubmitError,
   FormationSubmitResponse,
   GetFilingResponse,
+  InputFile,
   PaymentType,
   SignerTitle,
 } from "@shared/formationData";
@@ -27,8 +27,12 @@ type ApiConfig = {
 
 export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): FormationClient => {
   const logId = logger.GetId();
-  const form = (userData: UserData, returnUrl: string): Promise<FormationSubmitResponse> => {
-    const postBody = makePostBody(userData, returnUrl, config);
+  const form = (
+    userData: UserData,
+    returnUrl: string,
+    foreignGoodStandingFile: InputFile | undefined
+  ): Promise<FormationSubmitResponse> => {
+    const postBody = makePostBody(userData, returnUrl, config, foreignGoodStandingFile);
     logger.LogInfo(
       `Formation - NICUSA - Id:${logId} - Request Sent to ${
         config.baseUrl
@@ -148,7 +152,12 @@ export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): Fo
       });
   };
 
-  const makePostBody = (userData: UserData, returnUrl: string, config: ApiConfig): ApiSubmission => {
+  const makePostBody = (
+    userData: UserData,
+    returnUrl: string,
+    config: ApiConfig,
+    foreignGoodStandingFile: InputFile | undefined
+  ): ApiSubmission => {
     const formationFormData = userData.formationData.formationFormData;
 
     const isManual = formationFormData.agentNumberOrManual === "MANUAL_ENTRY";
@@ -221,12 +230,24 @@ export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): Fo
       return {};
     };
 
+    const getPracticesLaw = (): "Yes" | "No" | undefined => {
+      if (isForeignCorp && formationFormData.willPracticeLaw !== undefined) {
+        return formationFormData.willPracticeLaw ? "Yes" : "No";
+      }
+      return undefined;
+    };
+
     return {
       Account: config.account,
       Key: config.key,
       ReturnUrl: `${returnUrl}?completeFiling=true`,
       FailureReturnUrl: `${returnUrl}?completeFiling=false`,
-      ForeignGoodStandingFile: isForeignCorp ? formationFormData.foreignGoodStandingFile : undefined,
+      ForeignGoodStandingFile: foreignGoodStandingFile
+        ? {
+            Extension: foreignGoodStandingFile.fileType,
+            Content: foreignGoodStandingFile.base64Contents,
+          }
+        : undefined,
       Payer: {
         CompanyName: formationFormData.businessName,
         Address1: isForeign ? "" : formationFormData.addressLine1,
@@ -251,7 +272,7 @@ export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): Fo
           BusinessName: formationFormData.businessName,
           BusinessDesignator: formationFormData.businessSuffix as Exclude<BusinessSuffix, undefined>,
           Naic: naicsCode,
-          PracticesLaw: isForeignCorp ? false : undefined,
+          PracticesLaw: getPracticesLaw(),
           ForeignStateOfFormation: formationFormData.foreignStateOfFormation,
           ForeignDateOfFormation: formationFormData.foreignDateOfFormation
             ? parseDateWithFormat(formationFormData.foreignDateOfFormation, defaultDateFormat).format(
@@ -413,7 +434,7 @@ interface Formation extends Provisions {
     Business: BusinessType; //Business Type (DomesticLimitedLiabilityCompany)
     BusinessName: string; //The requested business name, must be available and not contain any restricted words. Name plus designator length must be less than 100 characters in length.
     BusinessDesignator: BusinessSuffix; //The designator - LLC. L.L.C etc
-    PracticesLaw: boolean | undefined;
+    PracticesLaw: "Yes" | "No" | undefined;
     Naic: string; // If supplied must be 6 digits
     BusinessPurpose: string | undefined; // Max 300 chars
     EffectiveFilingDate: string; // date mm/dd/yyyy
@@ -461,7 +482,10 @@ export interface ApiSubmission {
     Email?: string;
   };
   Formation: Formation;
-  ForeignGoodStandingFile?: ForeignGoodStandingFileObject;
+  ForeignGoodStandingFile?: {
+    Extension: string;
+    Content: string;
+  };
 }
 
 type MembersObjects = {
--- a/content/src/fieldConfig/business-formation.json
+++ b/content/src/fieldConfig/business-formation.json
@@ -205,6 +205,13 @@
         "label": "Original State of Formation",
         "error": "Select a State"
       },
+      "foreignGoodStandingFile": {
+        "label": "Foreign Certificate of Good Standing",
+        "helperText": "Select a PNG or PDF file. Max file size: 3MB.",
+        "errorMessageFileSize": "${fileName} exceeds maximum size of ${maxFileSize} MB.",
+        "errorMessageFileType": "${fileName} file format is not supported.",
+        "errorMessageRequired": "A Certificate of Good Standing is required."
+      },
       "createLimitedPartnerTerms": {
         "label": "Partnership Rights - Assign rights terms"
       },
@@ -293,6 +300,12 @@
             "description": "`Authorized representative(s)|authorized-representative` or `general partner(s)|general-partner` should sign below by adding their full name. You can have up to 10 signers listed."
           }
         }
+      },
+      "willPracticeLaw": {
+        "label": "Will you practice law?",
+        "radioNoText": "No",
+        "radioYesText": "Yes",
+        "error": "Select 'Yes' or 'No'."
       },
       "withdrawals": {
         "label": "Withdrawals",
--- a/shared/src/formationData.ts
+++ b/shared/src/formationData.ts
@@ -98,11 +98,6 @@ export interface FormationMember extends FormationAddress {
   readonly name: string;
 }
 
-export type ForeignGoodStandingFileObject = {
-  Extension: "PDF" | "PNG";
-  Content: string; // Binary contents converted to base64 string 4,000,000 string limit (3mb)
-};
-
 export interface FormationFormData extends FormationAddress {
   readonly legalType: FormationLegalType;
   readonly businessName: string;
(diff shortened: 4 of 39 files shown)


api/src/client/ApiFormationClient.test.ts
```diff
@@ -567,13 +568,22 @@ describe("ApiFormationClient", () => {
           formationFormData
         );
 
-        await client.form(userData, "hostname.com/form-business");
+        const foreignGoodStandingFile = generateInputFile({
+          fileType: "PNG",
+          base64Contents: "this is my string",
+        });
```
aloverso, line 574: I don't think we need to specify the fileType and base64Contents? We're never testing for them later on. I tend to push for reducing/removing any extraneous info from tests so that everything that is present has a reason to be. I think here we could just let the random generator do its work, we don't care the exact values.

api/src/client/ApiFormationClient.test.ts
```diff
@@ -593,7 +603,7 @@ describe("ApiFormationClient", () => {
 
             BusinessInformation: {
               CompanyOrigin: "Foreign",
-              PracticesLaw: false,
+              PracticesLaw: "Yes",
```
aloverso, line 606: Why do we expect this to be "Yes"? I don't see that anywhere in the setup. Shouldn't this come from setting it to `true` somewhere in the form data during setup?

api/src/client/ApiFormationClient.ts
```diff
@@ -251,7 +265,7 @@ export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): Fo
           BusinessName: formationFormData.businessName,
           BusinessDesignator: formationFormData.businessSuffix as Exclude<BusinessSuffix, undefined>,
           Naic: naicsCode,
-          PracticesLaw: isForeignCorp ? false : undefined,
+          PracticesLaw: isForeignCorp ? "Yes" : undefined,
```
aloverso, line 268: Re my comment in the test file. Looks like we hardcoded this to "Yes" temporarily and never came back to actually set this value based on the formationFormData. This should be updated

shared/src/formationData.ts
```diff
@@ -145,10 +142,11 @@ export interface FormationFormData extends FormationAddress {
   readonly contactPhoneNumber: string;
   readonly foreignStateOfFormation: StateNames | undefined;
   readonly foreignDateOfFormation: string | undefined; // YYYY-MM-DD
-  readonly foreignGoodStandingFile: ForeignGoodStandingFileObject | undefined;
+  readonly willPracticeLaw: "Yes" | "No" | undefined;
```
aloverso, line 145: why are we redefining this when we have the value up on line 101?
Also, I would argue we don't want to store this as "Yes" and "No" but rather as a boolean true/false. The Yes/No is how the API accepts it but since this is our domain layer, we have the freedom to choose a more appropriate representation for our type. And being able to write things like `if (willPracticeLaw)` is more in line with how we expect things to work, rather than having to do `if (willPracticeLaw === "Yes")`

shared/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "./tsconfig.paths.json",
   "compilerOptions": {
     "esModuleInterop": true,
-    "lib": ["ESNext"],
+    "lib": ["ESNext", "DOM"],
```
aloverso, line 5: What is this for?

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,144 @@
...
+  PNG: "image/png",
+};
+
+const InputStringToFileType: Record<string, AcceptedFileType> = (
+  Object.keys(FileTypeToInputString) as AcceptedFileType[]
+).reduce((acc, key) => {
+  const value = FileTypeToInputString[key];
+  acc[value] = key;
+  return acc;
+}, {} as Record<string, AcceptedFileType>);
```
aloverso, line 26: Might be nice to extract this into a generic "flip dictionary" helper method

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export const FileInput = (props: Props): ReactElement => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const fileInput = require("../../../../node_modules/@uswds/uswds/packages/usa-file-input/src");
```
aloverso, line 30: a) Is this really how the USWDS wants us to use this component?
b) Why does it need to be here instead of a regular top-level import?

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const BYTES_IN_A_MB = 1048576;
```
aloverso, line 32: could go outside the function so it's not constantly redefined. Small tweak

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,144 @@
...
+    // Prevent the user from selecting a file that is too large or of the wrong file type
+    const preventInvalidFileSelection = (e: Event): void => {
+      const eventTarget = e.target as HTMLInputElement;
+      if (eventTarget.files && eventTarget.files.length > 0) {
+        const selectedFile = eventTarget.files[0];
+        if (props.accept && !props.accept.includes(InputStringToFileType[selectedFile.type])) {
+          e.stopImmediatePropagation();
+          setFileSizeError({
+            hasError: true,
+            errorMessage: `${selectedFile.name} file format is not supported.`,
```
aloverso, line 78: I think this error message (and the one below it) should be externalized into the config files. They can still use the `${}` string interpolation. There's a helper called `templateEval` that allows us to do that with config strings

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,144 @@
...
+    // Event listener is added to the file input element ahead of USWDS listeners - React onChange will be added after
+    fileInputElement?.addEventListener("change", (e: Event) => {
+      preventInvalidFileSelection(e);
+    });
+    if (typeof fileInput.on === "function" && typeof fileInput.off === "function") {
+      fileInput.on(fileInputElement); // initialize USWDS fileInput component
+    }
+  }, [fileInput, props.accept, size, props.maxSizeInMegabytes]);
+
+  useEffect(() => {
```
aloverso, line 99: because this is a weird useEffect, it might be nice to use the named-function approach to a useEffect here, so that we can give it a function name to help make it more readable & explain what's going on for people reading it

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export const FileInput = (props: Props): ReactElement => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const fileInput = require("../../../../node_modules/@uswds/uswds/packages/usa-file-input/src");
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const BYTES_IN_A_MB = 1048576;
+  const size = props.maxSizeInMegabytes ? props.maxSizeInMegabytes * BYTES_IN_A_MB : undefined;
+
+  const [hasFileBeenRead, setHasFileBeenRead] = useState<boolean>(false);
+  const [fileSizeError, setFileSizeError] = useState({ hasError: false, errorMessage: "" });
```
aloverso, line 36: It seems a little misleading to call this a `fileSizeError` when it also represents a file type error as well. Maybe just `fileUploadError` or something similar?

web/src/components/tasks/business-formation/business/MainBusiness.tsx
```diff
@@ -45,26 +47,34 @@ export const MainBusiness = (): ReactElement => {
...
+              <FormationDate fieldName="foreignDateOfFormation" />
+            </WithErrorBar>
           </WithErrorBar>
-        </WithErrorBar>
+          {state.formationFormData.legalType === "foreign-c-corporation" && (
+            <PracticesLaw hasError={doesFieldHaveError("willPracticeLaw")} />
+          )}
+          {state.formationFormData.legalType === "foreign-c-corporation" && (
+            <ForeignCertificate hasError={doesFieldHaveError("foreignGoodStandingFile")} />
+          )}
```
aloverso, line 76: These could both go within the same logic check instead of repeating it?

aloverso: Also - I'm noticing that the other components like `FormationDate` and `ForeignStateOfFormation` are independent components, and then are wrapped in the WithErrorBar here in the parent-display layout. Whereas you included the WithErrorBar wrapper internal to the ForeignCertificate. 

I'd argue that makes more sense to keep it external because the error bar relies on assumptions about layout. Specifically the MOBILE/DESKTOP view. If the file input were side-by-side with another component, the chosen "ALWAYS" prop would have to change, and that would be hard to track internally. Hence my argument for keeping the ErrorBar at the top level like we do for the other components here.

web/src/components/tasks/business-formation/business/PracticesLaw.tsx
```diff
@@ -0,0 +1,72 @@
...
+        willPracticeLaw: value,
+      };
+    });
+  };
+
+  return (
+    <WithErrorBar hasError={hasError} type="ALWAYS">
+      <FormControl variant="outlined" fullWidth className="padding-bottom-2">
+        <RadioGroup
+          aria-label={"Foreign address type"}
```
aloverso, line 30: copy paste error? Usually we use `camelCaseToSentenceCase` of the field name for aria labels

web/src/components/tasks/business-formation/review/ReviewForeignCertificate.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <>
+      <ReviewSectionHeader
+        header="Foreign Certificate of Good Standing"
+        stepName="Business"
+        testId="foreign-certificate"
+      />
+      {props.foreignGoodStandingFile?.filename ? (
+        <div style={{ display: "flex", alignItems: "center" }}>
+          <img
+            src="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7"
```
aloverso, line 22: what is this?

web/src/components/tasks/business-formation/review/ReviewForeignCertificate.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface Props {
+  foreignGoodStandingFile: InputFile | undefined;
+}
+
+export const ReviewForeignCertificate = (props: Props): ReactElement => {
+  const { Config } = useConfig();
+  return (
+    <>
+      <ReviewSectionHeader
+        header="Foreign Certificate of Good Standing"
```
aloverso, line 15: should be based on the config label?

web/src/components/tasks/business-formation/review/ReviewStep.tsx
```diff
@@ -30,6 +32,12 @@ export const ReviewStep = (): ReactElement => {
       <div data-testid="review-step">
         <BusinessNameAndLegalStructure isReviewStep />
         <ReviewBusinessSuffixAndStartDate />
+        {isForeignCCorp && (
+          <>
+            <ReviewText header="Will you practice law?" fieldName="willPracticeLaw" stepName="Business" />
```
aloverso, line 37: should be based on the config label?

web/src/lib/api-client/apiClient.ts
```diff
@@ -30,8 +31,13 @@ export const checkLicenseStatus = (nameAndAddress: NameAndAddress): Promise<User
   return post(`/license-status`, nameAndAddress);
 };
 
-export const postBusinessFormation = (userData: UserData, returnUrl: string): Promise<UserData> => {
-  return post(`/formation`, { userData, returnUrl });
+// Pass file obj here
```
aloverso, line 34: delete comment

web/src/styles/base/global.scss
```diff
@@ -209,6 +209,10 @@ ol {
   }
 }
 
+.usa-file-input {
+  max-width: unset;
+}
```
aloverso, line 214: Is this overriding the default? If so, did you check for sure it works / is actually overridden in browser? usually in order to to override an njwds setting, we have to use `!important`. I'm surprised to see this would work like this

web/test/helpers/helpers-formation.tsx
```diff
@@ -180,6 +180,8 @@ export type FormationPageHelpers = {
   fillAndSubmitAddressModal: (overrides: Partial<FormationSignedAddress>, fieldName: string) => Promise<void>;
   clickSubmit: () => Promise<void>;
   selectDate: (value: DateObject, fieldType: "Business start date" | "Foreign date of formation") => void;
+  uploadFile: (file: File) => void;
+  willPracticeLaw: (response?: boolean) => void;
```
aloverso, line 184: I think it'd be more readable if this function was named something that clearly described the action being performed.

api/src/client/ApiFormationClient.test.ts
jphechter: Fair point. Included!

api/src/client/ApiFormationClient.ts
jphechter: 👀

shared/tsconfig.json
jphechter: I had added that to use the `File` type defined by React. I don't think I need this any more.

web/src/styles/base/global.scss
jphechter: Yup, this is overriding the default value. This is pulling styles from the USWDS package. I'm not sure how NJWDS implements USWDS.
Screenshot 2023-04-11 at 10 05 07 AM

web/src/components/njwds-extended/FileInput.tsx
jphechter: included

web/src/components/tasks/business-formation/review/ReviewForeignCertificate.tsx
jphechter: This is the preview image that's used by the USWDS file-input component. I used it here for continuity. I could extrapolate this to a global variable or reference it some other way if you'd like.

web/src/components/tasks/business-formation/business/MainBusiness.tsx
jphechter: That reasoning makes sense, but I'm leaving `WithErrorBar` nested. I see this pattern in other components like `BusinessFormationTextField`. These components really only exist to be a part of this form and aren't affected by the surrounding items. I think it just eliminates a little clutter from `MainBusiness`.

web/src/components/njwds-extended/FileInput.tsx
jphechter: included

web/src/components/njwds-extended/FileInput.tsx
jphechter: included

api/src/client/ApiFormationClient.test.ts
```diff
@@ -593,11 +600,11 @@ describe("ApiFormationClient", () => {
 
             BusinessInformation: {
               CompanyOrigin: "Foreign",
-              PracticesLaw: false,
               Business: "ForeignForProfitCorporation",
               BusinessName: formationFormData.businessName,
               BusinessDesignator: formationFormData.businessSuffix,
               Naic: userData.profileData.naicsCode,
+              PracticesLaw: "No",
```
aloverso, line 607: Same comment as before - shouldn't this be driven by `willPracticeLaw` being set to `false` somewhere in the setup section of this test?

content/src/fieldConfig/business-formation.json
```diff
@@ -205,6 +205,15 @@
...
         "error": "Select a State"
       },
+      "foreignGoodStandingFile": {
+        "label": "Foreign Certificate of Good Standing",
+        "helperText": "Select a PNG or PDF file. Max file size: 3MB.",
+        "errors": {
+          "fileSize": "${fileName} exceeds maximum size of ${maxFileSize} MB.",
+          "fileType": "${fileName} file format is not supported.",
+          "required": "A Certificate of Good Standing is required."
+        }
```
aloverso, line 215: Not a huge thing, but something for consideration: sometimes nesting in the CMS ui is helpful to organize things and close extraneous information. But the more nesting we do, the more grey bars and boxes that Netlify creates on the screen. 

So I tend to nest things when it is clear that you'd want to look at one thing while being able to collapse unrelated info. I would argue that a content team member looking at the Foreign Good Standing File would want to see all the fields, and there's not much value in being able to see label & helpertext but collapse the error fields. So at that point, the extra boxes and nesting in the CMS UI is more visually distracting than helpful. If you haven't looked at the CMS to see what I mean, I recommend it. Just a thoughtt.

web/src/components/njwds-extended/FileInput.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+        value={undefined}
+      />
+    );
+
+    const uploader = screen.getByTestId("file-input");
+
+    fireEvent.change(uploader, { target: { files: [file] } });
+
+    expect(mockOnChange).not.toHaveBeenCalled();
+    expect(screen.getByText("cool.png file format is not supported.")).toBeInTheDocument();
```
aloverso, line 83: If this is in config, we should use the config in this. If content changes the message, then this test will start failing

aloverso: and the one above

web/src/components/tasks/business-formation/business/PracticesLaw.tsx
```diff
@@ -0,0 +1,70 @@
...
+    setFormationFormData((previousState) => {
+      return {
+        ...previousState,
+        willPracticeLaw,
+      };
+    });
+  };
+
+  return (
+    <WithErrorBar hasError={hasError} type="ALWAYS">
```
aloverso, line 28: Same comment about withErrorBar as the other component

web/src/components/njwds-extended/FileInput.test.tsx
jphechter: I provided a default value to the `errorMessageFileSize` and `errorMessageFileType` values because these errors require multiple currently optional params. My thinking was that I was trying to test the defaults which is why I didn't pull from the Config, but that's really not what this test is currently doing.

Maybe I'm over-engineering this. It might not be entirely obvious where the default provided copy is coming from. I was trying to account for an edge case that might not really be a big deal.

api/src/client/ApiFormationClient.test.ts
jphechter: Fixed! Thank you.

content/src/fieldConfig/business-formation.json
jphechter: I haven't poked a ton around in there. I'll de-structure this.

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { ReactElement, useEffect, useLayoutEffect, useRef, useState } from "react";
+
+interface Props {
+  accept?: AcceptedFileType[];
+  errorMessageFileSize?: string;
+  errorMessageFileType?: string;
+  errorMessageRequired: string;
+  hasError?: boolean;
+  helperText: string;
+  maxSizeInMegabytes?: number; // filesize in bytes (1,048,576 = 1 MB)
```
aloverso, line 12: comment is incorrect, should be deleted

web/src/components/tasks/business-formation/review/ReviewForeignCertificate.tsx
aloverso: lets give it a variable name so other people can know this also

web/src/components/tasks/business-formation/review/ReviewForeignCertificate.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { Config } = useConfig();
+  return (
+    <ReviewSection
+      buttonText={Config.formation.general.editButtonText}
+      header={Config.formation.fields.foreignGoodStandingFile.label}
+      stepName="Business"
+      testId="edit-foreign-certificate-step"
+    >
+      {props.foreignGoodStandingFile?.filename ? (
+        <div style={{ display: "flex", alignItems: "center" }}>
```
aloverso, line 20: classname "fac" in flex.scss

web/src/components/tasks/business-formation/business/PracticesLaw.tsx
```diff
@@ -0,0 +1,67 @@
...
+interface Props {
+  hasError: boolean;
+}
+
+export const PracticesLaw = (props: Props): ReactElement => {
+  const { hasError } = props;
+  const { Config } = useConfig();
+  const { state, setFormationFormData } = useContext(BusinessFormationContext);
+
+  const onChange = (value: "true" | "false"): void => {
```
aloverso, line 16: Just noticed this - can we just use a boolean instead of string values?

api/src/client/ApiFormationClient.test.ts
```diff
@@ -665,6 +671,92 @@ describe("ApiFormationClient", () => {
...
+      describe("getPracticesLaw", () => {
+        let stubResponse: ApiResponse;
+
+        beforeEach(() => {
+          jest.resetAllMocks();
+          stubResponse = generateApiResponse({});
+          mockAxios.post.mockResolvedValue({ data: stubResponse });
+        });
+
+        it("PracticesLaw is undefined if not provided", async () => {
```
aloverso, line 684: not a big deal but just a note that we try to write test names as a sentence starting with "it..."

aloverso: The reason for this is that it's a technique to encourage behavioral testing. As a principle, we always want to be testing some _output_ or behavior of the thing under test. So if we can describe what "it" does - that's its behavior. And if there's something that we can't put in this format, it's probably not actually observable behavior and we should consider how to refactor the test

web/src/components/njwds-extended/FileInput.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+        filename: "cool.png",
+      });
+    });
+  });
+
+  it("refuses files larger than the provided max file size", async () => {
+    const mockOnChange = jest.fn();
+    const errorMessageFileSize = Config.formation.fields.foreignGoodStandingFile.errorMessageFileSize;
+
+    file = new File(["x".repeat(1048576 / 2)], "cool.png", { type: "image/png" }); // .5 MB file
```
aloverso, line 44: this would be an example where I recommend a variable named `const halfMegabyteFile = "x".repeat...` instead of using a comment, to have the code be "self-documenting"

web/src/components/njwds-extended/FileInput.test.tsx
aloverso: Is this comment still relevant? We want to test that this value came from config, right? Since there's no default anymore?

web/src/components/njwds-extended/FileInput.tsx
```diff
@@ -0,0 +1,170 @@
...
+        setFileUploadError({ hasError: false, errorMessage: "" });
+      };
+      reader.readAsDataURL(uploadedFile);
+    }
+  };
+
+  useLayoutEffect(() => {
+    const fileInputElement = fileInputRef.current;
+
+    // Prevent the user from selecting a file that is too large or of the wrong file type
```
aloverso, line 88: No need for this comment because the function name is self-evident

web/src/components/tasks/business-formation/review/ReviewForeignCertificate.tsx
```diff
@@ -0,0 +1,35 @@
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { InputFile } from "@businessnjgovnavigator/shared/formationData";
+import { ReactElement } from "rehype-react/lib";
+import { ReviewSection } from "./section/ReviewSection";
+
+interface Props {
+  foreignGoodStandingFile: InputFile | undefined;
+}
+
+const filePreviewImage = "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7"; // taken from USWDS
```
aloverso, line 10: could be called `uswdsFileImagePreview` to remove the need for a comment

jphechter:
Rebased, but will follow up on comments shortly.

PR #5612, opened 2023-04-10
chore: [#184305938] view more / less button added to formation review…

FarazA22:
## Description

I created a wrapper component that takes a string and can restrict the number of lines can wrap. The component has a view more and view less button that expands and collapses the text. 

### Ticket

[184305938](https://www.pivotaltracker.com/story/show/184305938)

### Approach

I leverage a popular yarn package.

## Code author checklist

### DIFF
--- a/web/src/components/ExpandCollapseString.tsx
+++ b/web/src/components/ExpandCollapseString.tsx
@@ -1,0 +1,27 @@
+import { ReactElement } from "react";
+import ClampLines from "react-clamp-lines";
+
+interface Props {
+  text: string;
+  dataTestId?: string;
+  viewMoreText: string;
+  viewLessText: string;
+  lines: number;
+}
+
+export const ExpandCollapseString = (props: Props): ReactElement => {
+  return (
+    <div {...(props.dataTestId ? { "data-testid": props.dataTestId } : {})}>
+      <ClampLines
+        text={props.text}
+        id={Math.random().toString().slice(2)}
+        lines={props.lines}
+        ellipsis="..."
+        moreText={props.viewMoreText}
+        lessText={props.viewLessText}
+        className="lines-ellipsis"
+        innerElement="p"
+      />
+    </div>
+  );
+};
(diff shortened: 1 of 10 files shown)


web/src/components/ExpandCollapseString.tsx
```diff
@@ -0,0 +1,27 @@
...
+  viewLessText: string;
+}
+
+export const ExpandCollapseString = (props: Props): ReactElement => {
+  return (
+    <div {...(props.dataTestId ? { "data-testid": props.dataTestId } : {})}>
+      <ClampLines
+        text={props.text}
+        id={props.id}
+        lines={2}
```
aloverso, line 18: In order for this to be more reusable, should the number of lines be a prop maybe?

web/src/components/ExpandCollapseString.tsx
```diff
@@ -0,0 +1,27 @@
+import { ReactElement } from "react";
+import ClampLines from "react-clamp-lines";
+
+interface Props {
+  text: string;
+  id: string;
```
aloverso, line 6: Why is it necessary to pass an ID? we don't often use ids on html elements

web/src/components/tasks/business-formation/review/ReviewText.tsx
```diff
@@ -14,6 +15,7 @@ interface Props {
   header: string;
   fieldName: FormationTextField;
   stepName: FormationStepNames;
+  isExpandCollapseStringEnabled?: boolean;
```
aloverso, line 18: could maybe be shortened to `isExpandable` or something?

aloverso:
I'm surprised this isn't failing the dependency-check, since it introduces a new library without updating the dependency checker file. Any idea why?

web/src/components/ExpandCollapseString.tsx
FarazA22: I was debating whether they should be a prop now or in the future when the need arose. I can make the update now.

aloverso: Yeah, I'm usually opposed to extra props until the need arises, totally. I just thought maybe in this case, it was important enough to the usage of the component that it seemed like a good candidate

FarazA22:
> I'm surprised this isn't failing the dependency-check, since it introduces a new library without updating the dependency checker file. Any idea why?

This is a really good question, I usually don't worry about the dependency check until it screams at me. I'll take a look.

PR #5613, opened 2023-04-10
Formation review page reorg 184305952

AnJuHyppolite:
### Ticket

## Code author checklist


content/src/fieldConfig/business-formation.json
```diff
@@ -368,7 +368,8 @@
       "c-corporation": "Domestic Corporation",
       "s-corporation": "Domestic Corporation",
       "foreignPrefaceText": "Foreign",
-      "reviewStepLabel": "`**Business Structure**|business-structure-learn-more`",
+      "reviewStepLabel": "Business Structure",
+      "reviewStepContextualInfo": "business-structure-learn-more",
```
aloverso, line 372: can we refactor this to just be `label` and `labelContextualInfo`? It's actually used on the main page, not just the review step

web/src/components/tasks/business-formation/review/ReviewMembers.tsx
```diff
@@ -27,19 +27,20 @@ export const ReviewMembers = (): ReactElement => {
       buttonText={Config.formation.general.editButtonText}
       header={getConfig().header}
       stepName="Contacts"
-      testId="edit-members-step"
+      datatest-id={"members-step"}
```
aloverso, line 30: what's going on here? Not a valid prop?

web/src/components/tasks/business-formation/review/section/ReviewSection.tsx
```diff
@@ -6,11 +6,13 @@ import { scrollToTop } from "@/lib/utils/helpers";
...
 interface Props {
-  buttonText: string;
+  buttonText?: string;
   handleButtonClick?: () => void;
   header: string;
   stepName: FormationStepNames;
-  testId: string;
+  testId?: string;
+  sectionType: "MAIN" | "SUBSECTION";
+  marginOverride?: string;
```
aloverso, line 15: Correct me if I'm wrong, but here's what it seems:
- Every "MAIN" component has a `buttonText` and no `marginOverride`
- Every "SUBSECTION" component has no `buttonText` and a `margin-y-4` override.

There's maybe one exception I could find (`combinedInvestment` is a subsection that's missing the margin. But I looked at the design and that seems to be an error, it looks like it _should_ have the margin).

This kind of clear delineation between 2 cases, where one case uses one set of props exclusively, and another case uses another set of props exclusively, looks to me like a component that is begging to actually be 2 components. If you want to reduce duplication, you can make use of a shared (abstract) base component, but that's optional and might be more trouble than its worth. 

Basically what I'm saying is, if it would never make sense to have SUBSECTION that has an edit button (and from the designs, it looks like that's the case) then the component should not support that as a configurable option. #makeImpossibleStatesImpossible 

Following this suggestion will make our components more readable, because they won't be full of ternary logic and instead be declarative. Here's what a subsection looks like. Here's what a Main section looks like. No need to juggle so many class names and have a developer struggle to figure out what it'll actually render as.

aloverso: Also, two side notes, in terms of reducing unnecessary props.
- It looks like every time we pass a value to `buttonText`, it's the same Config editText. Why is it a prop if it's always the same?
- It looks like `handleButtonClick` is never used. It's an optional prop and there is no usage with it ever being passed a value. Why do we need it?

web/src/components/tasks/business-formation/review/section/ReviewSection.tsx
```diff
@@ -33,9 +37,9 @@ export const ReviewSection: React.FC<PropsWithChildren<Props>> = ({
 
   return (
     <>
-      <div className="flex space-between">
-        <div className="maxw-mobile-lg margin-bottom-2">
-          <h2 className="h3-styling">{header}</h2>
+      <div className={marginOverride ? `flex space-between ${marginOverride}` : `flex space-between`}>
+        <div className={sectionType === "MAIN" ? "maxw-mobile-lg margin-bottom-2" : "maxw-mobile-lg"}>
```
aloverso, line 41: This comment is a note for the future, because if you follow my suggestion to make these two components, this ternary will go away altogether.

BUT:

If you do find yourself writing a className ternary like this and the one above in the future, then I recommend reducing duplication by only inserting the part that matters. Like so:
```
className={`maxw-mobile-lg ${sectionType === "MAIN" ? "margin-bottom-2" : ""}`}
```

This makes it more future-proof if we add more classes.

Imagine if you had like four more shared class names and the one that changed. In your style, it'd look like:

```
className={sectionType === "MAIN" ? "maxw-mobile-lg class-1 class-2 class-3 class-4 margin-bottom-2" : "maxw-mobile-lg class-1 class-2 class-3 class-4"}
```

in my suggestion, it'd look like this:

```
className={`maxw-mobile-lg class-1 class-2 class-3 class-4 ${sectionType === "MAIN" ? "margin-bottom-2" : ""}`}
```

web/src/components/tasks/business-formation/review/ReviewWillPracticeLaw.tsx
```diff
@@ -18,10 +18,10 @@ export const ReviewWillPracticeLaw = (props: Props): ReactElement => {
   };
   return (
     <ReviewSection
-      buttonText={Config.formation.general.editButtonText}
       header={Config.formation.fields.willPracticeLaw.label}
       stepName="Business"
-      testId="edit-will-practice-law-step"
+      testId="will-practice-law-step"
```
aloverso, line 23: what are these test ids being used for?

web/src/components/tasks/business-formation/review/ReviewStep.tsx
```diff
@@ -32,62 +32,86 @@ export const ReviewStep = (): ReactElement => {
...
         )}
-        <ReviewRegisteredAgent />
-        {hasMembers && !isLP && <ReviewMembers />}
-        <ReviewSignatures />
-        <ReviewBillingContact />
-        <ReviewBillingServices />
+        <hr className="margin-y-205" />
+        <ReviewSection
+          buttonText={Config.formation.general.editButtonText}
+          header={"Contacts"}
```
aloverso, line 96: All these headers should not be hardcoded; they should be pulled from the same Configuration values we use to power the stepper step names

web/src/components/tasks/business-formation/review/section/ReviewMainSection.tsx
```diff
@@ -1,49 +1,39 @@
...
+import { ReactElement, ReactNode, useContext } from "react";
 
 interface Props {
-  buttonText: string;
-  handleButtonClick?: () => void;
   header: string;
   stepName: FormationStepNames;
-  testId: string;
+  testId?: string;
+  children?: ReactNode;
```
aloverso, line 13: children shouldn't be optional, right? we would never have a MainSection without anything in it?

web/src/styles/base/flex.scss
```diff
@@ -92,6 +92,10 @@
   flex-direction: column-reverse;
 }
 
+.flex-column {
```
aloverso, line 95: this is already in this file as `.fdc`

web/src/components/tasks/business-formation/review/section/ReviewSubSection.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface Props {
+  header: string;
+  marginOverride?: string;
+  children?: ReactNode;
+}
+
+export const ReviewSubSection = (props: Props): ReactElement => {
+  return (
+    <>
+      <div className={`flex space-between ${props.marginOverride ?? ""}`}>
```
aloverso, line 12: given that the prop is called `marginOverride` and not `margin`, I think it would make sense to set a "default" margin here instead of empty-string. Still allows people to override margin, which it looks like you do in one or two places? But given that 90% of your usages of this component use "margin-y-4", I would argue for setting that as the default and allowing components that deviate from that norm to override.

web/src/components/tasks/business-formation/review/ReviewStep.tsx
```diff
@@ -32,62 +34,73 @@ export const ReviewStep = (): ReactElement => {
...
+          )}
+          {isLP && <ReviewPartnership />}
+          {isLP && (
+            <ReviewSubSection
+              header={Config.formation.fields.dissolution.label}
+              marginOverride="margin-top-4"
+            >
+              <ReviewText fieldName={"dissolution"} />
+            </ReviewSubSection>
+          )}
```
jphechter, line 76: Could we consolidate these `isLP` assertions?

web/src/components/tasks/business-formation/review/section/ReviewSubSection.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <div className={"maxw-mobile-lg"}>
+          <h2 className={"h3-styling"}>{props.header}</h2>
```
aloverso, line 14: Shouldn't this just be an `h3`? And then you don't need "h3-styling" as a class anyways.

From an accessibility standpoint, the `h` hierarchy should reflect the actual page hierarchy. h3s should be nested inside h2s, and my understanding is that these subsections are nested inside the h2 main sections?

web/src/components/tasks/business-formation/review/section/ReviewSubSection.tsx
```diff
@@ -0,0 +1,20 @@
+import { ReactElement, ReactNode } from "react";
+
+interface Props {
+  header: string;
+  marginOverride?: string;
+  children?: ReactNode;
```
aloverso, line 6: same comment as other - children should never be undefined, right?

web/src/components/tasks/business-formation/review/section/ReviewMainSection.tsx
```diff
@@ -1,49 +1,39 @@
...
 
-export const ReviewSection: React.FC<PropsWithChildren<Props>> = ({
-  children,
-  buttonText,
-  handleButtonClick,
-  header,
-  stepName,
-  testId,
-}) => {
+export const ReviewMainSection = (props: Props): ReactElement => {
```
jphechter, line 16: More of a general nit, but I prefer `ReviewSection` as a name for this component. `ReviewMainSection` to me feels like a generalized wrapper for the whole Review step. Just curious, what was the thought behind the name change?

web/src/styles/base/flex.scss
AnJuHyppolite: Not sure how this showed up in this story, b/c it already existed in the codebase & is used in various components.

web/src/components/tasks/business-formation/review/ReviewBillingContact.tsx
```diff
@@ -10,24 +11,21 @@ export const ReviewBillingContact = (): ReactElement => {
...
-    >
+    <ReviewSubSection header={Config.formation.sections.review.billingContactHeader}>
       <ReviewLineItem
-        label={Config.formation.fields.contactFirstName.label}
+        label={`${Config.formation.fields.contactFirstName.label}:`}
         value={state.formationFormData.contactFirstName || italicNotEnteredText}
       />
       <ReviewLineItem
-        label={Config.formation.fields.contactLastName.label}
+        label={`${Config.formation.fields.contactLastName.label}:`}
```
jphechter, line 20: I'm seeing a number of different places where we're manually adding a `:` to the label. I don't think I'm seeing any where we don't. Should this just be part of the component?

web/src/components/tasks/business-formation/review/section/ReviewSubSection.tsx
AnJuHyppolite: I agree. This was carried over from `ReviewSection` (nka `ReviewMainSection`). I should have brought it to you when I saw it then.

web/src/components/tasks/business-formation/review/ReviewStep.tsx
AnJuHyppolite: Consolidate? Should we create a chore for this versus addressing it in this story?

web/src/components/tasks/business-formation/review/section/ReviewMainSection.tsx
AnJuHyppolite: There was logic in place for a `sectionType` property that could be MAIN or SUBSECTION to differentiate the sections by button text or margin properties. In doing so, we created `ReviewSubsection` and renamed `ReviewSection`.

jphechter: Gotcha, I'd advocate for reverting back to `ReviewSection`. I think that name is still more representative.

web/src/components/tasks/business-formation/review/ReviewStep.tsx
jphechter: This is what I'm proposing.
```suggestion
          {isLP && (
            <ReviewSubSection
              header={Config.formation.fields.combinedInvestment.label}
              marginOverride="margin-top-4"
            >
              <ReviewText fieldName={"combinedInvestment"} />
            </ReviewSubSection>
            <ReviewSubSection
              header={Config.formation.fields.withdrawals.label}
              marginOverride="margin-top-4"
            >
              <ReviewText fieldName={"withdrawals"} />
            </ReviewSubSection>
          <ReviewPartnership />
            <ReviewSubSection
              header={Config.formation.fields.dissolution.label}
              marginOverride="margin-top-4"
            >
              <ReviewText fieldName={"dissolution"} />
            </ReviewSubSection>
          )}
```

AnJuHyppolite: I marked it resolved, b/c I updated the code. That's how I go through the feedback to ensure I have addressed all of the reviewer's comments. You can see the updates when I push the code again.

web/src/components/tasks/business-formation/review/section/ReviewMainSection.tsx
aloverso: @jphechter My concern about just `ReviewSection` is that the name on its own doesn't communicate that it is part of a "pair" and that it goes hand-in-hand with `ReviewSubSection`.

I'm not tied to "Main" in particular in the name, but I'd argue for something that indicates that this is the "top-level" part of a review section, as contrasted with the subsection level

web/src/components/tasks/business-formation/review/ReviewStep.tsx
jphechter: Makes sense. I just didn't see the change. Wanted to make sure this was addressed.

web/src/components/tasks/business-formation/review/section/ReviewMainSection.tsx
jphechter: Totally fair. Not a hill I need to die on, I just don't think "main" sounds like something that will be used multiple times side-by-side. At the end of the day, this does not affect how the code renders.

PR #5635, opened 2023-04-14
feat: [#184306248] formation remembers last user visited page

adidner:
## Description

We wanted to have the application remember the last formation page the user visited so that if they navigate away and come back they will still be on the same step. (This should be reset if the legal structure type is reset). 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/#184306248)

### Steps to Test

Go to the formation page as a user, navigate to a step via the stepper or the save and continue button

navigate away from this page, 

navigate back

you should not be on the first step but the step you left off on,

got to the profile page and change your legal structure (business structure on the page) (NOTE not all business structure have a formation process)

go to formation and see that the formation step has reset 

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
+++ b/web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
@@ -1,5 +1,8 @@
 import { BusinessFormation } from "@/components/tasks/business-formation/BusinessFormation";
-import { LookupStepIndexByName } from "@/components/tasks/business-formation/BusinessFormationStepsConfiguration";
+import {
+  BusinessFormationStepsConfiguration,
+  LookupStepIndexByName,
+} from "@/components/tasks/business-formation/BusinessFormationStepsConfiguration";
 import { getMergedConfig } from "@/contexts/configContext";
 import { MunicipalitiesContext } from "@/contexts/municipalitiesContext";
 import { IsAuthenticated } from "@/lib/auth/AuthContext";
@@ -7,6 +10,7 @@ import { FormationStepNames, TasksDisplayContent } from "@/lib/types/types";
 import analytics from "@/lib/utils/analytics";
 import {
   generateEmptyFormationData,
+  generateFormationData,
   generateFormationDbaContent,
   generateFormationSubmitError,
   generateFormationSubmitResponse,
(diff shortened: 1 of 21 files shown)


api/src/db/migrations/v110_add_last_visited_formation_page_to_formation_data.ts
```diff
@@ -0,0 +1,401 @@
...
+  taskItemChecklist: Record<string, boolean>;
+  licenseData: v110LicenseData | undefined;
+  preferences: v110Preferences;
+  taxFilingData: v110TaxFilingData;
+  formationData: v110FormationData;
+  lastUpdatedISO: string | undefined;
+  version: number;
+}
+
+export const migrate_v109_to_v110 = (v109Data: v109UserData): v110UserData => {
```
aloverso, line 17: need to rebase and fix migration version for whats on main right now

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2168,51 @@ describe("<BusinessFormationPaginator />", () => {
       page.fillText("Search business name", "Pizza Joint");
     };
   });
+
+  describe("remembers formation step", () => {
+    it("shows the first formation step with default formationStep", () => {
```
aloverso, line 2173: I am not sure what this means... what is "default formationStep" 
I assume that means index 0 but:
- I would specify that in the test name to be clearer
- We should set index 0 explicitly on the user data in the "given" part here to make the cause-and-effect of the test clearer

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2168,51 @@ describe("<BusinessFormationPaginator />", () => {
...
+      expect(screen.getByText(Config.formation.fields.businessName.header)).toBeInTheDocument();
+    });
+
+    it("shows the last formation step when lastVisitedPage has step index 4", () => {
+      const legalStructureId = "limited-liability-company";
+      const profileData = generateFormationProfileData({ legalStructureId });
+      const formationData = generateFormationData({ lastVisitedPageIndex: 4 });
+      displayContent = {
+        formationDbaContent: generateFormationDbaContent({}),
+      };
```
aloverso, line 2184: is there pre-created earlier in the test displaycontent we could use here?

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2168,51 @@ describe("<BusinessFormationPaginator />", () => {
...
   });
+
+  describe("remembers formation step", () => {
+    it("shows the first formation step with default formationStep", () => {
+      preparePage(initialUserData, displayContent);
+      expect(screen.getByText(Config.formation.fields.businessName.header)).toBeInTheDocument();
+    });
+
+    it("shows the last formation step when lastVisitedPage has step index 4", () => {
+      const legalStructureId = "limited-liability-company";
```
aloverso, line 2179: why LLC specifically? none of the other tests seem to set this

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2168,51 @@ describe("<BusinessFormationPaginator />", () => {
...
+      const profileData = generateFormationProfileData({ legalStructureId });
+      const formationData = generateFormationData({ lastVisitedPageIndex: 4 });
+      displayContent = {
+        formationDbaContent: generateFormationDbaContent({}),
+      };
+      initialUserData = generateUserData({ profileData, formationData });
+      preparePage(initialUserData, displayContent);
+      expect(screen.getByText(Config.formation.general.submitButtonText)).toBeInTheDocument(); // expect submit button
```
aloverso, line 2187: delete comment

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2168,51 @@ describe("<BusinessFormationPaginator />", () => {
...
+    it("formState.lastVistedPage updates on step change", async () => {
+      const page = preparePage(initialUserData, displayContent);
+      expect(screen.getByText(Config.formation.fields.businessName.header)).toBeInTheDocument();
+
+      await page.stepperClickToBillingStep();
+      await waitFor(() => {
+        expect(currentUserData().formationData.lastVisitedPageIndex).toEqual(3);
+      });
+
+      await page.stepperClickToBusinessStep();
```
aloverso, line 2199: both these step changes are via stepper. should we also test that next/prev button step switches also updates the variable?

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2168,51 @@ describe("<BusinessFormationPaginator />", () => {
...
+      await waitFor(() => {
+        expect(currentUserData().formationData.lastVisitedPageIndex).toEqual(1);
+      });
+    });
+
+    it("defaults to first formation step we go beyond page index allowed", () => {
+      const legalStructureId = "limited-liability-company";
+      const profileData = generateFormationProfileData({ legalStructureId });
+      const formationData = generateFormationData({ lastVisitedPageIndex: 999 });
```
aloverso, line 2208: i vote we test edge case (ie 5)

aloverso: BusinessStepConfiguration.length

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2168,51 @@ describe("<BusinessFormationPaginator />", () => {
...
+        expect(currentUserData().formationData.lastVisitedPageIndex).toEqual(3);
+      });
+
+      await page.stepperClickToBusinessStep();
+      await waitFor(() => {
+        expect(currentUserData().formationData.lastVisitedPageIndex).toEqual(1);
+      });
+    });
+
+    it("defaults to first formation step we go beyond page index allowed", () => {
```
aloverso, line 2205: it looks like we're protecting for edge case when greater than but not less than. can we add one for -1?

web/src/components/tasks/business-formation/NexusFormation.test.tsx
```diff
@@ -162,13 +162,19 @@ describe("<NexusFormationFlow />", () => {
     await page.clickSubmit();
 
     const base64String = Buffer.from("my cool file contents", "utf8").toString("base64");
+    const updatedForeignUserData = foreignUserData;
+    updatedForeignUserData.formationData.lastVisitedPageIndex = 1;
```
aloverso, line 166: should this be 4? If the last thing we did was click to review step

web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
```diff
@@ -81,6 +82,7 @@ export const DbaFormationPaginator = (): ReactElement => {
       stepperState[1].isComplete = stepIndex === 2;
       return stepperState;
     });
+    updateQueue.queueFormationData({ lastVisitedPageIndex: stepIndex }).update();
```
aloverso, line 85: we should have a test for this within the DBA flow

aloverso: Since DBA / Nexus always starts on the first step, do we even need/want to set this?

aloverso: Or should we add a note to update this once we do the upcoming Dakota story that *will* remember Nexus name availability?

web/test/pages/profile.test.tsx
```diff
@@ -2193,6 +2193,26 @@ describe("profile", () => {
     });
   });
 
+  it("resets the last visited page in form state when changing legal structure", async () => {
```
aloverso, line 2196: I might say "user data" instead of "form state" because form state implies we're looking at the internal state of the form (ie, ProfileDataContext) but that's not what's under test

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2171,92 @@ describe("<BusinessFormationPaginator />", () => {
...
+      const legalStructureId = "limited-liability-company";
+      const profileData = generateFormationProfileData({ legalStructureId });
+      const formationData = generateFormationData({ lastVisitedPageIndex: 4 });
+
+      initialUserData = generateUserData({ profileData, formationData });
+      preparePage(initialUserData, displayContent);
+      expect(screen.getByTestId("review-step")).toBeInTheDocument();
+    });
+
+    it("formState.lastVistedPage updates on step change via stepper", async () => {
```
somabadri, line 2194: Should it be renamed to say userData instead of formState because we are checking the currentUserData in the test and we don't refer to form state or the state in any of our tests so it might confusing down the line?

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -2167,4 +2171,92 @@ describe("<BusinessFormationPaginator />", () => {
...
+        expect(currentUserData().formationData.lastVisitedPageIndex).toEqual(3);
+      });
+
+      await page.stepperClickToBusinessStep();
+      await waitFor(() => {
+        expect(currentUserData().formationData.lastVisitedPageIndex).toEqual(1);
+      });
+    });
+
+    it("formState.lastVisitedPage updates on step change via the previous and next button steps", async () => {
```
somabadri, line 2209: same as above

aloverso (approved):

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
adidner: Good catch and yes you're totally right, that was an artifact from earlier

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
adidner: ALso yes

PR #5669, opened 2023-04-21
chore: [#182479557] add config search to cms search

aloverso:
### Ticket

## Code author checklist

### DIFF
--- a/web/src/lib/search/helpers.ts
+++ b/web/src/lib/search/helpers.ts
@@ -3,13 +3,17 @@ import { LabelledContent, LabelledContentList, Match } from "@/lib/search/typesF
 export const findMatchInBlock = (blockTexts: string[], term: string, match: Match): Match => {
   for (const blockText of blockTexts) {
     if (blockText.includes(term)) {
-      const index = blockText.indexOf(term);
-      const startIndex = index - 50 < 0 ? 0 : index - 50;
-      const endIndex = term.length + index + 50;
-      match.snippets.push(blockText.slice(startIndex, endIndex));
+      match.snippets.push(makeSnippet(blockText, term));
     }
   }
   return match;
+};
+
+export const makeSnippet = (text: string, term: string): string => {
+  const index = text.toLowerCase().indexOf(term.toLowerCase());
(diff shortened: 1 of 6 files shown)


web/src/lib/search/helpers.ts
```diff
@@ -3,15 +3,19 @@ import { LabelledContent, LabelledContentList, Match } from "@/lib/search/typesF
...
   }
   return match;
 };
 
+export const makeSnippet = (text: string, term: string): string => {
+  const index = text.toLowerCase().indexOf(term.toLowerCase());
+  const startIndex = index - 50 < 0 ? 0 : index - 50;
+  const endIndex = term.length + index + 50;
+  return text.slice(startIndex, endIndex);
+};
```
jphechter, line 17: Nice usage of the Extract Function approach!

web/src/lib/search/searchConfig.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (value.toLowerCase().includes(term)) {
+          matches = [
+            ...matches,
+            {
+              value: value,
+              keyPath: [...keyPaths, key],
+            },
+          ];
+        }
+      } else if (typeof value === "object" && value !== null && !Array.isArray(value)) {
```
jphechter, line 51: This needs to be validated, but can this just be an `else`? Because we're doing the same check on line 38 with a similar result I feel like this could be simplified.

web/src/lib/search/searchConfig.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  return Object.keys(groupedConfigMatches).map((key) => ({
+    groupLabelPath: key,
+    matches: groupedConfigMatches[key],
+  }));
+};
+
+const searchObj = (obj: any, term: string, matches: JsonMatch[], keyPaths: string[]): JsonMatch[] => {
+  if (typeof obj === "object" && obj !== null && !Array.isArray(obj)) {
```
jphechter, line 38: This check sucks, but I don't know a better way to do this.

jphechter (approved):
I think everything here makes sense! 👍

PR #5672, opened 2023-04-21
chore: [#185009333] refactor auto name check to not fire for users wh…

somabadri:
…o have formed successfully

## Description

Refactored the auto-namecheck to not fire if the user has formed successfully by checking `formationData.completedFilingPayment` and `formationData.formationResponse.success`. The auto-checkname only runs when both of the values are not true.

### Ticket

[#185009333](https://www.pivotaltracker.com/story/show/185009333)

## Code author checklist

### DIFF
--- a/api/src/api/userRouter.test.ts
+++ b/api/src/api/userRouter.test.ts
@@ -259,9 +259,52 @@ describe("userRouter", () => {
         expect(stubTimeStampBusinessSearch.search).not.toHaveBeenCalled();
       });
 
+      it("does not update businessNameAvailability if it's completedFilingPayment is true", async () => {
+        const userData = generateUserData({
+          formationData: generateFormationData({
+            completedFilingPayment: true,
+            businessNameAvailability: generateBusinessNameAvailability({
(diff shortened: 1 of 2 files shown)


api/src/api/userRouter.test.ts
```diff
@@ -259,9 +259,61 @@ describe("userRouter", () => {
         expect(stubTimeStampBusinessSearch.search).not.toHaveBeenCalled();
       });
 
+      it("does not update businessNameAvailability if it's completedFilingPayment and formationResponse success are true", async () => {
+        const userData = generateUserData({
+          formationData: generateFormationData({
+            completedFilingPayment: true,
+            formationResponse: generateFormationSubmitResponse({
```
aloverso, line 266: formationResponse isn't the right thing to check here. That doesn't indicate that they have formed/payed, just that they have advanced to the payment screen. 

I think we can remove all references to formationResponse here & in the code and JUST check if `completedFilingPayment` is true

somabadri: okay sounds good!

PR #5700, opened 2023-04-26
Move shared factories to shared file #183973645

adidner:
## Description

We wanted to consolidate a bunch of our factories that are repeated between the web and api factories into our shared factories.

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/183973645)

### Approach

Basically a lot of copying and pasting and re building and re running of tests. 

### Steps to Test

Make sure all tests, run. Glance at code and make sure the new position of code makes sense

### Notes

It seemed like the only way to get imports to work in the shared factories was via relative imports. 

Probably could and should have been broken up into more commits but it was a little late to fix when I realized this

Also because of how this is structured a lot of these are available via the @shared tag alone rather than the @shared/tests tag. So they can be referenced via slightly different imports. I thought this wasn't something that needed to be 100% consistent in every place. 

## Code author checklist

### DIFF
--- a/web/src/components/RegisteredForTaxesModal.test.tsx
+++ b/web/src/components/RegisteredForTaxesModal.test.tsx
@@ -1,7 +1,6 @@
 import { RegisteredForTaxesModal } from "@/components/RegisteredForTaxesModal";
 import { getMergedConfig } from "@/contexts/configContext";
 import { MunicipalitiesContext } from "@/contexts/municipalitiesContext";
-import { generateProfileData, generateUserData } from "@/test/factories";
 import { fillText, selectLocationByText } from "@/test/helpers/helpers-testing-library-selectors";
 import { markdownToText } from "@/test/helpers/helpers-utilities";
 import {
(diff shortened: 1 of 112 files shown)


web/src/components/RegisteredForTaxesModal.test.tsx
```diff
@@ -10,8 +9,13 @@ import {
...
 } from "@/test/mock/withStatefulUserData";
-import { LegalStructures, LookupOwnershipTypeById, UserData } from "@businessnjgovnavigator/shared";
-import { generateMunicipality } from "@businessnjgovnavigator/shared/test";
+import {
+  generateProfileData,
+  LegalStructures,
+  LookupOwnershipTypeById,
+  UserData,
+} from "@businessnjgovnavigator/shared";
+import { generateMunicipality, generateUserData } from "@businessnjgovnavigator/shared/test";
```
adidner, line 18: So I was debating about when and whether to divide up my imports between @shared and shared/test. I had no strong feelings about this so things are a bit inconsistent in some places based on what the file was already doing before I added things. I think this is fine but if someone heavily disagrees lmk

PR #5742, opened 2023-05-01
feat: [#184688179] Added Help/intercom Button to Formation Pages

adidner:
## Description

We wanted to a help button that pulls up intercom to all the formation pages

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/184688179)

### Steps to Test

Go to formation
See it has the help button on desktop and mobile
See that CMS field exists in the CMS

## Code author checklist


web/src/components/njwds-extended/UnStyledButton.tsx
```diff
@@ -41,6 +41,9 @@ export const UnStyledButton = forwardRef(
         style =
           "usa-button usa-tag bg-transparent text-normal text-base border-1px border-base-light hide-unhide-button";
         break;
+      case "basic":
+        style = "usa-button bg-transparent";
+        break;
```
adidner, line 46: So I tried using all 3 of the other styles outlined in this component first and ended up overriding multiple other things like the border or the hide-unhide-button or the usa-button--unstyled and some of those end up being hard to override because they're !important. So in the end this ended up being the most intuitive to understand

Although I'm open to name changes besides "basic"

adidner (commented):
So for .svg files it looks like the pattern we have in the code base currently, is to just make another .svg file with the different color which I've followed here. However this isn't great because it introduced multiple sources of truth (for instance if we golbally change the color for blue to a darker shade someone will need to remember to update this .svg manually.) So, their is a potential refactor here where we reformat the .svg's so they can be passed vars.scss color variables. (Would also mean we don't need to have multiple .svg's for different colors which is also good if we ever need to update the .svg body. In that you don't have to recreate the new file for a bunch of colors)

web/src/components/njwds-extended/UnStyledButton.tsx
jphechter: The goal of this case is to make an unstyled variant, correct? Should we call it that instead of `basic`?

I see what you're saying about `.usa-button--unstyled`. Not something you introduced, but I do find it funny that the unstyled class has an !important color.

web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
```diff
@@ -371,19 +373,35 @@ export const BusinessFormationPaginator = (): ReactElement => {
...
+                textBold={true}
+                intercomButton
+                noRightMargin={true}
+                dataTestid={"help-button"}
+              >
+                <ButtonIcon svgFilename="help-circle-blue" sizePx="25px" />
+                {Config.formation.general.helpButtonText}
+              </UnStyledButton>
+            </div>
+
```
jphechter, line 390: I know the ticket doesn't specifically call this out, but should this also in the DBA/Nexus flow as well?

web/src/styles/base/text-colors.scss
```diff
@@ -26,6 +26,10 @@
   color: $accent-cool-lightest !important;
 }
 
+.text-accent-cool-darker {
+  color: $accent-cool-darker !important;
```
jphechter, line 30: Is there a reason this has to be `!important`? I see that on hover the color is changing, but I think the right thing to do would be to target the problem instead of just emphasizing this.

web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
adidner: So it exists for Nexus by virtue of being here but you're right that I missed DBA. I will fix

web/src/styles/base/text-colors.scss
adidner: So the USWDS library has its colors done in a way that this is needed which is why all the colors in here have it this way for them to take effect correctly, initially when I used some of the other button varients this was required to create the effect desired (might not be true in current incarnation). 

I completely hate it because using `!important` is an anti-pattern and want to move away from it but for now consistency with the rest of the file and use against the library seems like a better way to do things

jphechter: I would rather see us override the problem than use `!important`. This worked for me.
```
.text-accent-cool-darker.usa-button:hover {
  color: $accent-cool-darker;
}
```
Screenshot 2023-05-01 at 4 24 32 PM

web/src/components/njwds-extended/UnStyledButton.tsx
adidner: Lol, like I said I'm down for a different name and that seems good. 

Yeah kinda absurd that unstyled has a !important color. Some BS right there

jphechter (approved):

PR #5765, opened 2023-05-01
feat: [#184998185] marketing-calendar-welcome-landing-page

AnJuHyppolite:
## Description

As a user visiting navigator.business.nj.gov/welcome, when viewing the welcome page, only the ‘Get Started’ and ‘File for Taxes’ tiles appear. 

### Ticket

[#184998185](https://www.pivotaltracker.com/story/show/184998185)

### Approach

N/A

### Steps to Test

Upon visiting navigator.business.nj.gov/welcome, when viewing the welcome page, only the ‘Get Started’ and ‘File for Taxes’ tiles appear. 

### Notes

Refactored `LandingPageTiles.tsx` and `LandingPageSlider.tsx` to ensure DRY code and use of props to create one source of truth for the tiles.

## Code author checklist

### DIFF
--- a/web/src/components/LandingPageSlider.tsx
+++ b/web/src/components/LandingPageSlider.tsx
@@ -1,10 +1,7 @@
 import { LandingPageActionTile } from "@/components/LandingPageActionTile";
 import { Icon } from "@/components/njwds/Icon";
-import { useConfig } from "@/lib/data-hooks/useConfig";
-import { QUERIES, ROUTES, routeWithQuery } from "@/lib/domain-logic/routes";
-import analytics from "@/lib/utils/analytics";
+import { ActionTile } from "@/lib/types/types";
 import { useMountEffect } from "@/lib/utils/helpers";
-import { useRouter } from "next/router";
 import { ReactElement, useRef, useState } from "react";
 import Slider, { CustomArrowProps } from "react-slick";
 
@@ -23,9 +20,11 @@ function PrevArrow(props: CustomArrowProps): ReactElement {
   );
 }
 
-export const LandingPageSlider = (): ReactElement => {
-  const router = useRouter();
-  const { Config } = useConfig();
+interface Props {
+  actionTiles: ActionTile[];
+}
+
+export const LandingPageSlider = (props: Props): ReactElement => {
   const [activeTile, setActiveTile] = useState(0);
   const [slideBreakPoint, setSlideBreakPoint] = useState(7);
   const cardRef = useRef<null | HTMLDivElement>(null);
@@ -53,18 +52,6 @@ export const LandingPageSlider = (): ReactElement => {
       window.removeEventListener("resize", handleResize);
     };
   });
-
-  const routeToOnboarding = (): void => {
-    router.push(ROUTES.onboarding);
-    analytics.event.landing_page_hero_get_started.click.go_to_onboarding();
-  };
-
-  const setFlowAndRouteUser = (flow: "starting" | "out-of-state" | "up-and-running"): void => {
-    routeWithQuery(router, {
-      path: ROUTES.onboarding,
-      queries: { [QUERIES.flow]: flow },
-    });
-  };
 
   const settings = {
     className: "slider variable-width",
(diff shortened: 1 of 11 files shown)


web/src/components/LandingPageTiles.test.tsx
```diff
@@ -14,26 +14,39 @@ describe("<LandingPageTiles />", () => {
...
-    render(<LandingPageTiles />);
+    render(<LandingPageTiles isWelcomePage={false} />);
 
     fireEvent.click(screen.getByText(Config.landingPage.landingPageTile3Text));
 
     expect(mockPush).toHaveBeenCalledWith({ pathname: "/onboarding", query: { flow: "up-and-running" } });
   });
+
+  it("shows only 'Get Started' and 'Pay Taxes' tiles when visiting welcome page URL", () => {
+    useMockRouter({ pathname: "/welcome" });
```
jphechter, line 41: Given that the logic that controls the welcome page is determined by `isWelcomePage`, I don't think you need `useMockRouter`.

web/src/components/LandingPageTiles.tsx
```diff
@@ -25,67 +29,95 @@ export const LandingPageTiles = (): ReactElement => {
...
   };
 
+  interface ActionTileProps {
+    imgPath: string;
+    tileText: string;
+    tileText2?: string;
+    dataTestId: string;
+    onClick: () => void;
+    isPrimary?: boolean;
+  }
```
jphechter, line 39: We define this same interface in 2 places. Here and in `LandingPageSlider`. The `Props` interface in `LandingPageActionTile` is also very similar to this shape.

I don't think we're necessarily using this enough to warrant it being added to `types.ts`, but maybe we could cut down on the duplication by renaming and exporting the `Props` interface in `LandingPageActionTile`.

web/src/lib/auth/signinHelper.ts
```diff
@@ -99,14 +99,28 @@ export const onGuestSignIn = async (
...
+      case ROUTES.loading: {
+        setRegistrationDimension("Began Onboarding");
+        push(ROUTES.onboarding);
+
+        break;
+      }
+      default: {
+        setRegistrationDimension("Not Started");
+        push(ROUTES.landing);
+      }
```
jphechter, line 123: Nice usage of a switch!

web/src/pages/welcome.tsx
```diff
@@ -0,0 +1,12 @@
+import Home from "@/pages/index";
+import { ReactElement } from "react";
+
+const WelcomePage = (): ReactElement => {
+  return (
+    <>
+      <Home isWelcomePage={true} />
+    </>
+  );
+};
```
jphechter, line 10: This is a nit-picking point, but you don't have to wrap a stand-alone element in fragment tags. In fact, `<Home>` could just be on the same line as the return, no parentheses.

web/src/pages/index.tsx
```diff
@@ -18,7 +18,11 @@ import { useMediaQuery } from "@mui/material";
 import { useRouter } from "next/router";
 import { ReactElement, useContext, useEffect, useRef, useState } from "react";
 
-const Home = (): ReactElement => {
+interface Props {
+  isWelcomePage: boolean;
```
jphechter, line 22: This should be optional.

web/test/pages/index.test.tsx
```diff
@@ -35,33 +35,33 @@ describe("HomePage", () => {
         onboardingFormProgress: "COMPLETED",
         profileData: generateProfileData({ businessPersona: "STARTING" }),
       });
-      render(withAuth(<Home />, { user: generateUser({}) }));
+      render(withAuth(<Home isWelcomePage={false} />, { user: generateUser({}) }));
```
jphechter, line 38: If we make `isWelcomePage` an optional param, you can drop it from these render statements.

web/src/components/LandingPageTiles.tsx
aloverso: I'd actually suggest putting it in `types` - I think it's generally an antipattern to have components importing things from other components other than the React Component itself.

aloverso: One reason for this is that there is no real reason for one component file to "own" the type definition. If both  `LandingPageTiles` and `LandingPageSlider` use this interface, why should LandingPageSlider own it and define it? And if someone is looking for that type definition, why would they expect it to be in Slider as opposed to the Tiles or anywhere else? Having a single `types` as a shared place for shared types addresses all this

aloverso: I would also rename the type to perhaps just `ActionTile`? It's an object that defines a tile, and "props" is a specific word for component props that doesn't really fit what's going on here.

web/src/components/LandingPageSlider.tsx
```diff
@@ -99,65 +94,22 @@ export const LandingPageSlider = (): ReactElement => {
...
+              key={`landing-page-slider-key-${index}`}
+              className={"landing-page-slide"}
+              imgPath={actionTile.imgPath}
+              dataTestId={actionTile.dataTestId}
+              tileText={actionTile.tileText}
+              tileText2={actionTile.tileText2}
+              onClick={actionTile.onClick}
+              isPrimary={actionTile.isPrimary}
+              isActive={activeTile === index}
+              reference={index === 0 ? cardRef : undefined}
```
aloverso, line 109: Per our convo about making the ActionTile type shared, I would also change `LandingPageActionTile` to take this same type as a prop, instead of having to split it out into so many individual variables here.

web/src/pages/welcome.tsx
AnJuHyppolite: Yes, I had to add them when I added the Welcome string to test functionality and forgot to remove them. Thanks!

web/src/components/LandingPageTiles.test.tsx
```diff
@@ -14,26 +14,38 @@ describe("<LandingPageTiles />", () => {
   });
 
   it("routes user to industry selection when the starting a business button is clicked", async () => {
-    render(<LandingPageTiles />);
+    render(<LandingPageTiles isWelcomePage={false} />);
```
aloverso, line 17: these new prop changes in the test here can all be removed per JP's comment about making the prop optional, right?

PR #5766, opened 2023-05-02
feat: [#184305929] review page highlighted empty fields

somabadri:
## Description

The empty fields on the review page are now highlighted and display the text `Not Entered`.

### Ticket

[184305929](https://www.pivotaltracker.com/story/show/184305929)

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/business/BusinessNameAndLegalStructure.tsx
+++ b/web/src/components/tasks/business-formation/business/BusinessNameAndLegalStructure.tsx
@@ -4,6 +4,7 @@ import { ModalTwoButton } from "@/components/ModalTwoButton";
 import { ModifiedContent } from "@/components/ModifiedContent";
 import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
 import { LookupStepIndexByName } from "@/components/tasks/business-formation/BusinessFormationStepsConfiguration";
+import { ReviewNotEntered } from "@/components/tasks/business-formation/review/section/ReviewNotEntered";
 import { BusinessFormationContext } from "@/contexts/businessFormationContext";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { useUserData } from "@/lib/data-hooks/useUserData";
@@ -62,6 +63,14 @@ export const BusinessNameAndLegalStructure = ({ isReviewStep = false }: Props): 
     return isReviewStep ? `${label}:` : label;
   };
 
+  const notEnteredBusinessName = (): ReactElement => {
+    if (isReviewStep) {
+      return <ReviewNotEntered />;
+    } else {
+      return <span className="text-accent-cool-darker">{Config.formation.general.notEntered}</span>;
+    }
+  };
+
   return (
     <>
       <div className="flex space-between margin-bottom-2 flex-align-center">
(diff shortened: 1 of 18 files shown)


web/src/components/tasks/business-formation/review/ReviewPartnership.tsx
```diff
@@ -12,9 +12,10 @@ export const ReviewPartnership = (): ReactElement => {
   const notEnteredText = (question: string): ReactElement => {
     return (
       <>
-        <div>
-          <em>{Config.formation.general.notEntered}</em> - {question}
-        </div>
+        <span className={"bg-accent-warm-extra-light text-italic"}>
+          {Config.formation.general.notEntered}
```
aloverso, line 16: can we use the <NotEntered> component here?

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
```diff
@@ -19,9 +20,43 @@ export const ReviewSignatures = (): ReactElement => {
     };
   };
 
+  const displayEmptySignersIfNotEntered = (): ReactElement | undefined => {
```
aloverso, line 23: I might recommend moving the if-statement out of this and into the place where this function gets called. Having an "if" in the name of the method is a little confusing, not least because it doesn't fully capture the extent of the "if" statement in here. Might be cleaner to name the function `displayEmptySigners` and move the if-statement to the render

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
```diff
@@ -19,9 +20,43 @@ export const ReviewSignatures = (): ReactElement => {
...
+              value={undefined}
+              marginOverride={"margin-top-0"}
+            />
+          )}
+        </div>
+      );
+    }
+  };
+
+  const displayEmptyIncorporatorsIfEmpty = (): ReactElement | undefined => {
```
aloverso, line 40: same notes as above

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
aloverso: It would have the added benefit of always returning a thing, never undefined

web/src/components/tasks/business-formation/review/ReviewPartnership.tsx
somabadri: The styling doesn't work because the `Not Entered` portion has to be a part of the first sentence but if we use the component then there are two sections, the left and right and it doesn't look like one sentence flowing into the second line anymore.

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
somabadri: I was debating between the two for the same reasons but decided to go with this because it felt cleaner in the code but agree with your points regarding the confusion it might lead to, making the change.

aloverso: looking at the code below, there's a lot going on with the mapping of signers and incorporators, and then these functions in there as well. What if we have a strcuture like this? (psuedocode)

```
const signers = () => {
if (hasNoSigners) {
  return displayEmptySigners()
 } else {
  return displaySigners()
 }
}

const incorporators = () => {
 if (hasNoIncorporators) {
  return displayEmptyIncorporators()
 } else {
  return displayIncorporators()
 }
}

const displayEmptyIncorporators = () => <>...</>
const displayEmptySigners = () => <>...</>
const displayIncorporators = () => <>...</>
const displaySigners = () => <>...</>


render (
  {signers()}
  {incorporators()}
)

```

somabadri: I like this, makes it easier to read. I will restructure it.

web/src/components/tasks/business-formation/business/BusinessNameAndLegalStructure.tsx
```diff
@@ -81,20 +82,26 @@ export const BusinessNameAndLegalStructure = ({ isReviewStep = false }: Props):
...
-            >
-              {Config.formation.general.editButtonText}
-            </UnStyledButton>
-          )}
+          <div className="flex">
+            <div className="margin-right-05">
+              {state.formationFormData.businessName ? (
+                <span className="text-accent-cool-darker">{state.formationFormData.businessName}</span>
+              ) : (
+                <ReviewNotEntered />
```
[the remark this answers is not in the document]
somabadri: Let me confirm with design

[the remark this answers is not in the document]
somabadri: They wanted to keep the original blue body text design, so updated it accordingly.

aloverso (approved):

PR #5768, opened 2023-05-03
feat: [#184858230] add business structure task for all industries

aloverso:
## Description

- Added a new task to all roadmaps for business structure
- Task starts as completed in the `emptyUserData` object (this will change with the next story in this chain)
- Task shows saved legal structure & allows user to edit and save it
- Task progress automatically updates when clicking edit and when clicking save
- Task status checkbox is locked, tooltip text depends on state

### Ticket

[#184858230](https://www.pivotaltracker.com/story/show/184858230)

### Approach

One interesting piece to flag here:

I used the updateQueue _without_ triggering a save for when the user clicks "edit" and the task status goes from Complete to In-Progress. Because if they are editing & leave the task without saving, it needs to revert back to the old, saved, completed state. So I queue up the task-progress change without actually saving it in the backend.

This meant I needed to change `TaskHeader` and `TaskProgressCheckbox` to decide their state (ie, the grey/blue/green header background color and the text it displays) based on the state of the queue, not the state of the saved userData. Hence the change in these files from `userData` to `updateQueue.current()`

### Notes

I made the same exact change to all 50+ industry json files so you can probably skip all that to make reviewing easier

## Code author checklist

### DIFF
--- a/content/src/roadmaps/industries/acupuncture.json
+++ b/content/src/roadmaps/industries/acupuncture.json
@@ -19,6 +19,12 @@
       "required": true
     },
     {
+      "step": 1,
+      "weight": 100,
+      "task": "business-structure",
+      "required": true
+    },
+    {
       "step": 2,
       "weight": 20,
       "task": "register-for-ein",
--- a/content/src/fieldConfig/business-structure-task.json
+++ b/content/src/fieldConfig/business-structure-task.json
@@ -1,0 +1,10 @@
+{
+  "businessStructureTask": {
+    "uncompletedTooltip": "Select and then save to complete this task.",
+    "completedTooltip": "This task is complete. You can edit your business structure selection below.",
+    "saveButton": "Save Business Structure",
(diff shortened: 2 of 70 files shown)


content/src/roadmaps/industries/acupuncture.json
```diff
@@ -18,6 +18,12 @@
       "task": "determine-naics-code",
       "required": true
     },
+    {
+      "step": 1,
+      "weight": 100,
+      "task": "business-structure",
+      "required": true
+    },
```
mikewolfd, line 26: Was there a reason we never had a default add-on for tasks all industries shared?

mikewolfd (approved):

content/src/roadmaps/industries/acupuncture.json
aloverso: we actually used to and Dave removed it because there were a couple exceptions to it.... but I think it might be a good idea

PR #5771, opened 2023-05-04
feat: created a separate welcome page content method

mikewolfd:
## Description

Added a new Landing Page content key that overrides primary landing page content when used

### Ticket

[#184998185](https://www.pivotaltracker.com/story/show/184998185)

### Approach

We imitated how we approached the ABExperience copy integration, note to remove the AB experience code as we no longer use that feature

### Steps to Test

Go to /welcome


AnJuHyppolite (commented):
Looks good!

PR #5773, opened 2023-05-04
feat: [#184125120] reset naics codes when changing Industry

FarazA22:
## Description
When the user changes their industry and saves their data, the naics code values in the user data are removed, also the determine your naics code task is reset back to not started.

### Ticket
https://www.pivotaltracker.com/story/show/184125120

## Code author checklist

### DIFF
--- a/web/test/pages/profile.test.tsx
+++ b/web/test/pages/profile.test.tsx
@@ -51,6 +51,7 @@ import {
   LookupOwnershipTypeById,
   LookupSectorTypeById,
   Municipality,
+  naicsCodeTaskId,
   OperatingPhase,
   OperatingPhaseId,
   OperatingPhases,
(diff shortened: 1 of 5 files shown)


web/test/pages/profile.test.tsx
```diff
@@ -1813,6 +1819,23 @@ describe("profile", () => {
...
+    renderPage({ userData });
+    chooseTab("info");
+    selectByValue("Industry", "e-commerce");
+    clickSave();
+
+    await waitFor(() => {
+      expect(currentUserData().taskProgress[naicsCodeTaskId]).toEqual("NOT_STARTED");
+    });
+  });
+
```
somabadri, line 1838: I know we are checking that the NAICS code is being reset in the OnboardingIndustry.test.tsx but it might make it more readable in the future if we are explicitly checking it here as well so we see that two changes are expected to happen? Just a thought.

PR #5798, opened 2023-05-11
fix: [#185157746] industy name instead of industry id in webflow

somabadri:
## Description

Webflow should now display the industry name instead of the industry id. 

### Ticket

[#185157746](https://www.pivotaltracker.com/story/show/185157746)

## Code author checklist


web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -40,7 +41,7 @@ const getLicenseFromMd = (licenseMd) => {
     "department-phone-2": licenseMd.divisionPhone,
     "license-certification-classification": licenseMd.licenseCertificationClassification,
     "form-name": licenseMd.formName,
-    "primary-industry": licenseMd.industryId ?? licenseMd.webflowIndustry,
+    "primary-industry": LookupIndustryById(licenseMd.industryId).name ?? licenseMd.webflowIndustry,
```
aloverso, line 44: I don't think `??` is the right thing here. `LookupIndustryById` returns an object with `name` set to `""` if the id passed is undefined.

Empty string doesn't count as undefined, so it would not actually cascade to the `webflowIndustry` part here

aloverso:
@somabadri once this is merged then make sure you trigger a webflow license sync via circle ci to update everything in webflow

PR #5800, opened 2023-05-12
chore: [#185149976] include webflow tasks in prod deploy job

jphechter:
## Description

Update production deploy job to include Webflow sync tasks

### Ticket

[185149976](https://www.pivotaltracker.com/story/show/185149976)

### Notes

Created in response to conversation at team retro.

## Code author checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -705,33 +705,26 @@ workflows:
       - production-hold:
           type: approval
           filters:
-            tags:
-              only: /^env-update.*/
             branches:
               only:
                 - /^release-.*/
           requires:
             - staging-deploy
       - prepare-release:
-          filters:
-            tags:
-              ignore: /^env-update.*/
-            branches:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -732,6 +732,14 @@ workflows:
           requires:
             - prepare-release
             - production-hold
+      - sync-fundings:
```
aloverso, line 735: should these require the release-branch filter like the other jobs above?

jphechter: I don't think so. The `requires` option needs a successful run. If the required task is filtering and won't run, filtering again is a redundancy. At least that's my understanding.

jphechter: I guess the right thing to do would be to drop the redundant `filters`.

aloverso: ah got it, sure makes sense

PR #5813, opened 2023-05-12
chore: [#184834138] update summaryDescription format to prevent breaking CMS

aloverso:
## Description

Fixes the following bug in CMS where multi-line summaries would entirely break the CMS editability of that task. See screenshots below where some tasks would be missing their content (no title in list view) and all of the internal content:

Screenshot 2023-05-12 at 3 54 16 PM

Screenshot 2023-05-12 at 3 54 06 PM

### Additional

Also fixes ensuring that the summary shows up all of the time, not just on the "default" task. See comments in code below

### DIFF
--- a/web/src/components/tasks/LicenseTask.tsx
+++ b/web/src/components/tasks/LicenseTask.tsx
@@ -133,6 +133,7 @@ export const LicenseTask = (props: Props): ReactElement => {
             <TabPanel value="0" sx={{ paddingX: 0 }}>
               <div className="margin-top-3">
                 <UnlockedBy task={props.task} />
+                <Content>{props.task.summaryDescriptionMd || ""}</Content>
                 <Content>{getModifiedTaskContent(roadmap, props.task, "contentMd")}</Content>
               </div>
               <div className="flex flex-column margin-top-4 margin-bottom-1">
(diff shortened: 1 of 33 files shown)


web/src/components/tasks/LicenseTask.tsx
```diff
@@ -133,6 +133,7 @@ export const LicenseTask = (props: Props): ReactElement => {
             <TabPanel value="0" sx={{ paddingX: 0 }}>
               <div className="margin-top-3">
                 <UnlockedBy task={props.task} />
+                <Content>{props.task.summaryDescriptionMd || ""}</Content>
```
aloverso, line 136: This line is necessary, otherwise the summary does not show up in the license tasks that are integrated with consumer affairs (cosmetology, HIC, etc)

web/src/pages/tasks/[taskUrlSlug].tsx
```diff
@@ -202,6 +202,7 @@ export const TaskElement = (props: { task: Task; children?: ReactNode | ReactNod
       <div>
         <TaskHeader task={props.task} />
         {props.children}
+        <Content>{props.task.summaryDescriptionMd || ""}</Content>
```
Comment (aloverso) [inline web/src/pages/tasks/[taskUrlSlug].tsx:205]: This change is necessary because otherwise the summary did not show up if the task contained location-dependent content or post-onboarding content.

This led to the Town Mercantile License task not having any content displayed to the user, because it has a location-dependent section

PR #5823, opened 2023-05-16
feat: [#185191912] updated landing page navbar to hamburger and sticky

somabadri:
## Description

Updated the landing page navbar to be a hamburger and sticky.

### Ticket

[#185191912](https://www.pivotaltracker.com/story/show/185191912)

## Code author checklist


web/src/components/navbar/NavBar.tsx
```diff
@@ -37,8 +37,21 @@ export const NavBar = (props: Props): ReactElement => {
...
+        <NavBarMobile
+          scrolled={scrolled}
+          task={props.task}
+          showSidebar={props.showSidebar}
+          hideMiniRoadmap={props.hideMiniRoadmap}
+          landing
+        />
+        <div className={!isLargeScreen && scrolled ? "padding-top-6" : ""} />
+      </>
+    );
```
jphechter, line 54: With the additional `isLargeScreen` check for the previous condition, this case should fall to the final else. This is redundant. You can kill this.

jphechter: I can't write a comment on line 66 but `!isLargeScreen` check in className can be removed because we're checking that before that ternary is evaluated.

```
className={scrolled ? "padding-top-6" : ""}
```

web/src/components/navbar/NavBarMobile.tsx
```diff
@@ -15,9 +15,16 @@ interface Props {
...
+  landing?: boolean;
 }
 
-export const NavBarMobile = ({ scrolled, task, showSidebar, hideMiniRoadmap }: Props): ReactElement => {
+export const NavBarMobile = ({
+  scrolled,
+  task,
+  showSidebar,
+  hideMiniRoadmap,
+  landing,
```
jphechter, line 26: I lean towards using deconstructed props, but I think we as a team are trying to use a props object to make it more obvious where a value is coming from. Not a necessary change, but this might be a good chance to change that pattern in this component to be consistent with the other components in our app.

web/src/components/navbar/NavBar.tsx
```diff
@@ -37,8 +37,21 @@ export const NavBar = (props: Props): ReactElement => {
 
   if (props.logoOnly) {
     return <NavBarLogoOnly />;
-  } else if (props.landingPage) {
+  } else if (props.landingPage && isLargeScreen) {
     return <NavBarLanding />;
+  } else if (props.landingPage && !isLargeScreen) {
```
aloverso, line 42: This becomes a little confusing naming-wise because previously `NavBarLanding` was on the landing page. 

Now, if we're mobile, `NavBarLanding` is NOT rendered, `NavBarMobile` is.

Maybe we should call it `NavBarLandingDesktop` or something?

web/src/components/navbar/NavBarMobile.tsx
aloverso: +1

My argument against deconstructed props is that seeing `props.scrolled` in the body of JSX makes it 100% obvious where that value comes from without "clicking through" to its definition.

Whereas if you deconstruct and you see `scrolled` variable being used, you have no idea if that is computed or passed it. You have to click through to see if it comes from props, or if it is defined somewhere in the component `const scrolled = ....`

It is useful to know at-a-glance whether something is computed or passed from props, thus my argument against deconstructing.

Agree with JP, this is a good time to apply the campground rule

aloverso, line 26: I have a slight preference for booleans named clearly to be boolean, ie `isLanding` instead of `landing`

web/src/components/navbar/NavSidebarUserSettings.tsx
```diff
@@ -11,7 +11,11 @@ import Link from "next/link";
 import { useRouter } from "next/router";
 import { ReactElement, ReactNode, useContext, useMemo } from "react";
 
-export const NavSidebarUserSettings = (): ReactElement => {
+interface Props {
+  landing?: boolean;
+}
+
+export const NavSidebarUserSettings = ({ landing }: Props): ReactElement => {
```
aloverso, line 18: same comments here about deconstructing and bool naming

web/src/components/navbar/NavSidebarUserSettings.tsx
```diff
@@ -26,16 +30,18 @@ export const NavSidebarUserSettings = (): ReactElement => {
...
-            <UnStyledButton
-              style="tertiary"
-              onClick={(): void => {
-                analytics.event.account_menu_my_profile.click.go_to_profile_screen();
-              }}
-            >
-              <span className="text-base">{Config.navigationDefaults.profileLinkText}</span>
-            </UnStyledButton>
-          </Link>
+          {!landing && (
```
aloverso, line 33: This logic should be tested

web/src/components/navbar/NavBarMobile.tsx
```diff
@@ -80,10 +87,12 @@ export const NavBarMobile = ({ scrolled, task, showSidebar, hideMiniRoadmap }: P
           data-testid="nav-sidebar-menu"
         >
           <h4 className={`margin-0 flex flex-align-center fdr fjc space-between text-${textColor}`}>
-            <div className="flex">
-              <Icon className="margin-top-2px margin-right-1 usa-icon--size-3 minw-3">{accountIcon}</Icon>
-              <div>{accountString}</div>
-            </div>
+            {!landing && (
+              <div className="flex">
```
aloverso, line 91: This logic should be tested

aloverso (approved):

PR #5843, opened 2023-05-24
Update profile dropdown single business ethan

adidner:
## Description

We wanted to redesign out nav to accomodate Ethans eventually

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/184857807)

### Approach

I wanted to try and consolidate the NavMenu concept because its now the same on desktop and mobile rather than distinct

### Steps to Test

On mobile and desktop
- checkout the nav bar during onboarding
- before formation and authentication
- before formation and after authentication
- after authentication and formation

Test for Keyboard accessability

### Notes

For accessability sake when you're not on the dashboard or profile which are selected and highlit, on desktop when you open the dropdown you still need to have an entry focused so I focused the first and treated the highlighting and focus as distinct things. 

## Code author checklist


web/src/components/navbar/NavBarPopupMenu.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  function handleListKeyDown(event: React.KeyboardEvent): void {
+    if (event.key === "Tab") {
+      event.preventDefault();
+      handleClose();
+    }
+  }
+
+  const loginMenuItem = (): ReactElement => {
+    return NavMenuItem({
```
adidner, line 58: As disscussed briefly during stand up these are in this function call format rather than the normal component invocation because thats the main way you get focus for accessibility to work
image

https://github.com/mui/material-ui/issues/31006

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -43,6 +55,38 @@ const setLargeScreen = (value: boolean): void => {
...
+    }),
+  });
+};
+
+const generateGuestUserData = (overrides?: Partial<UserData>): UserData => {
+  return generateUserData({
+    profileData: generateProfileData({
+      businessName: "",
+      industryId: "industryId",
+      legalStructureId: "legalStructureId",
```
aloverso, line 73: What's going on with these value? They're not valid industry / legal structure IDs and would cause code to break if we ever try to interpret them

aloverso: Also in the chunk of code below as well.

web/src/lib/utils/helpers.ts
```diff
@@ -144,6 +144,18 @@ export const getUserNameOrEmail = (userData: UserData | undefined): string => {
   }
 };
 
+export const getNavBarBusinessTitle = (userData: UserData | undefined): string => {
+  const Config = getMergedConfig();
+  if (userData?.profileData.industryId === undefined || userData.profileData.legalStructureId === undefined) {
+    return Config.navigationDefaults.navBarGuestText;
+  }
+  if (userData?.profileData.businessName === undefined || userData?.profileData.businessName === "") {
```
aloverso, line 152: This can be simplified to `!userData?.profileData.businessName` because both undefined and empty string are falsey.

Same with above, `=== undefined` can be replaced with a `!` instead

web/src/lib/utils/helpers.ts
```diff
@@ -144,6 +144,18 @@ export const getUserNameOrEmail = (userData: UserData | undefined): string => {
...
+    return `${Config.navigationDefaults.navBarUnnamedBusinessText} [${userData?.profileData?.industryId}][${userData?.profileData?.legalStructureId}]`;
```
aloverso, line 153: This is not correct. This will render these on the screen as the IDs that we use to represent these values; not the way design wants them to show.

`industryId` would be something like "home-improvement-contractor" and it looks like the designs want us to use the industry name ("Home Improvement Contractor")

For `legalStructureId`, this code would make it show up like "limited-liability-company". We can't use the name here, because the name would be "Limited Liability Company (LLC)" and it looks like the designs want us to show the abbreviation (ie, just "LLC") so we'll have to add something to legal structures to support this.

web/src/lib/utils/helpers.ts
```diff
@@ -144,6 +144,18 @@ export const getUserNameOrEmail = (userData: UserData | undefined): string => {
   }
 };
 
+export const getNavBarBusinessTitle = (userData: UserData | undefined): string => {
```
aloverso, line 147: This function should be tested. Also, I think it should not live here in "helpers" but rather in its own thing. It is not a generic-use helper function, it has a very specific purpose for the NavBar only (as the title suggests)

web/src/styles/sections/nav-bar.scss
```diff
@@ -6,3 +6,24 @@
 .nav-bar-override-usa-accordion__content {
   padding: 0px 0px 12px 16px !important;
 }
+
+.menu-item-title.Mui-disabled {
+  opacity: 1;
+}
+
+$icon-offset: 45px;
```
aloverso, line 14: Should we use rem instead?

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -172,36 +195,43 @@ describe("<NavBar />", () => {
...
-      const user = { name: "John Smith", email: "test@example.com" };
+      const userData = generateBusinessNamedUserData();
+      useMockUserData({ ...userData });
 
-      useMockUserData({
-        user: generateUser(user),
-      });
       renderDesktopNav();
-      const menuEl = screen.getByText(user.name);
+      const menuEl = screen.getByText(getNavBarBusinessTitle(userData));
```
aloverso, line 215: I would generally recommend against using a helper function like this in tests. Shouldn't we know what the name would be in the context of the test data? we can just select by the text directly

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -172,36 +195,43 @@ describe("<NavBar />", () => {
...
     });
 
     it("displays an open dropdown menu when clicked and closes when clicked again", async () => {
-      const user = { name: "John Smith", email: "test@example.com" };
-
-      useMockUserData({
-        user: generateUser(user),
-      });
+      const userData = generateBusinessNamedUserData();
+      useMockUserData({ ...userData });
```
aloverso, line 223: I see you changed a one-liner to this two-liner in many of these tests. Is there a reason to add the temp variable? Why not just `useMockUserData(generateBusinessNamedUserData())`

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -220,50 +250,66 @@ describe("<NavBar />", () => {
...
         expect(screen.getByText(Config.navigationDefaults.profileLinkText)).toBeInTheDocument();
       });
-      userEvent.click(menuEl);
+      fireEvent.click(menuEl);
       await waitFor(() => {
         expect(screen.queryByText(Config.navigationDefaults.profileLinkText)).not.toBeInTheDocument();
       });
     });
 
+    it("displays dashboard button when the menu is open", async () => {
```
aloverso, line 281: This is a little confusing because I don't see anything in the body of this test that asserts that there is a "dashboard button"

Looking at the designs, I can see that the business title is that button I think? But a test should be readable and understandable without that context.

If it's a dashboard button, should we be testing that it is clickable and that it does something (ie, route to dashboard?)

Otherwise, we could just write the test as "it displays business name title within menu" or something similar

web/src/components/navbar/NavBarMobile.tsx
```diff
@@ -34,15 +35,33 @@ export const NavBarMobile = (props: Props): ReactElement => {
...
+  };
+
+  const getMenuConfirguration = (): menuConfiguration => {
+    if (props.isLanding) {
+      return "login-getstarted";
+    }
+    if (currentlyOnboarding()) {
+      return "login";
+    } else {
+      if (isAuthenticated) {
```
aloverso, line 53: this could be another else-if right, doesn't need to be nested

web/src/components/navbar/NavBarPopupMenu.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { useUserData } from "@/lib/data-hooks/useUserData";
+import { ROUTES } from "@/lib/domain-logic/routes";
+import analytics from "@/lib/utils/analytics";
+import { getNavBarBusinessTitle, getUserNameOrEmail } from "@/lib/utils/helpers";
+import Config from "@businessnjgovnavigator/content/fieldConfig/config.json";
+import { MenuItem, MenuList } from "@mui/material";
+import { useRouter } from "next/router";
+import { ReactElement, useContext } from "react";
+
+export type menuConfiguration =
```
aloverso, line 17: types should always be PascalCase

web/src/components/navbar/NavBarPopupMenu.tsx
```diff
@@ -0,0 +1,188 @@
...
+  | "profile"
+  | "profile-register-login"
+  | "profile-myNJ-logout"
+  | "login"
+  | "login-getstarted";
```
aloverso, line 22: A few thoughts:
- did you consider naming these after states instead of render results? Looking at them, it's hard to decipher when I might want to use "profile" vs "profile-myNJ-logout" etc. I think it might be a lot easier for someone else to use these if they described when they get used. "onboarding" and "guest-mode" and so on. Otherwise, I have to find the actual logical if/else structure just to understand the difference between when they show up.
- in any case, it'd be nice to standardize the casing/style of these terms. It's almost kebab-case between words except for "getstarted". And it's almost all lowercase except for "myNJ"

web/src/components/navbar/NavBarMobile.tsx
```diff
@@ -34,15 +35,33 @@ export const NavBarMobile = (props: Props): ReactElement => {
...
-  const accountString = isAuthenticated ? userName : Config.navigationDefaults.navBarGuestText;
+
+  const currentlyOnboarding = (): boolean => {
+    if (!router) {
+      return false;
+    }
+    return router.pathname === ROUTES.onboarding;
+  };
+
+  const getMenuConfirguration = (): menuConfiguration => {
```
aloverso, line 46: spelling error

web/src/components/navbar/NavBarPopupMenu.tsx
```diff
@@ -0,0 +1,188 @@
...
+  hasCloseButton?: boolean;
+  menuConfiguration: menuConfiguration;
+}
+
+export const NavBarPopupMenu = ({
+  menuConfiguration,
+  handleClose,
+  open,
+  hasCloseButton,
+}: Props): ReactElement => {
```
aloverso, line 36: We typically avoid prop-spread notation on components

web/src/components/navbar/NavBarPopupMenu.tsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+    if (menuConfiguration === "profile-myNJ-logout") {
+      return [profileMenuItem(), myNjMenuItem(), logoutMenuItem()];
+    }
+    if (menuConfiguration === "profile-register-login") {
+      return [profileMenuItem(), registerMenuItem(), loginMenuItem()];
+    }
+    if (menuConfiguration === "login-getstarted") {
+      return [getStartedMenuItem(), loginMenuItem()];
+    }
```
aloverso, line 161: This is enough logic that seems like it could be unit-tested for this component independently of the integration-tests that are in NavBar.test

web/src/styles/sections/nav-bar.scss
adidner: So their might be a better way to write this if I gave it more thought but the idea was that the offset makes up for an icon not being with the title so as long as the Icon size is in PX we should keep the offset in PX.

web/src/lib/utils/helpers.ts
adidner: So I looked around a bit and it seems like we only have an abbreviator for LegalStructureId in the backend and it would need to be modified even more. And it seems like potentially good for a follow up,  as this entire case wasn't initially covered in the story. Thoughts?

For industryId is there already a function for Title case?

web/src/lib/utils/helpers.ts
adidner: Indeed.

web/src/components/navbar/NavBarPopupMenu.tsx
adidner: Sure

web/src/components/navbar/NavBarPopupMenu.tsx
adidner: So I had considered it the issue being largely that it isn't consistent across mobile and desktop in some but not all cases which would mean our division would be something like.

landing - mobile (landing desktop doesn't exist)
onboarding - mobile (onboarding desktop doesn't exist)
guest - mobile
guest - desktop 
authenticated - mobile & desktop

I also think this is more helpful to read in the NavBarMobile and NavBarDesktop components because you know when you render the NavBarPopupMenu exactly which parts you're getting based on the logic.

web/src/components/navbar/NavBarMobile.tsx
adidner: yes

web/src/components/navbar/NavBar.test.tsx
adidner: So the reason is that we need access to the userData to determine the name of the menu depending on businessName or industryId/LegalStructure and sometimes the names of buttons as well based on that

web/src/components/navbar/NavBar.test.tsx
adidner: Yeah, we can manually figure out what the name should be and look for that

web/src/components/navbar/NavBar.test.tsx
adidner: Yeah, I should probably use real values.

web/src/lib/utils/helpers.ts
adidner: This would proabbly live in the lib/utils not components/navbar right?

adidner: Also when you say this function should be tested. Initially I thought you meant in its own test file in utils but on second thought I think you probably mean in the NavMenuPopup test file implicitly. Is that correct?

adidner: When you say this should be tested. Do you mean impliclty in the tests for the NavBarPopupMenu? or do you mean explictly in its own test file? (Wouldn't testing in its own file be a bit of a weird practice since its just an individual function and we talked about how you don't rip out functions to test them alone?)

web/src/components/navbar/NavBarPopupMenu.tsx
```diff
@@ -0,0 +1,192 @@
...
+      itemText: Config.navigationDefaults.logoutButton,
+      key: "logoutMenuItem",
+    });
+  };
+
+  const myNjMenuItem = (): ReactElement => {
+    return NavMenuItem({
+      onClick: (): void => {
+        analytics.event.account_menu_myNJ_account.click.go_to_myNJ_home();
+        window.open(process.env.MYNJ_PROFILE_LINK || "", "_ blank");
```
adidner, line 83: So oddly enough this is the code thats in production for this line and it works as expected by opening a new tab but the "_ blank" suprises me as I thought it would be "_blank"

web/src/components/navbar/NavBarPopupMenu.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+    useMockUserData({});
+    renderNavBarPopupMenu({ hasCloseButton: false });
+    expect(screen.queryByTestId("close-button-nav-menu")).not.toBeInTheDocument();
+  });
+
+  describe("menuConfiguration prop's various menu configurations", () => {
+    describe("login configuration", () => {
+      it("shows login button", () => {
+        useMockUserData(generateOnboardingUser());
+        renderNavBarPopupMenu({ menuConfiguration: "login" });
```
adidner, line 106: So we have a weird situation here where these 2 lines are repeated a a few times each and I considered consolidating them but they each or making them one bigger function but render calls can only happen in it or describe blocks and beyond that I'm not really sure we gain much from doing so.

adidner: Similar situation with some of the checks I'm making in that we could consolidate but it feels to me like they don't actually happen that many times and it wouldn't increase.

Many of these checks happening 3 times which I think is perhaps in my range of acceptable copying? 

4 should be a function, 2 should be copied if right next to one another, 3 is a gray area for me? Curious as to others opinions

web/src/components/navbar/NavBarPopupMenu.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+jest.mock("@/lib/auth/signinHelper", () => {
+  const originalSigninHelper = jest.requireActual("@/lib/auth/signinHelper");
+  return {
+    ...originalSigninHelper,
+    onSignOut: jest.fn(),
+  };
+});
+jest.mock("@/lib/auth/sessionHelper", () => ({
+  triggerSignIn: jest.fn(),
+}));
```
adidner, line 38: So I mocked out, triggerSignIn and onSignOut, because they're tested in other parts of the code base independently. 

I'd already programmed the register section at that point so I figured I'd leave it but I can mock that out as well if this is bad for inconsistency reasons. 

I'm assuming mocks are fine because they're tested in depth in other places and these are unit tests but if others think I'm wrong I would love to know

adidner, line 38: Also might write a follow up because this is a bit weird how we have triggerSignIn and triggerSignOut and onSignOut and onSignIn and we're mixing them here which is also what we're doing in main I believe. Something to investigate and potentially refactor maybe?

web/src/components/navbar/NavBarPopupMenu.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+  it("renders close button when hasCloseButton prop is true", () => {
+    useMockUserData({});
+    renderNavBarPopupMenu({ hasCloseButton: true });
+    expect(screen.getByTestId("close-button-nav-menu")).toBeInTheDocument();
+  });
+
+  it("triggers handleClose when close button is pressed", () => {
+    useMockUserData({});
+    renderNavBarPopupMenu({ hasCloseButton: false });
+    expect(screen.queryByTestId("close-button-nav-menu")).not.toBeInTheDocument();
```
aloverso, line 99: I don't think this test name matches what is being tested

FYI, if you *did* want to test that it triggers handleClose (seems like a good idea, in addition to this close button false prop test here), the way to do that would be to pass a jest mock function into handleClose prop on line 78 and then expect it to have been called

web/src/components/navbar/NavBarPopupMenu.test.tsx
aloverso: this seems like an excellent situation to put these 2 (or even 3) lines in a beforeEach for the "login configuration" describe block

web/src/lib/domain-logic/getNavBarBusinessTitle.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+        businessName: "businessName",
+        industryId: "cannabis",
+        legalStructureId: undefined,
+      }),
+    });
+    const NavBarBusinessTitle = getNavBarBusinessTitle(userData);
+    expect(NavBarBusinessTitle).toBe(Config.navigationDefaults.navBarGuestText);
+  });
+
+  it("shows businessName if business name is defined adn legalStructureId and industryId are also defined", () => {
```
aloverso, line 33: typo

web/src/lib/domain-logic/getNavBarBusinessTitle.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+describe("getNavBarBusinessTitle", () => {
+  it("shows guest text if industryId is undefined", () => {
+    const userData = generateUserData({
+      profileData: generateProfileData({
+        businessName: "",
+        industryId: undefined,
+        legalStructureId: "limited-liability-company",
+      }),
+    });
+    const NavBarBusinessTitle = getNavBarBusinessTitle(userData);
```
aloverso, line 17: for this in all these tests, we should always camelCase and not PascalCase variable names

web/src/components/navbar/NavBarPopupMenu.tsx
```diff
@@ -0,0 +1,192 @@
...
+  };
+
+  const renderMenu = (): ReactElement[] | undefined | Array<ReactElement | ReactElement[]> => {
+    if (props.menuConfiguration === "login") {
+      return [loginMenuItem()];
+    }
+    if (props.menuConfiguration === "profile") {
+      return [profileMenuItem()];
+    }
+    if (props.menuConfiguration === "profile-myNJ-logout") {
```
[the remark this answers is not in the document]
adidner: Nope just forgot my bad, will fix now

web/src/lib/domain-logic/getNavBarBusinessTitle.test.tsx
adidner: Ah yes

web/src/components/navbar/NavBarPopupMenu.test.tsx
adidner: Yep yep yep, I totally wrote then and then meant to but forgot to come back to it

PR #5851, opened 2023-05-30
feat: [#185212073] adjust tooltip mobile behavior to stay open until …

adidner:
…clicked off

## Description

We wanted tooltips on mobile to open on click and not disappear until clicked off. 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185212073)

### Steps to Test

Go into the app and go to anywhere that we get tool tips. Some appear in profile after we finish formation because fields are locked and we have tool tips to explain. Also the tax calendar has some and a few other places. 
Confirm that on desktop you still hover as normal.
Then on mobile you click on the tool tip and it appears and stays forever and then you can click off to get rid of it. 

### Notes

So we realized that the focus state of the tool tips when clicked was a bit awkward because it was inside of the tool tip. So I made a minor change to take the margins and put them outside of the tool tips when they occur. 

## Code author checklist

### DIFF
--- a/web/src/components/ArrowTooltip.test.tsx
+++ b/web/src/components/ArrowTooltip.test.tsx
@@ -1,0 +1,64 @@
+import { ArrowTooltip } from "@/components/ArrowTooltip";
+import * as materialUi from "@mui/material";
+import { createTheme, Icon, ThemeProvider, useMediaQuery } from "@mui/material";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+
+jest.mock("@mui/material", () => mockMaterialUI());
+
+function mockMaterialUI(): typeof materialUi {
+  return {
+    ...jest.requireActual("@mui/material"),
+    useMediaQuery: jest.fn(),
+  };
+}
+
+const isMobile = (value: boolean): void => {
+  (useMediaQuery as jest.Mock).mockImplementation(() => {
+    return value;
+  });
+};
+
+const testString = "test_string";
+
+const renderToolTip = (): void => {
+  render(
+    <ThemeProvider theme={createTheme()}>
+      <ArrowTooltip title={testString}>
+        <div className="fdr fac font-body-lg text-green" data-testid="tooltip">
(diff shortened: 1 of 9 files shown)


web/src/components/onboarding/taxId/DisabledTaxId.tsx
```diff
@@ -103,13 +101,9 @@ export const DisabledTaxId = (props: Props): ReactElement => {
   const SimpleDiv = (props: { children: ReactNode }): ReactElement => (
     <div className="flex">
       <div>{props.children}</div>
```
adidner, line 103: Deleted because we have a tool tip in the header in this case and content confirmed we can delete.

web/src/components/ArrowTooltip.tsx
```diff
@@ -19,13 +21,47 @@ const useStylesBootstrap = makeStyles((theme: Theme) => {
...
+                disablePortal: true,
+              }}
+              onClose={(): void => setOpen(false)}
+              open={open}
+              disableFocusListener
+              disableHoverListener
+              disableTouchListener
+              {...props}
+            >
+              <div>
```
adidner, line 48: div wrapper is needed to make the MUI Tooltip component happy with its children.

web/src/components/ArrowTooltip.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it("shows onClick on mobile", async () => {
+    isMobile(true);
+
+    renderToolTip();
+
+    fireEvent.click(screen.getByTestId("tooltip"));
+    await waitFor(() => {
+      expect(screen.getByText(testString)).toBeInTheDocument();
```
aloverso, line 47: I think both this test and the one after it need a "before" case assertion to provide adequate confidence about the cause-and-effect here.

Basically: expect string to not be there, fire click, expect string to be there.

Only having the assertion afterwards doesn't prove that it came from the onclick (this test would still pass if the string is just permanently on the page always)

PR #5849, opened 2023-05-26
Fix ${oos} bug in Dakota onboarding flow

jphechter:
## Description

During the Dakota onboarding flow, the Industry question text was not replacing the ${oos} substring correctly. The hook responsible for this behavior has been updated to reference the `updateQueue` over the `userData` object.

Pasted Image at 2023_05_25_11_12 am

### Ticket

[185266819](https://www.pivotaltracker.com/story/show/185266819)

## Code author checklist

- [X] I have performed a self-review of my code
- [X] I have created and/or updated relevant documentation, if necessary
- [X] I have not used any relative imports
- [X] I have checked for and removed instances of unused config from CMS
- [X] I have pruned any instances of unused code
- [X] I have not added any markdown to labels, titles and button text in config

PR #5852, opened 2023-05-31
Fix ${oos} bug in Dakota onboarding flow

jphechter:
## Description

Follow up for #5849. Previous PR only addressed ${oos} for Dakota formation when user was logged in. Fixes state issue for users forming through marketing link containing industry query string.

### Ticket

[185266819](https://www.pivotaltracker.com/story/show/185266819)

### Steps to Test

1. Open incognito window
1. Go to https://navigator.business.nj.gov/onboarding?industry=cannabis
1. See ${oos} in the industry question (screenshot below)

## Code author checklist

### DIFF
--- a/web/src/pages/onboarding.tsx
+++ b/web/src/pages/onboarding.tsx
@@ -199,40 +199,42 @@ const OnboardingPage = (props: Props): ReactElement => {
       }
     })();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [router.isReady, state.user, state.isAuthenticated]);
+  }, [router.isReady, state.user, state.isAuthenticated, updateQueue]);
 
   const setIndustryAndRouteToPage = async (
     userData: UserData,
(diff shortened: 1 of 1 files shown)


web/src/pages/onboarding.tsx
```diff
@@ -216,23 +216,25 @@ const OnboardingPage = (props: Props): ReactElement => {
     } else {
       setPage({ current: 3, previous: 2 });
     }
+    await updateQueue?.queueProfileData(profileData);
```
aloverso, line 219: For these, I'd recommend using a local variable for both the setProfileData and the profileData that gets queued here.

SetState functions aren't actually synchronous so even though we're calling setProfileData up there, you can't guarantee that it will be set to the new value at this point in the code.

jphechter: Fair point. I'll make that update.

PR #5865, opened 2023-06-01
feat: [#185141116] naics code & location confirmed in tax access modal

somabadri:
### Ticket

## Code author checklist


api/src/client/ApiTaxFilingClient.ts
```diff
@@ -49,10 +49,26 @@ export const ApiTaxFilingClient = (config: ApiConfig, logger: LogWriterType): Ta
       logger.LogInfo(
         `TaxFiling Lookup - NICUSA - Id:${logId} - Response received: ${JSON.stringify(response.data)}`
       );
+
       const apiResponse = response.data as ApiTaxFilingLookupResponse;
+
+      const city = apiResponse.Data?.find((x) => {
+        if (x.Name === "tax-city") {
```
aloverso, line 56: Let's avoid using `x` as a variable and pick something else. Even a little more descriptive like `item` would be preferable to a single character.

web/src/components/filings-calendar/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -188,6 +188,45 @@ describe("<FilingsCalendarTaxAccess />", () => {
...
     mockApi.postTaxFilingsOnboarding.mockResolvedValue({
       ...userDataWithPrefilledFields,
+      profileData: {
+        ...userDataWithPrefilledFields.profileData,
+        naicsCode: "123456",
+        municipality: {
+          name: "Absecon",
+          county: "",
+          id: "",
+          displayName: "",
```
aloverso, line 198: Why do we need to add this new data to the mock? I don't see anything in the expectations that is expecting the municipality or naics code to appear on the page

aloverso: same question for the test below

web/src/components/filings-calendar/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -202,6 +241,18 @@ describe("<FilingsCalendarTaxAccess />", () => {
     });
     await screen.findByTestId("tax-success");
     expect(screen.getByText(Config.taxCalendar.snackbarSuccessHeader)).toBeInTheDocument();
+    expect(screen.getByTestId("close-icon-button")).toBeInTheDocument();
+    jest.useFakeTimers();
+    act(() => {
+      jest.advanceTimersByTime(6000);
+    });
```
aloverso, line 248: I'm not sure this is necessary to test.

"Doesn't auto close alert" is a little strange to test because it only makes sense in the context of how it USED to work.

Someone new coming into the code wouldn't have that context, and this test wouldn't make any sense. It's drawing on assumptions the code USED to make but doesn't anymore. They'd be like "Why 6000 seconds?" - Because it USED to work that way, but doesn't anymore. 

At a certain point in the future, this would make as much sense as testing it "does not display 100 digits of pi" or whatever random thing. You'd just be like, sure, it doesn't, but why are we testing that?

It would be reasonable here to test that the close button works. No reason to test that the alert is NOT timed

web/src/components/filings-calendar/tax-access-modal/TaxAccessStepTwo.tsx
```diff
@@ -159,8 +160,27 @@ export const TaxAccessStepTwo = (props: Props): ReactElement => {
...
+
+          userDataWithUpdatedMunicipality = {
+            ...userDataToSet,
+            profileData: {
+              ...userDataToSet.profileData,
+              municipality: {
+                name: record.townName,
+                county: record.countyName,
+                id: record.id,
+                displayName: record.townDisplayName,
```
aloverso, line 176: why are we not using the `queueProfileData` that already existed down below here?

web/src/components/filings-calendar/tax-access-modal/TaxAccessStepTwo.tsx
```diff
@@ -179,6 +199,7 @@ export const TaxAccessStepTwo = (props: Props): ReactElement => {
         setIsLoading(false);
         analytics.event.tax_calendar_modal.submit.tax_deadlines_added_to_calendar();
         props.onSuccess();
+        await updateQueue.queueTaskProgress({ "determine-naics-code": "COMPLETED" }).update();
```
aloverso, line 202: we need to use the `useUpdateTaskProgress` here instead so that the section-completion logic gets triggered as needed

web/src/components/onboarding/LockedProfileField.tsx
```diff
@@ -12,12 +12,17 @@ export const LockedProfileField = (props: Props): ReactElement => {
   const { state } = useContext(ProfileDataContext);
 
   let valueToDisplay = state.profileData[props.fieldName] as string;
+
+  if (props.fieldName === "municipality") {
```
aloverso, line 16: I'll be honest, I don't love this approach because I don't love the hardcoding of specific field names and their specialized logic into an if condition.

However, I can't think of a solution right now that's better. so 🤷

web/src/components/profile/ProfileNaicsCode.test.tsx
```diff
@@ -52,18 +55,33 @@ describe("<ProfileNaicsCode />", () => {
...
   });
 
-  it("Display Edit when there is naics code", () => {
+  it("display Edit when there is naics code", () => {
     renderComponent({ naicsCode: "624410" });
     expect(screen.getByText(configForField.editText)).toBeInTheDocument();
     expect(screen.queryByText(configForField.addText)).not.toBeInTheDocument();
   });
 
+  it("doesn't display Edit text and displays tooltip when there is a naics code and operating phase is up and running or up and running owning", () => {
```
aloverso, line 64: Is this correct?

If a user is a Poppy who has graduated to UP_AND_RUNNING and has not yet done their tax access lookup, shouldn't this still be editable?

web/src/components/profile/ProfileNaicsCode.test.tsx
```diff
@@ -52,18 +55,33 @@ describe("<ProfileNaicsCode />", () => {
...
+    expect(screen.queryByTestId("naics-code-tooltip")).not.toBeInTheDocument();
+  });
+
   it("displays Not Entered text when user has no NAICS code", () => {
     renderComponent({ naicsCode: "" });
     expect(screen.getByTestId("not-entered")).toBeInTheDocument();
   });
 
-  it("Displays Add text when naics code field is empty", () => {
+  it("displays Add text when naics code field is empty", () => {
```
aloverso, line 84: is this different between UP_AND_RUNNING_OWNING and others?

Because the Add text wouldn't link to anything for an Oscar - they don't have a roadmap

web/src/components/profile/ProfileNaicsCode.tsx
```diff
@@ -27,13 +29,34 @@ export const ProfileNaicsCode = (): ReactElement => {
     return `/tasks/${urlSlug}`;
   }, [roadmap]);
 
+  const lockField = (): boolean => {
```
aloverso, line 32: i would prefer to name this something like `shouldLockField`, which clearly communicates that it returns a boolean

Whereas `lockField` sounds like an action, ie, invoking this function will perform the action to lock the field

web/src/components/tasks/NaicsCodeDisplay.tsx
```diff
@@ -21,19 +24,32 @@ export const NaicsCodeDisplay = (props: Props): ReactElement => {
...
-        </UnStyledButton>
+        {props.lockField ? (
+          <ArrowTooltip
+            title={Config.profileDefaults.lockedFieldTooltipText}
+            data-testid="naics-code-tooltip"
+          >
+            <div className="fdr fac font-body-lg">
+              <Icon>help_outline</Icon>
+            </div>
+          </ArrowTooltip>
```
aloverso, line 41: As a note (maybe we should add this as a refactoring chore)

After Faraz merges his changes to the Business STructure task, he made some almost identical changes to lock the business structure alert versus showing the edit/remove text there.

After both these are merged, we should refactor these to use a shared component for the shared behavior

web/src/components/tasks/NaicsCodeTask.test.tsx
```diff
@@ -329,6 +341,33 @@ describe("<NaicsCodeTask />", () => {
...
     });
+
+    it("doesn't display the edit or remove button when the naics code is present and the operating phase is up and running", () => {
+      renderPage("UP_AND_RUNNING");
+      expect(screen.queryByText(Config.taskDefaults.editText)).not.toBeInTheDocument();
+      expect(screen.queryByText(Config.taskDefaults.removeText)).not.toBeInTheDocument();
+      expect(screen.getByTestId("naics-code-tooltip")).toBeInTheDocument();
+    });
+
+    it("displays the edit or remove button when the naics code is present and the operating phase is not up and running", () => {
```
aloverso, line 352: edit AND remove?

web/src/components/tasks/NaicsCodeTask.test.tsx
```diff
@@ -329,6 +341,33 @@ describe("<NaicsCodeTask />", () => {
...
+    });
+
+    it("displays the tooltip when the naics code is present and the operating phase is up and running", () => {
+      renderPage("UP_AND_RUNNING");
+      expect(screen.queryByText(Config.taskDefaults.editText)).not.toBeInTheDocument();
+      expect(screen.queryByText(Config.taskDefaults.removeText)).not.toBeInTheDocument();
+      expect(screen.getByTestId("naics-code-tooltip")).toBeInTheDocument();
+    });
+
+    it("doesn't display the tooltip when the naics code is present and the operating phase is not up and running", () => {
```
aloverso, line 365: might be easier to read all these tests if nested under a `describe("when the naics code is present")`

web/src/lib/roadmap/buildUserRoadmap.test.ts
```diff
@@ -10,6 +9,7 @@ import {
   emptyIndustrySpecificData,
   ProfileData,
 } from "@businessnjgovnavigator/shared/profileData";
+import * as fetchMunicipality from "../async-content-fetchers/fetchMunicipalities";
```
aloverso, line 12: relative import

web/src/pages/profile.tsx
```diff
@@ -275,6 +275,14 @@ const ProfilePage = (props: Props): ReactElement => {
     );
   };
 
+  const lockMunicipality = (): boolean => {
```
aloverso, line 278: same comment re function name for boolean

web/src/pages/profile.tsx
```diff
@@ -275,6 +275,14 @@ const ProfilePage = (props: Props): ReactElement => {
...
+    return (
+      !!profileData.municipality &&
+      (profileData.operatingPhase === "UP_AND_RUNNING_OWNING" ||
+        profileData.operatingPhase === "UP_AND_RUNNING")
```
aloverso, line 282: shouldn't this be based on whether or not tax access was success or not? This is just saying the field is locked for all Oscars and all up-and-running Poppys, always?

web/src/components/profile/ProfileNaicsCode.tsx
```diff
@@ -27,13 +29,34 @@ export const ProfileNaicsCode = (): ReactElement => {
...
+    return (
+      state.profileData.naicsCode !== "" &&
+      (state.profileData.operatingPhase === "UP_AND_RUNNING" ||
+        state.profileData.operatingPhase === "UP_AND_RUNNING_OWNING")
```
aloverso, line 36: shouldn't this be based on whether or not tax access was success or not? This is just saying the field is locked for all Oscars and all up-and-running Poppys, always?

web/src/pages/profile.tsx
```diff
@@ -427,6 +446,13 @@ const ProfilePage = (props: Props): ReactElement => {
             )}
           </div>
         </ProfileField>
+        {userData?.profileData.naicsCode && (
```
aloverso, line 449: I don't think we want to add this to foreign non-nexus profile?

web/src/components/profile/ProfileNaicsCode.test.tsx
somabadri: The add or edit text would never show up for an oscar because the field only shows up if their NAICS code is populated and it also always locked. I just changed the capitalization of the word displays in the test.

web/src/components/filings-calendar/FilingsCalendarTaxAccess.test.tsx
```diff
@@ -189,6 +192,9 @@ describe("<FilingsCalendarTaxAccess />", () => {
   it("displays alert on success", async () => {
     mockApi.postTaxFilingsOnboarding.mockResolvedValue({
       ...userDataWithPrefilledFields,
+      profileData: {
+        ...userDataWithPrefilledFields.profileData,
+      },
```
aloverso, line 197: This is all unnecessary, if it's just spreading, right? In other tests also

PR #5871, opened 2023-06-02
Calendar license #184752321

mikewolfd:
### Ticket

## Code author checklist


shared/src/license.ts
```diff
@@ -5,6 +5,7 @@ export type LicenseStatusItem = {
 
 export type LicenseStatusResult = {
   status: LicenseStatus;
+  expirationISO?: string;
```
aloverso, line 8: Why should this be potentially undefined? When would it ever not exist but the items and status still do?

web/public/mgmt/config.yml
```diff
@@ -14,6 +14,11 @@ local_backend: true
 media_folder: web/public/img
 logo_url: https://dev.navigator.business.nj.gov/img/Navigator-logo.svg
 public_folder: web/public
+slug:
+  encoding: "ascii"
+  clean_accents: true
+  sanitize_replacement: "_"
```
aloverso, line 20: What does this do?

web/src/components/filings-calendar/FilingsCalendarSingleGrid.tsx
```diff
@@ -41,10 +43,51 @@ export const FilingsCalendarSingleGrid = (props: Props): ReactElement => {
...
+              <span className="text-no-uppercase text-underline text-base-dark">
+                {licenseEvent.licenseType} {titles[licenseEvent.type]}
+              </span>
+            </a>
+          </Link>
+        </Tag>
+      </div>
+    ) : (
+      <></>
+    );
```
aloverso, line 88: This seems ripe to be its own component.

Also, we avoid `xx ? <stuff> : <></>` notation in favor of `xx && <stuff>`

web/src/components/filings-calendar/FilingsCalendarSingleGrid.tsx
```diff
@@ -41,10 +43,51 @@ export const FilingsCalendarSingleGrid = (props: Props): ReactElement => {
...
 
+  const licenseEvent = getLicenseCalendarEvent(
+    parseDate(props.userData?.licenseData?.expirationISO).add(4, "year").toISOString(),
+    date,
+    props.userData.profileData.industryId
+  );
+
+  if (licenseEvent && visibleFilings.length === NUM_OF_FILINGS_ALWAYS_VIEWABLE) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    remainingFilings.unshift(visibleFilings.pop()!);
```
aloverso, line 58: What is this doing?

web/src/lib/types/types.ts
```diff
@@ -381,6 +381,24 @@ export interface Filing {
   agency?: TaxAgency | null;
 }
 
+export interface License {
```
aloverso, line 384: it seems like we should call this a LicenseEvent or a LicenseFiling... "License" is pretty generic and I don't think it represents what this is.

web/src/lib/domain-logic/getLicenseCalendarEvent.ts
```diff
@@ -0,0 +1,33 @@
...
+import { parseDate } from "@businessnjgovnavigator/shared/dateHelpers";
+import { LookupIndustryById } from "@businessnjgovnavigator/shared/index";
+import { Dayjs } from "dayjs";
+
+export const getLicenseCalendarEvent = (
+  expirationISO: string | undefined,
+  date: Dayjs,
+  industryId: string | undefined
+): LicenseCalendarEvent | undefined => {
+  if (expirationISO !== undefined && industryId) {
```
aloverso, line 11: This should be a guard clause instead

web/src/components/filings-calendar/FilingsCalendarSingleGrid.tsx
```diff
@@ -41,10 +43,51 @@ export const FilingsCalendarSingleGrid = (props: Props): ReactElement => {
     );
   });
 
-  const firstTwoFilings = thisMonthFilings.slice(0, NUM_OF_FILINGS_ALWAYS_VIEWABLE);
+  const visibleFilings = thisMonthFilings.slice(0, NUM_OF_FILINGS_ALWAYS_VIEWABLE);
   const remainingFilings = thisMonthFilings.slice(NUM_OF_FILINGS_ALWAYS_VIEWABLE);
   const isOnCurrentYear = getCurrentDate().year().toString() === props.activeYear;
 
+  const licenseEvent = getLicenseCalendarEvent(
+    parseDate(props.userData?.licenseData?.expirationISO).add(4, "year").toISOString(),
```
aloverso, line 51: Why are we adding 4 years? And why adding it here, in a component, and not in the function? That absolutely seems like domain logic, not render logic.

If the expiration date we're getting from the backend is off by 4 years, shouldn't we even do this on the backend before saving it to begin with?

aloverso:
This is missing all of its frontend tests

web/public/mgmt/config.yml
mikewolfd: Stops people from putting weird characters in the slug field, like Unicode stuff, but I can remove that from here

web/src/components/filings-calendar/FilingsCalendarSingleGrid.tsx
aloverso: Is that part of the story? I didn't see anything about that in the AC

PR #5880, opened 2023-06-06
[184306224] Dakota Auto-Name Check During Formation

jphechter:
## Description

Updates Dakota formation flow to check if a submitted business name or DBA name has become unavailable since it was originally submitted.

### Ticket

[184306224](https://www.pivotaltracker.com/story/show/184306224)

### Approach

Builds off the work that currently exists to do this for Oscar and Poppy users. Additionally, includes a partial refactor to consolidate Nexus and DBA business name search.

## Code author checklist

### DIFF
--- a/api/src/api/userRouter.test.ts
+++ b/api/src/api/userRouter.test.ts
@@ -50,6 +50,9 @@ const cognitoPayload = ({ id }: { id: string }): any => {
   };
 };
 
+const fiftyNineMinutesAgo: string = getCurrentDate().subtract(1, "hour").add(1, "minute").toISOString();
+const sixtyOneMinutesAgo: string = getCurrentDate().subtract(1, "hour").subtract(1, "minute").toISOString();
+
 describe("userRouter", () => {
   let app: Express;
 
@@ -183,7 +186,7 @@ describe("userRouter", () => {
       it("does not update license if licenseData lastCheckedDate is within the last hour", async () => {
         const userData = generateUserData({
           licenseData: generateLicenseData({
-            lastUpdatedISO: getCurrentDate().subtract(1, "hour").add(1, "minute").toISOString(),
+            lastUpdatedISO: fiftyNineMinutesAgo,
           }),
         });
         stubUserDataClient.get.mockResolvedValue(userData);
@@ -198,7 +201,7 @@ describe("userRouter", () => {
             industryId: "home-contractor",
           }),
           licenseData: generateLicenseData({
-            lastUpdatedISO: getCurrentDate().subtract(1, "hour").subtract(1, "minute").toISOString(),
+            lastUpdatedISO: sixtyOneMinutesAgo,
           }),
         });
         stubUserDataClient.get.mockResolvedValue(userData);
@@ -216,7 +219,7 @@ describe("userRouter", () => {
             industryId: "home-contractor",
           }),
           licenseData: generateLicenseData({
-            lastUpdatedISO: getCurrentDate().subtract(1, "hour").subtract(1, "minute").toISOString(),
+            lastUpdatedISO: sixtyOneMinutesAgo,
           }),
         });
         stubUserDataClient.get.mockResolvedValue(userData);
@@ -248,7 +251,7 @@ describe("userRouter", () => {
           formationData: generateFormationData({
             businessNameAvailability: generateBusinessNameAvailability({
               status: "AVAILABLE",
-              lastUpdatedTimeStamp: getCurrentDate().subtract(1, "hour").add(1, "minute").toISOString(),
+              lastUpdatedTimeStamp: fiftyNineMinutesAgo,
             }),
           }),
         });
@@ -264,7 +267,7 @@ describe("userRouter", () => {
             completedFilingPayment: true,
             businessNameAvailability: generateBusinessNameAvailability({
               status: "AVAILABLE",
-              lastUpdatedTimeStamp: getCurrentDate().subtract(1, "hour").add(1, "minute").toISOString(),
+              lastUpdatedTimeStamp: fiftyNineMinutesAgo,
             }),
           }),
         });
@@ -280,7 +283,7 @@ describe("userRouter", () => {
             completedFilingPayment: false,
             businessNameAvailability: generateBusinessNameAvailability({
               status: "AVAILABLE",
-              lastUpdatedTimeStamp: getCurrentDate().subtract(1, "hour").subtract(1, "minute").toISOString(),
+              lastUpdatedTimeStamp: sixtyOneMinutesAgo,
             }),
           }),
         });
@@ -295,6 +298,7 @@ describe("userRouter", () => {
         const result = await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
         expect(stubTimeStampBusinessSearch.search).toHaveBeenCalled();
         expect(result.body.formationData.businessNameAvailability.status).toEqual("UNAVAILABLE");
+        expect(result.body.profileData.needsNexusDbaName).toEqual(false);
         expect(
           parseDate(result.body.formationData.lastUpdatedTimeStamp).isSame(getCurrentDate(), "minute")
         ).toEqual(true);
@@ -306,7 +310,7 @@ describe("userRouter", () => {
             completedFilingPayment: false,
             businessNameAvailability: generateBusinessNameAvailability({
               status: "AVAILABLE",
-              lastUpdatedTimeStamp: getCurrentDate().subtract(1, "hour").subtract(1, "minute").toISOString(),
+              lastUpdatedTimeStamp: sixtyOneMinutesAgo,
             }),
           }),
         });
@@ -321,6 +325,7 @@ describe("userRouter", () => {
         const result = await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
         expect(stubTimeStampBusinessSearch.search).toHaveBeenCalled();
         expect(result.body.formationData.businessNameAvailability.status).toEqual("UNAVAILABLE");
+        expect(result.body.profileData.needsNexusDbaName).toEqual(false);
         expect(
           parseDate(result.body.formationData.lastUpdatedTimeStamp).isSame(getCurrentDate(), "minute")
         ).toEqual(true);
@@ -332,7 +337,7 @@ describe("userRouter", () => {
             completedFilingPayment: false,
             businessNameAvailability: generateBusinessNameAvailability({
               status: "AVAILABLE",
-              lastUpdatedTimeStamp: getCurrentDate().subtract(1, "hour").subtract(1, "minute").toISOString(),
+              lastUpdatedTimeStamp: sixtyOneMinutesAgo,
             }),
           }),
         });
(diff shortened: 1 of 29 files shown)


web/src/components/tasks/business-formation/name/NexusSearchBusinessNameStep.test.tsx
```diff
@@ -52,34 +61,55 @@ function setupMockAnalytics(): typeof analytics {
...
-describe("<NexusSearchBusinessNameStep />", () => {
-  const initialUserData = generateUserData({
-    formationData: generateFormationData({ formationFormData: createEmptyFormationFormData() }),
-    profileData: generateProfileData({
-      businessName: "",
-      nexusDbaName: "",
-      needsNexusDbaName: false,
-    }),
-  });
+describe("Formation - NexusSearchBusinessNameStep", () => {
```
jphechter, line 64: This is essentially a rewrite of the file, but all of the tests should be the same.

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,40 +48,104 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+                })
+                .update();
+            }
+
+            setFormationFormData((previousFormationData) => {
+              return {
+                ...previousFormationData,
+                businessName: submittedName,
+              };
+            });
```
jphechter, line 142: This is logic is ultimately going to live in `useBusinessNameSearch` when Business Name Search also uses this component. Was keeping this together for quick comparative purposes. Given that this is happening before the refactor is done, maybe it should be broken apart.

api/src/api/userRouter.test.ts
```diff
@@ -343,6 +345,186 @@ describe("userRouter", () => {
         expect(stubTimeStampBusinessSearch.search).toHaveBeenCalledWith(userData.profileData.businessName);
         expect(stubUpdateOperatingPhase).toHaveBeenCalledWith(userData);
       });
+
+      describe("when businessPersona is 'FOREIGN'", () => {
+        it("does not update business name search if dbaBusinessNameAvailability is undefined", async () => {
```
aloverso, line 350: this is only true if `businessNameSearch` is ALSO undefined right? 
If `businessNameSearch` was defined and `dbaBusinessNameAvailability` was undefined then it WOULD do a search, right?

api/src/api/userRouter.test.ts
```diff
@@ -343,6 +345,186 @@ describe("userRouter", () => {
...
+              }),
+            }),
+          });
+          stubUserDataClient.get.mockResolvedValue(userData);
+
+          await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
+          expect(stubTimeStampBusinessSearch.search).not.toHaveBeenCalled();
+        });
+
+        it("updates businessNameAvailability and sets needsNexusDbaName", async () => {
```
aloverso, line 386: It would be nice to describe the logical flow here, like "it sets needsNexusDbaName to true when business name becomes unavailable"

api/src/api/userRouter.test.ts
```diff
@@ -343,6 +345,186 @@ describe("userRouter", () => {
...
+                  .subtract(1, "hour")
+                  .subtract(1, "minute")
+                  .toISOString(),
+              }),
+              dbaBusinessNameAvailability: generateBusinessNameAvailability({
+                status: "AVAILABLE",
+                lastUpdatedTimeStamp: getCurrentDate()
+                  .subtract(1, "hour")
+                  .subtract(1, "minute")
+                  .toISOString(),
```
aloverso, line 435:
```
getCurrentDate()
  .subtract(1, "hour")
  .subtract(1, "minute")
  .toISOString(),
```

is used all over the place and seems like a great opportunity for an extract function call

api/src/api/userRouter.ts
```diff
@@ -60,15 +61,23 @@ const shouldCheckLicense = (userData: UserData): boolean => {
...
-  );
+
+  const isDba = userData.profileData.needsNexusDbaName;
+  const shouldUpdateNameAvailability = isDba
+    ? userData.profileData.nexusDbaName !== undefined &&
+      userData.formationData.dbaBusinessNameAvailability !== undefined &&
+      hasBeenMoreThanOneHour(userData.formationData.dbaBusinessNameAvailability.lastUpdatedTimeStamp)
+    : userData.profileData.businessName !== undefined &&
+      userData.formationData.businessNameAvailability !== undefined &&
+      hasBeenMoreThanOneHour(userData.formationData.businessNameAvailability.lastUpdatedTimeStamp);
```
aloverso, line 78: This is a lot of &&

Maybe could be refactored and broken down into separate statements / variables to be more readable?

api/src/api/userRouter.ts
```diff
@@ -235,17 +245,35 @@ export const userRouterFactory = (
...
+
+      const nameAvailability: NameAvailability = {
+        status: response.status,
+        similarNames: response.similarNames ?? [],
+        lastUpdatedTimeStamp: response.lastUpdatedTimeStamp,
+        invalidWord: response.invalidWord,
+      };
+
+      const needsNexusDbaName =
+        isForeign && !isDba ? response.status !== "AVAILABLE" : userData.profileData.needsNexusDbaName;
```
aloverso, line 261: Again, if the variable assignment doesn't fit on a single line, it would be worth considering how we could break it down using temp variables maybe to make it more readable

web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
```diff
@@ -161,7 +161,11 @@ export const DbaFormationPaginator = (): ReactElement => {
   };
 
   const displayButtons = (): ReactNode => {
-    if (state.stepIndex === 0 && state.businessNameAvailability?.status === "AVAILABLE") {
+    if (
+      state.stepIndex === 0 &&
+      ((isNotDba && state.businessNameAvailability?.status === "AVAILABLE") ||
+        (!isNotDba && state.dbaBusinessNameAvailability?.status === "AVAILABLE"))
```
aloverso, line 167: I gotta say, `!isNotDba` .... could be renamed I think

aloverso: Again, I think things like this would be easier to read if it was extracted into a variable. Like, we want to be able to quickly read a line and understand the if-case for these buttons, and the added complexity makes it hard.

Something like `if (state.stepIndex === 0 &&& nameIsAvailable)`

web/src/components/tasks/business-formation/name/DbaNameSearch.tsx
```diff
@@ -7,51 +7,32 @@ import { useConfig } from "@/lib/data-hooks/useConfig";
...
 import { useMountEffect } from "@/lib/utils/helpers";
-import { NameAvailability } from "@businessnjgovnavigator/shared/";
 import { ReactElement, useContext } from "react";
 
 export const DbaNameSearch = (): ReactElement => {
   const { Config } = useConfig();
-  const { updateQueue } = useUserData();
-  const { setBusinessNameAvailability } = useContext(BusinessFormationContext);
+  const { state } = useContext(BusinessFormationContext);
+  const { updateQueue: userData } = useUserData();
```
aloverso, line 15: I don't understand this syntax, actually. What is this line doing with the colon?

aloverso: I see... below, we're using `userData?.current()` instead of `updateQueue.current()`

Why? This seems like a really confusing pattern for other developers, in terms of breaking expectations of what people expect when they see a variable named `userData`

web/src/components/tasks/business-formation/name/NexusSearchBusinessNameStep.test.tsx
```diff
@@ -89,115 +117,96 @@ describe("<NexusSearchBusinessNameStep />", () => {
...
-      profileData: generateProfileData({
-        businessName: "some cool name",
-        needsNexusDbaName: true,
-      }),
-    });
-    mockSearchReturnValue({ status: "UNAVAILABLE" });
-    renderTask();
-    await waitFor(() => {
-      expect(screen.getByText(markdownToText(Config.nexusNameSearch.dbaNameHeader))).toBeInTheDocument();
+  it("does not show DBA search immediately if it is needsNexusDbaName is false in profile", async () => {
```
aloverso, line 171: Why the deletion of the corresponding "if true" test?

web/src/components/tasks/search-business-name/SearchBusinessNameForm.test.tsx
```diff
@@ -9,102 +14,126 @@ import {
...
+      ),
+      formationResponse: undefined,
+      getFilingResponse: undefined,
+      completedFilingPayment: false,
+      businessNameAvailability: undefined,
+      dbaBusinessNameAvailability: undefined,
+      lastVisitedPageIndex: 0,
+    };
+    const page = preparePage(generateUserData({ profileData, formationData }), displayContent);
+    return page;
```
aloverso, line 81: It seems like this function is used in like all of these tests. Could it be extracted as a helper?

web/src/components/tasks/search-business-name/SearchBusinessNameForm.test.tsx
```diff
@@ -9,102 +14,126 @@ import {
...
+    const page = await getPageHelper();
+    await page.searchBusinessName({ status: "AVAILABLE" });
+
+    expect(true).toBe(true);
+  });
 
-  it("pre-fills the text field with the business name entered in onboarding", () => {
-    useMockProfileData({ businessName: "Best Pizza" });
-    renderForm();
+  it("pre-fills the text field with the business name entered in onboarding", async () => {
```
aloverso, line 91: maybe "business name in profile" might be more descriptive now that onboarding doesnt ask for Business Name anymore

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,40 +48,104 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+          } else {
+            if (!nameAvailability || !updateQueue || isInitialSubmit) {
+              return;
+            }
+            setFieldsInteracted([FIELD_NAME]);
+            if (nameAvailability.status === "AVAILABLE") {
+              await updateQueue
+                .queueFormationData({
+                  formationFormData: {
+                    ...updateQueue.current().formationData.formationFormData,
```
aloverso, line 110: For a nested field like this, I would rather we write a `queueFormationFormData` function that do the spreads inline like this

aloverso: Not the least of which because of the upcoming Ethan refactoring

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,40 +48,104 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+                    businessName: submittedName,
+                  },
+                  businessNameAvailability: nameAvailability,
+                })
+                .queueProfileData({
+                  businessName: submittedName,
+                  needsNexusDbaName: true,
+                })
```
aloverso, line 133: 90% of these updates are in common between the AVAILABLE and UNAVAILABLE case. It seems like this duplication could be refactored by queuing the shared changes first and then using the if/else structure only to change the things are are actually different between the two cases

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,40 +48,104 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+                  },
+                  dbaBusinessNameAvailability: nameAvailability,
+                })
+                .queueProfileData({
+                  nexusDbaName: submittedName,
+                })
+                .update();
+            }
+          } else {
+            if (!nameAvailability || !updateQueue || isInitialSubmit) {
```
aloverso, line 102: `!nameAvailability || !updateQueue` is a guard clause shared between the if/else cases and should be lifted to top-level to simplify all the nested if clauses

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,40 +48,104 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
-          setNameAvailability(nameAvailability);
-          setSubmittedName(submittedName);
-          if (onSubmit) {
-            onSubmit(submittedName, nameAvailability, isInitialSubmit);
+        .then(async ({ nameAvailability, submittedName }) => {
+          if (props.isDba) {
+            if (!nameAvailability || !updateQueue) {
+              return;
+            }
+            setNameAvailability(nameAvailability);
```
aloverso, line 87: How come we dont do this in the else case?

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -174,42 +234,55 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+                      ...updateQueue.current().formationData.formationFormData,
+                    },
+                    businessNameAvailability: emptyNameAvailability,
+                    dbaBusinessNameAvailability: emptyNameAvailability,
+                  })
+                  .queueProfileData({
+                    nexusDbaName: "",
+                  })
+                  .update();
+              }
```
aloverso, line 282: A function this large shouldn't be inline in JSX and instead should be extracted as its own named function

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,40 +48,104 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+            if (nameAvailability.status === "AVAILABLE") {
+              await updateQueue
+                .queueFormationData({
+                  formationFormData: {
+                    ...updateQueue.current().formationData.formationFormData,
+                  },
+                  dbaBusinessNameAvailability: nameAvailability,
+                })
+                .queueProfileData({
+                  nexusDbaName: submittedName,
```
aloverso, line 97: Why do we have this data structure where `businessName` exists in both `profileData` and `formationFormData` indepedently, and have to keep each in line with each other, whereas for `nexusDbaName` we don't store it in `formationFormData` at all and *only* update profile data when something is entered/searched here?

api/src/api/userRouter.test.ts
```diff
@@ -343,6 +345,186 @@ describe("userRouter", () => {
...
+            }),
+            formationData: generateFormationData({ dbaBusinessNameAvailability: undefined }),
+          });
+          stubUserDataClient.get.mockResolvedValue(userData);
+
+          await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
+          expect(stubTimeStampBusinessSearch.search).not.toHaveBeenCalled();
+        });
+
+        it("does not update dbaBusinessNameAvailability if needsNexusDbaName is false", async () => {
```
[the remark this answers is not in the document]
jphechter: I'm using `dbaBusinessNameAvailability.lastUpdatedTimeStamp` as a reference to recheck availability. I'd also have to recheck the availability every time we show the `SearchBusinessNameForm`.

I think `dbaBusinessNameSearch` and `needsNexusDbaName` serve unique enough purposes that it makes sense to have both.

api/src/api/userRouter.test.ts
jphechter: `businessNameSearch` is in reference to the object that used to contain `businessNameAvailability`, correct?

Yes, I can make this more explicit. If neither `businessNameAvailability` or `dbaBusinessNameAvailability` are defined, the search will not happen.

api/src/api/userRouter.test.ts
jphechter: Agreed. I couldn't connect that phrasing when I wrote this initially.

web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
jphechter: renamed and replaced with `isDba`
broke out the status comparison

web/src/components/tasks/business-formation/name/DbaNameSearch.tsx
jphechter: I don't remember my exact motivation. Changing that back to `updateQueue`.

web/src/components/tasks/business-formation/name/NexusSearchBusinessNameStep.test.tsx
jphechter: Bad copy paste job. It's back in there.

api/src/api/userRouter.ts
jphechter: included

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: I honestly couldn't tell you, and I think that should be looked at as a follow on to this.

jphechter: Based on our types it looks like `businessName` is also in `TaxFilingData`. I'm not looking super deep into the code, but there may be potentially 3 places we store the business name.

jphechter: [Refactor how businessName is stored](https://www.pivotaltracker.com/story/show/185356746)

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: I agree. I wasn't super concerned about this because I thought I was going to immediately handle it with the refactor of BusinessNameSearch that I was trying to tackle following this. Not really an excuse for keeping it like this though.

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: I think that was a mistake.

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: Good callout

web/src/components/tasks/search-business-name/SearchBusinessNameForm.test.tsx
jphechter: updated

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,40 +48,104 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+            if (!nameAvailability || !updateQueue) {
+              return;
+            }
+            setNameAvailability(nameAvailability);
+            if (nameAvailability.status === "AVAILABLE") {
+              await updateQueue
+                .queueFormationData({
+                  formationFormData: {
+                    ...updateQueue.current().formationData.formationFormData,
+                  },
```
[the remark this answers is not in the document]
jphechter: updated

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: included

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: simplified

web/src/components/tasks/search-business-name/SearchBusinessNameForm.test.tsx
jphechter: I started to pull this out and use same `preparePage` that's being exported by `testHelpers.tsx`. It's being used extensively which I didn't realize. I need to modify the function slightly to account for formationData and not just formationFormData. It touches enough that I think it's worth including separate from this PR.

api/src/api/userRouter.test.ts
```diff
@@ -343,6 +345,186 @@ describe("userRouter", () => {
...
+            })
+          );
+
+          const result = await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
+          expect(stubTimeStampBusinessSearch.search).toHaveBeenCalled();
+          expect(result.body.formationData.businessNameAvailability.status).toEqual("UNAVAILABLE");
+          expect(result.body.profileData.needsNexusDbaName).toEqual(true);
+        });
+
+        it("only updates businessNameAvailability if needsNexusDbaName is false", async () => {
```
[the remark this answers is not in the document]
aloverso: it's not my **favorite** because of the desire to make everything a sentence like: "it [x]" buuuut we can let it slide

web/src/lib/UpdateQueue.ts
```diff
@@ -48,11 +48,18 @@ export class UpdateQueueFactory implements UpdateQueue {
     return this;
   }
 
-  queueFormationData(formationData: Partial<FormationData>): UpdateQueue {
+  queueFormationData(
+    formationData: Partial<FormationData>,
+    formationFormData?: Partial<FormationFormData>
```
aloverso, line 53: I'm thinking I would rather write  function `queueFormationFormData` than add a second parameter to this one. 

Here's my reasons why:
This breaks the pattern of these queue functions, which so far are very consistent. they are named "queueType" and they take a parameter that is of type Partial<Type> and then they spread everything else in userdata while updating Type with the params passed in. It's a little weird to expect people to break with this pattern and have something take two parameters. It also opens us up to a whole world of muddying up these functions to take arbitrary variables. By this model, we might as well have any function take any number of sub-objects and then we'd wind up with the unreadableness of the original `update` function. I'd much rather keep them very explicit and declarative: it queues ONE subtype object and inserts it into the userData.

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,42 +48,128 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+        };
+      });
+    }
+  };
+
+  const handleDbaBusinessNameAvailability = async (
+    nameAvailability: NameAvailability,
+    submittedName: string
+  ): Promise<void> => {
+    if (updateQueue && nameAvailability.status === "AVAILABLE") {
```
aloverso, line 113: Is this right? We're only updating/saving the name availablity if available? Shouldn't we save it if it comes back unavailable also?

I'm a little confused by the "simplification"/breakdown of these two `handle` functions - these weren't here before, right? unfortunately I can't see the old version

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,42 +48,128 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
-  const onSubmit = props.onSubmit;
+
+  const handleBusinessNameAvailability = async (
+    nameAvailability: NameAvailability,
+    submittedName: string
+  ): Promise<void> => {
+    const validName = ["AVAILABLE", "UNAVAILABLE"].includes(nameAvailability.status ?? "");
+    const nameUnavailable = nameAvailability.status === "UNAVAILABLE";
+
+    if (updateQueue && validName) {
```
aloverso, line 83: We prefer guard clauses over if statements. The reason is to avoid having the "meat" of the internal logic of a function indented inwards... it should be as close to the top-level indentation as possible.

The difference between
```
const doThing = () => {
  if (thing) {
     doWhatever()
  }
}
```


```
const doThing = () => {
  if (!thing) return;
  doWhatever()
}
```

The reasoning here is that if you're skimming the doThing method and want to know what it does, then you can skim the top level and not have to search inwards to find that it calls doWhatever

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
aloverso: also, same note about guard clause for update queue

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,42 +48,128 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
       { isInitialSubmit }: { isInitialSubmit: boolean }
     ): Promise<void> => {
       searchBusinessName(event)
-        .then(({ nameAvailability, submittedName }) => {
-          setNameAvailability(nameAvailability);
-          setSubmittedName(submittedName);
-          if (onSubmit) {
-            onSubmit(submittedName, nameAvailability, isInitialSubmit);
+        .then(async ({ nameAvailability, submittedName }) => {
+          if (!nameAvailability || !updateQueue || isInitialSubmit) return;
```
aloverso, line 133: updateQueue isn't used here

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
```diff
@@ -51,42 +48,128 @@ export const SearchBusinessNameForm = (props: Props): ReactElement => {
...
+    [searchBusinessName]
   );
 
+  const resetNameAvailability = (): void => {
+    resetSearch();
+    if (updateQueue) {
+      updateQueue
+        .queueFormationData({
+          formationFormData: {
+            ...updateQueue.current().formationData.formationFormData,
```
aloverso, line 153: same point as before about spreading formData being unnecssary

web/src/lib/UpdateQueue.ts
jphechter: That's a fair point. I'll make the change.

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: Fair point. Restructured.

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: That's old

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: removed

web/src/components/tasks/search-business-name/SearchBusinessNameForm.tsx
jphechter: They were not separated before. It was all one thing. I was initially keeping it all together to make it easier for myself while I was considering the next part of the refactor. I just opted to break it out now.

The current behavior is that we don't want to save unavailable DBA name searches, and for biz name we don't want to save anything that isn't unavailable or available.

PR #5914, opened 2023-06-14
feat: [#184858078] Nav Bar Updates Ethan Mobile

adidner:
## Description

We wanted Ethans to have mobile nav bar with hamburger on the right side we corresponding animations (taken care of in a different commit already)

But now also truncated text with very long business names 

Also the business name being displayed on the mobile nav bar when not on the dashboard page. 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/184858078)

### Steps to Test

Register as a user, make a really long business name in profile, see that it truncates on desktop and mobile

See that on mobile, the hamburger opens from the right side and the hamburger starts on the right side

Also see that when on non-dashboard pages on mobile the business name is displayed on the left side of the nav menu opposite the hamburger menu

## Code author checklist


web/src/components/navbar/NavBar.test.tsx
```diff
@@ -94,6 +94,30 @@ describe("<NavBar />", () => {
...
+    setLargeScreen(false);
+    render(
+      withAuth(<NavBar landingPage={false} task={undefined} showSidebar={false} />, {
+        isAuthenticated: IsAuthenticated.TRUE,
+      })
+    );
+
+    expect(screen.queryByText(businessName)).not.toBeInTheDocument();
+  });
+
```
SirSamuelJoseph, line 120: Is it possible to add a test for the truncation behavior?

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -94,6 +94,30 @@ describe("<NavBar />", () => {
...
+    render(
+      withAuth(<NavBar landingPage={false} task={undefined} showSidebar={true} />, {
+        isAuthenticated: IsAuthenticated.TRUE,
+      })
+    );
+
+    expect(screen.getByText(businessName)).toBeInTheDocument();
+  });
+
+  it("does not show business name when showSidebar is false", () => {
```
FarazA22, line 109: [Pebble] Since are setting the media query to false and rendering a mobile view, we should update the test name so it clearly states this is in mobile

web/src/components/navbar/NavBar.test.tsx
adidner: Not meaningfully, I can check for the class or the style attributes but the way that jest renders tests its a css visual change not an underlying element change. Does that make sense?

SirSamuelJoseph: Yeah that makes sense

PR #5919, opened 2023-06-16
Dev design

FarazA22:
## Description

Re-factor the unstyled button, so that it's in line with the generic button that we use in the app. This included removing variables we don't use in the unstyled button in changing the prop names.

### Ticket

[185168345](https://www.pivotaltracker.com/story/show/185168345)

## Code author checklist

### DIFF
--- a/web/src/components/Header.tsx
+++ b/web/src/components/Header.tsx
@@ -49,12 +49,12 @@ export const Header = (): ReactElement => {
     <div className="bg-white margin-bottom-4">
       <h1 className="margin-top-0 break-word">{getHeader()}</h1>
       <UnStyledButton
-        style="tertiary"
-        underline
-        textBold
+        style="default"
+        isUnderline
+        isTextBold
         onClick={editOnClick}
         dataTestid="header-link-to-profile"
-        ariaLabel="Link To Business Profile"
+        isAriaLabelApplied="Link To Business Profile"
       >
         {getButtonText()}
       </UnStyledButton>
--- a/web/src/components/filings-calendar/FilingsCalendar.tsx
+++ b/web/src/components/filings-calendar/FilingsCalendar.tsx
@@ -17,7 +17,7 @@ import {
   LookupLegalStructureById,
   LookupOperatingPhaseById,
   UserData,
-} from "@businessnjgovnavigator/shared/index";
+} from "@businessnjgovnavigator/shared/";
 import { useMediaQuery } from "@mui/material";
 import { ReactElement, useState } from "react";
 
@@ -101,25 +101,24 @@ export const FilingsCalendar = (props: Props): ReactElement => {
     };
 
     if (displayToggleButton) {
-      return userData?.preferences.isCalendarFullView ? (
+      return (
         <UnStyledButton
-          style="light"
-          noRightMargin
-          className="font-body-2xs padding-y-05 padding-x-1"
+          style={"transparentBgColor"}
+          isRightMarginRemoved
+          className="font-body-2xs padding-x-1 text-normal text-base border-1px border-base-light hide-unhide-button padding-y-11px"
           onClick={handleCalendarOnClick}
         >
-          <Icon className="usa-icon--size-3 margin-right-05">list</Icon>
-          {Config.dashboardDefaults.calendarListViewButton}
-        </UnStyledButton>
-      ) : (
-        <UnStyledButton
-          style="light"
-          noRightMargin
-          className="font-body-2xs padding-y-05 padding-x-1"
-          onClick={handleCalendarOnClick}
-        >
-          <Icon className="usa-icon--size-3 margin-right-05">grid_view</Icon>
-          {Config.dashboardDefaults.calendarGridViewButton}
+          {userData?.preferences.isCalendarFullView ? (
+            <>
+              <Icon className="usa-icon--size-3 margin-right-05">list</Icon>
+              {Config.dashboardDefaults.calendarListViewButton}
+            </>
+          ) : (
+            <>
+              <Icon className="usa-icon--size-3 margin-right-05">grid_view</Icon>
+              {Config.dashboardDefaults.calendarGridViewButton}
+            </>
+          )}
         </UnStyledButton>
       );
     }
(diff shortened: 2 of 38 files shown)


web/src/components/filings-calendar/FilingsCalendar.tsx
```diff
@@ -103,19 +103,19 @@ export const FilingsCalendar = (props: Props): ReactElement => {
...
         </UnStyledButton>
       ) : (
         <UnStyledButton
-          style="light"
-          noRightMargin
-          className="font-body-2xs padding-y-05 padding-x-1"
+          style={"transparentBgColor"}
+          isRightMarginRemoved
+          className="font-body-2xs padding-x-1 text-normal text-base border-1px border-base-light hide-unhide-button padding-y-11px"
           onClick={handleCalendarOnClick}
```
adidner, line 119: [Dust] might make sense to consolidate these 2 that are the same into a variable in this file (I know its not he point of the story though hence dust)

web/src/components/njwds-extended/ButtonDropdown.tsx
```diff
@@ -56,7 +56,7 @@ export const ButtonDropdown = (props: Props): ReactElement => {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const DefaultButton = (props: { children: ReactNode; [key: string]: any }): ReactElement => (
-    <UnStyledButton style="tertiary" textBold smallText {...props} />
+    <UnStyledButton style="default" isTextBold isSmallerText {...props} />
```
adidner, line 59: [Dust] I think hasSmallerText would be more gramatically apealing to me

web/src/components/njwds-extended/HelpButton.tsx
```diff
@@ -9,11 +9,11 @@ export const HelpButton = (): ReactElement => {
...
       <UnStyledButton
-        style={"unstyled"}
+        style={"transparentBgColor"}
         className={"text-accent-cool-darker fjc padding-0"}
-        textBold={true}
-        intercomButton
-        noRightMargin={true}
+        isTextBold
+        isIntercomEnabled
+        isRightMarginRemoved={true}
```
adidner, line 16: [Dust] A bit inconsistent here to say `isRightMarginRemoved={true}` instead of just `isRightMarginRemoved` but I imagine you did this because its more readable?

FarazA22: Either scenario would work here, I noticed that my IDE automatically creates the curly braces so I don't go to the process of deleting them sometimes. I'll look into a setting to disable this on my IDE. I think we using either one interchangeably, the fact that you justify in the prop implies that it is true.

web/src/components/njwds-extended/UnStyledButton.tsx
```diff
@@ -1,87 +1,46 @@
...
-      widthAutoOnMobile,
-      noRightMargin,
-      underline,
-      smallText,
-      textBold,
-      intercomButton,
-      showDisabledClass,
-      fullWidth,
-    ]
+    const className = [style, props.className, noRightMargin, underline, smallText, textBold, intercomButton]
```
adidner, line 43: [Pebble] I think (but am not sure) that in this current version right now, the props based booleans will override the props.className and the style attributes. 

For example if I have className={"font-size-10"} and I also pass isSmallText={true} I think in the current version the isSmallText will override the class but I think in general the className should override the booleans.

Easy resolve by moving the className to the bottom of the list because last in wins (probably should double check that). Thoughts?

web/src/components/njwds-extended/ButtonDropdown.tsx
FarazA22: We previously ran into instances where the developers had a hard time figuring out what properties were available, so this "is-" prefix was added. It allows us to quickly discern what options are available using some of the tools that are built into VS code.

web/src/components/Header.tsx
```diff
@@ -49,12 +49,12 @@ export const Header = (): ReactElement => {
...
-        style="tertiary"
-        underline
-        textBold
+        style="default"
+        isUnderline
+        isTextBold
         onClick={editOnClick}
         dataTestid="header-link-to-profile"
-        ariaLabel="Link To Business Profile"
+        isAriaLabelApplied="Link To Business Profile"
```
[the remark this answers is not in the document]
FarazA22: We previously ran into instances where the developers had a hard time figuring out what properties were available, so this "is-" prefix was added. It allows us to quickly discern what options are available using some of the tools that are built into VS code.

web/src/components/njwds-extended/HelpButton.tsx
FarazA22: Added as a techtro discussion item

web/src/components/Header.tsx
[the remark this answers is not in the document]
jphechter: I know this is already merged, but I agree with Aaron. I would advocate for this remaining `ariaLabel`.

web/src/components/njwds-extended/UnStyledButton.tsx
```diff
@@ -1,87 +1,46 @@
-import { CircularProgress } from "@mui/material";
 import React, { forwardRef, ReactElement, useEffect, useRef, useState } from "react";
+
 interface Props {
-  style: "tertiary" | "light" | "narrow-light" | "unstyled";
+  style: "default" | "transparentBgColor";
```
jphechter, line 4: I'm seeing a number of instances where we explicitly set `style="default"`. Could `default` be applied by default? That makes a little more sense given the naming.

web/src/components/njwds-extended/UnStyledButton.tsx
```diff
@@ -1,87 +1,46 @@
...
-  className?: string;
-  align?: "start" | "end" | "center";
-  ariaLabel?: string;
-  fullWidth?: boolean;
+  isRightMarginRemoved?: boolean;
+  isUnderline?: boolean;
+  isSmallerText?: boolean;
+  isTextBold?: boolean;
+  isIntercomEnabled?: boolean;
+  isAriaLabelApplied?: string;
```
jphechter, line 14: This is the `ariaLabel`. It should be named `ariaLabel`.

PR #5925, opened 2023-06-20
feat: [#185214542] blue profile page alert should be within the busin…

adidner:
…ess info header

## Description

We wanted to have the blue profile header but shown in the Business Info header rather than above

before: 
image

after: 
image

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185214542)

## Code author checklist

### DIFF
--- a/web/src/pages/profile.tsx
+++ b/web/src/pages/profile.tsx
@@ -263,12 +263,9 @@ const ProfilePage = (props: Props): ReactElement => {
     return LookupLegalStructureById(userData.profileData.legalStructureId).hasTradeName;
   };
 
-  const displayOpportunityAlert = (): boolean => {
-    return (
-      profileTab === "info" &&
-      LookupOperatingPhaseById(profileData.operatingPhase).displayProfileOpportunityAlert
-    );
-  };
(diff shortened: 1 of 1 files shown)


web/src/pages/profile.tsx
```diff
@@ -264,10 +264,7 @@ const ProfilePage = (props: Props): ReactElement => {
   };
 
   const displayOpportunityAlert = (): boolean => {
-    return (
-      profileTab === "info" &&
-      LookupOperatingPhaseById(profileData.operatingPhase).displayProfileOpportunityAlert
-    );
+    return LookupOperatingPhaseById(profileData.operatingPhase).displayProfileOpportunityAlert;
   };
```
adidner, line 268: Left this as a function because I think its a bit cleaner to look at

adidner: Could be in-lined though if people want

cbrunefearless (approved):
lgtm

web/src/pages/profile.tsx
adidner: Make variable instead of function per mobbing

PR #5935, opened 2023-06-22
feat: [#185461947] minor styling QA for pre ethan nav

adidner:
## Description

Minor QA stuff for pre ethan nav re-design

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185461947)

## Code author checklist

### DIFF
--- a/web/src/styles/sections/nav-bar.scss
+++ b/web/src/styles/sections/nav-bar.scss
@@ -28,12 +28,31 @@ $icon-offset: 45px;
   z-index: 5;
 }
 
+.Mui-focusVisible {
+  &.nav-menu-item-container {
+    background-color: $transparent;
+  }
+
+  &.nav-menu-item-container.bg-primary-extra-light-mui-selected-override {
+    background-color: $primary-extra-light;
+  }
+
+  &.profile-menu-item {
+    background-color: #fbfdf9;
+  }
+}
+
 .nav-menu-item-container {
   min-height: 47px;
 }
 
-.profile-default-background-color {
+.profile-menu-item {
   background-color: #fbfdf9;
+
+  &:hover {
+    background-color: $primary-extra-light;
+    font-weight: 700;
+  }
 }
 
 .Mui-selected.bg-primary-extra-light-mui-selected-override {
--- a/web/src/components/navbar/NavBarPopupMenu.tsx
+++ b/web/src/components/navbar/NavBarPopupMenu.tsx
@@ -122,7 +122,7 @@ export const NavBarPopupMenu = (props: Props): ReactElement => {
         icon: <ButtonIcon svgFilename="business-green" sizePx="35px" />,
         itemText: navBarBusinessTitle,
         key: "dashboardMenuItem",
-        className: "profile-default-background-color",
+        className: "profile-menu-item",
       }),
 
       NavMenuItem({
@@ -134,7 +134,7 @@ export const NavBarPopupMenu = (props: Props): ReactElement => {
         itemText: Config.navigationDefaults.profileLinkText,
         key: "profileMenuItem",
         dataTestid: "profile-link",
-        className: "profile-default-background-color",
+        className: "profile-menu-item",
       }),
 
       <hr className="margin-0 hr-2px" key={"profile-break-2"} />,


web/src/styles/sections/nav-bar.scss
```diff
@@ -28,14 +28,31 @@ $icon-offset: 45px;
...
-.profile-default-background-color {
+.profile-non-selected {
   background-color: #fbfdf9;
 }
 
+.profile-non-selected:hover {
+  background-color: $primary-extra-light;
+  font-weight: 700;
+}
+
```
jphechter, line 55: With Sass you can nest related pseudo selectors
```suggestion
.profile-non-selected {
  background-color: #fbfdf9;
  &:hover {
    background-color: $primary-extra-light;
    font-weight: 700;
  }
}
```

web/src/styles/sections/nav-bar.scss
```diff
@@ -28,14 +28,31 @@ $icon-offset: 45px;
...
+.Mui-focusVisible.nav-menu-item-container.bg-primary-extra-light-mui-selected-override {
+  background-color: $primary-extra-light;
+}
+
+.Mui-focusVisible.profile-non-selected {
+  background-color: #fbfdf9;
+}
+
```
jphechter, line 42: Same thing here. All these classes can be nested.
```
.Mui-focusVisible{
  .nav-menu-item-container {
    background-color: $transparent;
    bg-primary-extra-light-mui-selected-override {
      background-color: $primary-extra-light;
    }
  }
 ...
}
```

web/src/styles/sections/nav-bar.scss
```diff
@@ -28,14 +28,31 @@ $icon-offset: 45px;
...
 .nav-menu-item-container {
   min-height: 47px;
 }
 
-.profile-default-background-color {
+.profile-non-selected {
```
jphechter, line 47: I don't love the naming convention `profile-non-selected`. Is there a better name for this than just saying "non-selected"?

adidner: So only the profile section (dashboard and profile buttons) has this slight green background other sections like the login / logout buttons / register / myNJ buttons don't so I like how profile reffers to a specific section at a glance

web/src/styles/sections/nav-bar.scss
adidner: Ya, will fix, forgot we were using scss

web/src/styles/sections/nav-bar.scss
jphechter: It's not the `profile` piece I'm concerned about. `non-selected` feels non-specific and in my opinion breaks specific naming conventions we're using elsewhere.

jphechter: I would expect to see something like `profile-button` and then `profile-button-selected`. This just feels a little backwards to me.

adidner: Ah, gotcha I misread it. 

I had it originally as `profile-default-background-color` would that be prefferable?

jphechter: I like that more. I'd probably argue for something like `profile-menu-item`. I don't really like the idea of using a CSS property as a class name. It just feels like a code smell to me.

adidner: Sure, I was imagining it might be clearer given that most of our classes are USWDS so they are descriptive of what they do and its kinda nice to not have to go into an scss file in some cases

web/src/styles/sections/nav-bar.scss
```diff
@@ -28,12 +28,31 @@ $icon-offset: 45px;
...
+  }
+
+  &.nav-menu-item-container.bg-primary-extra-light-mui-selected-override {
+    background-color: $primary-extra-light;
+  }
+
+  &.profile-menu-item {
+    background-color: #fbfdf9;
+  }
+}
```
jphechter, line 43: This is a total nit, but if you wanted you could take this a little further.
```suggestion
.Mui-focusVisible {
  &.nav-menu-item-container {
    background-color: $transparent;
    
    &.bg-primary-extra-light-mui-selected-override {
        background-color: $primary-extra-light;
    }
  }

  &.profile-menu-item {
    background-color: #fbfdf9;
  }
}
```

jphechter (approved):

PR #5939, opened 2023-06-23
Create script to get user locations

jphechter:
## Description

Create a script that can be used to populate user location information in BFS Users Table

### Ticket

[185396010](https://www.pivotaltracker.com/story/show/185396010)

### Approach

The script added in this PR scans DynamoDB and prints all users' email and provided location to a CSV file. This is sensitive information, so the output file is included in `.gitignore`.

Running this script requires AWS session keys which can be acquired through the AWS login console. Additionally, you must provide the name of the table you want to query.

I was running into issues with `ProvisionedThroughputExceededException`. Tried a couple different things, but the best solution I was able to come up with short of modifying our DynamoDB instance was to just add a pause in the script. Every 10000 records, the script pauses for 5 seconds before automatically continuing.

## Code author checklist

### DIFF
--- a/scripts/get-user-location.sh
+++ b/scripts/get-user-location.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+if [[ -z $AWS_ACCESS_KEY_ID ]] || [[ -z $AWS_SECRET_ACCESS_KEY ]] || [[ -z $AWS_SESSION_TOKEN ]]; then
+    echo "AWS session variables not set. Must set CLI short term credentials before proceeding."
+    exit 1
+fi
+
+aws_command="aws dynamodb scan"
+output_file="get-user-location-output.csv" # This filename is included in .gitignore to prevent accidental commits
+max_items=1000
+last_evaluated_key=""
+counter=0
+
+read -p "Enter the AWS region: " region
+read -p "Enter the table name: " table_name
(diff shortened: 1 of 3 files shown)


scripts/get-user-location.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+aws_command="aws dynamodb scan"
+output_file="get-user-location-output.csv" # This filename is included in .gitignore to prevent accidental commits
+max_items=1000
+last_evaluated_key=""
+counter=0
+
+read -p "Enter the AWS region: " region
+read -p "Enter the table name: " table_name
```
jphechter, line 15: This intentionally obfuscated.

adidner, line 15: [dust] Might be cool to have environment variables in our .env that this draws from by default? That way we don't type to type if the usual suspect here is almost always the same 

something like "hit enter to default to your env vars or type in the AWS region"

adidner (approved):
Looks good to me, (although I am no expert in AWS or shell scripting)

scripts/get-user-location.sh
jphechter: Good thought. These are in the `web/.env` file. I'll see about pulling from there.

PR #5941, opened 2023-06-26
refactor: [#184882700] replace userData calls in profile and related files

SirSamuelJoseph:
## Description

Continues the Ethan refactor work to stop directing fetching userData from the updateQueue

### Ticket

[184882700](https://www.pivotaltracker.com/story/show/)

### Approach

As part of the refactor to allow for more than one business for users, we need to stop relying on the updateQueue's definition of userData and rely on the explicit userData object from the hook. This moves almost all remaining references to updateQueue.current() to use the hook

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-structure/BusinessStructureTask.tsx
+++ b/web/src/components/tasks/business-structure/BusinessStructureTask.tsx
@@ -45,11 +45,10 @@ export const BusinessStructureTask = (props: Props): ReactElement => {
   } = useFormContextHelper(createProfileFieldErrorMap());
 
   useEffect(() => {
-    const current = userDataFromHook.userData;
     return () => {
-      if (current?.profileData.legalStructureId) {
+      if (userData?.profileData.legalStructureId) {
         queueUpdateTaskProgress(props.task.id, "COMPLETED");
-      } else if (!current?.profileData.legalStructureId) {
+      } else if (!userData?.profileData.legalStructureId) {
         queueUpdateTaskProgress(props.task.id, "NOT_STARTED");
       }
     };
@@ -99,7 +98,15 @@ export const BusinessStructureTask = (props: Props): ReactElement => {
     setShowRadioQuestion(true);
     await updateQueue.update();
 
-    setProfileData(updateQueue.current().profileData);
+    const updatedProfileState: ProfileData = {
+      ...profileData,
+      legalStructureId: undefined,
+      operatingPhase:
+        profileData.operatingPhase === "GUEST_MODE_WITH_BUSINESS_STRUCTURE"
+          ? "GUEST_MODE"
+          : profileData.operatingPhase,
+    };
+    setProfileData(updatedProfileState);
   };
 
   const preLookupContent = props.task.contentMd.split("${businessStructureSelectionComponent}")[0];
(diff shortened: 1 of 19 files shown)


web/src/pages/profile.tsx
```diff
@@ -186,30 +186,31 @@ const ProfilePage = (props: Props): ReactElement => {
...
 
       (async (): Promise<void> => {
-        updateQueue.queue(await postGetAnnualFilings(updateQueue.current()));
-        updateQueue.update().then(async () => {
-          setIsLoading(false);
-          setAlert("SUCCESS");
-          await redirect({ success: true });
-        });
+        updateQueue
+          .queue(await postGetAnnualFilings(updateQueue.current()))
```
SirSamuelJoseph, line 207: this instance of updateQueue.current() is different because it relies on updates that may have happened earlier in the function. This will be addressed in the next PR with BusinessFormation where there are similar issues

web/src/pages/tasks/[taskUrlSlug].tsx
```diff
@@ -90,7 +90,7 @@ const TaskPage = (props: Props): ReactElement => {
     }
 
     const hideNextUrlSlug =
-      props.task.id === businessStructureTaskId && !hasCompletedBusinessStructure(userData as UserData);
```
Comment (jphechter) [inline web/src/pages/tasks/[taskUrlSlug].tsx:93]: get that `as` outta here!

web/src/lib/auth/signinHelper.ts
```diff
@@ -11,7 +11,7 @@ import {
   setRegistrationDimension,
   setUserId,
 } from "@/lib/utils/analytics-helpers";
-import { createEmptyUser } from "@businessnjgovnavigator/shared/";
+import { createEmptyUser, UserData } from "@businessnjgovnavigator/shared/";
 import { Dispatch } from "react";
 import { UpdateQueue } from "../types/types";
 import { AuthAction } from "./AuthContext";
```
jphechter, line 17: I know you didn't do this, but can we swap relative imports?
```suggestion
import { UpdateQueue } from "@/lib/types/types";
import { AuthAction } from "@/lib/auth/AuthContext";
import * as session from "@/lib/auth/sessionHelper";
import { triggerSignOut } from "@/lib/auth/sessionHelper";
```

web/src/components/tasks/business-structure/BusinessStructureTask.tsx
```diff
@@ -45,7 +45,7 @@ export const BusinessStructureTask = (props: Props): ReactElement => {
   } = useFormContextHelper(createProfileFieldErrorMap());
 
   useEffect(() => {
-    const current = updateQueue?.current();
+    const current = userDataFromHook.userData;
```
jphechter, line 48: Why can't we just use `userData` for the instances of this variable?

web/src/pages/onboarding.tsx
```diff
@@ -161,7 +161,7 @@ const OnboardingPage = (props: Props): ReactElement => {
         return;
       }
 
-      let currentUserData = updateQueue?.current();
+      let currentUserData = userData;
```
jphechter, line 164: It looks like `currentUserData` was created to be more friendly than `updateQueue?.current()`. Can we just use `userData` in all of these locations?

web/src/components/tasks/business-structure/BusinessStructureTask.tsx
SirSamuelJoseph: We can for this one! Updating

web/src/pages/onboarding.tsx
SirSamuelJoseph: We actually can't, because this is onboarding we might need to create user data for new users (see line 170)

PR #5944, opened 2023-06-26
chore: [#185484678] consolidate fetchMunicipalityByName in tax filing

somabadri:
## Description

Consolidated updating municipality logic to the tax filing interface on the backend instead of frontend to reduce duplication.

### Ticket

[#185484678](https://www.pivotaltracker.com/story/show/185484678)

## Code author checklist


api/src/domain/tax-filings/taxFilingsInterfaceFactory.ts
```diff
@@ -34,15 +35,22 @@ export const taxFilingsInterfaceFactory = (apiTaxFilingClient: TaxFilingClient):
...
+    if (taxCity) {
+      const record = await fetchMunicipalityByName(taxCity);
+      profileDataToReturn = {
+        ...profileDataToReturn,
         municipality: {
-          name: taxCity,
-          county: "",
-          id: "",
-          displayName: "",
+          name: record?.townName || profileDataToReturn.municipality?.name || "",
```
aloverso, line 50: [pebble] I'm thinking this logic isn't quite correct. If `record` fails to return and is undefined, and the user currently has an undefined municipality, then it will cascade down to saving it as `""`. It seems antithetical to our patterns to have a municipality that with all-empty fields. I think that if record is undefined then it should just stay set to what the municipality currently is - which means that if the municipality is currently undefined, it should stay undefined instead of becoming defined but with empty strings.

I think there are a lot of cases in our app where we check if a user has an industry by checking if municipality is undefined. If we set it defined, then those logics would break

api/src/domain/user/fetchMunicipalityByName.test.ts
```diff
@@ -0,0 +1,23 @@
+import { fetchMunicipalityByName } from "./fetchMunicipalityByName";
+
+describe("fetchMunicipalityByName", () => {
+  it("returns a record when a valid name is supplied", async () => {
+    const record = await fetchMunicipalityByName("TRENTON");
+    expect(record.townName).toEqual("Trenton");
+  });
+
+  it("returns a empty object when an invalid name is supplied", async () => {
```
aloverso, line 9: [sand] On the same note as before - I know this is existing behavior but if these don't match, do we really want to save a user's municipality as empty strings? Should it return undefined instead?

api/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.paths.json",
   "compilerOptions": {
+    "module": "commonjs",
```
aloverso, line 4: [dust] Why this change?

shared/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.paths.json",
   "compilerOptions": {
+    "module": "esnext",
```
aloverso, line 4: [dust] Why this change?

web/src/components/filings-calendar/tax-access-modal/TaxAccessStepTwo.test.tsx
```diff
@@ -601,7 +576,7 @@ describe("<TaxAccessStepTwo />", () => {
       });
     });
 
-    it("marks the naics code task as complete on successful response", async () => {
+    it("updated naics code and marks the naics code task as complete on successful response", async () => {
```
aloverso, line 579: [pebble] `updates`

web/src/components/filings-calendar/tax-access-modal/TaxAccessStepTwo.test.tsx
```diff
@@ -629,6 +598,38 @@ describe("<TaxAccessStepTwo />", () => {
         return expect(currentUserData().taskProgress["determine-naics-code"]).toEqual("COMPLETED");
       });
     });
+
+    it("updates municipality on successful response", async () => {
```
aloverso, line 602: [pebble] this reads as though `it` (the component) updates municipality... could maybe be renamed to say that it updates with municipality **from the API response** or something like that

PR #5952, opened 2023-06-27
feat: [#185141979] export roadmap to pdf

FarazA22:
## Description

We created a new export page that allows users to preview their roadmap and all of the detail task pages of the roadmap, we are specifically rendering the standard task component and not non-standard task components on this page. The user is able to click a download button that will create a PDF file of all of the information that is on the page.

### Ticket

https://www.pivotaltracker.com/story/show/185141979

### Approach

We installed two yarn modules, jsPDF and html2canvas. The former compiled the PDF and the latter took screenshots of each section we wanted to create a PDF page for.

### Steps to Test

To test the functionality, you need to complete onboarding, and once you are on the dashboard page, you can manually navigate to the export page in the address bar. The download button will be available, you can then click the download button to generate the PDF.

## Code author checklist


web/src/lib/roadmap/exportRoadmapAsPdf.ts
```diff
@@ -0,0 +1,84 @@
...
+    images.push(result);
+  }
+  clone.remove();
+  return images;
+};
+
+export const exportComponentsAsPDF = async (
+  componentIds: string[],
+  prePdfProcessingCb: () => void,
+  postPdfProcessingCb: () => void
```
aloverso, line 43: [pebble] two notes
- I think `Cb` means "callback" here, I don't think that's a common abbreviation (took me a min) and might be clearer to just write out
- With so many arguments to this function, especially with the same type signature, I think this would be clearer to use if we did named parameters instead

web/src/pages/export.tsx
```diff
@@ -0,0 +1,60 @@
...
+                exportComponentsAsPDF(
+                  roadmap ? ["roadmap", ...roadmap.tasks.map((task) => task.id)] : [],
+                  () => setDownloadIndicator(true),
+                  () => setDownloadIndicator(false)
+                );
+              }}
+              dataTestId="downloadPdf"
+              isLargeButton
+            >
+              Download PDF
```
aloverso, line 39: [pebble/dust] should be Config?

web/src/pages/export.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Roadmap } from "@/components/dashboard/Roadmap";
+import { PrimaryButton } from "@/components/njwds-extended/PrimaryButton";
+import { TaskBody } from "@/components/TaskBody";
+import { useRoadmap } from "@/lib/data-hooks/useRoadmap";
+import { ROUTES } from "@/lib/domain-logic/routes";
+import { exportComponentsAsPDF } from "@/lib/roadmap/exportRoadmapAsPdf";
+import { useRouter } from "next/router";
+import { ReactElement, useEffect, useState } from "react";
+
+const ExportPage = (): ReactElement => {
```
aloverso, line 11: is this component being used / linked to anywhere?

web/src/pages/export.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const router = useRouter();
+  const FEATURE_EXPORT_PDF = process.env.FEATURE_EXPORT_PDF === "true";
+
+  useEffect(() => {
+    if (!FEATURE_EXPORT_PDF) router.push(ROUTES.dashboard);
+  }, [router, FEATURE_EXPORT_PDF]);
+
+  return (
+    <>
+      {FEATURE_EXPORT_PDF && roadmap ? (
```
aloverso, line 23: [sand] two notes:
- It seems strange that if we have this feature disabled, it will just have a forever loading indicator instead of just a 404 or text on the page that says "this is not enabled" or similar. I feel like loading forever will make it seem broken when it's not.
- In general, we prefer these to be a guard clause before the return statement rather than a ternary in the JSX. It makes it much much easier to read and understand what's being rendered when the body of the component isn't nested inside a ternary.

web/src/pages/export.tsx
FarazA22: This page is not linked anywhere in the application. Dillon mentioned that this would likely be under a future flag for the next several months, and will go through several iterations before we figure out the final state of the PDF. We decided to render the functionality on a separate page, as opposed to adding it to an existing page, such as the profile page.

web/src/pages/export.tsx
FarazA22:
```
useEffect(() => {
    if (!FEATURE_EXPORT_PDF) router.push(ROUTES.dashboard);
  }, [router, FEATURE_EXPORT_PDF]);

if (!FEATURE_EXPORT_PDF || !roadmap) return <CircularIndicator />;
```

I've added the suggested guard clause. I have a question about a situation where a user is on this page in an environment where the feature is disabled. They will see the loading indicator until the use effect hook routes them to the dashboard page. Do you have any comments about this behavior? Should we be routing them to the 404 page instead? @aloverso

aloverso: @FarazA22 looks good, makes sense to me. I missed the routing in the useEffect initially, sorry

PR #5961, opened 2023-06-28
chore: [#184198991] re-add zebra pattern to signers section

AnJuHyppolite:
## Description

Re-add zebra pattern to signers section and a line to the bottom of the table once the BOD and signers sections are complete.

### Ticket

[#184198991](https://www.pivotaltracker.com/story/show/184198991)

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/contacts/Addresses.tsx
+++ b/web/src/components/tasks/business-formation/contacts/Addresses.tsx
@@ -98,6 +98,8 @@ export const Addresses = <T extends FormationMember | FormationIncorporator>(
     props.setData([...[...props.addressData].slice(0, index), ...[...props.addressData].slice(index + 1)]);
   };
 
+  const zebraOnOdd = (index: number): string => (index % 2 === 1 ? "bg-base-extra-light" : "");
+
   const renderDesktopTable = (
     <table className={`addresses margin-top-2 margin-bottom-3`}>
       <thead>
@@ -121,7 +123,11 @@ export const Addresses = <T extends FormationMember | FormationIncorporator>(
           props.addressData.map((it, index) => {
             return (
               <Fragment key={index}>
-                <tr className="margin-bottom-1" key={index} data-testid={`${props.fieldName}-${index}`}>
+                <tr
+                  className={`margin-bottom-1 ${zebraOnOdd(index)}`}
+                  key={index}
+                  data-testid={`${props.fieldName}-${index}`}
+                >
                   <td className="break-word">{it.name}</td>
                   <td className="break-word">{formatAddress(it)}</td>
                   {"signature" in it ? (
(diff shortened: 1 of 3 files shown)


aloverso:
@AnJuHyppolite why change from the CSS-based pattern to doing it in JS? As I was scrolling through, I was going to leave a comment letting you know that this could be done using CSS nth-child. But it looks like we already had that, and are switching to calculating in javascript instead - how come?

web/src/components/tasks/business-formation/contacts/Addresses.tsx
```diff
@@ -121,7 +121,11 @@ export const Addresses = <T extends FormationMember | FormationIncorporator>(
           props.addressData.map((it, index) => {
             return (
               <Fragment key={index}>
-                <tr className="margin-bottom-1" key={index} data-testid={`${props.fieldName}-${index}`}>
+                <tr
+                  className={`margin-bottom-1 ${index % 2 === 1 ? "bg-base-lightest" : ""}`}
```
adidner, line 125: [sand] so it seems like we reuse this code many times `${index % 2 === 1 ? "bg-base-lightest" : ""}`

so we could probably make this a function `const zerbraOnOdd = (index) =>  index % 2 === 1 ? "bg-base-lightest" : ""`

then use it like so `className={`margin-bottom-1 ${zebraOnOdd(index)}`}`

AnJuHyppolite:
> @AnJuHyppolite why change from the CSS-based pattern to doing it in JS? As I was scrolling through, I was going to leave a comment letting you know that this could be done using CSS nth-child. But it looks like we already had that, and are switching to calculating in javascript instead - how come?

@aloverso: Because the error message is rendered as a row, the CSS-based pattern, error messages were that are even-number-indexed are being highlighted. If you have another suggestion, I'd like to try it.

adidner:
> > @AnJuHyppolite why change from the CSS-based pattern to doing it in JS? As I was scrolling through, I was going to leave a comment letting you know that this could be done using CSS nth-child. But it looks like we already had that, and are switching to calculating in javascript instead - how come?
> 
> @aloverso: Because the error message is rendered as a row, the CSS-based pattern, error messages were that are even-number-indexed are being highlighted. If you have another suggestion, I'd like to try it.

TLDR is that we're using <Table> which is having issues looking like the design and accommodating that CSS, should probably refactor to use grid or flex box which would allow us to do the CSS instead but I think this is also fine

aloverso:
> > @AnJuHyppolite why change from the CSS-based pattern to doing it in JS? As I was scrolling through, I was going to leave a comment letting you know that this could be done using CSS nth-child. But it looks like we already had that, and are switching to calculating in javascript instead - how come?
> 
> @aloverso: Because the error message is rendered as a row, the CSS-based pattern, error messages were that are even-number-indexed are being highlighted. If you have another suggestion, I'd like to try it.

@AnJuHyppolite oh I see, got it. Would it work to add a class name instead? Like, for any non-error row that should receive the zebra highlighting, you can just give it a className like `<tr className="zebra-highlighting">`. And then change the CSS to apply the nth-child styling to instances of that class, instead of instances of `tr`?

web/src/styles/base/background-colors.scss
```diff
@@ -204,3 +204,6 @@
 .bg-accent-semi-cool {
   background-color: $accent-semi-cool !important;
 }
+.bg-base-lightest {
```
FarazA22, line 207: [Pebble] Some of our colors are already available through USWDS/NJ's CSS file we import. I believe this can be removed.

PR #5962, opened 2023-06-28
fix: [#185355437] roadmap part of license tasks not rendering in the CMS

adidner:
## Description

part 1 of the changes regarding license tasks not rendering in the CMS

basically netlify doesn't allow us to have a dropdown that grabs from 2 different folders to populate its content so we ended up wanting to seperate the dropdown field into 2 fields (license-task and tasks) and update the json and logic accordingly 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185355437)

## Code author checklist


web/src/lib/roadmap/roadmapBuilder.test.ts
```diff
@@ -33,6 +33,30 @@ describe("roadmapBuilder", () => {
     ).toEqual(expect.arrayContaining(["tea-task-1-id"]));
   });
 
+  it("builds a roadmap with license add-ons only when industry is undefined", async () => {
```
FarazA22, line 36: [Dust] Consider updating test name so it states license task and not license

web/src/lib/roadmap/roadmapBuilder.test.ts
```diff
@@ -33,6 +33,30 @@ describe("roadmapBuilder", () => {
...
+      addOns: ["license-addon"],
+    });
+    expect(
+      roadmap.tasks.map((it) => {
+        return it.id;
+      })
+    ).toEqual(expect.arrayContaining(["license-task-1-id"]));
+  });
+
+  it("prioritizes a task over a license when building a roadmap with add-ons only when industry is undefined", async () => {
```
FarazA22, line 48: [Dust] Consider updating test name so it states license task and not license

web/src/lib/roadmap/roadmapBuilder.test.ts
```diff
@@ -33,6 +33,30 @@ describe("roadmapBuilder", () => {
...
+    const roadmap = await buildRoadmap({
+      industryId: undefined,
```
FarazA22, line 50: [Dust] I'm a little confused about this test. The reason for saying this is that I think a license task would be tied to the industry. If you have no industry, how could you possibly have a license task? Am I off? I might be off. I think this test is OK, I don't know if we need another.

web/public/mgmt/config.yml
```diff
@@ -1500,14 +1500,22 @@ collections:
...
-          - label: Task
+          - label: Task (this or License Task is required)
             name: task
             widget: "relation"
             collection: "tasks"
             search_fields: ["{{filename}}", "name", "urlSlug"]
             value_field: "{{filename}}"
             display_fields: ["{{filename}}"]
             options_length: 500
+          - label: License Task (this or Task is required)
```
FarazA22, line 1511: `
  name: validation
    label: Custom Validation
    widget: hidden
    required: true
    script: path/to/validation-script.js
`

[Dust] perhaps this is a larger discussion. In an instance like this, where one of two fields is required, but we are not making them both required, should we have a custom validation?

web/public/mgmt/config.yml
```diff
@@ -1500,14 +1500,22 @@ collections:
           - label: Weight
             name: weight
             widget: number
-          - label: Task
+          - label: Task (this or License Task is required)
```
FarazA22, line 1503: [Sand] I might be wrong, but I think in Netlify, all of the fields are required by default, I think you need to pass in an attribute where are you set required to false

jphechter: Based on what I'm seeing in the docs I think it does default to optional, but being explicit wouldn't hurt especially in this case.

web/public/mgmt/config.yml
adidner: Totally, we totally should, wasn't sure if that was possible in Netlify but looks like you answered that question, nice

web/src/lib/roadmap/roadmapBuilder.test.ts
adidner: So I'm not super familiar with the section of the app to be honest, and the way the code is written just for roadmapBuilder it looks like that is a possible state, but if you're saying that from a logical perspective it shuldn't happen in the app when the users are taking their flows that makes sense. 

I just defined the industry ID and moved the test down a bit.

web/public/mgmt/config.yml
adidner: I'll look into it now

web/public/mgmt/config.yml
adidner: I think you're right @FarazA22 , I will make them not required

web/public/mgmt/config.yml
adidner: So I played with this quite a bit and it seems like it should totally be possible but might be a bit of a pain. So after time boxing it I'm going to write a follow up chore for it.

adidner: https://www.pivotaltracker.com/story/show/185532198

PR #5963, opened 2023-06-28
feat: [#184822307] add nonprofit onboarding and roadmap

aloverso:
## Description

Added nonprofit 

### Ticket

[#184822307](https://www.pivotaltracker.com/story/show/#184822307)

### Approach

The additions to Business Formation with `searchOnly` prop are intended to be quick & temporary, since that will be replaced and removed in the upcoming nonprofit formation story

## Code author checklist

### DIFF
--- a/shared/src/profileData.ts
+++ b/shared/src/profileData.ts
@@ -94,6 +94,7 @@ export interface ProfileData extends IndustrySpecificData {
   readonly nexusDbaName: string;
   readonly needsNexusDbaName: boolean;
   readonly operatingPhase: OperatingPhaseId;
+  readonly isNonprofitOnboardingRadio: boolean;
 }
 
 export const emptyProfileData: ProfileData = {
@@ -122,6 +123,7 @@ export const emptyProfileData: ProfileData = {
   needsNexusDbaName: false,
   nexusLocationInNewJersey: undefined,
   operatingPhase: "GUEST_MODE",
+  isNonprofitOnboardingRadio: false,
   ...emptyIndustrySpecificData,
 };
 
--- a/web/src/components/onboarding/OnboardingNonprofit.tsx
+++ b/web/src/components/onboarding/OnboardingNonprofit.tsx
@@ -1,0 +1,73 @@
+import { ProfileDataContext } from "@/contexts/profileDataContext";
+import { profileFormContext } from "@/contexts/profileFormContext";
+import { useConfig } from "@/lib/data-hooks/useConfig";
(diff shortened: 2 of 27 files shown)


web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -121,6 +121,31 @@ describe("<BusinessFormationPaginator />", () => {
     initialUserData = generateUserData({ profileData, formationData });
   });
 
+  describe("search business name only", () => {
+    it("does not show buttons or stepper", () => {
+      preparePage(initialUserData, displayContent, undefined, undefined, undefined, undefined, true);
```
aloverso, line 126: I know this is pretty awful, but I also know that the in-progress Ethan work changes this function to named params, so I don't want to redo that work. It'll get changed as soon as that is merged

web/src/components/onboarding/OnboardingNonprofit.tsx
```diff
@@ -0,0 +1,73 @@
...
+          name={camelCaseToKebabCase(fieldName)}
+          value={state.profileData[fieldName]?.toString() ?? ""}
+          onChange={handleChange}
+          row
+        >
+          <>
+            {["true", "false"].map((val) => {
+              return (
+                <FormControlLabel
+                  key={val.toString()}
```
adidner, line 54: Isn't the val here already a string? or am I missing something?

adidner (approved):
Seems good to me,

minor question about the use of toString() on strings.

web/src/components/onboarding/OnboardingNonprofit.tsx
aloverso: you're right; my mistake from using copy-pasted code. will fix

shared/src/profileData.ts
```diff
@@ -94,6 +94,7 @@ export interface ProfileData extends IndustrySpecificData {
   readonly nexusDbaName: string;
   readonly needsNexusDbaName: boolean;
   readonly operatingPhase: OperatingPhaseId;
+  readonly isNonprofitOnboardingRadio: boolean;
```
aloverso, line 97: The reason for this name is to be overly specific that this field does NOT represent if a user `isNonprofit`

They can change their legal structure at any point later, in the task itself. If they do that, their legal structure will be `nonprofit` but this would still be set to `false` if they selected "No" during onboarding.

So this is named this way to show that this field should NOT be used to logically check if a user is a nonprofit - it ONLY represents their radio selection during onboarding

PR #5972, opened 2023-06-29
fix: [#185512025] do not show biz structure prompt for foreign remote seller/worker

aloverso:
### Ticket

[#185512025](https://www.pivotaltracker.com/story/show/#185512025)

## Code author checklist


api/src/db/migrations/v117_add_onboarding_nonprofit.ts
```diff
@@ -1,3 +1,4 @@
+import { randomInt } from "@shared/intHelpers";
```
jphechter, line 1: If this is going out as a hotfix I don't know that the `v117` migration should be in here.

PR #6000, opened 2023-07-06
feat: [#185543521] remove landing page redirect to webflow

somabadri:
## Description

Removed the code that handles the redirect to the webflow landing page.

### Ticket

[#185543521](https://www.pivotaltracker.com/story/show/185543521)

### Notes

Follow up: remove the following environment variables in Circle Ci: `FEATURE_LANDING_PAGE_REDIRECT` and  `ALTERNATE_LANDING_PAGE_URL`

## Code author checklist

### DIFF
--- a/web/src/pages/index.tsx
+++ b/web/src/pages/index.tsx
@@ -75,9 +75,6 @@ const Home = (props: Props): ReactElement => {
     };
   }
   useEffect(() => {
-    const landingPageAlternateUrl = process.env.ALTERNATE_LANDING_PAGE_URL || ROUTES.landing;
-    const alternateLandingPageEnabled = process.env.FEATURE_LANDING_PAGE_REDIRECT === "true";
-
     if (state.isAuthenticated === IsAuthenticated.TRUE) {
       if (userData?.onboardingFormProgress === "COMPLETED") {
         router.replace(ROUTES.dashboard);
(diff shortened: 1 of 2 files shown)


web/src/pages/index.tsx
```diff
@@ -75,9 +75,6 @@ const Home = (props: Props): ReactElement => {
     };
   }
   useEffect(() => {
-    const landingPageAlternateUrl = process.env.ALTERNATE_LANDING_PAGE_URL || ROUTES.landing;
```
FarazA22, line 78: [Pebble] I think we should remove the environment variables from the .env-template. I think we can also remove it from bit warden, and the environment variables that are saved in circles ci

web/src/pages/index.tsx
```diff
@@ -75,9 +75,6 @@ const Home = (props: Props): ReactElement => {
...
-    const alternateLandingPageEnabled = process.env.FEATURE_LANDING_PAGE_REDIRECT === "true";
```
FarazA22, line 79: Same comment as above

web/src/pages/index.tsx
somabadri: yep, made a note of it above in the pr comment.

web/src/pages/index.tsx
```diff
@@ -86,12 +83,6 @@ const Home = (props: Props): ReactElement => {
       } else if (userData === undefined && error !== undefined) {
```
FarazA22, line 83: [Dust] Should we remove the else if conditional here and just keep what is inside of the code block?

FarazA22, line 83: [Dust] Is the `else if` conditional necessary? Can we keep what is inside the code block?

somabadri: The tests still pass without it but I like that it is explicitly spelled out, it makes it easier to read.

web/src/pages/index.tsx
FarazA22: The changes to .env-template should be part of this PR

somabadri: They are not there by default in .env-template file.

PR #6007, opened 2023-07-07
refactor: [#185362865] revise NavBarPopupMenu to simplify left side i…

adidner:
…con or space alignment

## Description

initially wanted to transition this into using grid but I don't think grid is a particularly superior solution here. So instead i made a minor modification for readability / simplification. 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185362865)

## Code author checklist


web/src/components/navbar/NavMenuItem.tsx
```diff
@@ -30,11 +30,8 @@ export const NavMenuItem = ({
       selected={selected}
       key={key}
     >
-      {icon && <div className="icon-width">{icon}</div>}
-      <div
-        className={`${icon ? "" : "no-icon-offset"} text-wrap truncate-long-business-names-desktop`}
-        data-testid={dataTestid}
-      >
+      <div className="icon-width">{icon ?? ""}</div>
+      <div className={`text-wrap truncate-long-business-names-desktop`} data-testid={dataTestid}>
```
jphechter, line 34: Does this also need `truncate-long-business-names-mobile`?

If so, why don't we have a single class `truncate-long-business-name` that uses a media query to change the value of `max-width`?

PR #6014, opened 2023-07-10
License tasks not rendering in cms 2 #185355437

adidner:
## Description

We had issues where we want a dropdown to pull from both the license task folder and task folder. Since this is not possible in the CMS we will simply have 2 drop downs but this requires some code and test and CMS changes outlined in this story. 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185355437)

## Code author checklist

### DIFF
--- a/content/src/roadmaps/task-dependencies.json
+++ b/content/src/roadmaps/task-dependencies.json
@@ -1,8 +1,8 @@
 {
   "dependencies": [
     {
-      "name": "register-for-ein",
-      "dependencies": [
+      "task": "register-for-ein",
+      "taskDependencies": [
         "form-business-entity",
         "form-business-entity-nj",
         "form-business-entity-foreign",
@@ -10,14 +10,14 @@
       ]
     },
     {
-      "name": "form-business-entity",
-      "dependencies": [
+      "task": "form-business-entity",
+      "taskDependencies": [
         "register-for-ein"
       ]
     },
     {
-      "name": "register-for-taxes",
-      "dependencies": [
+      "task": "register-for-taxes",
+      "taskDependencies": [
         "form-business-entity",
         "form-business-entity-nj",
         "form-business-entity-foreign",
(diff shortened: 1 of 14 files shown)


web/src/lib/static/loadTasks.ts
```diff
@@ -60,7 +60,7 @@ export const loadTaskByUrlSlug = (urlSlug: string): Task => {
   }
 };
 
-const loadTaskByFileName = (fileName: string, directory: string): Task => {
+export const loadTaskByFileName = (fileName: string, directory: string): Task => {
```
adidner, line 63: So loadTaskByFileName and fetchTaskByFilename are extremely similar and should probably be consolidated into 1 function. However it seems like most of the logic in the fetchTaskByFilename is similar to logic found in loadTasks.ts as a result I think it would be best to have a follow ticket to consolidate all similar logic between the 2 rather than doing some here and some in a follow up ticket. 

https://www.pivotaltracker.com/story/show/185588269

web/public/mgmt/config.yml
```diff
@@ -1567,13 +1567,23 @@ collections:
           - label: Weight
             name: weight
             widget: number
-          - label: Task
+          - label: Task (this or License Task is required)
```
AnJuHyppolite, line 1570: I think there's a typo here.

content/src/roadmaps/task-dependencies.json
```diff
@@ -28,186 +28,188 @@
...
+      "task": "apply-scorp-state",
+      "taskDependencies": [
         "apply-scorp-federal"
       ]
     },
     {
-      "name": "apply-for-shop-license",
-      "dependencies": [
+      "licenseTask": "apply-for-shop-license",
+      "taskDependencies": [
```
AnJuHyppolite, line 82: Shouldn't the key be 'licenseTaskDependencies`?

web/src/lib/static/loadTasks.ts
```diff
@@ -69,11 +69,14 @@ const loadTaskByFileName = (fileName: string, directory: string): Task => {
 
   const taskWithoutLinks = convertTaskMd(fileContents);
   const fileNameWithoutMd = fileName.split(".md")[0];
-  const unlockedByTaskLinks = (
-    dependencies.find((dependency) => {
-      return dependency.name === fileNameWithoutMd;
-    })?.dependencies || []
-  ).map((dependencyFileName) => {
+
+  const current = dependencies.find((dependency) => {
```
somabadri, line 73: [dust] same comment about `current` from above?

content/src/roadmaps/task-dependencies.json
```diff
@@ -28,186 +28,188 @@
...
+      "licenseTask": "annual-license-cannabis",
+      "taskDependencies": [
         "zoning-cannabis",
         "priority-status-cannabis"
       ]
     },
     {
-      "name": "conditional-permit-cannabis",
-      "dependencies": [
+      "task": "conditional-permit-cannabis",
```
AnJuHyppolite, line 115: Shouldn't this key be `licenseTask`?

somabadri: + 1

content/src/roadmaps/task-dependencies.json
```diff
@@ -28,186 +28,188 @@
...
+      "task": "site-safety-permits-cosmetology",
+      "taskDependencies": [
         "sign-lease",
         "building-permit"
       ]
     },
     {
-      "name": "town-mercantile-license-liquor",
-      "dependencies": [
+      "task": "town-mercantile-license-liquor",
```
AnJuHyppolite, line 50: `licenseTask`?

content/src/roadmaps/task-dependencies.json
adidner: What soma said @AnJuHyppolite in this case a task blocks a license.

content/src/roadmaps/task-dependencies.json
adidner: Unclear? Their is a town-mercantile-license.md in both the license-task and tasks folder. I'll follow up with content to make sure they're using the right one. Good catch. 

image
image

content/src/roadmaps/task-dependencies.json
adidner: Nice, totally missed that one. Good catch guys

web/src/lib/async-content-fetchers/fetchTaskByFilename.ts
```diff
@@ -5,13 +5,16 @@ export const fetchTaskByFilename = async (filename: string): Promise<Task> => {
...
   const dependencies = await fetchDependenciesFile();
-  const unlockedByTaskLinks = await Promise.all(
-    (
-      dependencies.find((dependency) => {
-        return dependency.name === filename;
-      })?.dependencies || []
-    ).map(fetchTaskLinkByFilename)
-  );
+
+  const current = dependencies.find((dependency) => {
```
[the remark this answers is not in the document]
adidner: Totally, a bit of a lazy habit I wasn't thinking about to write it this way. You're totally right.

content/src/roadmaps/task-dependencies.json
adidner: JK, so it looks t o me like the liquor verision is part of task? 
image

adidner: Ignore my first comment, now deleted, I can't read.

AnJuHyppolite: Also, there is a `town-mercantile-license-liquor.md` in the **tasks** folder, but I think it should be in **license-tasks** folder & I am wondering why it isn't. 
Screenshot 2023-07-12 at 1 19 02 PM

adidner: So I know that Anne told me the answer to this question before

and I think it has to do with webflow syncs and some content not being ready to be moved from a task to a license task as we only support 1 of those 2 things for webflow syncs and their is like extra content that needs to be added before we can move it. 

I'm not super confident in that answer though so perhaps we should rope her into the conversation for a re-fresher @aloverso

AnJuHyppolite: Also, along the lines of `town-mercantile-license`, we should update the key to `licenseTask`.
Screenshot 2023-07-12 at 1 29 40 PM

adidner: So I believe there are 2 town-mercantile-license files, one in task and one in license task so I think its ambigious as to which the content team actually wants so I will ask.

adidner: Vera mentioned its probably Task but she will check again and modify once its live on dev

PR #6023, opened 2023-07-11
Nonprofit formation #185153256

AnJuHyppolite:
### Ticket

[#185153256](https://www.pivotaltracker.com/story/show/185153256)

## Code author checklist


.circleci/config.yml
```diff
@@ -189,6 +189,7 @@ commands:
             echo 'export FEATURE_BUSINESS_FLP=$FEATURE_BUSINESS_FLP_DEV' >> $BASH_ENV
             echo 'export FEATURE_BUSINESS_FLC=$FEATURE_BUSINESS_FLC_DEV' >> $BASH_ENV
             echo 'export FEATURE_BUSINESS_FCORP=$FEATURE_BUSINESS_FCORP_DEV' >> $BASH_ENV
+            echo 'export FEATURE_BUSINESS_FNP=$FEATURE_BUSINESS_FNP_DEV' >> $BASH_ENV
```
aloverso, line 192: I see you have this set to false in CircleCI right now for dev... that's gonna make it hard for folks to test when this gets merged ;)

api/src/client/ApiFormationClient.test.ts
```diff
@@ -1223,6 +1223,514 @@ describe("ApiFormationClient", () => {
...
+          { legalStructureId: "nonprofit" }
+        );
+
+        // const userData = generateUserData({
+        //   profileData: generateProfileData({
+        //     legalStructureId: "nonprofit",
+        //     businessPersona: "STARTING",
+        //   }),
+        //   formationData: generateFormationData({ formationFormData }),
+        // });
```
aloverso, line 1440: [pebble] remove commented code

api/src/client/ApiFormationClient.test.ts
```diff
@@ -1223,6 +1223,514 @@ describe("ApiFormationClient", () => {
...
+            agentNumberOrManual: "MANUAL_ENTRY",
+            hasNonprofitBoardMembers: true,
+            nonprofitBoardMemberQualificationsSpecified: "IN_BYLAWS",
+            nonprofitBoardMemberQualificationsTerms: "LOREM IPSUM DOLOR SIT AMET",
+            nonprofitBoardMemberRightsSpecified: "IN_BYLAWS",
+            nonprofitBoardMemberRightsTerms: "LOREM IPSUM DOLOR SIT AMET",
+            nonprofitTrusteesMethodSpecified: "IN_BYLAWS",
+            nonprofitTrusteesMethodTerms: "LOREM IPSUM DOLOR SIT AMET",
+            nonprofitAssetDistributionSpecified: "IN_BYLAWS",
+            nonprofitAssetDistributionTerms: "LOREM IPSUM DOLOR SIT AMET",
```
aloverso, line 1255: [pebble] I would remove all these LOREM IPSUMs because (if I understand correctly) when the user selects "IN_BYLAWS" then they are NOT given a text box and an opportunity to type text. So this isn't really a state that is possible?

api/src/client/ApiFormationClient.test.ts
```diff
@@ -1223,6 +1223,514 @@ describe("ApiFormationClient", () => {
...
+            agentNumberOrManual: "MANUAL_ENTRY",
+            hasNonprofitBoardMembers: true,
+            nonprofitBoardMemberQualificationsSpecified: "IN_FORM",
+            nonprofitBoardMemberQualificationsTerms: "LOREM IPSUM DOLOR SIT AMET",
+            nonprofitBoardMemberRightsSpecified: "IN_FORM",
+            nonprofitBoardMemberRightsTerms: "LOREM IPSUM DOLOR SIT AMET",
+            nonprofitTrusteesMethodSpecified: "IN_FORM",
+            nonprofitTrusteesMethodTerms: "LOREM IPSUM DOLOR SIT AMET",
+            nonprofitAssetDistributionSpecified: "IN_FORM",
+            nonprofitAssetDistributionTerms: "LOREM IPSUM DOLOR SIT AMET",
```
aloverso, line 1426: [pebble] In order for this to be a test that gives confidence, these LOREM IPSUM should all be four different strings.

If I coded it with a mistaken copy-paste error so that whatever the user types into `nonprofitBoardMemberQualificationsTerms` is what is sent to the API for as the value for all four fields `MemberTerms`, `MemberClassPermissions`, etc - this test would still pass. 

Explicitly expecting each field to be used in its respective API field would give much more confidence that this is working as expected.

api/src/client/ApiFormationClient.test.ts
```diff
@@ -1223,6 +1223,514 @@ describe("ApiFormationClient", () => {
...
+              legalStructureId: "nonprofit",
+              businessPersona: "STARTING",
+            }),
+            formationData: generateFormationData({ formationFormData }),
+          })
+        );
+        const currentBusiness = getCurrentBusiness(userData);
+        await client.form(userData, "hostname.com/form-business");
+
+        expect(mockAxios.post).toHaveBeenCalledWith(
```
aloverso, line 1618: There's a LOT going on in this test and its expectation that makes it hard to read and understand the cause-and-effect.

One easy way to make it easier to read would be to just expect the one thing we care about here:

`expect(getLastCalledWith(mockAxios.post)[1].Formation.BusinessInformation).toEqual("DomesticNonProfitVeteranCorporation")`

`getLastCalledWith` is currently a helper function on the frontend; you'd have to move it into shared test helpers instead to use it on the backend

api/src/client/ApiFormationClient.ts
```diff
@@ -226,8 +234,43 @@ export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): Fo
...
+              TrusteeElectionProcess:
+                formationFormData.nonprofitTrusteesMethodSpecified === "IN_FORM"
+                  ? formationFormData.nonprofitTrusteesMethodTerms
+                  : undefined,
+              AssetDistributionProvisionLocation:
+                formationFormData.nonprofitAssetDistributionSpecified === "IN_FORM" ? "Herein" : "Bylaw",
+              AssetDistribution:
+                formationFormData.nonprofitAssetDistributionSpecified === "IN_FORM"
+                  ? formationFormData.nonprofitAssetDistributionTerms
+                  : undefined,
```
aloverso, line 267: [sand] a lot of copy-paste in this section. Could be refactored as a helper method instead of copying the same logic so many times?

content/src/fieldConfig/business-formation.json
```diff
@@ -26,7 +26,9 @@
       "viewMoreButtonText": "View more",
       "amendmentInfo": "#### Review your information before submitting to the New Jersey Division of Revenue and Enterprise Services.\n If you need to make a change after submitting, you must file an [amendment online.](https://www.njportal.com/dor/businessamendments)",
       "viewLessButtonText": "View less",
-      "genericErrorText": "This field requires information."
+      "genericErrorText": "This field requires information.",
+      "readMoreText": "Read More",
+      "readLessText": "Read Less"
```
aloverso, line 31: [pebble] We already have "viewMoreButtonText" and "viewLessButtonText" right here. Are you sure this isn't a design oversight? I can't imagine they sometimes want some fields to say "View More" and some to say "Read More" inconsistently. I think this needs to be confirmed with design and/or content and then (assuming they want it all to be one consistent copy) remove the addition. 

Bear in mind that the whole point of the JSON structure is that the keys don't have to change when the values do. So if they want it to say "Read More" everywhere, we should continue to use the existing "viewMoreButtonText" and change its value rather than adding a new key just because the text has changed.

.circleci/config.yml
jphechter: [mountain] lol

content/src/fieldConfig/business-formation.json
```diff
@@ -242,6 +262,23 @@
...
+        "radioYesText": "Yes",
+        "error": "Select 'Yes' or 'No'.",
+        "domNpVetCorpReviewText": "**Yes**, my Domestic Non-Profit is a Veteran Corporation.",
+        "yesBoardMembersReviewText": "**Yes**, my Non-Profit will have board members.",
+        "noBoardMembersReviewText": "**No**, my Non-Profit will not have board members.",
+        "undefinedBoardMembersReviewText": "**My Non-Profit will or will not have board members**",
+        "boardMembersQualificationsReviewText": "**Board member qualifications are specified**",
+        "rightsAndLimitationsReviewText": "**Rights and limitations of the** board **members are specified**",
+        "choosingTrusteesReviewText": "**The method for choosing trustees are specified**",
+        "distributingAssetsReviewText": "**The method for distributing assets are specified**",
```
aloverso, line 278: [pebble] For these 5 keys where the entire string is bold, this violates the "I have not added any markdown to labels, titles and button text in config" item. If this text should be bold always, then we should control that in the code with a <strong> tag, not here in the markdown.

The ones that are only partially bold like `"**Yes**, my Non-Profit will have board members.` are fine.

api/src/client/ApiFormationClient.test.ts
jphechter: That's correct, and testing around that _shouldn't_ allow this to happen. I helped AnJu write this one. I was thinking that by including this we would be verifying that the `ApiFormationClient` wouldn't include this information even if it was supplied. Passing the fields in this state to the NICUSA API will return an error.

@aloverso, would you test for that case? Or would you be confident enough in other testing that this state of the data wouldn't be possible?

.circleci/config.yml
AnJuHyppolite: This is foreign non-profit, so I was thinking we would set it to true when we enable formation for FNP. Thoughts?

shared/src/legalStructure.ts
```diff
@@ -132,6 +132,22 @@ export const LegalStructures: LegalStructure[] = [
       "dbaName",
```
jphechter, line 132: [dust] If we have a dedicated `foreign-nonprofit` legal type, we might be able to drop `dbaName` from this type.

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -122,23 +122,6 @@ describe("<BusinessFormationPaginator />", () => {
     business = generateBusiness({ profileData, formationData });
   });
 
-  describe("search business name only", () => {
```
jphechter, line 125: LOVE removing code. 👍

web/src/components/tasks/business-formation/BusinessFormationTextBox.tsx
```diff
@@ -45,7 +45,7 @@ export const BusinessFormationTextBox = (props: Props): ReactElement => {
   return (
     <>
       <div className="flex flex-column mobile-lg:flex-row mobile-lg:flex-align-center margin-bottom-2">
-        <h2 className="h3-styling margin-bottom-0">
+        <h2 className="h3-styling margin-0-override">
```
jphechter, line 48: [dust] We do have access to a `margin-0` class. Any reason we opted for the override here?

Reason I'm asking is there are a lot of places it looks like we may be excessively using `important` in our CSS. `margin-0-override` seems to be one of them. I do see both used in the code.

web/src/components/tasks/business-formation/business/BusinessStep.test.tsx
```diff
@@ -1045,5 +1079,156 @@ describe("Formation - BusinessStep", () => {
...
     });
+
+    it("isVeteranNonprofit", async () => {
+      const page = await getPageHelper({ legalStructureId: "nonprofit" }, { isVeteranNonprofit: undefined });
+      await attemptApiSubmission(page);
+      expect(screen.getByRole("alert")).toHaveTextContent(Config.formation.fields.isVeteranNonprofit.label);
+      expect(screen.getByText(Config.formation.fields.isVeteranNonprofit.error)).toBeInTheDocument();
+    });
+
+    it("hasNonprofitBoardMembers", async () => {
```
jphechter, line 1090: [pebble] I know what these tests are doing, but typically we're trying to be a little more explicit with our test names. Is there a better name for the tests in this file than just the associated formation data field names?

jphechter: Applies to all new tests in this file.

web/src/components/tasks/business-formation/business/BusinessStep.tsx
```diff
@@ -1,69 +1,111 @@
...
+      state.formationFormData.legalType !== "foreign-limited-partnership"
+    )
+      return null;
+
+    return <Provisions />;
+  };
+
+  const defaultOrder: (ReactElement | null)[] = [
+    <MainBusiness key={"main-business"} />,
+    <hr className="margin-bottom-2 margin-top-0" aria-hidden={true} key={"business-line-1"} />,
```
jphechter, line 83: [pebble] I don't think these `<hr>` elements need a key unless they're in a map.

jphechter: This is kindof nit-picky, but if you wanted, you could move all of the `<hr>`s into the corresponding function call, the same way you have it with `lpSection()`. It might make this a little cleaner.

web/src/components/tasks/business-formation/business/BusinessStep.tsx
```diff
@@ -1,69 +1,111 @@
...
-      state.formationFormData.legalType !== "foreign-limited-partnership" ? (
-        <></>
-      ) : (
-        <>
-          <hr className="margin-bottom-2 margin-top-0" aria-hidden={true} />
-          <Provisions />
-        </>
-      )}
+      {elementsToDisplay.map((element, i) => (
+        <Fragment key={i}>{element}</Fragment>
```
jphechter, line 107: [pebble] You can do this without the fragment. Supply the key directly to the element.

The current implementation also makes the keys you added in `nonprofitOrder` and `defaultOrder` redundant.

web/src/components/tasks/business-formation/business/BusinessStep.tsx
```diff
@@ -1,69 +1,111 @@
...
+    () => state.formationFormData.businessLocationType !== "NJ",
+    [state.formationFormData.businessLocationType]
+  );
+
+  const isNonprofit = state.formationFormData.legalType === "nonprofit";
+
+  const addressSection = (): ReactElement =>
+    isForeign ? <MainBusinessForeignAddressFlow /> : <MainBusinessAddressNj />;
+
+  const lpSection = (): ReactElement | null => {
```
jphechter, line 25: [pebble] I think the pattern we've used elsewhere in the app is to return an empty element instead of `null`.

shared/src/legalStructure.ts
```diff
@@ -132,6 +132,22 @@ export const LegalStructures: LegalStructure[] = [
...
     ]),
   },
+  {
+    id: "foreign-nonprofit",
```
aloverso, line 136: [pebble] Why is this a new entry here? We don't have a an entry for foreign-limited-liability-company or something. 

My understanding is that this list is used to supply the values of the legal structure radio buttons. If a user is onboarded as Poppy, we definitely don't want them to be able to select "Foreign Nonprofit" in their business structure. Why does this entry exist?

shared/src/legalStructure.ts
aloverso: I don't think we *should* have `foreign-nonprofit` though so I think this needs to stay

web/public/mgmt/config.yml
```diff
@@ -658,13 +658,85 @@ collections:
...
+                      - { label: "Error", name: "error", widget: "string" }
+                      - {
+                          label: "Domestic nonprofit veteran corp review text",
+                          name: "domNpVetCorpReviewText",
+                          widget: "string",
+                        }
+                      - {
+                          label: "Board members yes review text",
+                          name: "yesBoardMembersReviewText",
+                          widget: "string",
```
aloverso, line 706: [pebble] No just this item, but ANY config that contains markdown syntax like the `**` bolding needs to be widget type `markdown`

content/src/fieldConfig/business-formation.json
```diff
@@ -242,6 +262,23 @@
...
         "error": "Select 'Yes' or 'No'."
       },
+      "isVeteranNonprofit": {
+        "label": "Non-Profit Veteran Corporation",
+        "description": "Is your Domestic Non-Profit a `Veteran Corporation|veteran-corporation`?",
+        "radioNoText": "No",
+        "radioYesText": "Yes",
+        "error": "Select 'Yes' or 'No'.",
+        "domNpVetCorpReviewText": "**Yes**, my Domestic Non-Profit is a Veteran Corporation.",
+        "yesBoardMembersReviewText": "**Yes**, my Non-Profit will have board members.",
```
aloverso, line 272: [boulder] Why are all these board member provision review text keys nested under `isVeteranNonprofit`? They are not related to the veteran question and should be in a key for the correct question.

aloverso: I see that the keys like `hasNonprofitBoardMembers` exists below - I assume the review text should go with its correct question, right? 

We can talk more about organizing this config properly if needed.

web/src/components/tasks/business-formation/business/IsVeteranNonprofit.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <WithErrorBar hasError={hasError} type="ALWAYS">
+      <FormControl variant="outlined" fullWidth className="padding-bottom-2">
+        <label className="margin-right-3 text-bold">
+          <Content>{Config.formation.fields.isVeteranNonprofit.description}</Content>
+        </label>
+        <RadioGroup
+          aria-label={camelCaseToSentence(fieldName)}
+          className="fac"
+          key={`${state.formationFormData.isVeteranNonprofit}-key`}
```
jphechter, line 26: [pebble] We don't need this

content/src/fieldConfig/business-formation.json
```diff
@@ -242,6 +262,23 @@
...
+        "noBoardMembersReviewText": "**No**, my Non-Profit will not have board members.",
+        "undefinedBoardMembersReviewText": "**My Non-Profit will or will not have board members**",
```
aloverso, line 274: [pebble] Why does this exist? I can see in your code elsewhere that the boardMembers is a required field and undefined is not a valid answer - they won't be allowed to submit if it is undefined. I think this should have the "Not Entered" highlighted text when it is undefined, right?

web/public/mgmt/config.yml
```diff
@@ -658,13 +658,85 @@ collections:
...
                               - { label: "Description", name: "description", widget: "markdown" }
+                  - label: Trustees
+                    name: trustees
+                    collapsed: true
+                    widget: object
+                    fields:
+                      - { label: "Label", name: "label", widget: "string" }
+                      - { label: "Add Button Text", name: "addButtonText", widget: "string" }
+                      - { label: "Error", name: "error", widget: "string" }
+                      - { label: "Success snackbar - body", name: "successSnackbarBody", widget: "string" }
```
aloverso, line 669: [pebble] Not just here but in all your CMS changes - if something is longer than a few words, it should be a "text" widget instead of a string widget so that the content team can edit in a text area. 

(obviously the comment about Markdown takes precedence here though)

Overall, I recommend looking opening up the CMS to review your changes. If YOU can't edit the whole text in the small "string" box or if you can't edit the Contextual Info / bolding via rich text editor, the widget may need to be updated.

content/src/fieldConfig/business-formation.json
```diff
@@ -242,6 +262,23 @@
...
+        "boardMembersQualificationsReviewText": "**Board member qualifications are specified**",
+        "rightsAndLimitationsReviewText": "**Rights and limitations of the** board **members are specified**",
+        "choosingTrusteesReviewText": "**The method for choosing trustees are specified**",
+        "distributingAssetsReviewText": "**The method for distributing assets are specified**",
+        "inForm": "in this form.",
+        "inByLaws": "in the bylaws."
```
aloverso, line 280: [pebble] I see these are being used to construct the sentence on the review page

1. if something is used for review only, the key should indicate that
2. (overriding item 1) I don't think these are necessary. Why not just use the text we already have for `radioInFormText`, and apply a `toLowercase()` on it and add a period? That way, if the content team wants to change the text, they don't need to make the change in two places to keep it consistent.

web/src/components/tasks/business-formation/business/BusinessStep.test.tsx
aloverso: I think this is [sand] aka follow-up refactor because every test in this "required field" describe section is just named on the field name alone. These new ones are following the pattern and it should be an independent separate change if we want to make this across the whole file (and across not just this file but the tests for the other steps as well)

web/src/components/tasks/business-formation/business/BusinessStep.tsx
aloverso: I think I disagree - I think it would be harder to manage because the way the order is changing and things are appearing or disappearing, I feel like it's more likely that you might wind up with doubled-up <hr>s. Idk

web/src/components/tasks/business-formation/business/NonprofitProvisions.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+      const page = await getPageHelper(
+        { legalStructureId: "nonprofit" },
+        { hasNonprofitBoardMembers: true, nonprofitAssetDistributionSpecified: undefined }
+      );
+      expect(screen.queryByLabelText("Nonprofit asset distribution terms")).not.toBeInTheDocument();
+      page.chooseRadio("nonprofitAssetDistributionSpecified-IN_BYLAWS");
+      expect(screen.queryByLabelText("Nonprofit asset distribution terms")).not.toBeInTheDocument();
+    });
+  });
+});
```
aloverso, line 152: [boulder] This whole test appears to be missing a describe block describing the behavior of `hasNonprofitBoardMembers`, if I'm not missing something? There's a missing test for what happens if that field is true, false, and not yet specified, which is huge because it controls the behavior of all the other fields on the component.

web/src/components/tasks/business-formation/business/NonprofitProvisions.tsx
```diff
@@ -0,0 +1,169 @@
...
+import { WithErrorBar } from "@/components/WithErrorBar";
+import { BusinessFormationContext } from "@/contexts/businessFormationContext";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { useFormationErrors } from "@/lib/data-hooks/useFormationErrors";
+import { camelCaseToSentence } from "@/lib/utils/cases-helpers";
+import { FormationFields, FormationTextField } from "@businessnjgovnavigator/shared";
+import { FormControl, FormControlLabel, FormHelperText, Radio, RadioGroup } from "@mui/material";
+import { ReactElement, ReactNode, useContext } from "react";
+
+type RadioValues = "true" | "false" | "IN_FORM" | "IN_BYLAWS";
```
aloverso, line 12: [pebble] I don't love this type definition, there is no radio button that has these 4 values. We should have 2 types describing the 2 different types of radio buttons. You can still use the same functions like `getRadio`, it will probably just need to take a generic type so that the caller can specify which type to use

aloverso: (Thinking about this from the perspective of our tech talk we watched a while ago ["Making impossible states impossible"](https://www.youtube.com/watch?v=IcgmSRJHu_8))

web/src/components/tasks/business-formation/business/NonprofitProvisions.tsx
```diff
@@ -0,0 +1,169 @@
...
+      {state.formationFormData.hasNonprofitBoardMembers &&
+        getRadio({
+          fieldName: "nonprofitAssetDistributionSpecified",
+          title: Config.formation.fields.nonprofitAssetDistributionSpecified.body,
+          values: ["IN_BYLAWS", "IN_FORM"],
+        })}
+      {state.formationFormData.hasNonprofitBoardMembers &&
+        state.formationFormData.nonprofitAssetDistributionSpecified === "IN_FORM" &&
+        getTextField("nonprofitAssetDistributionTerms")}
+    </>
```
aloverso, line 167: [sand] A lot of duplicated logical comparisons here, like `{state.formationFormData.hasNonprofitBoardMembers &&` for one, and 

```
{state.formationFormData.hasNonprofitBoardMembers &&
        state.formationFormData.X === "IN_FORM" &&
```

Solid refactoring opportunity to make this less repetitive and more readable.

web/src/components/tasks/business-formation/contacts/Addresses.tsx
```diff
@@ -186,7 +187,18 @@ export const Addresses = <T extends FormationMember | FormationIncorporator>(
...
           <tr>
-            <td className={"text-error-dark text-bold"}>{props.displayContent.error}</td>
+            <td colSpan={3} className={"text-error-dark text-bold"}>
+              {props.displayContent.error}
+            </td>
+          </tr>
+        ) : props.hasError &&
+          props.fieldName === "members" &&
+          props.legalType === "nonprofit" &&
+          props.addressData.length < 3 ? (
```
aloverso, line 197: [boulder] This logic should not be necessary (here and further down in this file). This should be able to be controlled solely by `props.hasError`.

At a code design level, we know you already coded the logic for the fact that "members" has an error for fewer than 3 for nonprofits, and for fewer than 1 for not nonprofits. You tested that explicitly in `getErrorStateForField.test.ts` because the logic lives right there in `getErrorStateForField`. So at a code design level, when you are rewriting error handling logic that has already been logically defined elsewhere, that should be an alarm bell.

From a code perspective, we can trace this back. `props.hasError` comes here into `<Addresses>` from `<Members>`. Taking a look at `Members`, I can see on line 69 that the prop value comes from *its* props. Okay, who renders Members? Looks like now-line-49 of `ContactsStep`. And the value of that prop therefore comes from `doesFieldHaveError("members")`. And if we trace backward into how `doesFieldHaveError` makes its decision (inside of `useFormationErrors`) we can see that value is calculated in `getErrorStateForField` - which of course, already handles the logical decision about the minimum length being 3 or 1 depending on legalType!

So far as I can see, redoing this logic here should not be necessary (and thus, neither is adding the new legalType prop)

web/src/components/tasks/business-formation/review/ReviewNonprofitProvisions.tsx
```diff
@@ -0,0 +1,160 @@
...
+      <>
+        <Content style={{ display: "inline-block" }}>
+          {fillInByLawFormText(questionString, outputText)}
+        </Content>
+        {questionString === undefined && (
+          <div style={{ display: "inline-block" }}>
+            &nbsp;
+            <span className={"bg-accent-warm-extra-light text-italic"}>
+              {Config.formation.general.notEntered}
+            </span>
```
aloverso, line 43: [pebble] Can we use the `<ReviewNotEntered />` component here?

web/src/components/tasks/business-formation/review/ReviewNonprofitProvisions.tsx
```diff
@@ -0,0 +1,160 @@
...
+  };
+
+  return (
+    <>
+      <hr className="margin-y-205" />
+      <ReviewSubSection header={"Provisions"}>
+        <div className="margin-top-2">
+          {state.formationFormData.hasNonprofitBoardMembers === undefined &&
+            showQuestionAnswer(
+              state.formationFormData.hasNonprofitBoardMembers,
```
aloverso, line 57: [pebble/sand] I think this whole thing would be 10x easier to do read if you did some cheeky little destructuring 

`const { hasNonprofitBoardMembers, ...OTHERS } = state.formationFormData.`

That way I don't have to scan my eyes over the `state.formationFormData` text repeated a thousand times to try to parse out the actual logic in here

web/src/components/tasks/business-formation/review/ReviewNonprofitProvisions.tsx
```diff
@@ -0,0 +1,160 @@
...
+            text={state.formationFormData.nonprofitBoardMembersTerms}
+            viewMoreText={Config.formation.general.readMoreText}
+            viewLessText={Config.formation.general.readLessText}
+            lines={2}
+          />
+        </div>
+
+        <div className="margin-top-2">
+          {state.formationFormData.hasNonprofitBoardMembers &&
+            state.formationFormData.hasNonprofitBoardMembers !== undefined &&
```
aloverso, line 80: [pebble] I think `state.formationFormData.hasNonprofitBoardMembers && state.formationFormData.hasNonprofitBoardMembers !== undefined` is redundant. 

If `state.formationFormData.hasNonprofitBoardMembers` evaluates to truthy then it cannot be undefined, and the second condition is unnecessary.

web/src/components/tasks/business-formation/business/BusinessStep.tsx
AnJuHyppolite: I think they're necessary because we call `.map` on `elementsToDisplay` which is set to either `defaultOrder` or `nonprofitOrder`.

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
```diff
@@ -120,7 +121,7 @@ export const ReviewSignatures = (): ReactElement => {
   };
 
   const signers = (): ReactElement => {
-    if (!isCorp && hasNoSigners) {
+    if (!isCorp && !isNonprofit && hasNoSigners) {
```
aloverso, line 124: [pebble/sand] Just imo, this would be easier to read if reversed logic. The if (not x and not y and not z) pattern is pretty mind bending.

Obviously our tests should confirm this but I *think* it'd be:

```
if (isCorp || isNonprofit || hasSigners) {
  return displaySigners();
} else {
  return displayEmptySigners();
}
```

aloverso: btw, if you do this refactor, I'd argue that the good ol campground rule suggests you should improve the `incorporators` method below as well

web/src/components/tasks/business-formation/review/ReviewStep.test.tsx
```diff
@@ -569,4 +569,180 @@ describe("Formation - ReviewStep", () => {
...
+      const getByMarkup = withMarkup(screen.getByText);
+      expect(
+        getByMarkup(
+          markdownToText(
+            `${Config.formation.fields.isVeteranNonprofit.distributingAssetsReviewText} ${Config.formation.fields.isVeteranNonprofit.inForm}`
+          )
+        )
+      ).toBeInTheDocument();
+      expect(screen.getByText(assetTerms)).toBeInTheDocument();
+    });
```
aloverso, line 739: [sand] Since all these tests are copy-paste identical except for the field names, this is an excellent opportunity to improve readability by writing them in a loop or an `it.each` instead of writing the same three tests 4 times over.

shared/src/legalStructure.ts
AnJuHyppolite: We experienced buggy behavior with the `Suffix Dropdown` and found that this approach helped , but we had to filter out the foreign-nonprofit so that it doesn't appear. I'd love to find a different approach/hear suggestions.

web/src/components/tasks/business-formation/review/ReviewStep.tsx
```diff
@@ -32,18 +33,32 @@ export const ReviewStep = (): ReactElement => {
...
               <ReviewForeignCertificate foreignGoodStandingFile={state.foreignGoodStandingFile} />
             </>
           )}
+          {isNonProfit && <ReviewNonprofitProvisions />}
+          {isNonProfit && hasProvisions && <ReviewProvisions />}
+          {isNonProfit && hasPurpose && (
+            <ReviewSubSection header={Config.formation.fields.businessPurpose.label}>
+              <ReviewText fieldName={"businessPurpose"} isExpandable={true} />
+            </ReviewSubSection>
+          )}
```
aloverso, line 61: I'd personally prefer these to be grouped together rather than redoing the same logic on every line:
```
{isNonprofit && (
  <>
    <ReviewNonprofitProvisions />
    {hasProvisions && <ReviewProvisions />}
    ....
  </>
)}
```

web/src/components/tasks/business-formation/validatedFieldsForUser.test.ts
```diff
@@ -47,163 +47,276 @@ describe("validatedFieldsForUser", () => {
...
+        "nonprofitAssetDistributionSpecified",
+      ];
+
+      const validatedFieldsArray = validatedFieldsForUser(formationFormData);
+      expect(validatedFieldsArray).toEqual(expect.arrayContaining(expected));
+      expect(validatedFieldsArray).toContain("nonprofitBoardMemberQualificationsSpecified");
+      expect(validatedFieldsArray).toContain("nonprofitBoardMemberQualificationsSpecified");
+      expect(validatedFieldsArray).toContain("nonprofitBoardMemberRightsSpecified");
+      expect(validatedFieldsArray).toContain("nonprofitTrusteesMethodSpecified");
+      expect(validatedFieldsArray).toContain("nonprofitAssetDistributionSpecified");
```
aloverso, line 281: [pebble] `nonprofitAssetDistributionSpecified` and all the others are in the `expected` array, so it seems like all these additional expect statements are redundant? Plus, lines 277 and 278 are also the same

web/src/components/tasks/business-formation/validatedFieldsForUser.test.ts
```diff
@@ -47,163 +47,276 @@ describe("validatedFieldsForUser", () => {
...
+      expect(validatedFieldsForUser(formationFormData)).toContain("createLimitedPartnerTerms");
+      expect(validatedFieldsForUser(formationFormData)).toContain("getDistributionTerms");
+      expect(validatedFieldsForUser(formationFormData)).toContain("makeDistributionTerms");
+    });
+  });
+
+  describe("nonprofit", () => {
+    const legalStructureId = "nonprofit";
+
+    it("isVeteranNonprofit and Provisions radio questions does not have a response", () => {
```
aloverso, line 213: [pebble] This and the following tests:
1. Our pattern is to write these test names as sentences
2. I think `isVeteranNonprofit` should be removed from this test name, it's misleading because its value does not influence the outcome of the expectations.
3. Should be `hasNonprofitBoardMembers` radio questions not `Provisions` right?

web/src/components/tasks/business-structure/LegalStructureRadio.tsx
```diff
@@ -93,7 +93,9 @@ export const LegalStructureRadio = (props: Props): ReactElement => {
               value={state.profileData.legalStructureId || ""}
               onChange={handleLegalStructure}
             >
-              {LegalStructuresOrdered.map((legalStructure) => {
+              {/*           Changed map to not have foreign-nonprofit in legal structures loop.
+                Needed foreign-nonprofit added to legal structures for suffix tests to pass*/}
+              {LegalStructuresOrdered.filter((l) => l.id !== "foreign-nonprofit").map((legalStructure) => {
```
aloverso, line 98: As I commented above, I don't think `"foreign-nonprofit"` should be in the LegalStructures array at all.

api/src/client/ApiFormationClient.test.ts
aloverso: @jphechter That's a good point, but I think you actually do get to test that behavior "for free" by removing them here. Because this chunk of code is part of the overrides being passed to `generateFormationFormData` - anything not overridden will get a default random string generated for it.

So if these LOREM IPSUMs get removed from the overrides in the test setup here, they'll actually still have values in the object that gets operated in on the ApiClient, so you're still implicitly testing that it is not included.

My rationale for removing them is that they clog up your visual filter of someone skimming and understanding the code. It seems to me the important part of this test is that it's saying what happens when these fields are IN_BYLAWS. So making that part of the setup stand alone without adding unnecessary strings that don't affect its outcome makes the test tighter overall. imo

shared/src/legalStructure.ts
aloverso: I think that since you defined `foreign-nonprofit` as a FormationLegalType, you will need to supply a value to the `BusinessSuffixMap` if the suffix tests aren't working?

aloverso: If you can't figure it out, we can chat. I don't think that adding it to this array of Legal Structures is a valid approach because it violates our domain boundaries.

For foreign legal structures, we don't think of them as a different LegalStructure type as their non-foreign counterparts. That's why there aren't any other `foreign-x` in this array. We don't want to break that pattern

.circleci/config.yml
aloverso: wait.... i am actually a dummy and didn't read fully. Y'all were right to have it false because it's **F**NP - foreign nonprofit. Which is of course not yet fully built yet. Yeah, this should be false in dev. I was wrong

web/src/components/tasks/business-formation/business/MainBusiness.tsx
```diff
@@ -82,6 +82,15 @@ export const MainBusiness = (): ReactElement => {
           )}
         </>
       )}
+      {state.formationFormData.legalType === "nonprofit" && (
+        <>
+          <div className="grid-row">
+            <IsVeteranNonprofit />
+          </div>
```
jphechter, line 89: [pebble] Do we need this div?

web/src/components/tasks/business-formation/business/BusinessStep.tsx
jphechter: @AnJuHyppolite right, but the key isn't relevant until you're inside the map. You shouldn't need to supply that ahead of time.

web/src/components/tasks/business-formation/contacts/ContactsStep.test.tsx
```diff
@@ -209,6 +210,108 @@ describe("Formation - ContactsStep", () => {
...
+          name: "Donald Whatever",
+          addressCity: "Miami",
+          addressLine1: "160 Something Ave NW",
+          addressLine2: "Office of Whatever",
+          addressState: { name: "Florida", shortCode: "FL" },
+          addressCountry: "US",
+          addressZipCode: "20501",
+          businessLocationType: "US",
+          title: "Incorporator",
+          signature: true,
```
jphechter, line 246: I don't think this is a huge deal, but given this is an open source project that others will likely see and could fork we should be careful about making political remarks.

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
```diff
@@ -722,16 +724,31 @@ describe("getErrorStateForField", () => {
...
+      expect(getErrorStateForField({ field: "members", formationFormData }).hasError).toEqual(true);
+    });
+
+    it("has no error when members is at least 3 for nonprofits", () => {
+      const members = [generateFormationMember({}), generateFormationMember({}), generateFormationMember({})];
+      const formationFormData = generateFormationFormData({ legalType: "nonprofit", members });
+      expect(getErrorStateForField({ field: "members", formationFormData }).hasError).toEqual(false);
+    });
+
+    it("has no error when members is at least 1 for non-nonprofits", () => {
```
jphechter, line 744: [dust]
```suggestion
    it("has no error when members is at least 1 for for-profits", () => {
```

web/src/components/tasks/business-formation/business/BusinessStep.tsx
aloverso: This came up in discussion with @FarazA22 on my ticket as well, and wanted to jump in because I *do* think we return null all over the place in our app instead of empty element.

Everywhere in our code, we prefer to use `(shouldShowX && <Element />)` over `(shouldShowX ? <Element /> : <></>)`

When `shouldShowX` is `false` or `undefined`, nothing renders, by usage of the "JSX hole" which [the docs say are logically the same](https://react.dev/learn/conditional-rendering#logical-and-operator-) as returning `undefined` or `null` in your JSX tree.

So rewriting `shouldShowX && <Element />` into a function with a return type DOES make sense to have that function return ReactElement or null, imo. It's a closer mirror to the precise situation it was refactoring from.

In addition, I'd argue it's preferable. I don't think empty element is a good / useful thing to have in our code, which is why I prefer `&&` over the empty-element ternary above.

There is, of course, [a pretty good argument](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx) for ternaries over `&&` because of the danger with putting a `0` mistakenly on the left-hand-side. If we're concerned over that (validly), we should discuss as a team. Even if we do switch to ternary though, I think `(shouldShowX ? <Element /> : null)` is a more clear representation of what we want, and is preferable.

web/src/components/tasks/business-formation/contacts/ContactsStep.test.tsx
aloverso: Lol, a good point but should be a follow up. This is already in the codebase, it's not a new addition from this PR

web/src/components/tasks/business-formation/business/IsVeteranNonprofit.tsx
AnJuHyppolite: @jphechter: What is the "this" you are referring to? Can you be more specific?

web/src/components/tasks/business-formation/review/ReviewStep.tsx
cbrunefearless: Changed

web/src/components/tasks/business-structure/LegalStructureRadio.tsx
cbrunefearless: Taken out

web/src/components/tasks/business-formation/business/MainBusiness.tsx
cbrunefearless: Yes sir

content/src/fieldConfig/business-formation.json
cbrunefearless: Changed

web/src/components/tasks/business-formation/business/NonprofitProvisions.tsx
AnJuHyppolite: Updated

web/src/components/tasks/business-formation/contacts/ContactsStep.test.tsx
cbrunefearless: Follow up out of this PR

web/src/components/tasks/business-formation/business/NonprofitProvisions.tsx
cbrunefearless: Done

web/src/components/tasks/business-formation/validatedFieldsForUser.test.ts
AnJuHyppolite: Done

web/src/components/tasks/business-formation/review/ReviewNonprofitProvisions.tsx
cbrunefearless: Done

web/src/components/tasks/business-formation/review/ReviewNonprofitProvisions.tsx
cbrunefearless: done

web/src/components/tasks/business-formation/business/BusinessStep.tsx
cbrunefearless: Moved HR tags up inside the functions

web/src/components/tasks/business-formation/review/ReviewNonprofitProvisions.tsx
cbrunefearless: Done

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
AnJuHyppolite: done

content/src/fieldConfig/business-formation.json
AnJuHyppolite: Still not fixed yet, @cbrunefearless.

web/src/components/tasks/business-formation/business/BusinessStep.test.tsx
AnJuHyppolite: Will refactor this separately.

web/src/components/tasks/business-formation/validatedFieldsForUser.test.ts
AnJuHyppolite: done

web/src/components/tasks/business-formation/review/ReviewStep.test.tsx
AnJuHyppolite: Will fix this afterward

content/src/fieldConfig/business-formation.json
cbrunefearless: Change has been made.. just need unit tests to pass.

web/src/components/tasks/business-formation/business/NonprofitProvisions.test.tsx
cbrunefearless: Done

web/src/components/tasks/business-formation/review/ReviewSignatures.tsx
AnJuHyppolite: going to do as a separate refactor

PR #6088, opened 2023-07-18
feat: [#184882700] allow user to add multiple businesses

aloverso:
## Description

Add multiple businesses and toggle between them in the nav bar

### Ticket

[#184882700](https://www.pivotaltracker.com/story/show/184882700)

## Code author checklist


web/cypress/e2e/multiple-businesses.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    // original business has no calendar
+    cy.get('[data-testid="filings-calendar"]').should("not.exist");
+  });
+
+  it("exits out of additional business onboarding without saving", () => {
+    completeNewBusinessOnboarding({});
+    cy.url().should("contain", "/dashboard");
+
+    onDashboardPage.getDropdown().click();
```
FarazA22, line 45: [sand/ dust]: I think the name of this method could be more specific. I think it could be left alone since we are planning to have conversations with Marcus to figure around cypress

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -240,6 +246,7 @@ describe("<NavBar />", () => {
       expect(menuEl).toBeInTheDocument();
       expect(screen.queryByText(Config.navigationDefaults.logoutButton)).not.toBeInTheDocument();
       expect(screen.queryByText(Config.navigationDefaults.myNJAccountText)).not.toBeInTheDocument();
+      expect(screen.queryByText(Config.navigationDefaults.addBusinessButton)).not.toBeInTheDocument();
     });
 
     it("displays an open dropdown menu when clicked and closes when clicked again", async () => {
```
FarazA22, line 252: [sand] I noticed the description mentions the drop-down menu is closed when it's clicked for the second time, it looks like the click action to close the menu doesn't happen in this test

web/src/lib/domain-logic/removeBusiness.ts
```diff
@@ -0,0 +1,24 @@
...
+  userData,
+  idToRemove,
+  newCurrentBusinessId,
+}: {
+  userData: UserData;
+  idToRemove: string;
+  newCurrentBusinessId: string;
+}): UserData => {
+  if (!Object.keys(userData.businesses).includes(newCurrentBusinessId)) {
+    throw "Business does not exist";
```
FarazA22, line 16: [dust] Should there be an error object that gets created or a console log? What happens when you throw a string?

web/src/lib/domain-logic/switchCurrentBusiness.ts
```diff
@@ -0,0 +1,12 @@
+import { UserData } from "@businessnjgovnavigator/shared/userData";
+
+export const switchCurrentBusiness = (userData: UserData, newCurrentBusinessId: string): UserData => {
+  if (!Object.keys(userData.businesses).includes(newCurrentBusinessId)) {
+    throw "Business does not exist";
```
FarazA22, line 5: [dust] Should there be an error object that gets created or a console log? What happens when you throw a string?

web/src/components/onboarding/ReturnToPreviousBusinessBar.tsx
```diff
@@ -0,0 +1,60 @@
...
+interface Props {
+  previousBusiness: Business | undefined;
+}
+
+export const ReturnToPreviousBusinessBar = (props: Props): ReactElement | null => {
+  const { updateQueue, userData } = useUserData();
+  const router = useRouter();
+  const businessName = getNavBarBusinessTitle(props.previousBusiness);
+
+  if (!props.previousBusiness) return null;
```
FarazA22, line 22: [dust] should this be a new pattern we follow (returning null instead of react fragments)?

web/src/styles/base/border-colors.scss
```diff
@@ -29,6 +29,10 @@
   border-color: $secondary-light !important;
 }
 
+.border-accent-cool-light {
```
FarazA22, line 32: [dust] I went through an exercise of creating elements with background color, text color, and border color for all of the colors that are defined in the design system within [storybook](http://nj-bfs-design-storybook.s3-website-us-east-1.amazonaws.com/?path=/story/atoms-colors--colors), and then manually created the css attribute names for the ones that did not exist.

I think this attribute is not needed. If it is, any idea why it's not rendering before you added it?

web/src/lib/domain-logic/switchCurrentBusiness.ts
aloverso: This is a good catch.

Really, it doesn't matter at all because the only reason to throw here is to prevent a situation that should never occur in the first place. This throw should never actually get thrown or have to be caught.

And technically you can throw anything you want in JS, including a string or even nothing at all. It throws just the same. But you're right, it's considered good form to throw an `Error` object instead of just a raw string.

web/src/components/onboarding/ReturnToPreviousBusinessBar.tsx
aloverso: I don't think it matters? React Fragments don't get inserted into the DOM, and neither do null values.

aloverso: Do you think this is something we should make consistent?

web/src/styles/base/border-colors.scss
aloverso: I'm not sure I understand - I only added it because it didn't exist yet in this file. Are you saying the color is missing from storybook?

web/src/components/onboarding/ReturnToPreviousBusinessBar.tsx
FarazA22: I don't think there is a need to make it consistent.

PR #6103, opened 2023-07-21
chore: [#185598851] control outage bar via github

aloverso:
### Ticket

[#185598851](https://www.pivotaltracker.com/story/show/#185598851)

### Approach

Outage bar controlled by changing [this config file in the business-files repo](https://github.com/newjersey/business-files/blob/master/navigator/config.json) which can be done from browser via the Github UI.

Note: I didn't currently distinguish between environments (dev/prod, etc) in the outage bar. If it's enabled, it will show up in all environments. I think this is fine because it seemed like unnecessary complexity to add it just for prod. It doesn't matter if it shows up in dev too because no one uses dev except us.

## Code author checklist


web/src/components/OutageAlertBar.tsx
```diff
@@ -1,15 +1,30 @@
...
+  });
+
+  const isEnabled = outageConfig.FEATURE_ENABLE_OUTAGE_ALERT_BAR;
+  const message = outageConfig.OUTAGE_ALERT_MESSAGE;
+  const alertType = outageConfig.OUTAGE_ALERT_TYPE as OutageAlertType;
+  const urlToFetchFrom = process.env.OUTAGE_ALERT_CONFIG_URL || "";
+
+  useMountEffect(() => {
+    if (!urlToFetchFrom) return;
+    axios.get(urlToFetchFrom).then((response) => {
```
jphechter, line 24: [pebble] I think we need to handle for unsuccessful calls. If for some reason we can't resolve this wouldn't a error show up in the UI?

jphechter: Literally the only reason I'm thinking about this is because GH had downtime earlier today.

aloverso: @jphechter I didn't think so? The default value is set on lines 11-15 and defaults to NOT showing the Outage Bar. So there's no way that the FEATURE_ENABLE_OUTAGE_ALERT_BAR value would be set to true unless the call succeeds, right? A failed call should just fail silently

jphechter: I wasn't thinking the alert bar would show up. I was thinking about the React errors, but that won't show up in prod any how. Never mind.

Screenshot 2023-07-21 at 3 47 09 PM

aloverso: Oh hm. That's fair, I can throw in a catch

PR #6112, opened 2023-07-25
Update Broken Link Checker

adidner:
## Description

We wanted to update the broken link checker which is now updated

### Ticket

### Steps to Test

spin up local and then go to this URL

http://localhost:3000/mgmt/deadurls

NOTE it will take like 10 - 12 minutes to load / run 

## Code author checklist

### DIFF
--- a/web/src/lib/static/admin/findDeadLinks.ts
+++ b/web/src/lib/static/admin/findDeadLinks.ts
@@ -10,10 +10,14 @@ const roadmapsDir = path.join(process.cwd(), "..", "content", "src", "roadmaps")
 const displayContentDir = path.join(process.cwd(), "..", "content", "src", "display-content");
 const filingsDir = path.join(process.cwd(), "..", "content", "src", "filings");
 const tasksDir = path.join(roadmapsDir, "tasks");
+const licenseTasksDir = path.join(roadmapsDir, "license-tasks");
 const industriesDir = path.join(roadmapsDir, "industries");
 const addOnsDir = path.join(roadmapsDir, "add-ons");
 const contextualInfoDir = path.join(displayContentDir, "contextual-information");
 const fieldConfigDir = path.join(process.cwd(), "..", "content", "src", "fieldConfig");
+const fundingsDir = path.join(process.cwd(), "..", "content", "src", "fundings");
+const certificationsDir = path.join(process.cwd(), "..", "content", "src", "certifications");
+const licensesDir = path.join(process.cwd(), "..", "content", "src", "licenses");
 
 type Filenames = {
   tasks: string[];
@@ -23,6 +27,10 @@ type Filenames = {
   contextualInfos: string[];
   displayContents: string[];
   fieldConfigs: string[];
+  fundings: string[];
+  certifications: string[];
+  licenses: string[];
+  licenseTasks: string[];
 };
 
 type FileContents = {
@@ -61,6 +69,10 @@ const getFilenames = (): Filenames => {
       return it.endsWith(".md");
     }),
     fieldConfigs: fs.readdirSync(fieldConfigDir),
+    fundings: fs.readdirSync(fundingsDir),
+    certifications: fs.readdirSync(certificationsDir),
+    licenses: fs.readdirSync(licensesDir),
+    licenseTasks: fs.readdirSync(licenseTasksDir),
   };
 };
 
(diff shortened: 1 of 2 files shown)


web/src/lib/static/admin/findDeadLinks.ts
```diff
@@ -238,29 +265,42 @@ export const findDeadLinks = async (): Promise<Record<string, string[]>> => {
...
+    for (const page of batch) {
+      const promise = new Promise((resolve) => {
+        const htmlUrlChecker = new HtmlUrlChecker(
+          {},
+          {
+            link: (result: any): void => {
+              if (
+                result.broken &&
+                !isTemplateLink(result.url.original) &&
+                result.url.original !== "https://twitter.com/i/lists/1148625562895826945"
```
adidner, line 284: So the way twitter works now, is that you need an account to access any page without a 403, so we run into this issue where the page just has this link for every page and it ended up taking much longer, so I've filtered it out.

web/src/lib/static/admin/findDeadLinks.ts
```diff
@@ -212,15 +223,31 @@ export const findDeadLinks = async (): Promise<Record<string, string[]>> => {
     ...filenames.filings.map((it) => {
       return `/filings/${it.split(".md")[0]}`;
     }),
+    "/welcome",
```
jphechter, line 226: This kinda stinks that populating pages is a manual process. I can see this getting looked over easily as other application changes are made.

I don't have a solution right now, but it'd be cool if there's some way that we could use logic from routes to initialize this variable.

web/src/lib/static/admin/findDeadLinks.ts
jphechter: Twitter is only ever mentioned in our footer, correct?

I'm comfortable with this.

web/src/lib/static/admin/findDeadLinks.ts
```diff
@@ -238,29 +265,42 @@ export const findDeadLinks = async (): Promise<Record<string, string[]>> => {
...
-      );
-      const url = new URL(process.env.REDIRECT_URL || "");
-      htmlUrlChecker.enqueue(`${url.origin}${page}`, {});
-    });
+  const numberOfBatches = Math.ceil(pages.length / 5);
+
+  for (let j = 0; j < numberOfBatches; j++) {
+    const pagePromises = [];
+    const startIndex = j * 5;
+    const batch = pages.slice(startIndex, startIndex + 5);
```
jphechter, line 273: I have a vague memory of you mentioning batching in standup. What was the reasoning for this again? Just curious.

web/src/lib/static/admin/findDeadLinks.ts
```diff
@@ -238,29 +265,42 @@ export const findDeadLinks = async (): Promise<Record<string, string[]>> => {
...
+        );
+        const url = new URL(process.env.REDIRECT_URL || "");
+        htmlUrlChecker.enqueue(`${url.origin}${page}`, {});
+      });
+
+      pagePromises.push(promise);
+    }
 
-    pagePromises.push(promise);
+    await Promise.all(pagePromises);
```
jphechter, line 301: Is there a reason we moved this inside the loop?

web/src/lib/static/admin/findDeadLinks.ts
adidner: I totally second this and might be good as a refactor ticket in the back log for the future. I'll make one right now

web/src/lib/static/admin/findDeadLinks.ts
adidner: So the reasoning is that we believe that starting jobs is faster than finishing jobs so what was happening is it would eventually get to the point where it was trying to do like 200 job simultaneously which ends up tanking the computer

web/src/lib/static/admin/findDeadLinks.ts
adidner: Yes, thats my understanding, in the footer on every page.

web/src/lib/static/admin/findDeadLinks.ts
adidner: Yeah, so the call here attempts to resolve all the promises simultaneously so previously, this meant every page all at once. 

Now that we're batching it means resolve each batch of 5

jphechter (approved):

PR #6119, opened 2023-07-27
fix: [#185362965] green box bug fix

adidner:
## Description

We talked with content and to get around the weird bug where we have back to back green block and switching between markdown and causing them to combine. They preffered a more robust solution to using <br/>, so I added a block for the green text

### Ticket

### Steps to Test

Go into the app and add the code block via the markdown section 

image
see that the preview and such works as expected

### Notes

I wanted to try and do try and do this with just an inner block and no outer block, but while it looks like its probably possible, not worth the effort

## Code author checklist

### DIFF
--- a/web/src/components/Content.tsx
+++ b/web/src/components/Content.tsx
@@ -44,6 +44,9 @@ export const Content = (props: ContentProps): ReactElement => {
       return <HorizontalLine />;
     },
     blockquote: GreenBox,
+    greenBox: (props: any): ReactElement => {
+      return <GreenBox>{props.children}</GreenBox>;
+    },
     infoAlert: (props: any): ReactElement => {
       return (
         <Alert variant="info" heading={props.header}>
--- a/web/src/components/PureMarkdownContent.tsx
+++ b/web/src/components/PureMarkdownContent.tsx
@@ -42,6 +42,13 @@ function customRemarkPlugin():
             header: node.attributes.header,
           };
         }
+        if (node.name === "greenBox") {
+          const data = node.data || (node.data = {});
+          data.hName = "greenBox";
+          data.hProperties = {
+            header: node.attributes.header,
+          };
+        }
         if (node.name === "icon") {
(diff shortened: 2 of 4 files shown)


web/src/components/Content.tsx
```diff
@@ -44,6 +44,9 @@ export const Content = (props: ContentProps): ReactElement => {
       return <HorizontalLine />;
     },
     blockquote: GreenBox,
+    greenBox: (props: any): ReactElement => {
```
adidner, line 47: Is there a particulalry easy way to check that this is working in the app? I was gonna let content test this out on their own.

I'm curious about saving changes in the CMS locally and seeing them locally but I'm not super confident in how to do that at the moment without screwing anything up. 

Like in the localhost version of the CMS can I just publish new changes and see them only on local or do I need to change the branch in my local config.yml first or something else entirley? Would love a bit more clarity if a reviewer has the answers here.

cbrunefearless (approved):
LGTM, would love to know about the CMS

web/src/components/Content.tsx
jphechter: @adidner I don't know how to do this locally. If you wanted to see this rendered, you could just manually modify one of the `md` files and add the syntax.
```
:::greenBox 
 This is a test of the emergency alert system.
:::
```

It does seem to be working nicely though. I accidentally [created a PR](https://github.com/newjersey/navigator.business.nj.gov/pull/6131/files) with the change.

jphechter (approved):

web/src/components/Content.tsx
adidner: Thats interesting, I'm gonna try and re-produce locally

adidner: It might not be working because of what we said regarding the CMS content pointing at the content branch

adidner: Works for me when I run the proxy server

PR #6133, opened 2023-07-28
chore: [#185575277] remove supporting text from radio labels in busin…

FarazA22:
Remove supporting text from radio button labels in the business structure task, so that links within the supporting task are clickable and do not select the radio button.

### DIFF
--- a/web/src/components/tasks/business-structure/LegalStructureRadio.tsx
+++ b/web/src/components/tasks/business-structure/LegalStructureRadio.tsx
@@ -62,14 +62,9 @@ export const LegalStructureRadio = (props: Props): ReactElement => {
   };
 
   const makeLabel = (legalStructureId: string): ReactElement => {
-    const supportingText =
-      (contentFromConfig.optionContent as Record<string, string>)[legalStructureId] ?? "";
     return (
-      <div data-value={legalStructureId}>
-        <div className={supportingText === "" ? "" : "text-bold"}>
-          {LookupLegalStructureById(legalStructureId).name}
-        </div>
-        <Content>{supportingText}</Content>
+      <div data-value={legalStructureId} className={"text-bold"}>
+        {LookupLegalStructureById(legalStructureId).name}
       </div>
     );
   };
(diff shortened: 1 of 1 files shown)


web/src/components/tasks/business-structure/LegalStructureRadio.tsx
```diff
@@ -95,16 +90,26 @@ export const LegalStructureRadio = (props: Props): ReactElement => {
...
+                      labelPlacement="end"
+                      key={legalStructure.id}
+                      data-testid={legalStructure.id}
+                      value={legalStructure.id}
+                      control={<Radio color={isFormFieldInValid ? "error" : "primary"} />}
+                      label={makeLabel(legalStructure.id)}
+                    />
+
+                    {(contentFromConfig.optionContent as Record<string, string>)[legalStructure.id] && (
+                      <div className="margin-left-4 padding-left-05">
```
jphechter, line 106: I've seen this a couple times in the app where we use `padding/margin-05` and `padding/margin-5`. As far as I can tell there's no difference in the style rules they apply. Is there a difference between these? It doesn't really bother me a ton, just wondering if there was a reason.

Not calling out instances of this specific class. Moreso the pattern of the class name containing a leading 0 in some cases.

jphechter (approved):

web/src/components/tasks/business-structure/LegalStructureRadio.tsx
FarazA22: For this case, I needed 36 pixels on the left and leveraged both the margin and padding to achieve it. If you look at the [Box Model](https://www.w3.org/TR/CSS2/box.html), you'll notice that padding adds spacing between the content and border, and that margin adds spacing between the border in other elements. 

CSS styling like background color well only add color to the contact and padding sections of the box model, the color will not apply to areas that are outside of the border.

jphechter: Sorry if I was unclear. I'm not talking about the combination of these 2 classes. I'm saying there is a `padding-left-05` and a `padding-left-5`. Both are used in the application. Both has the same rules associated with them.

FarazA22: My bad, they are different. `-05` will add 4px, while `-5` will add 40. Each whole number increments by a factor of 8. These css properties are free b/c of the NJ/USWDS import.
https://designsystem.digital.gov/utilities/margin-and-padding/

jphechter: Interesting. I don't know why I thought they had the same rules. That makes sense.

PR #6136, opened 2023-07-31
feat: [#185307864] non essential questions for poppy and dakota nexus

somabadri:
## Description

Create CMS infrastructure which allows the content folks to add non essential questions to industries without the aid of dev.

### Ticket

[#185307864](https://www.pivotaltracker.com/story/show/185307864)

### Approach

There is now a `NonEssentialQuestions.json` file which has an array of questions with an `id` field, `addOn` field and `question` text field, and content can add new entries to the file using the CMS. Additionally, there is a new field in the industry object known as `nonEssentialQuestionsIds` which is an array of non essential questions ids. Leveraging all of this, content can independently add a new question, decide which industry gets that question and what addOn goes along with it to determine what task is added and where.

## Code author checklist


shared/src/industry.ts
```diff
@@ -9,6 +9,7 @@ export interface Industry {
   readonly additionalSearchTerms?: string;
   readonly defaultSectorId?: string;
   readonly roadmapSteps: AddOn[];
+  readonly nonEssentialQuestions: NonEssentialQuestion[];
```
adidner, line 12: This is kinda confusing in that both Industry.ts and profileData.ts both have something called nonEssentialQuestions and one is a record while the other is an Array, could we consider more specific names here?

web/src/components/profile/ProfileNonEssentialQuestion.tsx
```diff
@@ -0,0 +1,62 @@
...
+                labelPlacement="end"
+                data-testid={`${props.essentialQuestionId}-radio-no`}
+                value={false}
+                control={<Radio color="primary" />}
+                label={Config.profileDefaults.fields.nonEssentialQuestions.default.noText}
+              />
+            </RadioGroup>
+          </FormControl>
+        </>
+      )}{" "}
```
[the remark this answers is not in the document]
somabadri: I think my editor added that automatically, will get rid of it.

shared/src/industry.ts
somabadri: thoughts on nonEssentialQuestionsArray and nonEssentialQuestionsRecord?

somabadri: or should I name the array to be nonEssentialQuestionMappings and leave the record as nonEssentialQuestions?

web/src/components/profile/ProfileNonEssentialQuestions.tsx
```diff
@@ -0,0 +1,39 @@
+import { ProfileNonEssentialQuestion } from "@/components/profile/ProfileNonEssentialQuestion";
```
adidner, line 1: Isn't it an anti-pattern to have 2 things named so similarly? (ProfileNonEssentialQuestion vs ProfileNonEssentialQuestions)

web/src/lib/domain-logic/getNonEssentialQuestionText.test.ts
```diff
@@ -0,0 +1,24 @@
...
+      questionText: "Test Question?",
+    },
+  ],
+}));
+
+describe("getNonEssentialQuestionText", () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
```
adidner, line 16: [dust]: could be nice to test that it can render multiple

web/src/lib/roadmap/buildUserRoadmap.test.ts
```diff
@@ -462,6 +488,34 @@ describe("buildUserRoadmap", () => {
...
+      });
+
+      it("doesn't add addOn if nonEssentialsQuestion is value is false", async () => {
+        await buildUserRoadmap(
+          generateStartingProfile({
+            industryId: "test-industry",
+            nonEssentialQuestions: {
+              "test-question-1": true,
+              "test-question-2": false,
+            },
```
adidner, line 513: [dust] This could be one test I think with just both assertions

web/src/pages/profile.tsx
```diff
@@ -487,6 +489,7 @@ const ProfilePage = (props: Props): ReactElement => {
         </ProfileField>
         <ProfileField fieldName="industryId">
```
adidner, line 490: Shouldn't we have 3 additions to this file for Poppy, Oscar and Dakota Nexus? Where is Oscar? Am I blind?

somabadri: The ticket is only for Poppy and Dakota Nexus because those are the only ones with industries, Oscar has sectors so this isn't applicable for it.

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -271,4 +310,53 @@ describe("profile - shared", () => {
...
+          businessPersona: businessPersona as BusinessPersona,
+        });
+        renderPage({ business, setRegistrationModalIsVisible });
+        selectByValue("Industry", "test-industry");
+        clickSave();
+        await waitFor(() => {
+          expect(currentBusiness().profileData.nonEssentialQuestions).toStrictEqual({
+            "test-question-1": false,
+            "test-question-2": false,
+          });
```
adidner, line 341: Would it be good to have a test here where the questions are true? Like the non essential question exist / are true

web/src/lib/roadmap/buildUserRoadmap.test.ts
somabadri: I was thinking of that as well, but I remember discussing with @aloverso at some point that we shouldn't have an `and` in a test because then you are testing for two different behaviors and rewriting this would require an `and`? I don't have strong preference, could go either way.

web/test/pages/profile/profile-shared.test.tsx
somabadri: I didn't add that test in this file because the `true` and `false` accurately being updated is being tested at the component level in `ProfileNonEssentialQuestion.test.tsx`?

web/public/mgmt/config.yml
```diff
@@ -4308,6 +4355,19 @@ collections:
...
+                    collapsed: true
+                    widget: object
+                    fields:
+                      - label: Default
+                        name: "default"
+                        collapsed: false
+                        widget: object
+                        fields:
+                          - { label: "Header", name: "header", widget: "string" }
+                          - { label: "Yes Text", name: "yesText", widget: "string" }
```
adidner, line 4369: I'm not 100% certain this is possible but I think you can give this a default text of yes via the default: property which would probably be nice for content

adidner (approved):
Seems quite good. 

A few optional potential name changes or added tests for ever so slightly better coverage. As well as a few questions about various minor things.

web/src/components/profile/ProfileNonEssentialQuestions.tsx
somabadri: Thoughts on ProfileNonEssentialQuestion and ProfileNonEssentialQuestionsSection respectively?

shared/src/industry.ts
adidner: I'm down either, the second seems a bit better to me

web/src/pages/profile.tsx
adidner: Gotcha, weird that the ticket calls out Oscar then

shared/src/industry.ts
aloverso: @somabadri It's generally considering an antipattern to name something based on its type like `nonEssentialQuestionsArray` or `displayString` or something, because your IDE can tell you the type, so naming it for its type is redundant.

I'd recommend trying to name it based on the difference between them, domain-wise, like what is their purpose?

also, just be careful with the `nonEssentialQuestionMappings` idea because a `Map` is traditionally a synonym for a Record, so if anything is called Mappings, it should be a record, not an array

web/src/lib/roadmap/buildUserRoadmap.test.ts
aloverso: @somabadri @adidner I do stand by that rule of thumb. In this case, if you *wanted* you could probably name the test as something like "only adds addOn when nonEssentialsQuestion value is true" and the "only" implies the not-add-if-false behavior, and then you could put both assertions in the same test without it feeling like it's two separate "and" tests. I don't have a strong preference - I think simplicity and clarity are the important things to optimize for, and either option is fairly simple so so strong preference.

adidner (approved):
LGTM

PR #6146, opened 2023-08-02
feat: [#184770173] update postOnboarding to change task status with checkboxes

aloverso:
### Ticket

[#184770173](https://www.pivotaltracker.com/story/show/#184770173)

### Steps to Test

This pertains to the post-onboarding question in "Prepare for Site Renovation, if Applicable" task in Step 3.

To get this task, be a location with a physical location, either by being a physical-location-only industry, or by selecting "No" for home-based-business

## Code author checklist

### DIFF
--- a/web/src/components/post-onboarding/PostOnboardingRadioQuestion.test.tsx
+++ b/web/src/components/post-onboarding/PostOnboardingRadioQuestion.test.tsx
@@ -1,0 +1,161 @@
+import { PostOnboardingRadioQuestion } from "@/components/post-onboarding/PostOnboardingRadioQuestion";
+import * as fetchPostOnboardingModule from "@/lib/async-content-fetchers/fetchPostOnboarding";
+import {
+  currentBusiness,
+  setupStatefulUserDataContext,
+  WithStatefulUserData,
+} from "@/test/mock/withStatefulUserData";
+import { ProfileData } from "@businessnjgovnavigator/shared/profileData";
+import {
+  generateBusiness,
+  generateProfileData,
+  generateUserDataForBusiness,
+} from "@businessnjgovnavigator/shared/test";
+import { Business } from "@businessnjgovnavigator/shared/userData";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+
+import { IsAuthenticated } from "@/lib/auth/AuthContext";
+import { generatePostOnboarding } from "@/test/factories";
+import { withAuthAlert } from "@/test/helpers/helpers-renderers";
+
+jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
+jest.mock("@/lib/data-hooks/useRoadmap", () => ({ useRoadmap: jest.fn() }));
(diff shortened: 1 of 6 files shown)


web/src/components/post-onboarding/PostOnboardingRadioQuestion.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    it("updates task progress to In-Progress if radio is Yes and one or more checkboxes completed", async () => {
+      await setupAndRenderWithCheckboxes();
+      fireEvent.click(getInputByName("content1"));
+      expect(getInputByName("content1").checked).toBe(true);
+      expect(currentBusiness().taskItemChecklist["checkbox-1"]).toEqual(true);
+      expect(currentBusiness().taskProgress[taskId]).toEqual("IN_PROGRESS");
+    });
+
+    it("updates task progress to Completed if radio is Yes and all checkboxes completed", async () => {
```
AnJuHyppolite, line 115: I noticed the test makes reference to ALL checkboxes, but we're only testing for two. Should we be adding input for two add'l checkboxes since there are a total of four?

aloverso: It *is* testing all of them. There are only 2 in the context of this test. The entire test is never making use of the "real" markdown. It is testing against a mocked task with 2 checkboxes: `checkbox-1` and `checkbox-2`, which is set up on line 95 of this file.

web/src/components/post-onboarding/PostOnboardingRadioQuestion.tsx
```diff
@@ -1,17 +1,19 @@
...
+import { postOnboardingCheckboxes } from "@/lib/domain-logic/postOnboardingCheckboxes";
 import { PostOnboarding } from "@/lib/types/types";
 import { ProfileData } from "@businessnjgovnavigator/shared/";
 import { FormControl, FormControlLabel, Radio, RadioGroup } from "@mui/material";
-import React, { ReactElement, useEffect, useState } from "react";
+import React, { ReactElement, useEffect, useMemo, useState } from "react";
 
 interface Props {
-  id: string;
+  postOnboardingQuestionId: string;
```
jphechter, line 11:
```
postOnboardingQuestionId: keyof typeof postOnboardingCheckboxes;
```
[dust] If you wanted to be super specific about what kind of string we allow, we could limit this to just the keys of `postOnboardingCheckboxes`. But that is a little verbose and clunky to read. The juice might not be worth the squeeze.

jphechter (approved):
I think this all makes sense. 👍

web/src/components/post-onboarding/PostOnboardingRadioQuestion.test.tsx
AnJuHyppolite: Yes, I saw the input was only two checkboxes and thought it may be okay b/c you were "feeding" the tests two checkboxes, but when I checked the UI, I noticed that they were four checkboxes. I do understand the need to be flexible, etc. Thanks, y'all!

aloverso: @AnJuHyppolite The test is using a mock, it is not testing the task that appears in the application UI

PR #6151, opened 2023-08-03
Update broken link checker revision 2

adidner:
## Description

We resolved to have the link checker run locally by a dev once a quater or at the request of content. So this PR is to add a downloadable option to the link checker to facilitate hand off to content as well as reverting to the version where it worked on local (as opposed to the way we hoped might work on dev)

### Ticket

### Notes

If you test this locally it will take 10 minutes to run 

## Code author checklist

### DIFF
--- a/web/src/pages/mgmt/deadurls.tsx
+++ b/web/src/pages/mgmt/deadurls.tsx
@@ -1,4 +1,5 @@
 import { MgmtAuth } from "@/components/auth/MgmtAuth";
+import { PrimaryButton } from "@/components/njwds-extended/PrimaryButton";
 import { SingleColumnContainer } from "@/components/njwds/SingleColumnContainer";
 import { PageSkeleton } from "@/components/PageSkeleton";
 import { findDeadLinks } from "@/lib/static/admin/findDeadLinks";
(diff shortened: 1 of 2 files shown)


web/src/pages/mgmt/deadurls.tsx
```diff
@@ -15,9 +16,50 @@ const DeadUrlsPage = (props: Props): ReactElement => {
...
+    link.remove();
+  };
+
   const authedView = (
     <>
       <h1>Dead URLs</h1>
+      <h2>
+        NOTE: this only works locally, (ask a dev to run this locally for results
+        https://www.pivotaltracker.com/story/show/185355762)
+      </h2>
```
adidner, line 59: I wanted to link to the pivotal ticket as the comments contain vital context if we ever want to update this to be available in dev again. 

It feels kinda like a comment in this regard but I would much rather have this context than not.

web/src/pages/mgmt/deadurls.tsx
```diff
@@ -15,9 +16,50 @@ const DeadUrlsPage = (props: Props): ReactElement => {
   const [isAuthed, setIsAuthed] = useState<boolean>(false);
   const [password, setPassword] = useState<string>("");
 
+  const generateContent = (): string => {
+    let deadLinkFileOutput = "Possibly Blocked Or Deadlinks \n\n";
```
cbrunefearless, line 20: Do we have a config to store this?

web/src/pages/mgmt/deadurls.tsx
cbrunefearless: No way to store this message elsewhere in a shared place?  I guess you dont have to if it is used once.

web/src/pages/mgmt/deadurls.tsx
aloverso: We don't because this isn't actually user-facing so we don't need to set it so content can change it (because it doesn't matter)

cbrunefearless (approved):
lgtm

PR #6164, opened 2023-08-08
chore: [#184643463] prefactor to move profileDefaults from config.jso…

FarazA22:
Prefactor to move profileDefaults from config.json to profile.json

## Code author checklist

### DIFF
--- a/web/src/components/onboarding/OnboardingEssentialQuestion.tsx
+++ b/web/src/components/onboarding/OnboardingEssentialQuestion.tsx
@@ -48,7 +48,7 @@ export const OnboardingEssentialQuestion = <T,>(props: Props<T>): ReactElement =
         />
         {isFormFieldInValid && (
           <div className="text-error-dark text-bold margin-top-05">
-            {Config.profileDefaults.essentialQuestionInlineText}
+            {Config.profileDefaults.default.essentialQuestionInlineText}
           </div>
         )}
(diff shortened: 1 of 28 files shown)


web/src/components/onboarding/OnboardingEssentialQuestion.tsx
```diff
@@ -48,7 +48,7 @@ export const OnboardingEssentialQuestion = <T,>(props: Props<T>): ReactElement =
         />
         {isFormFieldInValid && (
           <div className="text-error-dark text-bold margin-top-05">
-            {Config.profileDefaults.essentialQuestionInlineText}
+            {Config.profileDefaults.default.essentialQuestionInlineText}
```
aloverso, line 51: Did you consider putting the variables top-level instead of nested inside the `default` object? Then, all these changes would not be necessary because the path would remain the same.

cbrunefearless: Good call Anne!

FarazA22: Yes, the thought did come up. I decided to go this route because I wanted to store the data in a consistent manner. The business formation JSON file was one file that I look that closely, it has nested objects and does not store variables are the top level.

cbrunefearless (approved):
lgtm

PR #6171, opened 2023-08-10
feat: [#185167184] add Note block to cms

adidner:
## Description

We wanted to add a Note block to the CMS capabilities

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185167184)

## Code author checklist

### DIFF
--- a/web/src/components/Content.test.tsx
+++ b/web/src/components/Content.test.tsx
@@ -53,4 +53,18 @@ describe("<Content />", () => {
       expect(screen.getByText("body text")).toBeInTheDocument();
     });
   });
+
+  it("renders greenBox content", () => {
+    const mdString = ":::greenBox\n" + "body text\n" + ":::";
+
+    render(<Content>{mdString}</Content>);
+    expect(screen.getByText("body text")).toBeInTheDocument();
+  });
+
+  it("renders note content", () => {
+    const mdString = ":::note \n" + "body text\n" + ":::";
+
+    render(<Content>{mdString}</Content>);
+    expect(screen.getByText("body text")).toBeInTheDocument();
+  });
 });
--- a/web/src/components/PureMarkdownContent.tsx
+++ b/web/src/components/PureMarkdownContent.tsx
@@ -29,33 +29,29 @@ export const PureMarkdownContent = (props: Props): ReactElement => {
   return <>{markdown}</>;
 };
 
+type hProperties = {
+  header?: any;
+  type?: any;
+};
+
+function setNodeValues(node: any, nodeName: string, hProperties: hProperties): void {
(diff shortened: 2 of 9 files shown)


web/src/styles/components/alerts.scss
```diff
@@ -2,7 +2,43 @@
...
 }
 
+.usa-alert--note::before {
+  content: url(/img/push-pin.svg);
+  height: 2.5rem;
+  width: 1.5rem;
+  display: block;
+  position: absolute;
+  left: 0.91rem;
+}
```
adidner, line 12: So this isn't exactly how I would have wanted to do this but, we're using USWDS stuff for most of our alert types and this matches their patterns.

web/src/styles/components/alerts.scss
```diff
@@ -2,7 +2,43 @@
...
+  color: $base-darkest;
+}
+
+.usa-alert.usa-alert--note {
+  @include note();
+}
+
+.usa-alert.usa-alert--note.alert-border-small {
+  @include note();
+}
```
adidner, line 32: Doing this twice is also not my preference but if we want the CSS priority / hierarchy to be correct with all the combinations of different border sizes and including right border properties that I added to alert this is necessary.

jphechter: Couldn't we comma delineate them? Or does that have some side-effect that I'm not considering?

web/src/components/Content.test.tsx
```diff
@@ -53,4 +53,18 @@ describe("<Content />", () => {
...
     });
   });
+
+  it("renders greenBox content", () => {
+    const mdString = ":::greenBox\n" + "body text\n" + ":::";
+
+    render(<Content>{mdString}</Content>);
+    expect(screen.getByText("body text")).toBeInTheDocument();
+  });
+
```
jphechter, line 63: Thanks for retroactively adding this!

web/src/components/PureMarkdownContent.tsx
```diff
@@ -49,6 +49,13 @@ function customRemarkPlugin():
             header: node.attributes.header,
           };
         }
+        if (node.name === "note") {
+          const data = node.data || (node.data = {});
+          data.hName = "note";
+          data.hProperties = {
+            header: node.attributes.header,
+          };
+        }
```
jphechter, line 58: [dust] This section might be a good candidate for a `switch`.

jphechter: [pebble] A lot of this (lines 38-66) is redundant. We could rewrite this to be a little more efficient and possibly easier to read.

PR #6176, opened 2023-08-10
[#185307864] [#185815263] non essential questions fixes and add to search

somabadri:
## Description

- fix to only add a task corresponding to a non essential question if the id is present in the industry file
- fix for string to boolean typing on the radio question answers
- added test for displaying both question text and option text suffix
- added non essential questions to search

### Ticket

[#185307864](https://www.pivotaltracker.com/story/show/185307864)

## Code author checklist

### DIFF
--- a/web/src/lib/roadmap/buildUserRoadmap.test.ts
+++ b/web/src/lib/roadmap/buildUserRoadmap.test.ts
@@ -21,7 +21,7 @@ jest.mock("@/lib/domain-logic/getNonEssentialQuestionAddOn", () => ({
   getNonEssentialQuestionAddOn: jest.fn(),
 }));
 
-const mockGetNonEssentiaQuestionAddOn = (
+const mockGetNonEssentialQuestionAddOn = (
   getNonEssentialAddOnModule as jest.Mocked<typeof getNonEssentialAddOnModule>
 ).getNonEssentialQuestionAddOn;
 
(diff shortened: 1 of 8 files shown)


web/src/lib/roadmap/buildUserRoadmap.test.ts
```diff
@@ -490,30 +490,46 @@ describe("buildUserRoadmap", () => {
...
+        expect(getLastCalledWith(mockRoadmapBuilder)[0].addOns).not.toContain("non-essential-add-on-2");
+      });
+
+      it("doesn't add addOn if question id is not in industry nonEssentialQuestions", async () => {
+        mockGetNonEssentialQuestionAddOn.mockReturnValue("non-essential-add-on-3");
+        await buildUserRoadmap(
+          generateStartingProfile({
+            industryId: "non-essential-question-industry",
+            nonEssentialRadioAnswers: {
+              "non-essential-question-3": true,
```
aloverso, line 528: The test would be easier to read, imo, if you name this something descriptive rather than non-essential-question-3.

Reading this test on its own, it's not obvious why the expectation - why _shouldn't_ the addOns contain this one, when we're doing a test with "non-essential-question-industry" - which _sounds like_ it should have something like "non-essential-question-3" in it.

I have to scroll to the top of the file to understand that oh, that industry contains "non-essential-question-1" and "non-essential-question-2" but not "non-essential-question-3" which is a very subtle difference.

Anyways, all that to say, this would be much more obvious and require less investigation if you name it something like "question-not-contained-in-industry" or something. Names can be used to descriptively give intuition to the reader of what the purpose is.

cbrunefearless (approved):
lgtm

PR #6194, opened 2023-08-11
feat: 185681556 sidebar test removal

cbrunefearless:
## Description
Removing welcome and welcome up and running from sodebarcontainertests

### Ticket

[#185681556 ](https://www.pivotaltracker.com/story/show/)

## Code author checklist

### DIFF
--- a/web/src/components/dashboard/SidebarCardsContainer.test.tsx
+++ b/web/src/components/dashboard/SidebarCardsContainer.test.tsx
@@ -77,6 +77,51 @@ describe("<SidebarCardsContainer />", () => {
   };
 
   describe("nudges", () => {
+    it("renders content for visible card", () => {
+      const sidebarCards = {
+        fakeCard: generateSidebarCardContent({ contentMd: "FakeCardContent" }),
+      };
+
+      useMockBusiness({ preferences: generatePreferences({ visibleSidebarCards: ["fakeCard"] }) });
+
+      renderPage({ sidebarCards });
+      expect(screen.getByText("FakeCardContent")).toBeInTheDocument();
+    });
+
+    it("renders a card in the top section if set to above-opportunities", () => {
+      const sidebarCards = {
+        "fake-above-card": generateSidebarCardContent({
+          id: "fake-above-card",
+          section: "above-opportunities",
+        }),
+      };
+
+      useMockBusiness({
+        preferences: generatePreferences({ visibleSidebarCards: ["fake-above-card"] }),
+      });
+
+      renderPage({ sidebarCards });
+      const topCardSection = screen.getByTestId("top-cards");
+      expect(within(topCardSection).getByTestId("fake-above-card")).toBeInTheDocument();
+    });
+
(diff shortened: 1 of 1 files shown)


web/src/components/dashboard/SidebarCardsContainer.test.tsx
```diff
@@ -77,6 +77,40 @@ describe("<SidebarCardsContainer />", () => {
   };
 
   describe("nudges", () => {
+    it("renders a card in the top section if set to above-opportunities", () => {
+      const sidebarCards = {
+        "funding-nudge": generateSidebarCardContent({
+          id: "funding-nudge",
```
aloverso, line 83: Maybe we should invent an ID instead of using an existing ID, so we don't run into the same confusion again in the future regarding whether this test is testing the "funding-nudge"?

web/src/components/dashboard/SidebarCardsContainer.test.tsx
```diff
@@ -77,6 +77,40 @@ describe("<SidebarCardsContainer />", () => {
...
+      };
+
+      useMockBusiness({
+        preferences: generatePreferences({ visibleSidebarCards: ["formation-nudge"] }),
+      });
+
+      renderPage({ sidebarCards });
+      const bottomCardSection = screen.getByTestId("bottom-cards");
+      expect(within(bottomCardSection).getByTestId("formation-nudge")).toBeInTheDocument();
+    });
```
aloverso, line 112: I think we're still missing the third test, which was called "renders the welcome card" but in the body, it was just testing that the content of a card rendered on the page, so I'd argue it should be added back and renamed something like "it renders content for visible card"

somabadri: oh that's a good catch. I think to your point above, we should create test content so it's not confusing because I read renders welcome card and fully thought it was only specific to that card only.

cbrunefearless: @aloverso can you re review :)

web/src/components/dashboard/SidebarCardsContainer.test.tsx
cbrunefearless: I just called it fake-above-card and fake-below-card... Always up for opinions on naming

web/src/components/dashboard/SidebarCardsContainer.test.tsx
```diff
@@ -77,6 +77,51 @@ describe("<SidebarCardsContainer />", () => {
   };
 
   describe("nudges", () => {
+    it("renders content for visible card", () => {
+      const sidebarCards = {
+        welcome: generateSidebarCardContent({ contentMd: "WelcomeCardContent" }),
```
somabadri, line 82: [pebble] could we name this fake card or test card, to decouple it a realistic card like we did below?

cbrunefearless: done

somabadri (approved):

PR #6233, opened 2023-08-11
Cannabis location validation alert

jphechter:
## Description

Adds an alert message to formation and the profile page for new (Poppy and Dakota) cannabis businesses.

### Ticket

[184992648](https://www.pivotaltracker.com/story/show/184992624) [184992624](https://www.pivotaltracker.com/story/show/184992648)

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/business/MainBusinessAddressNj.tsx
+++ b/web/src/components/tasks/business-formation/business/MainBusinessAddressNj.tsx
@@ -1,3 +1,4 @@
+import { CannabisLocationAlert } from "@/components/CannabisLocationAlert";
 import { Content } from "@/components/Content";
 import { ModifiedContent } from "@/components/ModifiedContent";
 import { Alert } from "@/components/njwds-extended/Alert";
@@ -61,6 +62,7 @@ export const MainBusinessAddressNj = (): ReactElement => {
       </div>
       {isExpanded && (
         <>
+          <CannabisLocationAlert />
           <BusinessFormationTextField
             label={Config.formation.fields.addressLine1.label}
             fieldName="addressLine1"
--- a/web/src/lib/utils/helpers.ts
+++ b/web/src/lib/utils/helpers.ts
@@ -45,6 +45,8 @@ export const useMountEffectWhenDefined = (fun: () => void, thingToBeDefined: unk
 
 export const useScrollToPathAnchor = (): void => {
   useEffect(() => {
+    if (window.location.pathname === "/mgmt/cms") return;
+
     const path = window.location.hash;
     if (path && path.includes("#")) {
       const id = path.replace("#", "");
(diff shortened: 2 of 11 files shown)


web/src/lib/utils/helpers.ts
```diff
@@ -45,6 +45,8 @@ export const useMountEffectWhenDefined = (fun: () => void, thingToBeDefined: unk
 
 export const useScrollToPathAnchor = (): void => {
   useEffect(() => {
+    if (window.location.pathname === "/mgmt/cms") return;
```
jphechter, line 48: Scroll hook was trying to use the admin panel url as an on page id. This prevents that.

web/src/components/tasks/business-formation/business/MainBusinessAddressNj.tsx
```diff
@@ -61,6 +64,11 @@ export const MainBusinessAddressNj = (): ReactElement => {
       </div>
       {isExpanded && (
         <>
+          {industryId === "cannabis" && (
+            <Alert variant="warning">
+              <Content>{Config.profileDefaults.default.cannabisLocationAlert}</Content>
+            </Alert>
+          )}
```
aloverso, line 71: This 5-line code chunk is duplicated a bunch and could be its own component

web/src/pages/profile.tsx
```diff
@@ -516,6 +520,11 @@ const ProfilePage = (props: Props): ReactElement => {
         </ProfileField>
 
         <ProfileField fieldName="municipality" locked={shouldLockMunicipality()}>
+          {profileData.industryId === "cannabis" && (
```
adidner, line 523: [dust] the thought occurs that maybe this would be a cool concept to make super flexible and reusable at some point with regards to letting content choose to have industry specific alerts in profile when and where ever they want somehow. 

Not relevant at the moment but if they request something like this done another time something to consider maybe

web/test/pages/profile/profile-starting.test.tsx
```diff
@@ -41,7 +42,11 @@ import {
   OperatingPhases,
   UserData,
 } from "@businessnjgovnavigator/shared/index";
-import { generateFormationData, generateTaxFilingData } from "@businessnjgovnavigator/shared/test";
+import {
+  generateFormationData,
```
adidner, line 46: So I wonder if maybe these tests belong in profile-shared with an it each to test against each persona. As they all feel like they're repeating the same test.

adidner: I can see a situation in the future where we want to change this slightly and have to modify it across all these files as opposed to 1 test in shared

somabadri: I agree that if there is a way to consolidate the repeated tests, then we should do it and place it in the shared file to decrease the amount of places you would have to go to update logic if the behavior were to change.

somabadri (approved):

web/src/pages/profile.tsx
jphechter: That could be a cool feature and it wouldn't be too hard.

web/src/components/tasks/business-formation/business/MainBusinessAddressNj.tsx
jphechter: included

PR #6236, opened 2023-08-14
feat: remove-account-creation-card-#185681589

AnJuHyppolite:
### Ticket

[#185681589](https://www.pivotaltracker.com/story/show/185681589)

### Steps to Test

When you are prompted to created an account, close the **Link with myNewJersey to Save Your Progress** snack bar. A create an account card should appear. Once you create your account, the card confirmation that once appeared in the **For You*** section, will no longer appear.

## Code author checklist


web/src/components/dashboard/SidebarCardsContainer.test.tsx
```diff
@@ -121,36 +121,6 @@ describe("<SidebarCardsContainer />", () => {
       const bottomCardSection = screen.getByTestId("bottom-cards");
       expect(within(bottomCardSection).getByTestId("fake-below-card")).toBeInTheDocument();
     });
-
-    it("removes successful registration card when it's closed", async () => {
```
aloverso, line 125: [boulder] Same as my comment regarding the welcome card story last week, this test should not be removed. It is not testing the success card, it is testing the "close button" functionality and just happens to use the success card as its test subject. Without this test, there is no test for the behavior that happens when a card has `hasCloseButton` = true and what happens when that close button is clicked. 

This test should continue to exist, it should just be a made-up card instead of the success card.

web/src/components/dashboard/SidebarCardsContainer.test.tsx
```diff
@@ -125,30 +125,28 @@ describe("<SidebarCardsContainer />", () => {
     it("removes successful registration card when it's closed", async () => {
```
adidner, line 125: I think the test should be renamed as well right?

PR #6237, opened 2023-08-14
chore: [#185153256] add nonprofit formation feature flag

AnJuHyppolite:
## Description

Added feature flag for non-profit formation

### Ticket

[185153256](https://www.pivotaltracker.com/story/show/185153256)

## Code author checklist


web/src/lib/domain-logic/allowFormation.ts
```diff
@@ -17,6 +17,7 @@ export const allowFormation = (
     "foreign-s-corporation": process.env.FEATURE_BUSINESS_FCORP === "true",
     "foreign-c-corporation": process.env.FEATURE_BUSINESS_FCORP === "true",
     "foreign-nonprofit": process.env.FEATURE_BUSINESS_FNP === "true",
+    nonprofit: process.env.FEATURE_BUSINESS_NP === "false",
```
jphechter, line 20: Shouldn't this be `true`?

aloverso:
@AnJuHyppolite did you test this locally by setting the feature flag and ensuring it is disabled? I don't see how this could work, given that the logic in `allowFormation` only checks feature flags for FOREIGN businesses and assumes true otherwise.

web/src/lib/domain-logic/allowFormation.ts
cbrunefearless: yes

content/src/roadmaps/add-ons/nonprofit.json
```diff
@@ -5,6 +5,11 @@
       "weight": 3,
       "task": "nonprofit-business-trustees"
     },
+    {
+      "step": 2,
+      "weight": 0,
+      "task": "search-business-name"
```
cbrunefearless, line 11: is this correct?

web/src/pages/tasks/[taskUrlSlug].tsx
```diff
@@ -126,6 +126,14 @@ const TaskPage = (props: Props): ReactElement => {
               <BusinessFormation
                 task={getTaskFromRoadmap(roadmap, props.task.id)}
                 displayContent={props.displayContent}
+                searchOnly={
+                  allowFormation(
+                    business?.profileData.legalStructureId,
+                    business?.profileData.businessPersona
+                  )
+                    ? false
+                    : true
```
Comment (aloverso) [inline web/src/pages/tasks/[taskUrlSlug].tsx:135]: This should be simplified, ternary not necessary

web/test/helpers/helpers-formation.tsx
```diff
@@ -88,6 +89,7 @@ export const preparePage = ({
   isAuthenticated,
   setRegistrationModalIsVisible,
   user,
+  searchOnly,
```
aloverso, line 92: I don't think we need to add this if we're not adding the tests for `searchOnly` back in as well, then this is never being used

aloverso:
I'm not seeing this working when the feature flag is enabled as `true`.

image

When `true`, the "Search for Available Business Names" task should not exist, and the formation task should be called "Check for Available Names and Form Your Business".

This is happening because the change to `nonprofit.json` is happening no matter what. Those two changes are:
- Adding the search task
- The modification that sets the task to the "Authorize" task name for Formation.

These are correct.... only for when the feature flag is `false`. When true, we will need to remove these.

...Or vice versa... you could NOT have them in the json, and dynamically add them based on the value of the feature flag. Either one works. In fact... I think the best approach would be to have these two changes be their OWN add-on, like a `nonprofit-legacy-formation.json`. Then, depending on the value of the feature flag, we either add that add-on to the roadmap as well, or don't. 

This should be some logic in `buildUserRoadmap` that checks if the feature flag is enabled/disabled and adds those tasks dynamically. The feature isn't working right now when `true` because they're currently not dynamic.

.circleci/config.yml
```diff
@@ -190,6 +190,7 @@ commands:
             echo 'export FEATURE_BUSINESS_FLC=$FEATURE_BUSINESS_FLC_DEV' >> $BASH_ENV
             echo 'export FEATURE_BUSINESS_FCORP=$FEATURE_BUSINESS_FCORP_DEV' >> $BASH_ENV
             echo 'export FEATURE_BUSINESS_FNP=$FEATURE_BUSINESS_FNP_DEV' >> $BASH_ENV
+            echo 'export FEATURE_BUSINESS_NP=$FEATURE_BUSINESS_NP_DEV' >> $BASH_ENV
```
aloverso, line 193: [sand - fix this as a follow-up commit]

In order for the `feature-flags` page to work, we need to pass ALL environment variables into the dev env so that it can display their values.

So *in addition to* this line, which is good as is, we need to set the STAGING and PROD env vars here in the dev section. You can see this happening below on lines 195-198 and lines 200-203. We need to do this for both NP and FNP. Note that the "live" variable on this line here gets renamed without the environment tag (ie, `FEATURE_BUSINESS_NP_DEV` becomes `FEATURE_BUSINESS_NP_DEV`) but for the additional staging/prod variables, it keeps its original name with the `_STAGING` or `_PROD`

PR #6247, opened 2023-08-17
When adding the business in the start section bug

adidner:
## Description

So for this bug we found 3 things
1. The Nexus DBA we found that the error alert when you have an invalid name and have to create a DBA is confusing -> change to info alert
2. Even if the name was unavailable and we put you into the DBA flow we still update the nave dropdown with the business name -> should be "unamed DBA"
3. After a DBA name is chosen the nav bar drop down still displays normal business name rather than DBA name -> dropdown shows DBA name

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185698190)

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/config.json
+++ b/content/src/fieldConfig/config.json
@@ -50,6 +50,7 @@
   "navigationDefaults": {
     "navBarUnnamedOwnedBusinessText": "Unnamed Business",
     "navBarUnnamedForeignRemoteSellerWorkerText": "Unnamed Out-of-State Business",
+    "navBarUnnamedDbaBusinessText": "Unnamed DBA",
     "guestModalTitle": "## Link with myNewJersey to Save Your Progress",
     "guestAlertBody": "Now that you've filled out your business information, you can save your progress and come back later to [link your Navigator account with myNewJersey](/account-setup?source=guest_snackbar). ",
     "guestSuccessTitle": "### Welcome to your Business.NJ.gov account!",
--- a/web/src/lib/domain-logic/getNavBarBusinessTitle.test.ts
+++ b/web/src/lib/domain-logic/getNavBarBusinessTitle.test.ts
@@ -141,6 +141,38 @@ describe("getNavBarBusinessTitle", () => {
         });
       });
     });
+
+    describe("Nexus DBA", () => {
+      it("shows unnamed dba when the nexusDbaName is empty and needsNexusDbaName is true", () => {
+        const business = generateBusiness({
+          profileData: generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            businessName: "test business Name",
+            needsNexusDbaName: true,
+            nexusDbaName: "",
+          }),
+        });
+        const navBarBusinessTitle = getNavBarBusinessTitle(business, IsAuthenticated.TRUE);
+        expect(navBarBusinessTitle).toEqual(Config.navigationDefaults.navBarUnnamedDbaBusinessText);
+      });
+
+      it("shows the dba name when the user has entered a DBA name", () => {
+        const dbaName = "dbaName";
+
+        const business = generateBusiness({
+          profileData: generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            businessName: "test business Name",
+            needsNexusDbaName: true,
+            nexusDbaName: dbaName,
+          }),
+        });
+        const navBarBusinessTitle = getNavBarBusinessTitle(business, IsAuthenticated.TRUE);
+        expect(navBarBusinessTitle).toEqual(dbaName);
+      });
+    });
   });
 
   describe("when name undefined, legal structure defined, and industry defined", () => {
--- a/web/src/lib/domain-logic/getNavBarBusinessTitle.ts
+++ b/web/src/lib/domain-logic/getNavBarBusinessTitle.ts
@@ -11,19 +11,36 @@ export const getNavBarBusinessTitle = (
     return Config.navigationDefaults.navBarGuestText;
   }
 
-  const { businessName, tradeName, legalStructureId, industryId, businessPersona, foreignBusinessType } =
-    business.profileData;
+  const {
+    businessName,
+    tradeName,
+    legalStructureId,
+    industryId,
+    businessPersona,
+    foreignBusinessType,
+    needsNexusDbaName,
+    nexusDbaName,
+  } = business.profileData;
 
   const isRemoteWorkerOrSeller = (): boolean => {
     return foreignBusinessType === "REMOTE_SELLER" || foreignBusinessType === "REMOTE_WORKER";
   };
 
   const determineName = (): string => {
+    if (needsNexusDbaName) {
+      if (nexusDbaName === "") {
+        return Config.navigationDefaults.navBarUnnamedDbaBusinessText;
+      }
+      if (nexusDbaName !== "") {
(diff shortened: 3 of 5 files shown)


web/src/lib/domain-logic/getNavBarBusinessTitle.ts
```diff
@@ -26,6 +34,15 @@ export const getNavBarBusinessTitle = (
...
   };
 
+  if (foreignBusinessType === "NEXUS") {
+    if (needsNexusDbaName && nexusDbaName === "") {
+      return Config.navigationDefaults.navBarUnnamedDbaBusinessText;
+    }
+    if (needsNexusDbaName && nexusDbaName !== "") {
+      return nexusDbaName;
+    }
+  }
```
adidner, line 44: Looking at the rest of this file it seems like there might be a way to better structure this, but I'm not totally sure what that looks like even after a bit of thought. Advice is appreciated.

web/src/lib/domain-logic/getNavBarBusinessTitle.test.ts
```diff
@@ -141,6 +141,38 @@ describe("getNavBarBusinessTitle", () => {
         });
       });
     });
+
+    describe("Nexus DBA", () => {
+      it("shows unamed dba when the user has an unavailable name", () => {
```
aloverso, line 146: [pebble] I suggest renaming this test to describe the behavior at play here. Although it is caused by an unavailable name, yes, the actual test and logic never looks at availablity at all.

More specifically, what causes it to show unnamed DBA? It is when the user needsNexusDbaName and the nexusDbaName is empty. We should say that.

also, typo "unnamed"

web/src/lib/domain-logic/getNavBarBusinessTitle.test.ts
```diff
@@ -141,6 +141,38 @@ describe("getNavBarBusinessTitle", () => {
...
+            businessName: "test business Name",
+            needsNexusDbaName: true,
+            nexusDbaName: "",
+          }),
+        });
+        const navBarBusinessTitle = getNavBarBusinessTitle(business, IsAuthenticated.TRUE);
+        expect(navBarBusinessTitle).toEqual(`${Config.navigationDefaults.navBarUnnamedDbaBusinessText}`);
+      });
+
+      it("shows the dba name when the user has a valid dba name", () => {
```
aloverso, line 160: [pebble] the word "valid" is doing some unspecified work here. what does "valid" mean?

If we mean "when the user has entered a DBA name" then we should just say that.

web/src/lib/domain-logic/getNavBarBusinessTitle.test.ts
```diff
@@ -141,6 +141,38 @@ describe("getNavBarBusinessTitle", () => {
...
+          profileData: generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            businessName: "test business Name",
+            needsNexusDbaName: true,
+            nexusDbaName: dbaName,
+          }),
+        });
+        const navBarBusinessTitle = getNavBarBusinessTitle(business, IsAuthenticated.TRUE);
+        expect(navBarBusinessTitle).toEqual(`${dbaName}`);
```
aloverso, line 173: [pebble] `\`${dbaName}\`` is redundant, you can just say `dbaName`

web/src/lib/domain-logic/getNavBarBusinessTitle.test.ts
```diff
@@ -141,6 +141,38 @@ describe("getNavBarBusinessTitle", () => {
...
+          profileData: generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            businessName: "test business Name",
+            needsNexusDbaName: true,
+            nexusDbaName: "",
+          }),
+        });
+        const navBarBusinessTitle = getNavBarBusinessTitle(business, IsAuthenticated.TRUE);
+        expect(navBarBusinessTitle).toEqual(`${Config.navigationDefaults.navBarUnnamedDbaBusinessText}`);
```
aloverso, line 157: [pebble] `` `${...}\` `` is redundant, you can just say the thing. It's already a string

content/src/fieldConfig/config.json
```diff
@@ -50,6 +50,7 @@
   "navigationDefaults": {
     "navBarUnnamedOwnedBusinessText": "Unnamed Business",
     "navBarUnnamedForeignRemoteSellerWorkerText": "Unnamed Out-of-State Business",
+    "navBarUnnamedDbaBusinessText": "Unnamed DBA",
```
aloverso, line 53: [sand] I know this was already decided, but I think it was decided kind of quickly without a lot of thought and should be confirmed (maybe a slack to the content channel? I can do it if you want)

I don't think this should say "Unnamed DBA" because it's grammatically incorrect and does not make sense. A DBA is not a type of business, like Nexus or something. It's not a noun, it's an adjective.

DBA stands for "Doing Business As" and is an adjective that describes a name - you **HAVE** a "DBA Name", you cannot **BE** "a DBA".

I think this content needs to be revised.

web/src/lib/domain-logic/getNavBarBusinessTitle.ts
aloverso: I think this new code should go in the "determineName" function.

The overall flow of logic is "if they have entered a name, use that (via some logic to figure out what to use). Otherwise use a string via some other logic"

This new code falls under the "they have entered a name, use that via some logic" bucket and should therefore go in `determineName`

jphechter, line 44: [dust] You shouldn't actually need the check `foreignBusinessType === "NEXUS"`. `needsNexusDbaName` being `true` implies that this is a Nexus business. You could restructure this to be:
```
if (needsNexusDbaName) {
  if ( nexusDbaName === "") {
  } else {
  }
}
```

jphechter: This is completely unrelated to the work you're doing with this PR.

Now that name availability is data stored and associated with a business object, `needsNexusDbaName` is somewhat redundant. I would make an argument that this information can be inferred by other data and keeping both increases the likelihood of inconsistencies in our data.

content/src/fieldConfig/config.json
adidner: I can do it

web/src/lib/domain-logic/getNavBarBusinessTitle.ts
adidner: Yeah, that makes sense, I was thinking it might be cool to somehow get this next the other "Foreign" user code but I don't think its worth it

adidner: Sure but isn't that more just something to take into account for a refactor that eliminates the needsNexusDbaName object?

If not do you have specific fields or structure for how you would want to infer that?

jphechter: Yeah, the 2nd part is just me thinking out loud via the comments section.

Like I said, we track business name availability. If the business name is unavailable, and you are a foreign business, then you would need a dba name. Because we can calculate this from these other pieces of information, I would argue it shouldn't be tracked as part of the object.

adidner: Gotcha, I'll write a follow up refactor for the notion listed above

PR #6249, opened 2023-08-18
feat: [#184662027] profile error alert

somabadri:
## Description

Added an error alert at the top of the profile page which displays the headers of the fields that need to be fixed.

### Ticket

[#184662027](https://www.pivotaltracker.com/story/show/184662027)

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/profile.json
+++ b/content/src/fieldConfig/profile.json
@@ -322,7 +322,7 @@
           "header": "Industry",
           "headerContextualInfo": "",
           "description": "",
-          "errorTextRequired": "Select a industry.",
+          "errorTextRequired": "Select an industry.",
           "searchAffirmation": "_\"**${searchText}**\" falls under the industries below:_"
         },
         "onboarding": {
@@ -392,6 +392,9 @@
       }
     },
     "default": {
+      "profileErrorAlert": "Complete ${fieldText} to access your dashboard:",
+      "profileErrorAlertOneField": "the field",
+      "profileErrorAlertMultipleFields": "these fields",
       "lockedFieldTooltipText": "This information cannot be changed because it was submitted to the NJ Department of Treasury.",
       "errorTextHeader": "Error",
       "backButtonText": "Cancel",
@@ -408,7 +411,6 @@
       "errorTextBody": "Review your information in the fields marked below.",
       "escapeModalHeader": "Are you sure you want to continue?",
       "profileTabNoteTitle": "Notes",
-      "essentialQuestionInlineText": "Make a selection by clicking on a button.",
       "saveButtonText": "Save",
       "escapeModalReturn": "Continue without saving",
       "successTextHeader": "Business Profile Saved",
--- a/web/src/components/onboarding/OnboardingEssentialQuestion.tsx
+++ b/web/src/components/onboarding/OnboardingEssentialQuestion.tsx
@@ -48,7 +48,7 @@ export const OnboardingEssentialQuestion = <T,>(props: Props<T>): ReactElement =
         />
         {isFormFieldInValid && (
           <div className="text-error-dark text-bold margin-top-05">
-            {Config.profileDefaults.default.essentialQuestionInlineText}
+            {Config.siteWideErrorMessages.errorRadioButton}
           </div>
         )}
       </WithErrorBar>
--- a/web/src/components/profile/ProfileErrorAlert.test.tsx
+++ b/web/src/components/profile/ProfileErrorAlert.test.tsx
@@ -1,0 +1,40 @@
+import { ProfileErrorAlert } from "@/components/profile/ProfileErrorAlert";
+import { getMergedConfig } from "@/contexts/configContext";
+import { templateEval } from "@/lib/utils/helpers";
+import { render, screen, within } from "@testing-library/react";
+
+const Config = getMergedConfig();
+
+describe("<ProfileErrorAlert/>", () => {
+  it("displays single field text in header if there is only one error", () => {
+    render(<ProfileErrorAlert fieldErrors={["industryId"]} />);
+    const profileAlert = screen.getByTestId("profile-error-alert");
+    const headerText = templateEval(Config.profileDefaults.default.profileErrorAlert, {
+      fieldText: Config.profileDefaults.default.profileErrorAlertOneField,
+    });
+    expect(within(profileAlert).getByText(headerText)).toBeInTheDocument();
+  });
+
+  it("displays multiple fields text in header if there is only one error", () => {
+    render(<ProfileErrorAlert fieldErrors={["industryId", "sectorId"]} />);
+    const profileAlert = screen.getByTestId("profile-error-alert");
+    const headerText = templateEval(Config.profileDefaults.default.profileErrorAlert, {
+      fieldText: Config.profileDefaults.default.profileErrorAlertMultipleFields,
+    });
+    expect(within(profileAlert).getByText(headerText)).toBeInTheDocument();
+  });
+
+  it("displays nothing if there are no errors", () => {
+    render(<ProfileErrorAlert fieldErrors={[]} />);
+    expect(screen.queryByTestId("profile-error-alert")).not.toBeInTheDocument();
+  });
+
+  it("displays the pet care header when willSellPetCareItems has an error", () => {
+    render(<ProfileErrorAlert fieldErrors={["willSellPetCareItems"]} />);
+    const profileAlert = screen.getByTestId("profile-error-alert");
+    expect(profileAlert).toBeInTheDocument();
+    expect(
+      within(profileAlert).getByText(Config.profileDefaults.fields.petCareHousing.default.header)
+    ).toBeInTheDocument();
+  });
+});
--- a/web/src/components/profile/ProfileErrorAlert.tsx
+++ b/web/src/components/profile/ProfileErrorAlert.tsx
@@ -1,0 +1,55 @@
+import { Alert } from "@/components/njwds-extended/Alert";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { getProfileConfig } from "@/lib/domain-logic/getProfileConfig";
+import { ProfileContentField } from "@/lib/types/types";
+import { templateEval } from "@/lib/utils/helpers";
+import { ReactElement } from "react";
+
+interface Props {
+  fieldErrors: string[];
+}
+
+export const ProfileErrorAlert = (props: Props): ReactElement | null => {
+  const { Config } = useConfig();
+
+  const displayProfileErrorAlert = (): boolean => props.fieldErrors.length > 0;
+
+  const getLabel = (field: string): string => {
+    return getProfileConfig({
+      config: Config,
+      fieldName: field as ProfileContentField,
+    }).header;
+  };
+
+  const errorFieldsIds = (): string[] => {
+    return props.fieldErrors.map((error) => {
+      if (error === "willSellPetCareItems" && !props.fieldErrors.includes("petCareHousing")) {
+        return "petCareHousing";
+      } else {
+        return error;
+      }
+    });
+  };
+
+  const profileErrorAlertText = (): string =>
+    errorFieldsIds().length === 1
+      ? templateEval(Config.profileDefaults.default.profileErrorAlert, {
+          fieldText: Config.profileDefaults.default.profileErrorAlertOneField,
+        })
+      : templateEval(Config.profileDefaults.default.profileErrorAlert, {
+          fieldText: Config.profileDefaults.default.profileErrorAlertMultipleFields,
+        });
+
+  if (!displayProfileErrorAlert()) return null;
+
+  return (
+    <Alert variant="error" dataTestid={"profile-error-alert"}>
+      <div>{profileErrorAlertText()}</div>
+      <ul>
+        {errorFieldsIds().map((id) => (
+          <li key={id}>{getLabel(id)}</li>
+        ))}
+      </ul>
+    </Alert>
+  );
+};
--- a/web/src/components/profile/ProfileField.tsx
+++ b/web/src/components/profile/ProfileField.tsx
@@ -1,5 +1,8 @@
 import { FieldLabelProfile } from "@/components/onboarding/FieldLabelProfile";
 import { LockedProfileField } from "@/components/onboarding/LockedProfileField";
+import { WithErrorBar } from "@/components/WithErrorBar";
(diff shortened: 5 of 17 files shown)


content/src/fieldConfig/profile.json
```diff
@@ -392,6 +392,9 @@
       }
     },
     "default": {
+      "profileErrorAlert": "Complete ${fieldText} to access your dashboard:",
+      "profileErrorAlertOneField": "the field",
+      "profileErrorAlterMultipleFields": "these fields",
```
jphechter, line 397:
```suggestion
      "profileErrorAlertMultipleFields": "these fields",
```
Spelling

web/src/components/profile/ProfileErrorAlert.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  it("displays multiple fields text in header if there is only one error", () => {
+    render(<ProfileErrorAlert fieldErrors={["industryId", "sectorId"]} />);
+    const profileAlert = screen.getByTestId("profile-error-alert");
+    const headerText = templateEval(Config.profileDefaults.default.profileErrorAlert, {
+      fieldText: Config.profileDefaults.default.profileErrorAlterMultipleFields,
+    });
+    expect(within(profileAlert).getByText(headerText)).toBeInTheDocument();
+  });
+});
```
jphechter, line 26: [pebble] Should there be a 3rd test in here to check that nothing renders when there are no errors?

web/src/components/profile/ProfileErrorAlert.tsx
```diff
@@ -0,0 +1,61 @@
...
+            <ul>
+              {errorFieldsIds().map((id) => (
+                <li key={id}>{getHeader(id)}</li>
+              ))}
+            </ul>
+          </Alert>
+        </>
+      )}
+    </>
+  );
```
jphechter, line 60: [dust] There's nothing wrong with this, but I probably would have written this return differently. The main reason being that It's more explicit that nothing gets displayed when `displayProfileErrorAlert()` returns `false`. This is just preference though.

There is an additional React fragment in here you can remove, but that's just me nit picking.
```suggestion
return displayProfileErrorAlert() ? (
  <Alert variant="error" dataTestid={"profile-error-alert"}>
    <div>{profileErrorAlertText()}</div>
    <ul>
      {errorFieldsIds().map((id) => (
        <li key={id}>{getHeader(id)}</li>
      ))}
    </ul>
  </Alert>
) : <></>
```

web/src/components/profile/ProfileErrorAlert.tsx
```diff
@@ -0,0 +1,61 @@
...
+      config: Config,
+      fieldName: field as ProfileContentField,
+    }).header;
+  };
+
+  const errorFieldsIds = (): string[] => {
+    const errorFieldsIds: string[] = [];
+    for (const error of props.fieldErrors) {
+      if (error === "willSellPetCareItems" && !errorFieldsIds.includes("petCareHousing")) {
+        errorFieldsIds.push("petCareHousing");
```
jphechter, line 28: I'm assuming `perCareHousing` is the only field where the error and the field name don't match?

web/src/lib/data-hooks/useFormContextHelper.ts
```diff
@@ -87,6 +88,16 @@ export const useFormContextHelper = <
       .filter((k) => (k as FieldStatus<FieldError>).invalid && (k as FieldStatus<FieldError>).updated)
       .flatMap((k) => (k as FieldStatus<FieldError>).errorTypes ?? []);
 
+  const getFieldErrors = (): string[] => {
+    const fieldErrors: string[] = [];
+    for (const field in fieldStates) {
+      if (fieldStates[field].invalid) {
+        fieldErrors.push(field);
+      }
```
jphechter, line 96: [dust] A `filter()` might be more readable.

web/src/components/profile/ProfileErrorAlert.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const getHeader = (field: string): string => {
+    return getProfileConfig({
+      config: Config,
+      fieldName: field as ProfileContentField,
+    }).header;
+  };
+
+  const errorFieldsIds = (): string[] => {
+    const errorFieldsIds: string[] = [];
+    for (const error of props.fieldErrors) {
```
jphechter, line 26: [dust] Same as other for of loop, this could be a filter

web/src/components/onboarding/OnboardingEssentialQuestion.tsx
```diff
@@ -48,7 +48,7 @@ export const OnboardingEssentialQuestion = <T,>(props: Props<T>): ReactElement =
         />
         {isFormFieldInValid && (
           <div className="text-error-dark text-bold margin-top-05">
-            {Config.profileDefaults.default.essentialQuestionInlineText}
```
FarazA22, line 51: [Pebble] We should delete this from the config.yml file

web/src/components/profile/ProfileErrorAlert.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface Props {
+  fieldErrors: string[];
+}
+
+export const ProfileErrorAlert = (props: Props): ReactElement => {
+  const { Config } = useConfig();
+
+  const displayProfileErrorAlert = (): boolean => props.fieldErrors.length > 0;
+
+  const getHeader = (field: string): string => {
```
FarazA22, line 17: [Dust] would a variable name like `getLabel` be better? We are not grabbing the header.

web/src/components/profile/ProfileErrorAlert.tsx
FarazA22: Separate thought, we had previously talked about how we should separate the UI logic from the conditional logic on whether it should display. I think we agreed it should be defined at the parent level. Therefore, `fieldErrors` does not need to be a prop.

web/src/components/profile/ProfileErrorAlert.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { ReactElement } from "react";
+
+interface Props {
+  fieldErrors: string[];
+}
+
+export const ProfileErrorAlert = (props: Props): ReactElement => {
+  const { Config } = useConfig();
+
+  const displayProfileErrorAlert = (): boolean => props.fieldErrors.length > 0;
```
FarazA22, line 15: I copied and pasted this comment in two places

We had previously talked about how we should separate the UI logic from the conditional logic on whether it should display. I think we agreed it should be defined at the parent level. Therefore, `fieldErrors` does not need to be a prop.

web/src/pages/profile.tsx
```diff
@@ -293,6 +295,7 @@ const ProfilePage = (props: Props): ReactElement => {
     info: (
       <>
         <ProfileTabHeader tab="info" />
+        <ProfileErrorAlert fieldErrors={getFieldErrors()} />
```
FarazA22, line 298: See the next line for an example on how we do it in other places

I copied and pasted this comment in three places

[Pebble] We had previously talked about how we should separate the UI logic from the conditional logic on whether it should display. I think we agreed it should be defined at the parent level. Therefore, fieldErrors does not need to be a prop.

web/src/pages/profile.tsx
```diff
@@ -403,6 +406,7 @@ const ProfilePage = (props: Props): ReactElement => {
...
```
FarazA22, line 409: Should be fixed for all instances

web/test/pages/profile/profile-starting.test.tsx
```diff
@@ -1066,25 +1066,6 @@ describe("profile - starting business", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("prevents user from saving when petcare is selected as industry, but essential question is not answered", async () => {
```
FarazA22, line 1069: [Dust] shouldn't we keep this test? the essential question is still required right?

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -400,4 +400,55 @@ describe("profile - shared", () => {
       ).not.toBeInTheDocument();
     });
   });
+
+  describe("profile error alert", () => {
```
FarazA22, line 404: [dust] I think it would be helpful to have a test for the error alert appearing across the tabs, and include `owning/ oscar`

web/src/components/profile/ProfileErrorAlert.tsx
aloverso: Why don't they match? This seems like a weird thing to have a special exception for, epsecially in a generic component like this. 

And if we really do need this, it should be tested

web/src/components/profile/ProfileErrorAlert.tsx
aloverso: I disagree, I'd rather see this as a guard clause

web/src/components/profile/ProfileField.tsx
```diff
@@ -16,28 +19,32 @@ interface Props {
...
-                isAltDescriptionDisplayed={props.displayAltDescription}
-                hideHeader={props.hideHeader}
-                boldAltDescription={props.boldAltDescription}
-              />
-            )}
-            <div className={props.noLabel ? "margin-bottom-05" : ""}>{props.children}</div>
-          </>
-        )}
-      </div>
+      <WithErrorBar hasError={isFormFieldInValid} type={"ALWAYS"}>
```
aloverso, line 29: [dust] Should this be wrapped at this level? It's not like the locked field could ever have an error, right. Maybe it should be wrapped at line 35 instead?

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -400,4 +400,55 @@ describe("profile - shared", () => {
...
+    it.each(["STARTING", "FOREIGN"])(
+      "displays alert with the header of field that has an error when %s",
```
aloverso, line 406: [pebble] I think the test name should include "industry" since that's what's being tested

web/src/components/profile/ProfileErrorAlert.tsx
aloverso: I see now it was tested in the profile-shared but if I'd argue it belongs in the unit test for this component

web/test/pages/profile/profile-starting.test.tsx
```diff
@@ -1066,25 +1066,6 @@ describe("profile - starting business", () => {
...
-    renderPage({ business, setRegistrationModalIsVisible });
-    clickSave();
-
-    await waitFor(() => {
-      expect(
-        screen.getByText(Config.profileDefaults.default.essentialQuestionInlineText)
-      ).toBeInTheDocument();
-    });
-  });
-
```
aloverso, line 1087: is this behavior still tested?

web/src/pages/profile.tsx
```diff
@@ -293,6 +295,7 @@ const ProfilePage = (props: Props): ReactElement => {
...
         {displayOpportunityAlert && <ProfileOpportunitiesAlert />}
```
FarazA22, line 299: @aloverso I have a question related to the comment immediately above and your response it - `I disagree, I'd rather see this as a guard clause`

If want to add guard clauses within the component, then shouldn't examples like `displayOpportunityAlert` be embedded within the component as well? That logic currently resides at the parent level (profile page).

``` 
const displayOpportunityAlert = LookupOperatingPhaseById(
    profileData.operatingPhase
  ).displayProfileOpportunityAlert;
...
{displayOpportunityAlert && <ProfileOpportunitiesAlert />}
```

In Soma's example we are passing the errors as a prop. The pattern above is one we use often. I'm down to change the approach, it's confusing to me because it seems inconsistent.

Profile Page
`<ProfileErrorAlert fieldErrors={getFieldErrors()} />`

Component
```
interface Props {
  fieldErrors: string[];
}

export const ProfileErrorAlert = (props: Props): ReactElement => {
...
const displayProfileErrorAlert = (): boolean => props.fieldErrors.length > 0;
...
     {displayProfileErrorAlert() && (
        <>
          <Alert variant="error" dataTestid={"profile-error-alert"}>
            <div>{profileErrorAlertText()}</div>
            <ul>
              {errorFieldsIds().map((id) => (
                <li key={id}>{getHeader(id)}</li>
              ))}
            </ul>
          </Alert>
        </>
      )}
...
```

web/src/components/profile/ProfileErrorAlert.tsx
somabadri: so basically only `petCareHousing` has the header field in the config because both the questions `willSellPetCareItems` and `petCareHousing` go under the same Petcare Services section. I will make it a unit test though.

web/test/pages/profile/profile-starting.test.tsx
somabadri: It's being tested on line 1109.

web/test/pages/profile/profile-starting.test.tsx
somabadri: It's being tested on line 1109.

web/src/components/profile/ProfileErrorAlert.tsx
aloverso:
```
if (!displayProfileErrorAlert()) return null;
return <Alert>...
```

web/src/lib/data-hooks/useFormContextHelper.ts
```diff
@@ -20,6 +20,7 @@ export const useFormContextHelper = <
   onTabChange: (tab: Tab) => void;
   isValid: () => boolean;
   getErrors: () => FieldError[];
+  getFieldErrors: () => string[];
```
FarazA22, line 23: Aren't these the same function?
```
  getErrors: () => FieldError[];
  getFieldErrors: () => string[];
```

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -400,4 +400,30 @@ describe("profile - shared", () => {
       ).not.toBeInTheDocument();
     });
   });
+
+  describe("profile error alert", () => {
+    it.each(["STARTING", "FOREIGN"])(
```
FarazA22, line 405: [sand] I think we can include the `owning` tests here as well. Can we create a refactor ticket for this?

web/src/lib/data-hooks/useFormContextHelper.ts
somabadri: addressing in a refactor

PR #6290, opened 2023-08-22
feat: [#185757071] EIN task updated to trigger modal on edit

somabadri:
## Description

Changed the EIN task to display registration modal when the field is edited or the save button is clicked.

### Ticket

[#185757071](https://www.pivotaltracker.com/story/show/185757071)

## Code author checklist

### DIFF
--- a/web/src/components/tasks/EinInput.tsx
+++ b/web/src/components/tasks/EinInput.tsx
@@ -1,5 +1,6 @@
 import { GenericTextField } from "@/components/GenericTextField";
 import { SecondaryButton } from "@/components/njwds-extended/SecondaryButton";
+import { AuthAlertContext } from "@/contexts/authAlertContext";
 import { IsAuthenticated } from "@/lib/auth/AuthContext";
 import { useConfig } from "@/lib/data-hooks/useConfig";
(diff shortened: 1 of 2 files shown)


web/src/components/tasks/EinInput.tsx
```diff
@@ -21,24 +22,29 @@ export const EinInput = (props: Props): ReactElement => {
...
+      return;
+    }
     setEmployerId(value);
   };
 
   const save = async (): Promise<void> => {
+    if (isAuthenticated === IsAuthenticated.FALSE) {
+      setRegistrationModalIsVisible(true);
+      return;
+    }
```
cbrunefearless, line 46: should we put this in a separate function since this is repeated above?

somabadri: I tried doing that but we need the return statement to exit the function so even after extraction it ends up being 
```
if(unauthenticated){
 functionToOpen();
 return
}
```
I could rewrite the logic so the it's an `if/else` but even then the function would still just be opening the modal? Do you have any suggestions how to reduce the repetition?

cbrunefearless: ok i am with anne

PR #6304, opened 2023-08-23
fix: [#184643298] fix errors showing when go to profile nudge is displayed

cbrunefearless:
## Description
Fix errors showing when go to profile nudge is displayed

### Ticket

[#184643298](https://www.pivotaltracker.com/story/show/)

## Code author checklist

### DIFF
--- a/api/src/domain/updateSidebarCards.ts
+++ b/api/src/domain/updateSidebarCards.ts
@@ -1,5 +1,5 @@
 import { getCurrentBusiness } from "@shared/domain-logic/getCurrentBusiness";
-import { isFieldAnswered, OPPORTUNITY_FIELDS } from "@shared/domain-logic/opportunityFields";
+import { getFieldsForProfile, isFieldAnswered } from "@shared/domain-logic/opportunityFields";
 import { LookupOperatingPhaseById } from "@shared/operatingPhase";
 import { modifyCurrentBusiness } from "@shared/test";
 import { UserData } from "@shared/userData";
@@ -52,9 +52,11 @@ export const updateSidebarCards: UpdateSidebarCards = (userData: UserData): User
   }
 
   if (LookupOperatingPhaseById(operatingPhase).displayGoToProfileNudge) {
-    const isEveryOpportunityFieldAnswered = OPPORTUNITY_FIELDS.every((field) => {
-      return isFieldAnswered(field, currentBusiness.profileData);
-    });
+    const isEveryOpportunityFieldAnswered = getFieldsForProfile(currentBusiness.profileData).every(
+      (field) => {
+        return isFieldAnswered(field, currentBusiness.profileData);
+      }
+    );
 
     if (isEveryOpportunityFieldAnswered) {
       hideCard("go-to-profile");
--- a/shared/src/domain-logic/opportunityFields.test.ts
+++ b/shared/src/domain-logic/opportunityFields.test.ts
@@ -1,0 +1,106 @@
+import { generateProfileData } from "../test";
+import { getFieldsForProfile } from "./opportunityFields";
+
+describe("opportunityFields", () => {
+  describe("getFieldsForProfile", () => {
+    describe("when starting a business", () => {
+      it("when legal structure is a trade name", async () => {
+        const startingBusiness = generateProfileData({
+          businessPersona: "STARTING",
+          legalStructureId: "general-partnership",
+        });
+
+        const profileFields = getFieldsForProfile(startingBusiness);
+        expect(profileFields).not.toContain("dateOfFormation");
+        expect(profileFields).toContain("municipality");
+        expect(profileFields).toContain("existingEmployees");
+        expect(profileFields).toContain("homeBasedBusiness");
+        expect(profileFields).toContain("ownershipTypeIds");
+      });
+
+      it("when legal structure is not a trade name", async () => {
+        const startingBusiness = generateProfileData({
+          businessPersona: "STARTING",
+          legalStructureId: "limited-partnership",
+        });
+
+        const profileFields = getFieldsForProfile(startingBusiness);
+        expect(profileFields).toContain("dateOfFormation");
+        expect(profileFields).toContain("municipality");
+        expect(profileFields).toContain("existingEmployees");
+        expect(profileFields).toContain("homeBasedBusiness");
+        expect(profileFields).toContain("ownershipTypeIds");
+      });
+    });
+
+    describe("when foreign nexus", () => {
+      describe("when location in NJ", () => {
+        it("when legal structure that is a trade name", async () => {
+          const foreignNexusBusiness = generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            nexusLocationInNewJersey: true,
+            legalStructureId: "general-partnership",
+          });
+
+          const profileFields = getFieldsForProfile(foreignNexusBusiness);
+          expect(profileFields).toContain("municipality");
+          expect(profileFields).not.toContain("dateOfFormation");
+          expect(profileFields).not.toContain("existingEmployees");
+          expect(profileFields).not.toContain("homeBasedBusiness");
+          expect(profileFields).not.toContain("ownershipTypeIds");
+        });
+
+        it("when legal structure that is not a trade name", async () => {
+          const foreignNexusBusiness = generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            nexusLocationInNewJersey: true,
+            legalStructureId: "limited-partnership",
+          });
+
+          const profileFields = getFieldsForProfile(foreignNexusBusiness);
+          expect(profileFields).toContain("dateOfFormation");
+          expect(profileFields).toContain("municipality");
+          expect(profileFields).not.toContain("existingEmployees");
+          expect(profileFields).not.toContain("homeBasedBusiness");
+          expect(profileFields).not.toContain("ownershipTypeIds");
+        });
+      });
+
+      describe("when location not in NJ", () => {
+        it("when legal structure that is a trade name", async () => {
+          const foreignNexusBusiness = generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            nexusLocationInNewJersey: false,
+            legalStructureId: "general-partnership",
+          });
+
+          const profileFields = getFieldsForProfile(foreignNexusBusiness);
+          expect(profileFields).toContain("homeBasedBusiness");
+          expect(profileFields).not.toContain("municipality");
+          expect(profileFields).not.toContain("dateOfFormation");
+          expect(profileFields).not.toContain("existingEmployees");
+          expect(profileFields).not.toContain("ownershipTypeIds");
+        });
+
+        it("when legal structure that is not a trade name", async () => {
+          const foreignNexusBusiness = generateProfileData({
+            businessPersona: "FOREIGN",
+            foreignBusinessType: "NEXUS",
+            nexusLocationInNewJersey: false,
+            legalStructureId: "limited-partnership",
+          });
+
+          const profileFields = getFieldsForProfile(foreignNexusBusiness);
+          expect(profileFields).toContain("dateOfFormation");
+          expect(profileFields).toContain("homeBasedBusiness");
+          expect(profileFields).not.toContain("municipality");
+          expect(profileFields).not.toContain("existingEmployees");
+          expect(profileFields).not.toContain("ownershipTypeIds");
+        });
+      });
+    });
+  });
+});
--- a/shared/src/domain-logic/opportunityFields.ts
+++ b/shared/src/domain-logic/opportunityFields.ts
@@ -1,18 +1,39 @@
 import { LookupLegalStructureById } from "../legalStructure";
 import { ProfileData } from "../profileData";
 
-export const OPPORTUNITY_FIELDS: (keyof ProfileData)[] = [
+const OPPORTUNITY_FIELDS: (keyof ProfileData)[] = [
   "existingEmployees",
   "municipality",
   "homeBasedBusiness",
   "ownershipTypeIds",
 ];
 
-export const REPORT_FIELDS: (keyof ProfileData)[] = ["dateOfFormation"];
+const REPORT_FIELDS: (keyof ProfileData)[] = ["dateOfFormation"];
 
-export const FIELDS_FOR_PROFILE = [...REPORT_FIELDS, ...OPPORTUNITY_FIELDS];
-export const getFieldsForProfile = (legalStructureId?: string): (keyof ProfileData)[] => {
-  return FIELDS_FOR_PROFILE.filter((field) => {
+const FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION: (keyof ProfileData)[] = ["municipality"];
+const FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION: (keyof ProfileData)[] = ["homeBasedBusiness"];
+const ALL_FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION = [...FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION, ...REPORT_FIELDS];
+const ALL_FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION = [
+  ...FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION,
+  ...REPORT_FIELDS,
+];
+const FIELDS_FOR_PROFILE = [...REPORT_FIELDS, ...OPPORTUNITY_FIELDS];
+
+export const getFieldsForProfile = (profileData: ProfileData): (keyof ProfileData)[] => {
+  if (profileData.businessPersona === "FOREIGN" && profileData.foreignBusinessType === "NEXUS") {
+    return profileData.nexusLocationInNewJersey
+      ? filterByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION, profileData.legalStructureId)
+      : filterByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION, profileData.legalStructureId);
+  } else {
+    return filterByLegalStructure(FIELDS_FOR_PROFILE, profileData.legalStructureId);
+  }
+};
+
+const filterByLegalStructure = (
+  profileKeys: (keyof ProfileData)[],
+  legalStructureId?: string
+): (keyof ProfileData)[] => {
+  return profileKeys.filter((field) => {
     if (field === "dateOfFormation" && legalStructureId) {
       return LookupLegalStructureById(legalStructureId).elementsToDisplay.has("formationDate");
     }
--- a/web/src/components/profile/ProfileOpportunitiesAlert.test.tsx
+++ b/web/src/components/profile/ProfileOpportunitiesAlert.test.tsx
@@ -1,0 +1,250 @@
+import { ProfileOpportunitiesAlert } from "@/components/profile/ProfileOpportunitiesAlert";
+import { getMergedConfig } from "@/contexts/configContext";
+import { WithStatefulProfileData } from "@/test/mock/withStatefulProfileData";
+import { setupStatefulUserDataContext } from "@/test/mock/withStatefulUserData";
+import {
+  createEmptyProfileData,
+  generateMunicipality,
+  generateProfileData,
+  ProfileData,
+} from "@businessnjgovnavigator/shared";
+import { render, screen, within } from "@testing-library/react";
+
+jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
+jest.mock("@/lib/data-hooks/useRoadmap", () => ({ useRoadmap: jest.fn() }));
+
+const Config = getMergedConfig();
+
+describe("<ProfileOpportunitiesAlert />", () => {
+  const renderComponent = (profileData?: ProfileData): void => {
+    render(
+      <WithStatefulProfileData initialData={profileData || createEmptyProfileData()}>
+        <ProfileOpportunitiesAlert />
+      </WithStatefulProfileData>
+    );
+  };
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    setupStatefulUserDataContext();
+  });
+
+  describe("when starting a business", () => {
+    it("when all opportunity questions are  answered", () => {
+      const profileData = generateProfileData({
+        businessPersona: "STARTING",
+        dateOfFormation: "2023-03-01",
+        municipality: generateMunicipality({}),
+        homeBasedBusiness: true,
+        ownershipTypeIds: ["none"],
+        existingEmployees: "1",
+        legalStructureId: "limited-partnership",
+      });
+      renderComponent(profileData);
+      expect(screen.queryByTestId("opp-alert")).not.toBeInTheDocument();
+    });
+
+    it("when all opportunity questions are not answered", () => {
+      const profileData = generateProfileData({
+        businessPersona: "STARTING",
+        dateOfFormation: undefined,
+        municipality: undefined,
+        homeBasedBusiness: undefined,
+        ownershipTypeIds: [],
+        existingEmployees: undefined,
+        legalStructureId: "limited-partnership",
+      });
+      renderComponent(profileData);
+      expect(
+        within(screen.getByTestId("opp-alert")).getByText(
+          Config.profileDefaults.fields.dateOfFormation.default.header
+        )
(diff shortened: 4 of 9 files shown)


web/test/pages/profile/profile-foreign.test.tsx
```diff
@@ -449,4 +455,140 @@ describe("profile-foreign", () => {
...
+      renderPage({ business });
+      expect(screen.getByTestId("opp-alert")).toBeInTheDocument();
+    });
+
+    it("lists each unanswered funding/certification question", () => {
+      const business = generateBusinessForProfile({
+        profileData: generateProfileData({
+          operatingPhase: "UP_AND_RUNNING",
+          businessPersona: "FOREIGN",
+          ...getProfileDataMixin("FOREIGN"),
```
aloverso, line 481: [pebble] I've never seen the word "mixin" in code outside of the Sass framework, so this one made me have to go look up the implementation to understand what was going on. Maybe we could name it something more descriptive like `getForeignNexusProfileFields`

web/test/pages/profile/profile-helpers.tsx
```diff
@@ -161,3 +164,20 @@ export const expectLocationNotSavedAndError = (): void => {
...
+export const getProfileDataMixin = (businessPersona: BusinessPersona): Partial<ProfileData> => {
+  return businessPersona === "FOREIGN"
+    ? {
+        foreignBusinessType: "NEXUS",
+        foreignBusinessTypeIds: ["NEXUS"],
+        nexusLocationInNewJersey: true,
+      }
+    : {};
+};
+export const phasesWhereGoToProfileShow = OperatingPhases.filter(
```
aloverso, line 177: [pebble] This is not grammatically correct, should be `Shows`

web/test/pages/profile/profile-helpers.tsx
```diff
@@ -161,3 +164,20 @@ export const expectLocationNotSavedAndError = (): void => {
...
+  (it) => it.displayProfileOpportunityAlert
+).map((it) => it.id);
+
+export const phasesWhereGoToProfileDoNotShow = OperatingPhases.filter(
```
aloverso, line 181: [pebble] This is not grammatically correct, should be DoesNotShow

web/test/pages/profile/profile-owning.test.tsx
```diff
@@ -477,150 +478,181 @@ describe("profile - owning existing business", () => {
...
-        }),
-      });
-      renderPage({ business });
-      expect(screen.getByTestId("opp-alert")).toHaveTextContent(
-        ProfileConfig.existingEmployees.overrides.OWNING.header
-      );
-      fillText("Existing employees", "12");
-      expect(screen.queryByTestId("opp-alert")).not.toBeInTheDocument();
-    });
+    it.each(phasesWhereGoToProfileShow)(
```
aloverso, line 484: [pebble] The use of `it.each` here (and in all these oscar tests) doesn't feel right because the phases where it shows includes UP_AND_RUNNING, which is never a phase that an Oscar can be in, because they are in UP_AND_RUNNING_OWNING specifically.

web/test/pages/profile/profile-owning.test.tsx
```diff
@@ -477,150 +478,181 @@ describe("profile - owning existing business", () => {
     });
   });
 
-  describe("profile opportunities alert", () => {
+  describe("profile opportunities alert for Oscar", () => {
```
aloverso, line 481: this is unnecessary because the test is `profile-owning`

web/test/pages/profile/profile-helpers.tsx
cbrunefearless: resolved

web/test/pages/profile/profile-foreign.test.tsx
cbrunefearless: Renamed

web/test/pages/profile/profile-starting.test.tsx
```diff
@@ -1329,6 +1329,98 @@ describe("profile - starting business", () => {
...
+          dateOfFormation: undefined,
+        }),
+      });
+      renderPage({ business });
+      expect(screen.getByTestId("opp-alert")).toBeInTheDocument();
+      chooseTab("numbers");
+      expect(screen.queryByTestId("opp-alert")).not.toBeInTheDocument();
+    });
+
+    it("lists each unanswered funding/certification question", () => {
```
aloverso, line 1349: [boulder] We have the same duplicative tests repeated three times over for each of these personas. It seems like everything would be a lot easier to read if we instead had a unit test for `ProfileOpportunitiesAlert` that described the specific behavior of the alert. It would have the following benefits:
- No need to copy-paste and duplicate the same tests across personas because one test describes the behavior
- When things DO differ by persona, like for Foreign, makes it much more obvious and easier to understand
- Simplifies the profile tests. The only tests that would need to remain in profile would be about which tab it shows on, and what happens to it when fields get interacted with, because all that is within a larger context than just the alert and its unit test

web/test/pages/profile/profile-owning.test.tsx
cbrunefearless: JUst setting that one phase UP_AND_RUNNING_OWNING for oscar

web/test/pages/profile/profile-owning.test.tsx
cbrunefearless: relabeled

web/test/pages/profile/profile-starting.test.tsx
somabadri, line 1349: [Dust] I understand after reading the test that funding/certification questions refer to `existing employees, homeBasedBusiness and ownershipType` but prior to that I assumed these were new questions being asked to the user and not just pre existing profile fields, so does it make more sense name the tests along the line of `lists existing employees, homeBasedBusiness, and ownershipType to determine fundings and certifications`, it's specific that we are expecting these exact fields to show up and the `it` block captures the behavior in it's entirety? @aloverso thoughts on this?

shared/src/domain-logic/opportunityFields.ts
```diff
@@ -42,3 +57,15 @@ export const isFieldAnswered = (field: keyof ProfileData, data: ProfileData): bo
...
 };
+
+const areForeignNexusProfileFieldsAnswered = (field: keyof ProfileData, data: ProfileData): boolean => {
+  if (field === "dateOfFormation") return !!data.dateOfFormation;
+
+  if (data.nexusLocationInNewJersey) {
+    if (field === "municipality") return data.municipality !== undefined;
+  } else {
+    if (field === "homeBasedBusiness") return data.homeBasedBusiness !== undefined;
+  }
```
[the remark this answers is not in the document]
cbrunefearless: Poppy and Dakota have the 5 fields we are checking, for Dakota Nexus with land in NJ, it has dateOfFormation and municipality. For Dakota Nexus with no land in NJ, it has dateOfFormation and homeBaseBusiness.  The ownershipIds and existing employees dont show for either.

web/test/pages/profile/profile-owning.test.tsx
```diff
@@ -480,10 +480,11 @@ describe("profile - owning existing business", () => {
   describe("profile opportunities alert", () => {
     const ProfileConfig = Config.profileDefaults.fields;
 
-    it("only displays alert on info tab", () => {
+    it("only displays alert on info tab when phase is %s", () => {
```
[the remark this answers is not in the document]
cbrunefearless: Took that out

web/test/pages/profile/profile-starting.test.tsx
cbrunefearless: done

shared/src/domain-logic/opportunityFields.ts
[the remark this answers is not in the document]
aloverso: Got it, I didn't realize that, and I think that learning this only underscores my point about the distinction between `isQuestionAnswered` and `getFieldsForProfile`.

This code is "working" in that the fields are showing up correctly, but only because the logic is saying that for a field like "existingEmployees" for nexus - it is in the array returned from `getFieldsForProfile`. And then when we map over that array to find out if each question is answered, it hits the default `return true` on line 56. Which is saying, logically: "this field exists in the Nexus profile but it has already been answered" which is two statements that are both untrue. It's creating the correct outcome in that it does not show in the alert, sure, but the logic that's causing that outcome is not correct.

web/test/pages/profile/profile-starting.test.tsx
cbrunefearless: taken out

web/test/pages/profile/profile-foreign.test.tsx
```diff
@@ -449,4 +455,140 @@ describe("profile-foreign", () => {
...
+          existingEmployees: undefined,
+          municipality: undefined,
+          homeBasedBusiness: undefined,
+        }),
+      });
+      renderPage({ business });
+      expect(screen.getByTestId("opp-alert")).toBeInTheDocument();
+    });
+
+    it("lists each unanswered funding/certification question when nexus location is not in NJ", () => {
```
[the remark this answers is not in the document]
cbrunefearless: moved to unit test

shared/src/domain-logic/opportunityFields.ts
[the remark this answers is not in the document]
cbrunefearless: got it

api/src/domain/updateSidebarCards.ts
```diff
@@ -52,9 +52,7 @@ export const updateSidebarCards: UpdateSidebarCards = (userData: UserData): User
   }
 
   if (LookupOperatingPhaseById(operatingPhase).displayGoToProfileNudge) {
-    const isEveryOpportunityFieldAnswered = OPPORTUNITY_FIELDS.every((field) => {
-      return isFieldAnswered(field, currentBusiness.profileData);
-    });
+    const isEveryOpportunityFieldAnswered = isEveryFundingFieldAnswered(currentBusiness.profileData);
```
aloverso, line 55: [pebble] Let's keep the names consistent here. In our domain model, a Funding and an Opportunity is not the same thing. Opportunity is the umbrella term for both Fundings and Certifications. And the rest of the logic here should make it clear that we're really handling Opportunity fields here, so this new function name should reflect that.

shared/src/domain-logic/opportunityFields.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        expect(profileFields).not.toContain("existingEmployees");
+        expect(profileFields).not.toContain("ownershipTypeIds");
+      }
+    );
+
+    it(
+      "returns only date of formation and municipality keys when it is a " +
+        "foreign business with a business type of nexus and a legal structure that is not a trade name and the location is in NJ",
+      async () => {
+        const foreignNexusBusinessWithoutLocationInNJ = generateProfileData({
```
aloverso, line 30: [pebble] Copy paste error? This says "without" location in NJ, but it's true

shared/src/domain-logic/opportunityFields.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        expect(profileFields).not.toContain("municipality");
+        expect(profileFields).not.toContain("dateOfFormation");
+        expect(profileFields).not.toContain("existingEmployees");
+        expect(profileFields).not.toContain("ownershipTypeIds");
+      }
+    );
+
+    it(
+      "returns municipality, existing employees, home based business, and ownership type id keys when it is " +
+        "a starting business and the legal structure id is general-partnership",
```
aloverso, line 88: [pebble] the other tests say Trade Name, which is accurate. I think saying "when legal structure is general-partnership" is too specific for the behavior we want to describe

shared/src/domain-logic/opportunityFields.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      "returns false when municipality, existing employees, home based business, and ownership type id are all undefined and it is  " +
+        "a starting business",
+      async () => {
+        const startingBusiness = generateProfileData({
+          businessPersona: "STARTING",
+          legalStructureId: "limited-partnership",
+          existingEmployees: undefined,
+          homeBasedBusiness: undefined,
+          ownershipTypeIds: undefined,
+          municipality: undefined,
```
aloverso, line 134: [pebble] Let's be careful with these. You should look at `emptyProfileData` to match what is the empty/default value for these fields. For example, `ownershipTypeIds` is an array and is never undefined - it starts out empty. I'm kind of surprised the compiler allows us to give it an undefined value.

shared/src/domain-logic/opportunityFields.test.ts
```diff
@@ -0,0 +1,240 @@
+import { generateMunicipality, generateProfileData } from "../test";
+import { getFieldsForProfile, isEveryFundingFieldAnswered } from "./opportunityFields";
+
+describe("opportunityFields", () => {
+  describe("getFieldsForProfile", () => {
+    it(
```
somabadri, line 6: [pebble] this test and the ones below that are related to `business type of nexus and a legal structure that is a trade name` can be grouped into a describe block to make it easier to read.  

``` 
describe("when business type is nexus and legal structure is a trade name ", () => {
it("returns only date of formation and home based business keys when location is not in NJ"){...}
....
})
```

shared/src/domain-logic/opportunityFields.test.ts
```diff
@@ -0,0 +1,240 @@
...
+          municipality: generateMunicipality({}),
+        });
+
+        const response = isEveryFundingFieldAnswered(startingBusiness);
+        expect(response).toEqual(false);
+      }
+    );
+
+    it(
+      "returns true when home based business is true and it is  " +
```
aloverso, line 221: [pebble] Again, the test name should describe the _behavior_ not the actual test instance (same feedback as the GP/trade name comment)

The _test_ has `homeBasedBusiness` set to true, yes, but the behavior under test is that it returns true if the `homeBasedBusiness` question is _answered_ (ie, either true OR false would be a valid value)

So we should say that - when it has been answered. Or we should test explicitly for the false case as well. Right now the way it's describing the test is leaving a gap in the understanding... we know what happens if it's undefined and if it's true, but what if it's false? The tests don't say. Whereas testing for if undefined and if not undefined does cover every possible behavior.

api/src/domain/updateSidebarCards.ts
aloverso: Overall, what's the reason for rewriting this logic in 2 different ways?

Here, you create a function called `isEveryFundingFieldAnswered` based on profileData.

In the `ProfileOpportunitiesAlert` we use two functions:
```
 getFieldsForProfile(state.profileData).filter((field) => {
    return isFieldAnswered(field, state.profileData);
  });
```

It seems like we're replicating the same logic in two different ways. And the `getFieldsForProfile` has some logical overlap with `isEveryFundingFieldAnswered` --- and `isEveryFundingFieldAnswered` isn't actually tested?

Can we just remove `isEveryFundingFieldAnswered` altogether and use the same logic in both places?

shared/src/domain-logic/opportunityFields.ts
```diff
@@ -10,16 +10,55 @@ export const OPPORTUNITY_FIELDS: (keyof ProfileData)[] = [
...
 };
 
+export const isEveryFundingFieldAnswered = (profileData: ProfileData): boolean => {
+  let fields;
+  if (profileData.businessPersona === "FOREIGN" && profileData.foreignBusinessType === "NEXUS") {
+    fields = profileData.nexusLocationInNewJersey
+      ? FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION
+      : FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION;
+  } else {
+    fields = OPPORTUNITY_FIELDS;
```
aloverso, line 54: [pebble[ On the same comment thread as in `updateSidebarCards` regarding removing this helper function... 

I think another option would be to keep this function if you want (abstracting away the .every call) but use `getFieldsForProfile` inside of here instead of redeclaring the logic?

I'm really confused why we're redeclaring how to define `fields` when we have this exact same logic already written 20 lines up in `getFieldsForProfile` - am I missing something? Please let me know.

It seems like this rewritten logic is also incomplete - like, wouldn't it return the wrong set of fields for a Poppy LLC? It would be missing the `dateOfFormation` field because this logic isn't a complete duplication

shared/src/domain-logic/opportunityFields.ts
```diff
@@ -10,16 +10,55 @@ export const OPPORTUNITY_FIELDS: (keyof ProfileData)[] = [
...
+    return profileData.nexusLocationInNewJersey
+      ? lookupByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION, profileData.legalStructureId)
+      : lookupByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION, profileData.legalStructureId);
+  } else {
+    return lookupByLegalStructure(FIELDS_FOR_PROFILE, profileData.legalStructureId);
+  }
+};
+
+const lookupByLegalStructure = (
+  key: (keyof ProfileData)[],
```
aloverso, line 36: [pebble] Variable naming conventions - if something is an array, it should always be named in plural. This isn't `key` it is `keys` - or even better, to be truly descriptive - (`keys` begs the question "what keys?") - you could go with `profileKeys`

web/src/components/profile/ProfileOpportunitiesAlert.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    );
+  };
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    setupStatefulUserDataContext();
+  });
+
+  it(
+    "does not display date of formation, home based business, ownership type id, existing employees," +
```
aloverso, line 33: [pebble] This test name does not match the outcome. 

1. It says "when those questions are not answered" but they are clearly all answered
2. It says it "does not display [fields]..." but the outcome is actually that it doesn't display the alert at all! This is a big piece of behavior and the test name should clearly reflect that.

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -290,7 +286,7 @@ describe("profile - shared", () => {
       expect(screen.getByTestId("opp-alert")).toBeInTheDocument();
     });
 
-    it.each(phasesWhereAlertFalse)("does not display alert for %s", (operatingPhase) => {
+    it.each(phasesWhereGoToProfileDoesNotShow)("does not display alert for %s", (operatingPhase) => {
```
aloverso, line 289: [pebble] I notice that we no longer have changes to the `profile-owning.test` file. I know there's a ton of tests for the profile alert component in there. Did they get moved to the `ProfileOpportunitiesAlert.test` file?

If so, the ones that are already covered by the new test file should be deleted. And the ones that are NOT covered by the new test file should remain... and maybe they should be relevant across all three personas? For example, "removes question from alert when it gets answered" is important behavior and that has to be tested at the profile level, not the alert level. So it has to stay... and maybe be moved to shared instead of owning because now that's shared behavior across all personas. Whereas "does not display alert if all funding/certification questions are answered" is clearly already covered in ProfileOpportunities alert (well, once the name updates) and should be deleted from the profile-owning test.

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -357,7 +343,7 @@ describe("profile - shared", () => {
               businessPersona: businessPersona as BusinessPersona,
               municipality: generateMunicipality({ displayName: "Trenton" }),
               industryId: "cannabis",
-              ...getProfileDataMixin(businessPersona as BusinessPersona),
+              ...getForeignNexusProfileFields(businessPersona as BusinessPersona),
```
aloverso, line 346: [pebble] We should avoid removing our type-safety by using `as`. If you misspelled "STARTING" or something invalid, this would swallow up and hide the valid compiler error.

You can instead declare types explicitly.

```
const personas: BusinessPersona[] = ["STARTING", "FOREIGN"]
it.each(personas)("....", 
  async (businessPersona: BusinessPersona) => {
```

With this, no `as` is necessary

aloverso:
As a note, I see three commits with the same name - these should be squashed into one commit before merging

shared/src/domain-logic/opportunityFields.test.ts
```diff
@@ -0,0 +1,240 @@
+import { generateMunicipality, generateProfileData } from "../test";
+import { getFieldsForProfile, isEveryFundingFieldAnswered } from "./opportunityFields";
+
+describe("opportunityFields", () => {
+  describe("getFieldsForProfile", () => {
+    it(
+      "returns only date of formation and home based business keys when it is a " +
+        "foreign business with a business type of nexus and a legal structure that is not a trade name and the location is not in NJ",
```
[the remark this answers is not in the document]
cbrunefearless: grouped together

shared/src/domain-logic/opportunityFields.test.ts
cbrunefearless: changed to foreignNexusBusiness for all of the,

shared/src/domain-logic/opportunityFields.test.ts
cbrunefearless: changed to trade name

shared/src/domain-logic/opportunityFields.test.ts
cbrunefearless: changed to an empty array

shared/src/domain-logic/opportunityFields.test.ts
cbrunefearless: Done

shared/src/domain-logic/opportunityFields.ts
cbrunefearless: So the previous logic had date of formation not in the array when displaying the goToProfileNude bc it was only using the OpportunityFields array.  I thought we wanted to keep that logic.

shared/src/domain-logic/opportunityFields.ts
cbrunefearless: changed to profileKeys

shared/src/domain-logic/opportunityFields.ts
cbrunefearless: That might of been a bug as well

shared/src/domain-logic/opportunityFields.ts
```diff
@@ -10,16 +10,55 @@ export const OPPORTUNITY_FIELDS: (keyof ProfileData)[] = [
...
+  if (profileData.businessPersona === "FOREIGN" && profileData.foreignBusinessType === "NEXUS") {
+    return profileData.nexusLocationInNewJersey
+      ? lookupByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION, profileData.legalStructureId)
+      : lookupByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION, profileData.legalStructureId);
+  } else {
+    return lookupByLegalStructure(FIELDS_FOR_PROFILE, profileData.legalStructureId);
+  }
+};
+
+const lookupByLegalStructure = (
```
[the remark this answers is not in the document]
cbrunefearless: changed to filterByLegalStructure

shared/src/domain-logic/opportunityFields.test.ts
cbrunefearless: yea i grouped them all together

web/test/pages/profile/profile-shared.test.tsx
cbrunefearless: I understand! Replaced it.

api/src/domain/updateSidebarCards.ts
cbrunefearless: done

shared/src/domain-logic/opportunityFields.ts
cbrunefearless: done

shared/src/domain-logic/opportunityFields.ts
```diff
@@ -10,9 +10,33 @@ export const OPPORTUNITY_FIELDS: (keyof ProfileData)[] = [
...
+  if (profileData.businessPersona === "FOREIGN" && profileData.foreignBusinessType === "NEXUS") {
+    return profileData.nexusLocationInNewJersey
+      ? filterByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NJ_LOCATION, profileData.legalStructureId)
+      : filterByLegalStructure(ALL_FIELDS_FOR_FOREIGN_NEXUS_NOT_NJ_LOCATION, profileData.legalStructureId);
+  } else {
+    return filterByLegalStructure(FIELDS_FOR_PROFILE, profileData.legalStructureId);
+  }
+};
+
+export const filterByLegalStructure = (
```
aloverso, line 35: [sand] I think most of these functions & variables are only used in this file, right? For anything that is intended only for use in this file, which I think would include this function and most (all?) of the constant variables definitions, they should not include the `export` keyword

PR #6305, opened 2023-08-24
fix: [#185902802] wrote migration to deal with deleted successful-reg…

adidner:
…istration card

## Description

So we were getting an error on dev, 
image

Given this context I tracked down this https://github.com/newjersey/navigator.business.nj.gov/pull/6236/files
and https://www.pivotaltracker.com/n/projects/2573455/stories/185681589

and saw we had not written a migration to remove this card that no longer exists from user's data. (As on the prod DB we still have users with this card in the visible cards)

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185902802)

## Code author checklist

### DIFF
--- a/api/src/db/migrations/v123_remove_account_creation_sidecards.test.ts
+++ b/api/src/db/migrations/v123_remove_account_creation_sidecards.test.ts
@@ -1,0 +1,20 @@
+import {
+  generateV122Business,
+  generateV122Preferences,
+  generateV122UserData,
+} from "./v122_remove_welcome_and_welcomeupandrunning_sidecards";
+import { migrate_v122_to_v123 } from "./v123_remove_account_creation_sidecards";
+
+describe("migrate_v122_to_v123", () => {
+  it("removes the successful-registration card from sidebarCards", () => {
+    const id = "biz-1";
+    const v122Preferences = generateV122Preferences({
+      visibleSidebarCards: ["other-card", "successful-registration"],
+    });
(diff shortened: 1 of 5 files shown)


jphechter (approved):

api/src/db/migrations/v123_remove_account_creation_sidecards.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  generateV122Preferences,
+  generateV122UserData,
+} from "./v122_remove_welcome_and_welcomeupandrunning_sidecards";
+import { migrate_v122_to_v123 } from "./v123_remove_account_creation_sidecards";
+
+describe("migrate_v122_to_v123", () => {
+  it("removes the successful-registration card from sidebarCards", () => {
+    const id = "biz-1";
+    const v122Preferences = generateV122Preferences({
+      visibleSidebarCards: ["welcome", "welcome-up-and-running", "other-card", "successful-registration"],
```
aloverso, line 12: [pebble] while this technically works / passes as a unit test, I think it's kind of confusing because v122 was to remove "welcome" and "welcome-up-and-running" cards, so a user going into this function should never have those cards in their userData, and it feels weird/misleading to include them in the test setup. We could test pretty much anything else.

adidner: Ah, you are totally right. I will fix right away

PR #6310, opened 2023-08-28
refactor: [#184779281] prefactor to update auth context user

aloverso:
## Description

We were previously tracking `user` inside the AuthState, which represented the user that was signed in to Cognito, but as a full `BusinessUser` with all the BusinessUser things on it, like tracking if that user was signed up for the newsletter and so on. This was inaccurate because none of that information was correct. 

Before this refactor, there would be a `state.user` from the Auth Context, and a `userData.user` from the userData context, and they would both be of type `BusinessUser` and have the same ID, but half the fields were just wildly incorrect and did not actually have matching values between the two "identical" objects.

This refactor removes that issue by making the `user` inside the Auth state represent the values that we get from Cognito and nothing more.

This will make it easier to finish the myNJ existing user story because that story will add a new value to `ActiveUser` in AuthState that does NOT belong on `BusinessUser`, so separating these types is an important pre-factor.

### Ticket

## Code author checklist

### DIFF
--- a/web/src/components/navbar/NavBar.test.tsx
+++ b/web/src/components/navbar/NavBar.test.tsx
@@ -390,7 +390,7 @@ describe("<NavBar />", () => {
           <WithStatefulUserData initialUserData={userData}>
             <NavBar landingPage={false} showSidebar={false} />
           </WithStatefulUserData>,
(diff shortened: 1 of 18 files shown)


cbrunefearless (approved):
good refactor!! I like it...

PR #6311, opened 2023-08-28
Add editable display name field

adidner:
## Description

So the context here is we wanted to do a bunch of things, specifically we wanted to create a read-only field to be used for ID and filename

We also wanted to create a field called displayname to be used by the content team for the purposes of renaming and searching for things at their leisure without screwing up the actual filenames of files or webflow sync. 

This ended up getting extremely complicated and tedious as this change means I needed to go and add the displayname field to all of these files and if the filename field and actual filename didn't match because content changed the filename field we needed that modified filename field to become the new displayname field

Also making sure this worked everywhere was a bit of a hassel

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/184890110)

## Code author checklist


content/add_displayname_to_json_files.py
```diff
@@ -0,0 +1,25 @@
+import os
+import json
```
adidner, line 2: These are some of the script I wrote to help add fields to .md and .json files, as well as tell when files had different filename fields an actual filenames. 

Are these worth keeping and if so where should they live?

content/add_displayname_to_json_files.py
```diff
@@ -0,0 +1,25 @@
...
+                    file.write(updated_content)
+                    file.truncate()
+                    print(f"Added 'displayname' to {filename}")
+                except json.JSONDecodeError:
+                    print(f"Error decoding JSON in {filename}")
+
+# Specify the directory containing the JSON files
+json_directory = "./src/roadmaps/license-tasks"
+
+add_displayname_to_json_files(json_directory)
```
adidner, line 25: Unrelated to this code but here for visability , I did not add to the licenseSync.mjs file as I could add fields to the object that we're sending to webflow but webflow wouldn't use them so I don't think there's a point.

adidner:
Note I plan on squashing this before I merge it

content/add_displayname_to_json_files.py
aloverso: I'm neutral on keeping the scripts, but sure, might as well. You could put them in a `content/scripts' directory

PR #6314, opened 2023-08-29
feat: [#184779281] show modified content for users who encounter existing myNJ account issue

aloverso:
## Description

- Uses new browser storage component to remember if the user has encountered this error across pages of the app

### Ticket

[#184779281](https://www.pivotaltracker.com/story/show/#184779281)

## Code author checklist


api/src/domain/updateSidebarCards.test.ts
```diff
@@ -34,14 +34,35 @@ describe("updateRoadmapSidebarCards", () => {
...
+
+      const updatedUserData = updateSidebarCards(userData);
+      expect(getCurrentBusiness(updatedUserData).preferences.visibleSidebarCards).not.toContain(
+        "not-registered-existing-account"
+      );
+      expect(getCurrentBusiness(updatedUserData).preferences.visibleSidebarCards).toContain(
+        "formation-nudge"
+      );
+    });
+
```
jphechter, line 57: [dust] Do we need this test given on line 80 we check for this exact logic, and that no other cards are modified? I understand wanting to be explicit about the behavior, but these seems like one test could cover everything here.

aloverso: It's the same test but for the new different card. How would you suggest covering it all in one test?

jphechter: The only difference is that the second test has an additional card `other-card`. I'm saying the coverage is already there. In both cases we're validating that `not-registered-existing-account` is replaced with `formation-nudge`. Unless I'm missing something we already are covering it all in one test.

aloverso: Oh wait I looked at the wrong line. That's true but the reason for separating was to describe the 2 different pieces of behavior. I think the test name would be really long and wordy if it was all one test

jphechter: Totally fair.

web/test/pages/loading.test.tsx
```diff
@@ -105,18 +86,19 @@ describe("loading page", () => {
     expect(mockPush).toHaveBeenCalledWith("/tasks/some-task");
   });
 
-  it("shows modal when user has signin error and redirects user to onboarding", async () => {
+  it("saves triggers guest signin with encounteredError param when user has signin error", async () => {
```
jphechter, line 89: I'm leaving this note to come back and look at this again. Something about this test name is weird to me, but I think I need to look again.

web/src/lib/auth/signinHelper.ts
```diff
@@ -80,30 +81,40 @@ export const onSelfRegister = (
 export const onGuestSignIn = async (
   push: (url: string) => Promise<boolean>,
   pathname: string,
-  dispatch: Dispatch<AuthAction>
+  dispatch: Dispatch<AuthAction>,
+  params?: { encounteredMyNjLinkingError: boolean }
```
jphechter, line 85: [sand] Should all of these positional params be included in the new `params` object? It feels weird to me to have competing styles here.

aloverso: I could go either way. The reason for this style is following the "config" pattern that a lot of JS libraries use. Like for example axios, you do 
```
axios.post("/endpoint", data, { configKey: "value" }
```

Where the last argument is an optional object of additional config parameters, but the primary parameters to the function get their top-level importance.

web/src/components/auth/SignUpSnackbar.test.tsx
```diff
@@ -30,64 +40,143 @@ const setLargeScreen = (value: boolean): void => {
...
+          isAuthenticated,
+          {
+            registrationAlertIsVisible: true,
+            setRegistrationAlertIsVisible,
+          }
+        ),
+        { activeUser, isAuthenticated }
+      )
     );
   };
```
jphechter, line 95: These 2 functions seem pretty similar. I don't really mind them being distinct given this is a test file and I personally like being more verbose when it comes to tests, but I feel like there is enough overlap here that we could combine if we wanted.

web/src/lib/auth/sessionHelper.ts
```diff
@@ -108,16 +109,21 @@ export const getActiveUser = async (): Promise<ActiveUser> => {
...
-  return cognitoPayloadToActiveUser(cognitoPayload);
+  const encounteredMyNjLinkingError = AccountLinkingErrorStorageFactory().getEncounteredMyNjLinkingError();
+  return cognitoPayloadToActiveUser(cognitoPayload, { encounteredMyNjLinkingError });
 };
 
-const cognitoPayloadToActiveUser = (cognitoPayload: CognitoIdPayload): ActiveUser => {
+const cognitoPayloadToActiveUser = (
+  cognitoPayload: CognitoIdPayload,
+  params: { encounteredMyNjLinkingError: boolean | undefined }
+): ActiveUser => {
```
jphechter, line 119: [sand] I said this somewhere else, too. Should all of these positional params be included in the new params object? It feels weird to me to have competing styles here.

jphechter (approved):

PR #6317, opened 2023-08-29
fix: [#185834867] make sector required for Oscar personas

AnJuHyppolite:
## Description

Make sector selection required for Oscar personas.
### Ticket

[185834867](https://www.pivotaltracker.com/story/show/185834867)

### Approach

Users start out in GUEST_MODE, when sector selection is not a requirement. I updated the code such that when a user is an OWNING persona, then they will be in GUEST_MODE_OWNING: the point at which sector selection is a requirement.

## Code author checklist

### DIFF
--- a/web/src/components/onboarding/OnboardingBusinessPersona.tsx
+++ b/web/src/components/onboarding/OnboardingBusinessPersona.tsx
@@ -25,6 +25,8 @@ export const OnboardingBusinessPersona = <T,>(props: FormContextFieldProps<T>): 
     Validate(false);
     setProfileData({
       ...state.profileData,
+      operatingPhase:
+        (event.target.value as BusinessPersona) === "OWNING" ? "GUEST_MODE_OWNING" : "GUEST_MODE",
       businessPersona: event.target.value as BusinessPersona,
     });
   };
--- a/web/test/pages/onboarding/onboarding-owning.test.tsx
+++ b/web/test/pages/onboarding/onboarding-owning.test.tsx
@@ -17,7 +17,7 @@ import {
   generateUserDataForBusiness,
 } from "@businessnjgovnavigator/shared/test";
 import { UserData } from "@businessnjgovnavigator/shared/userData";
-import { screen, waitFor, within } from "@testing-library/react";
+import { fireEvent, screen, waitFor, within } from "@testing-library/react";
 
 jest.mock("next/router", () => ({ useRouter: jest.fn() }));
 jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
@@ -68,18 +68,17 @@ describe("onboarding - owning a business", () => {
       expect(screen.getByLabelText("Sector")).toBeInTheDocument();
     });
 
-    it("prevents user from moving after Step 1 if you have not entered a sector", async () => {
-      const userData = generateTestUserData({ sectorId: undefined });
-      useMockRouter({ isReady: true, query: { page: "1" } });
-      const { page } = renderPage({ userData });
-      page.clickNext();
+    it("does not allow OWNING user persona to move past Step 1 if user has not entered a sector", async () => {
+      const { page } = renderPage({ userData: undefined });
+      page.chooseRadio("business-persona-owning");
+      fireEvent.click(screen.getByTestId("next"));
       await waitFor(() => {
-        expect(screen.getByTestId("step-1")).toBeInTheDocument();
+        expect(
+          screen.getByText(Config.profileDefaults.fields.sectorId.default.errorTextRequired)
+        ).toBeInTheDocument();
       });
-      expect(
-        screen.getByText(Config.profileDefaults.fields.sectorId.default.errorTextRequired)
-      ).toBeInTheDocument();
       expect(screen.getByTestId("banner-alert-REQUIRED_REVIEW_INFO_BELOW")).toBeInTheDocument();
+      expect(screen.getByTestId("step-1")).toBeInTheDocument();
     });
 
     it("allows user to move past Step 1 if you have entered a sector", async () => {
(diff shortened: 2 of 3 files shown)


web/src/components/onboarding/OnboardingBusinessPersona.tsx
```diff
@@ -25,6 +25,10 @@ export const OnboardingBusinessPersona = <T,>(props: FormContextFieldProps<T>):
     Validate(false);
     setProfileData({
       ...state.profileData,
+      operatingPhase:
+        (event.target.value as BusinessPersona) === "OWNING"
+          ? "GUEST_MODE_OWNING"
+          : state.profileData.operatingPhase,
```
jphechter, line 31: [boulder] I think we need additional logic here. Right now, once the user indicates that they have an "up and running" business, they are stuck with a BusinessPersona `GUEST_MODE_OWNING`.

I know we paired on this yesterday, but there's key logic here that we over looked.

I'm logging the persona in this screenshot below. I clicked "starting a business" -> "up and running" -> "starting a business".
Screenshot 2023-08-29 at 3 48 42 PM

jphechter: Additionally, while the end result is that we require the user to set a sector, because we're conditionally setting the `BusinessPersona` we want a test to validate that as well.

web/test/pages/onboarding/onboarding-owning.test.tsx
```diff
@@ -82,6 +82,15 @@ describe("onboarding - owning a business", () => {
       expect(screen.getByTestId("banner-alert-REQUIRED_REVIEW_INFO_BELOW")).toBeInTheDocument();
```
adidner, line 82: So the test above this is "prevents user from moving after Step 1 if you have not entered a sector" but this isn't accurate anymore and should only apply to Guest Oscars right? So this test should be deleted I believe?

jphechter: This test definitely should have been failing before. If I remember correctly, seeing that this existed was the reason we wrote this bug.

This might be able to be deleted, but it highlights that we should be distinctly covering for both guest and authenticated users.

adidner: Okay, so I was thinking that but if you're authenticated don't you already have had to go through onboarding? Like isn't the onboarding only available for guests in the first place? @jphechter

Or is that not the case

somabadri (approved):
paired with you on it and looked through it again, looks good to me!

web/test/pages/onboarding/onboarding-owning.test.tsx
```diff
@@ -68,17 +68,15 @@ describe("onboarding - owning a business", () => {
...
       await waitFor(() => {
-        expect(screen.getByTestId("step-1")).toBeInTheDocument();
+        expect(
+          screen.getByText(Config.profileDefaults.fields.sectorId.default.errorTextRequired)
+        ).toBeInTheDocument();
       });
-      expect(
-        screen.getByText(Config.profileDefaults.fields.sectorId.default.errorTextRequired)
-      ).toBeInTheDocument();
       expect(screen.getByTestId("banner-alert-REQUIRED_REVIEW_INFO_BELOW")).toBeInTheDocument();
```
adidner, line 80: [dust] might be cool to some how check that we're still on the page at the end (right now we're only checking that there is an error so their might be a weird case where we move pages and show an error) but I get it if you think this is overkill.

somabadri: oh that's a good idea

web/test/pages/onboarding/onboarding-shared.test.tsx
```diff
@@ -43,6 +43,19 @@ describe("onboarding - shared", () => {
...
+
+      page.chooseRadio(radioOption);
+      fireEvent.click(screen.getByTestId("next"));
+      await waitFor(() => {
+        expect(screen.getByTestId("step-2")).toBeInTheDocument();
+      });
+    }
+  );
+
   it("routes to the second onboarding page when they have answered the first question and we route them to page 2", async () => {
```
adidner, line 59: [sand] I don't think this needs to be a part of this story (and if people think its not important I'm down to drop it) but it might be a good idea to have a test for the case JP was talking about to make sure that if we use the user data is reset and personas are updated accordingly if you switch between them on this screen. 

Example if I set myself as Oscar and choose a sector and then switch to poppy the sector I chose should probably be reset

adidner (approved):
Approved, some ideas for making tests a bit more robust but nothing vital

web/test/pages/onboarding/onboarding-owning.test.tsx
AnJuHyppolite: @adidner, can you share more about what you mean when you write, "weird case where we move pages and show an error"? There are no other required fields for the OWNING persona in onboarding, and once the user selects the OWNING person and a sector, they are directed to the dashboard.

somabadri: @AnJuHyppolite I think @adidner  is saying that we could have another expect statement like: 
`expect(screen.getByTestId("step-2")).toBeInTheDocument();`
to confirm that the error stops the user from navigating forward

adidner: Yeah, 

So imagine a situation where the user, selects oscar, doesn't enter a sector, hits the next button and the app goes to the next onboarding page and then pulls up the error. 

This would be wrong as we don't want them to move past the first page without a sector but the condition I described will get your test to pass. 

The test title says they cannot move past page 1 but you don't actually check that we're still on page 1, you're just checking for an error popup, so I'm just saying it might be cool to have 1 more expect to confirm we are still on page 1. 

Does that help clarify things?

PR #6320, opened 2023-08-30
fix: [#185940288] allow Outside of the USA selection for foreign state of formation

aloverso:
### Ticket

[185940288](https://www.pivotaltracker.com/story/show/#185940288)

## Code author checklist

### DIFF
--- a/web/src/components/StateDropdown.tsx
+++ b/web/src/components/StateDropdown.tsx
@@ -14,6 +14,7 @@ interface Props {
   validationText?: string;
   excludeNJ?: boolean;
   excludeTerritories?: boolean;
+  includeOutsideUSA?: boolean;
   useFullName?: boolean;
   validationLabel?: string;
   autoComplete?: boolean;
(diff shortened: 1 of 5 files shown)


web/src/components/StateDropdown.tsx
```diff
@@ -14,6 +14,7 @@ interface Props {
   validationText?: string;
   excludeNJ?: boolean;
   excludeTerritories?: boolean;
+  includeOutsideUSA?: boolean;
   useFullName?: boolean;
```
cbrunefearless, line 18: should we make it consistent for the file and it be excludeOutsideUSA?  I mean either way is fine, its just a question and not a hill for me to die on or even stand on.

aloverso: @cbrunefearless The reason for the include/exclude is to set a reasonable default. The default when passing no props should be: to include NJ and territories, and to NOT include "Outside the USA"

I think it would be weird behavior for the state dropdown to include "Outside the USA" by default and have to explicitly specify for it to be excluded. This behavior is unusual and should be the exception, not the rule

cbrunefearless (approved):
lgtm

jphechter (approved):

PR #6323, opened 2023-08-31
Add formation task error alert link

jphechter:
## Description

- Adds links to the error alert in business formation driving to the fields lower on the page.
- Makes all `usa-links` blue

### Ticket

[#184432488](https://www.pivotaltracker.com/story/show/184432488)
[#184432499](https://www.pivotaltracker.com/story/show/184432499)

### Approach

Componentize the alert used on profile to highlight opportunity fields, and implement in both the profile and in formation task.

### Steps to Test

Testing this in the UI requires triggering errors in formation and then clicking the generated links.

## Code author checklist

### DIFF
--- a/web/src/components/FieldEntryAlert.tsx
+++ b/web/src/components/FieldEntryAlert.tsx
@@ -1,0 +1,30 @@
+import { Content } from "@/components/Content";
+import { Alert, AlertVariant } from "@/components/njwds-extended/Alert";
+import { ReactElement } from "react";
+
+interface Props {
+  alertMessage: string;
+  fields: {
+    name: string;
+    label: string;
+  }[];
+  testId?: string;
+  variant: AlertVariant;
+}
+
+export const FieldEntryAlert = (props: Props): ReactElement => {
+  if (props.fields.length === 0) return <></>;
+
+  return (
+    <Alert variant={props.variant} dataTestid={props.testId}>
+      <Content>{props.alertMessage}</Content>
+      <ul>
+        {props.fields.map((field) => (
+          <li key={field.name} data-testid={`question-${field.name}-alert-text`}>
+            <a href={`#question-${field.name}`}>{field.label}</a>
+          </li>
+        ))}
+      </ul>
+    </Alert>
+  );
+};
--- a/web/src/components/FieldEntryAlert.tsx
+++ b/web/src/components/FieldEntryAlert.tsx
@@ -1,12 +1,13 @@
 import { Content } from "@/components/Content";
 import { Alert, AlertVariant } from "@/components/njwds-extended/Alert";
-import { ReactElement } from "react";
+import { ReactElement, ReactNode } from "react";
 
 interface Props {
   alertMessage: string;
   fields: {
     name: string;
     label: string;
+    children?: ReactNode;
   }[];
   testId?: string;
   variant: AlertVariant;
@@ -22,6 +23,7 @@ export const FieldEntryAlert = (props: Props): ReactElement => {
         {props.fields.map((field) => (
           <li key={field.name} data-testid={`question-${field.name}-alert-text`}>
             <a href={`#question-${field.name}`}>{field.label}</a>
+            {field.children}
           </li>
         ))}
       </ul>
--- a/web/src/components/FieldEntryAlert.tsx
+++ b/web/src/components/FieldEntryAlert.tsx
@@ -1,23 +1,22 @@
 import { Content } from "@/components/Content";
-import { Alert, AlertVariant } from "@/components/njwds-extended/Alert";
+import { Alert, AlertProps } from "@/components/njwds-extended/Alert";
 import { ReactElement, ReactNode } from "react";
 
 interface Props {
   alertMessage: string;
+  alertProps: AlertProps;
   fields: {
     name: string;
     label: string;
     children?: ReactNode;
   }[];
-  testId?: string;
-  variant: AlertVariant;
 }
 
 export const FieldEntryAlert = (props: Props): ReactElement => {
   if (props.fields.length === 0) return <></>;
 
   return (
-    <Alert variant={props.variant} dataTestid={props.testId}>
+    <Alert {...props.alertProps}>
       <Content>{props.alertMessage}</Content>
       <ul>
         {props.fields.map((field) => (
--- a/web/src/components/njwds-extended/Alert.tsx
+++ b/web/src/components/njwds-extended/Alert.tsx
@@ -1,8 +1,7 @@
 import { ReactElement, ReactNode } from "react";
 
-interface Props {
+export interface AlertProps {
   variant: AlertVariant;
-  children: ReactNode;
   noIcon?: boolean;
   heading?: string;
   rounded?: boolean;
@@ -10,6 +9,10 @@ interface Props {
   className?: string;
   borderRight?: boolean;
   borderSmall?: boolean;
+}
+
+interface Props extends AlertProps {
+  children: ReactNode;
 }
 
 export const AlertVariants = ["info", "success", "warning", "error", "note"] as const;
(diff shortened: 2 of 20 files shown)


web/src/components/tasks/business-formation/billing/BillingStep.tsx
```diff
@@ -23,37 +24,43 @@ export const BillingStep = (): ReactElement => {
...
         <div className="grid-row grid-gap-1">
           <div className="margin-top-2 tablet:grid-col-6">
-            <BusinessFormationTextField
-              label={Config.formation.fields.contactFirstName.label}
-              fieldName="contactFirstName"
-              errorBarType="MOBILE-ONLY"
-              required={true}
-              validationText={getFieldErrorLabel("contactFirstName")}
-            />
+            <FormationField fieldName="contactFirstName">
```
jphechter, line 27: Most of this PR is whitespace changes due to wrapping fields in this new `FormationField` component.

web/src/components/njwds-extended/Alert.tsx
```diff
@@ -12,7 +12,9 @@ interface Props {
   borderSmall?: boolean;
 }
 
-export type AlertVariant = "info" | "success" | "warning" | "error" | "note";
+export const AlertVariants = ["info", "success", "warning", "error", "note"];
+
+export type AlertVariant = (typeof AlertVariants)[number];
```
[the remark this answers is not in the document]
FarazA22: [pebble] When using the IDE to help me identify the types, I get an unexpected response. I don't think the `initial type: string` is right.  

Screenshot 2023-09-01 at 10 04 50 AM
Screenshot 2023-09-01 at 10 05 31 AM

web/src/styles/base/global.scss
```diff
@@ -87,7 +87,7 @@ ol {
 
 .usa-link:visited,
 .usa-prose > a:visited {
-  color: $primary-darker !important;
+  color: $accent-cool-darkest !important;
```
FarazA22, line 90: [pebble] In storybook, the colors are rendered with the lightest colors on the left, and the shades get darker as you move to the right.

I'd assume `primary` would get swapped with `accent-cool-darker`, `primary-dark` with `accent-cool-darkest` and `primary-darker` with `accent-cool-more-dark`

Screenshot 2023-09-01 at 10 13 10 AM
Screenshot 2023-09-01 at 10 13 23 AM

http://nj-bfs-design-storybook.s3-website-us-east-1.amazonaws.com/?path=/story/atoms-colors--colors

web/src/styles/base/global.scss
```diff
@@ -232,22 +232,23 @@ ol {
...
 .usa-link:hover,
 .usa-prose > a:hover {
-  color: $primary-dark !important;
+  color: $accent-cool-more-dark !important;
 }
 
 .usa-link:active,
 .usa-prose > a:active {
-  color: $primary-darker !important;
+  color: $accent-cool-darkest !important;
```
FarazA22, line 251: [pebble] We've only addressed anchor elements here. You'll need to update the unstyled button as well. There are instances in the app where Product desired analytics to fire when a link is clicked, to enable this, we used the unstyled button.

web/src/components/profile/ProfileOpportunitiesAlert.tsx
```diff
@@ -1,11 +1,10 @@
...
-import { Alert } from "@/components/njwds-extended/Alert";
 import { ProfileDataContext } from "@/contexts/profileDataContext";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { getProfileConfig } from "@/lib/domain-logic/getProfileConfig";
 import { ProfileContentField } from "@/lib/types/types";
 import { getFieldsForProfile, isFieldAnswered } from "@businessnjgovnavigator/shared";
 import { ReactElement, useContext } from "react";
+import { FieldEntryAlert } from "../FieldEntryAlert";
 
 export const ProfileOpportunitiesAlert = (): ReactElement => {
```
FarazA22, line 9: [dust] I think this component is unnecessary because there is not a lot of logic here. What are your thoughts on deleting it?

web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
```diff
@@ -458,7 +459,35 @@ export const BusinessFormationPaginator = (props: Props): ReactElement => {
...
+          label,
+          children: getApiErrorMessage(fieldError.field) && (
+            <ul>
+              <li>{getApiErrorMessage(fieldError.field)}</li>
+            </ul>
+          ),
+        };
+      });
+
+      const swap = true;
```
FarazA22, line 482: [dust] This shouldn't be hard coded

web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
```diff
@@ -458,7 +459,35 @@ export const BusinessFormationPaginator = (props: Props): ReactElement => {
...
+
+      return swap ? (
+        <FieldEntryAlert
```
FarazA22, line 485: [dust] I don't think we can use this component here because it renders anchor elements that would link to nothing.

web/src/styles/base/global.scss
jphechter: The AC of this ticket specifically says to use `accent-cool-darkest` and `accent-cool-more-dark` on hover. Happy to message design about this, but that's not how I read this.

web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
jphechter: That's actually a [boulder]. I had added that to quickly switch between the 2 options in the ternary below.

web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
jphechter: [boulder] this was a convenience that I left behind

web/src/styles/base/global.scss
jphechter: The ticket says hyperlinks and opportunity cards. Unless I'm misremembering I don't think we discussed this applying to buttons.

web/src/components/njwds-extended/Alert.tsx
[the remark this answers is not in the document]
jphechter: Even if the IDE didn't catch this, it should throw a compilation error if an incorrect value is provided.

This looks like it should fix this issue:
```
export const AlertVariants = ["info", "success", "warning", "error", "note"] as const;
```

web/src/components/profile/ProfileOpportunitiesAlert.tsx
jphechter: Completely misread this.

We use this component 4 times in `profile.tsx`. From an organization perspective, this makes sense to leave it as a separate component.

A week or so ago we were talking about when logic should be extracted from parent into a dedicated component. I think this is a good case. `profile.tsx` is 807 lines. Keeping these 40 lines in their own file creates a separation of logic.

web/src/styles/base/global.scss
jphechter: Gotcha. My misunderstanding. Included this.

web/src/components/FieldEntryAlert.tsx
```diff
@@ -0,0 +1,32 @@
+import { Content } from "@/components/Content";
+import { Alert, AlertVariant } from "@/components/njwds-extended/Alert";
+import { ReactElement, ReactNode } from "react";
+
+interface Props {
```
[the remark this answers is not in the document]
jphechter: I'm down to change the name of the component. I don't have a better name. Do you have a suggestion?

I extended `FieldEntryAlert` to take accept all of the `Alert` props. I'm not sure how much I like that right now, but it does make this more flexible if we needed it to be in the future.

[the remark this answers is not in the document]
FarazA22: @jphechter You do have the flexibility to exclude or override props. You could consider removing unnecessary props. For a name, maybe `AlereWithLinksById`?

PR #6324, opened 2023-09-05
Dev design

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/185709167
https://www.pivotaltracker.com/story/show/185250601
https://www.pivotaltracker.com/story/show/185268488

### Approach

Three separate issues were addressed in this PR. 

- One commit updated the headers in the requirements tab component since the existing H2 was being removed. 
- One commit added new colors in storybook and updating the border colors for several colors because they did not exist.
- One commit changed the Z index on the NAV bar since it currently has a higher value than the overlay that appears when the contextual link opens.

## Code author checklist


web/src/styles/base/global.scss
```diff
@@ -399,3 +399,7 @@ ol {
 .margin-bottom-05-override {
   margin-bottom: 0.5rem !important;
 }
+
+.margin-y-3-override {
```
aloverso, line 403: [dust] Why the need for a new class for this?

PR #6325, opened 2023-09-05
feat: [#185455538] Error alert when email is taken is now more consis…

adidner:
…tent with other alerts

## Description

Before: 

image

After:
image

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185455538)

## Code author checklist

### DIFF
--- a/web/src/components/auth/SelfRegSnackbar.tsx
+++ b/web/src/components/auth/SelfRegSnackbar.tsx
@@ -5,8 +5,8 @@ import { Icon } from "@/components/njwds/Icon";
 import { AuthAlertContext } from "@/contexts/authAlertContext";
 import { AuthContext } from "@/contexts/authContext";
 import { IsAuthenticated } from "@/lib/auth/AuthContext";
+import { useConfig } from "@/lib/data-hooks/useConfig";
 import analytics from "@/lib/utils/analytics";
-import Config from "@businessnjgovnavigator/content/fieldConfig/config.json";
 import { RegistrationStatus } from "@businessnjgovnavigator/shared/";
 import { IconButton } from "@mui/material";
 import { ReactElement, useContext, useEffect } from "react";
@@ -15,6 +15,7 @@ export const SelfRegSnackbar = (): ReactElement => {
   const { isAuthenticated, registrationAlertStatus, setRegistrationAlertStatus } =
     useContext(AuthAlertContext);
   const { state } = useContext(AuthContext);
+  const { Config } = useConfig();
 
   useEffect(() => {
     if (registrationAlertStatus === "IN_PROGRESS" && isAuthenticated === IsAuthenticated.TRUE) {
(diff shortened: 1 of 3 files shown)


web/src/components/auth/SelfRegSnackbar.tsx
```diff
@@ -69,7 +76,7 @@ export const SelfRegSnackbar = (): ReactElement => {
         )}
         <div>
           {alertMap[registrationAlertStatus] === "success" ? (
-            <h3 className="padding-right-2">{getTitle()}</h3>
+            <h3 className="padding-right-2">{getSuccessTitle()}</h3>
```
FarazA22, line 79: [pebble] it seems like the we are using a header for the success snackbar. We should be able to pass it into the `heading` on line 58

web/src/components/auth/SelfRegSnackbar.tsx
```diff
@@ -83,7 +90,13 @@ export const SelfRegSnackbar = (): ReactElement => {
               color: "#757575",
             }}
           >
-            <Icon className="usa-icon--size-4">close</Icon>
+            <Icon
```
FarazA22, line 93: [dust] I'd ask design it the sizing change is intentional. I'd expect consistency especially since they are required to create a variation of the snackbar in the design library with the different icon size and they haven't.

jphechter:
Part of the AC on the ticket says `Add hyperlink to "login".` Is that still relevant? I'm not seeing that here.

web/src/components/auth/SelfRegSnackbar.tsx
adidner: I believe its a special case because of the alternate style, so it needs to be out here to look as desired. Its the very bottom custom one. 
image

web/src/components/auth/SelfRegSnackbar.tsx
```diff
@@ -5,8 +5,8 @@ import { Icon } from "@/components/njwds/Icon";
 import { AuthAlertContext } from "@/contexts/authAlertContext";
 import { AuthContext } from "@/contexts/authContext";
 import { IsAuthenticated } from "@/lib/auth/AuthContext";
+import { useConfig } from "@/lib/data-hooks/useConfig";
 import analytics from "@/lib/utils/analytics";
-import Config from "@businessnjgovnavigator/content/fieldConfig/config.json";
```
adidner, line 9: Should this be a refactor to replace the file import with the hook?

adidner: I'm imagining yes for good cleanness but its also not breaking or problematic in any way

FarazA22: That's a great idea. I went ahead and created the ticket. We can discuss it during an upcoming gprooming session to see if it's needed.

https://www.pivotaltracker.com/story/show/185992659

web/src/components/auth/SelfRegSnackbar.tsx
```diff
@@ -46,48 +47,57 @@ export const SelfRegSnackbar = (): ReactElement => {
     return <></>;
   }
 
+  const isSuccessAlert = alertMap[registrationAlertStatus] === "success";
+
+  if (isSuccessAlert) {
```
FarazA22, line 52: This refactor looks really good. It's easy to read. One small suggestion, consider inlining line 50 into the conditional?

PR #6327, opened 2023-09-06
feat: [#184415884] analytics added to cert and funding cards

FarazA22:
## Description

Added analytics to funding and certification opportunity cards. The analytics fire when the show or hide button within the card is clicked and when the accordion of the hidden card is opened.

I also updated the logic around when the hidden card accordion is rendered. Previously, it only relied on the displayCertifications prop, the conditional was updated. It should rely on both since it contains both opportunity cards.

### Ticket
https://www.pivotaltracker.com/story/show/184415884

## Code author checklist

### DIFF
--- a/web/src/components/dashboard/SidebarCardsList.tsx
+++ b/web/src/components/dashboard/SidebarCardsList.tsx
@@ -4,6 +4,7 @@ import { SidebarCard } from "@/components/dashboard/SidebarCard";
 import { Icon } from "@/components/njwds/Icon";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { Certification, Funding, SidebarCardContent } from "@/lib/types/types";
+import analytics from "@/lib/utils/analytics";
 import { templateEval } from "@/lib/utils/helpers";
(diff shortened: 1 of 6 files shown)


somabadri (approved):

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -44,14 +45,17 @@ export const SidebarCardsList = (props: Props): ReactElement => {
   };
 
   const hiddenCardsAccordion = (): ReactNode => {
-    if (props.displayCertifications) {
+    if (props.displayCertifications || props.displayFundings) {
```
somabadri, line 48: [Pebble] does this need a test, it's new logic?

FarazA22: I added the test

somabadri: I was wondering if there are any tests regarding what's being returned based on this conditional that need to be updated or newly added

PR #6331, opened 2023-09-07
chore: [#185788142] for you analytics

cbrunefearless:
## Description
Add custom analytics events for when clicking the title link for any Funding opportunities and Certifications cards in the For You section.

Analytics Events
legacy_event_category: opportunity_card
legacy_event_action: click
legacy_event_label: go_to_opportunity_screen
event: link_clicks
item: opportunity_card

Insights
Suggested Analytics (New):

Clickthrough rate of opportunities (certs and funding)
Suggested Analytics (Existing):

Clickthrough rate of phase action cards

### Ticket

[#185788142](https://www.pivotaltracker.com/story/show/)

## Code author checklist

### DIFF
--- a/web/src/components/dashboard/OpportunityCard.test.tsx
+++ b/web/src/components/dashboard/OpportunityCard.test.tsx
@@ -2,10 +2,12 @@ import { OpportunityCard } from "@/components/dashboard/OpportunityCard";
 import { getMergedConfig } from "@/contexts/configContext";
 import analytics from "@/lib/utils/analytics";
 import { generateOpportunity } from "@/test/factories";
+import { useMockRouter } from "@/test/mock/mockRouter";
 import { useMockBusiness } from "@/test/mock/mockUseUserData";
 import { generatePreferences } from "@businessnjgovnavigator/shared/";
 import { fireEvent, render, screen } from "@testing-library/react";
 
+jest.mock("next/router", () => ({ useRouter: jest.fn() }));
 jest.mock("@/lib/utils/analytics", () => setupMockAnalytics());
 jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
 
@@ -26,6 +28,11 @@ function setupMockAnalytics(): typeof analytics {
           unhide_card: jest.fn(),
         },
       },
+      opportunity_card: {
+        click: {
+          go_to_opportunity_screen: jest.fn(),
+        },
+      },
     },
   };
 }
@@ -34,6 +41,7 @@ describe("OpportunityCard", () => {
   beforeEach(() => {
     jest.resetAllMocks();
     useMockBusiness({});
+    useMockRouter({});
   });
 
   it("renders the name", () => {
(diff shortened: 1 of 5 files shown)


web/src/components/dashboard/OpportunityCard.test.tsx
```diff
@@ -111,6 +119,17 @@ describe("OpportunityCard", () => {
...
 
+  it("fires opportunity_card analytics when title with link is clicked", () => {
+    render(
+      <OpportunityCard
+        opportunity={generateOpportunity({ name: "fundingName", urlSlug: "slug" })}
+        urlPath="funding"
+      />
+    );
+    fireEvent.click(screen.getByText("fundingName"));
+    expect(mockAnalytics.event.opportunity_card.click.go_to_opportunity_screen).toHaveBeenCalledTimes(1);
```
aloverso, line 130: is there a reason for the explicit `toHaveBeenCalledTimes(1)` instead of just `toHaveBeenCalled`?

web/src/components/dashboard/OpportunityCard.tsx
```diff
@@ -72,6 +74,12 @@ export const OpportunityCard = (props: Props): ReactElement => {
       .update();
   };
 
+  const routeToPage = (): void => {
+    const url = `/${props.urlPath}/${props.opportunity.urlSlug}`;
```
aloverso, line 78: [sand] it'd be nice to have a test that the router push is called with the correct url. we weren't doing it before (probably because it's a little more annoying to test when it's an `a` href) but it seems like a good opportunity to add test coverage

web/src/components/dashboard/SidebarCardsContainer.test.tsx
```diff
@@ -223,11 +226,12 @@ describe("<SidebarCardsContainer />", () => {
       expect(screen.queryByText("Cert 3")).not.toBeInTheDocument();
     });
 
-    it("links to task page for certifications", () => {
+    it("routes to task page for certifications", () => {
```
aloverso, line 229: seems like this test was copy-pasted / misnamed... it's not a "task" page and we should edit that.

web/src/components/dashboard/OpportunityCard.test.tsx
jphechter: This might be a little excessive, but given that we also want to validate the integrity of the data we're capturing, maybe it makes sense to validate this is only called once?

cbrunefearless: ok

web/src/components/dashboard/SidebarCardsContainer.test.tsx
cbrunefearless: renamed

web/src/components/dashboard/OpportunityCard.test.tsx
aloverso: @jphechter @cbrunefearless It's fair, I guess I feel like it doesn't matter much so I'd rather just keep with the pattern. Unless we should change the other analytics event tests to be CalledTimes(1) also?

PR #6332, opened 2023-09-07
Add accent-cool-darkest to additional links and unstyled buttons

jphechter:
## Description

Updates based on acceptance feedback.

### Ticket

[184432499](https://www.pivotaltracker.com/story/show/184432499)

## Code author checklist


web/public/img/hide.svg
jphechter, line 1: As far as I can tell, we only use the `show` and `hide` svgs in one location and we want to use this new blue color in all instances.

PR #6333, opened 2023-09-08
feat: [#183536833] signers and checkbox error message

somabadri:
## Description

Updated errors for signers in formation.

### Ticket

[#183536833](https://www.pivotaltracker.com/story/show/183536833)

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/business-formation.json
+++ b/content/src/fieldConfig/business-formation.json
@@ -131,7 +131,6 @@
       "signers": {
         "addButtonText": "Add New Signer",
         "signColumnLabel": "Sign*",
-        "errorInlineFirstSignerName": "Enter a signer.",
         "overrides": {
           "limited-liability-partnership": {
             "description": "Authorized Partner(s) should sign below by adding their full name. You can have up to 10 signers listed."
@@ -153,7 +152,10 @@
         "tableHeader": "Name,Address,Sign*,Actions",
         "titleLabel": "Title",
         "description": "Authorized Representative(s) should sign below by adding their full name. You can have up to 10 signers listed.",
-        "errorInlineAdditionalSignerName": "Enter the first and last name of each signer."
+        "errorInlineSignerName": "Enter a signer's name.",
+        "errorInlineAdditionalSignerName": "Enter the first and last name of each signer.",
+        "errorInlineNameAndSignature": "Enter a signer’s name and click the checkbox to authorize their signature.",
+        "errorInlineSignature": "Click the checkbox to authorize the signer’s signature."
       },
       "addressLine1": {
         "label": "Address Line 1",
--- a/web/public/mgmt/config.yml
+++ b/web/public/mgmt/config.yml
@@ -681,12 +681,22 @@ collections:
                       - { label: "Error banner - checkbox", name: "errorBannerCheckbox", widget: "string" }
                       - {
                           label: "Error inline - first signer name",
-                          name: "errorInlineFirstSignerName",
+                          name: "errorInlineSignerName",
                           widget: "string",
                         }
                       - {
                           label: "Error inline - additional signer name",
                           name: "errorInlineAdditionalSignerName",
+                          widget: "string",
+                        }
+                      - {
+                          label: "Error inline - Name and Signature Missing",
+                          name: "errorInlineNameAndSignature",
+                          widget: "string",
+                        }
+                      - {
+                          label: "Error inline - Signature Missing",
+                          name: "errorInlineSignature",
                           widget: "string",
                         }
                       - {
--- a/web/src/components/tasks/business-formation/contacts/ContactsStep.tsx
+++ b/web/src/components/tasks/business-formation/contacts/ContactsStep.tsx
@@ -1,11 +1,9 @@
-import { Alert } from "@/components/njwds-extended/Alert";
 import { Addresses } from "@/components/tasks/business-formation/contacts/Addresses";
 import { createSignedEmptyFormationObject } from "@/components/tasks/business-formation/contacts/helpers";
 import { Members } from "@/components/tasks/business-formation/contacts/Members";
 import { RegisteredAgent } from "@/components/tasks/business-formation/contacts/RegisteredAgent";
 import { Signatures } from "@/components/tasks/business-formation/contacts/Signatures";
 import { FormationField } from "@/components/tasks/business-formation/FormationField";
-import { getErrorStateForField } from "@/components/tasks/business-formation/getErrorStateForField";
 import { BusinessFormationContext } from "@/contexts/businessFormationContext";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { useFormationErrors } from "@/lib/data-hooks/useFormationErrors";
@@ -53,11 +51,6 @@ export const ContactsStep = (): ReactElement => {
           </>
         )}
         <hr className="margin-top-0 margin-bottom-3" />
-        {doesFieldHaveError("signers") && (
-          <Alert variant="error">
-            {getErrorStateForField({ field: "signers", formationFormData: state.formationFormData }).label}
-          </Alert>
-        )}
         {incorporationLegalStructures.includes(state.formationFormData.legalType) ? (
           <FormationField fieldName="incorporators">
             <Addresses<FormationIncorporator>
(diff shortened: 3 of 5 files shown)


content/src/fieldConfig/business-formation.json
```diff
@@ -50,7 +50,8 @@
       "genericApiError": "Review the information in the fields below for errors:",
       "errorOnStep": "Review the information in the fields below for errors or omissions:",
       "incompleteStepsError": "Check the steps below for missing information or errors:",
-      "apiResubmitText": "After you make the correction, resubmit the form."
+      "apiResubmitText": "After you make the correction, resubmit the form.",
+      "containsMissingSignature": "A signature is missing"
```
jphechter, line 54: [dust] should this just be `missingSignature`? It's not really containing anything.

web/src/components/FieldEntryAlert.tsx
```diff
@@ -25,6 +26,9 @@ export const FieldEntryAlert = (props: Props): ReactElement => {
             {field.children}
           </li>
         ))}
+        {props.additionalErrors?.map((error) => {
+          return <li key={error}>{error}</li>;
+        })}
```
jphechter, line 31: [sand] It's minor, but it might be nice to add a test for this.

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -582,6 +583,34 @@ describe("<BusinessFormationPaginator />", () => {
         expect(screen.getByRole("alert")).toHaveTextContent(Config.formation.fields.trustees.label);
       });
 
+      it.each(["limited-liability-company", "limited-liability-partnership"])(
+        "shows the signature is missing error text in the alert when %s",
```
jphechter, line 587: [dust]
```suggestion
        "shows the signature is missing error text in the alert when legalStructure is %s",
```

content/src/fieldConfig/business-formation.json
```diff
@@ -153,7 +154,10 @@
         "tableHeader": "Name,Address,Sign*,Actions",
         "titleLabel": "Title",
         "description": "Authorized Representative(s) should sign below by adding their full name. You can have up to 10 signers listed.",
-        "errorInlineAdditionalSignerName": "Enter the first and last name of each signer."
+        "errorInlineAdditionalSignerName": "Enter the first and last name of each signer.",
+        "errorInLineNameAndSignatureMissing": "Enter a signer’s name and click the checkbox to authorize their signature.",
```
AnJuHyppolite, line 158: [pebble] Can we update this to `errorInlineSignerNameAndCheckbox`? I don't think we need `Missing` since the `error` prefix denotes something is missing or incorrect? Also, this naming convention holds to the `Inline` and `SignerName` patterns that already exist?

content/src/fieldConfig/business-formation.json
```diff
@@ -153,7 +154,10 @@
...
+        "errorInLineNameMissing": "Enter a signer’s name.",
```
AnJuHyppolite, line 159: I see that the copy is from the design, but we already have `"errorInlineFirstSignerName": "Enter a signer.",` in our code? Are we sure the Content Team wants to add this?

[pebble] OAN: If we are keeping this new copy, can we change it to `errorInlineAdditionalSignerName`, omitting the `Missing` and having `SignerName` as our pattern (as mentioned previously)?

content/src/fieldConfig/business-formation.json
```diff
@@ -153,7 +154,10 @@
...
+        "errorInLineSignatureMissing": "Click the checkbox to authorize the signer’s signature."
```
AnJuHyppolite, line 160: [pebble] Can we update this to `errorInlineSignerCheckbox` (for the same aforementioned reasons)?

content/src/fieldConfig/business-formation.json
somabadri: Let me double check with content!

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -582,6 +583,32 @@ describe("<BusinessFormationPaginator />", () => {
         expect(screen.getByRole("alert")).toHaveTextContent(Config.formation.fields.trustees.label);
       });
 
+      it.each(["limited-liability-company", "limited-liability-partnership"])(
+        "shows the signature is missing error text in the alert when legalStructure is %s",
```
AnJuHyppolite, line 587: Can we update the description so that's it's clear? Something like: `"shows an error text when the signature is missing'`?

web/src/components/FieldEntryAlert.tsx
jphechter, line 31: I'm looking at this again, and I think I'm a little confused. Right now this error message seems to be breaking the current implementation. We already have a "signers" error. Is this new message supposed to be a detail associated with the current bullet? If so should we be implementing this change as a sort of a nested list structure?

If this is going to be a standalone error, should it drive to the location on the page where we're missing information?
Screenshot 2023-09-11 at 2 30 07 PM

jphechter: I'm looking at the design and it does look like this is the intended structure. I feel like this should probably drive to the field where the error is present.

web/src/components/tasks/business-formation/contacts/ContactsStep.test.tsx
```diff
@@ -369,6 +375,51 @@ describe("Formation - ContactsStep", () => {
       expect(page.getInputElementByLabel("Signer 1").value).toBe("signer 2");
       expect(page.getInputElementByLabel("Signer 2").value).toBe("signer 3");
     });
+
+    describe("signers", () => {
+      it("displays error to both enter name and sign if both are missing", async () => {
```
AnJuHyppolite, line 380: Can we update description to  `"displays error to both enter name and signature if both are missing"`?

web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
```diff
@@ -407,6 +407,10 @@ export const BusinessFormationPaginator = (props: Props): ReactElement => {
     return !allApiErrorsHaveMappings;
   };
 
+  const hasMissingSignature = (): boolean => {
+    return !!state.formationFormData.signers?.find((signer) => !signer.signature);
+  };
+
```
jphechter, line 413: [pebble] This error check feels like it should probably be part of the `useFormationErrors` hook. Is there a reason we want this to be closer to component? If we move it out we can probably pick up the error in the `allCurrentErrorsForStep` function. This might allow you to take out the `additionalErrors` prop from `FieldEntryAlert` and just render it with the existing map.

jphechter: Soma and I talked about this. This is the only error that we have where the field name is not also the error messaged displayed in the alert. `useFormationErrors` is probably the right place for this logic, but because this is a one off and `useFormationErrors` is pretty generic fitting this in would break a bunch of assumptions we have in that logic. We're going to keep this as is for now. If this is a pattern that we will use in other places, we'll need to make changes to `useFormationErrors`.

Additionally, Soma is following up with the design team to make sure that this is still a pattern we want to pursue. This ticket is from October 2022. If so, should this error message also hyperlink to the Signers field?

content/src/fieldConfig/business-formation.json
```diff
@@ -153,7 +152,10 @@
         "tableHeader": "Name,Address,Sign*,Actions",
         "titleLabel": "Title",
         "description": "Authorized Representative(s) should sign below by adding their full name. You can have up to 10 signers listed.",
-        "errorInlineAdditionalSignerName": "Enter the first and last name of each signer."
+        "errorInlineSignerName": "Enter a signer's name.",
+        "errorInlineAdditionalSignerName": "Enter the first and last name of each signer.",
+        "errorInLineNameAndSignature": "Enter a signer’s name and click the checkbox to authorize their signature.",
+        "errorInLineSignature": "Click the checkbox to authorize the signer’s signature."
```
AnJuHyppolite, line 158: Can we update `InLine` so the `'I'` is lower case instead of a capital letter: `'L'`?

somabadri: oh yes, good catch, I have been looking at it for so long, didn't notice the difference.

jphechter (approved):

PR #6340, opened 2023-09-08
chore: [#185992659] replace all imports of config.json with useConfig

FarazA22:
## Description
They are instances in the application where we are importing the config.json file. The intent of this re-factoring ticket is to replace that import with the use config data hook, so that that config variable gets all of the merged content.

### Ticket

https://www.pivotaltracker.com/story/show/185992659

## Code author checklist

### DIFF
--- a/web/src/components/TaskProgressTagLookup.tsx
+++ b/web/src/components/TaskProgressTagLookup.tsx
@@ -1,7 +1,9 @@
 import { Tag } from "@/components/njwds-extended/Tag";
-import Config from "@businessnjgovnavigator/content/fieldConfig/config.json";
+import { getMergedConfig } from "@/contexts/configContext";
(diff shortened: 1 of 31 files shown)


web/src/components/TaskProgressTagLookup.tsx
```diff
@@ -1,22 +1,28 @@
 import { Tag } from "@/components/njwds-extended/Tag";
-import Config from "@businessnjgovnavigator/content/fieldConfig/config.json";
+import { useConfig } from "@/lib/data-hooks/useConfig";
```
FarazA22, line 2: Changes in this file are "significant"

aloverso: If we wanted to keep the same pattern here, we could use `getMergedConfig` instead

somabadri (approved):

PR #6358, opened 2023-09-12
chore: [#184814757] combine onboarding and profile components

somabadri:
## Description

Consolidated onboarding and profile components into one folder and removed prefixes.

### Ticket

[#184814757](https://www.pivotaltracker.com/story/show/184814757)

## Code author checklist


web/src/components/profile/LockedProfileField.tsx
jphechter, line 1: Wow. A profile field in the onboarding directory?

Good job. Nice catch.

AnJuHyppolite:
This is great work, Soma! 

Re: Anne's points:
- Can the `FieldLabel...` components can go in `datafields`?
- re: `OnboardingBusinessPersona` and `OnboardingNonprofit` - I think I recall an initial conversation about grouping based on functionality, not whether they are used in  a particular part of the site (in this case `onboarding` or `profile`).  
- `ProfileNotes` is a data-field. It invokes `OnboardingField`....not sure whether we want to take that into account.
- Perhaps `OnboardingNameAndEmail` could live in data-fields.
- Should we rename `EscapeModal` to `ProfileEscapeModal`?

jphechter (approved):
Overall, this is a great refactor! 10/10

Anne makes some solid points. It'd be cool to see those additions. Preemptively approving.

aloverso:
Looks good. I'm still noticing a few opportunities for changes, but feel free to do as a follow up commit:

image
`ProfileNumericField` is used by EntityId, TaxId, and others and is clearly not profile-only related. Perhaps just NumericField?

image
`DataField` defines and exports `OnboardingProps` but I think that could be renamed like `DataFieldProps` or similar.

I see that `RadioQuestion` has moved to `/components` but I'd argue it belongs in `/data-fields` because it modifies ProfileDataContext like all the other components in there. That's my gut, at least? What do you think. 

I still feel like ProfileNotes is a data-field. It's really no different from TaxPin or anything else that just shows up in profile and lets a user enter profile data. I think it should just be Notes and live in data-field.

aloverso (approved):

PR #6363, opened 2023-09-15
feat: [#185801983] register for taxes task should not allow edits or …

adidner:
…submition in guest mode

## Description

The register state taxes task was allowing modifications and saving in guest mode in some cases (and copy paste into it)

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185801983)

### Steps to Test

Go to the task tax-register page after signing in as any persona and choosing a business structure. ,

try to type, paste into the input and try to hit the register and save button, see that you cannot and get a popup telling you to link to my NJ to save. 

## Code author checklist

### DIFF
--- a/web/src/components/tasks/TaxInput.tsx
+++ b/web/src/components/tasks/TaxInput.tsx
@@ -24,7 +24,7 @@ interface Props {
 
 export const TaxInput = (props: Props): ReactElement => {
   const { business, updateQueue } = useUserData();
-  const { isAuthenticated } = useContext(NeedsAccountContext);
+  const { isAuthenticated, setShowNeedsAccountModal } = useContext(NeedsAccountContext);
   const { Config } = useConfig();
   const [profileData, setProfileData] = useState<ProfileData>(
     business?.profileData ?? createEmptyProfileData()
@@ -106,6 +106,23 @@ export const TaxInput = (props: Props): ReactElement => {
     </div>
   );
 
+  const getNeedsAccountModalFunction = (): (() => void) | undefined => {
+    if (isAuthenticated === IsAuthenticated.FALSE) {
+      return () => {
+        return setShowNeedsAccountModal(true);
+      };
+    }
+    return undefined;
+  };
+
+  const onSave = (): void => {
+    if (isAuthenticated === IsAuthenticated.FALSE) {
+      setShowNeedsAccountModal(true);
+    } else {
+      onSubmit();
+    }
+  };
+
   return (
     <profileFormContext.Provider value={formContextState}>
       <ProfileDataContext.Provider
@@ -126,11 +143,14 @@ export const TaxInput = (props: Props): ReactElement => {
               </Alert>
             ) : (
               <>
-                <TaxId required />
+                <TaxId
+                  required={isAuthenticated === IsAuthenticated.TRUE}
+                  handleChangeOverride={getNeedsAccountModalFunction()}
+                />
                 <div className="tablet:margin-top-05 tablet:margin-left-2">
                   <SecondaryButton
                     isColor="primary"
-                    onClick={onSubmit}
+                    onClick={onSave}
                     isLoading={isLoading}
                     isSubmitButton={true}
                     isRightMarginRemoved={true}
--- a/web/src/components/data-fields/tax-id/SingleTaxId.tsx
+++ b/web/src/components/data-fields/tax-id/SingleTaxId.tsx
@@ -2,9 +2,7 @@
 
 import { DataField, DataFieldProps } from "@/components/data-fields/DataField";
 import { TaxIdDisplayStatus } from "@/components/data-fields/tax-id/TaxIdHelpers";
-import { NeedsAccountContext } from "@/contexts/needsAccountContext";
 import { ProfileDataContext } from "@/contexts/profileDataContext";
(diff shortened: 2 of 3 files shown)


web/src/components/onboarding/taxId/OnboardingSingleTaxId.tsx
```diff
@@ -24,12 +22,8 @@ export const OnboardingSingleTaxId = ({
 
   const isTabletAndUp = useMediaQuery(MediaQueries.tabletAndUp);
   const { state, setProfileData } = useContext(ProfileDataContext);
-  const { isAuthenticated, setShowNeedsAccountModal } = useContext(NeedsAccountContext);
 
   const handleChange = (value: string): void => {
-    if (isAuthenticated === IsAuthenticated.FALSE) {
-      setShowNeedsAccountModal(true);
-    }
     if (handleChangeOverride) {
```
adidner, line 27: So for this ticket I had a choice, 

This component is called in 2 places, in the tax task via taxInput via OnboardingTaxId and in the profile directly via OnboardingTaxId. 

Profile has this function declared in itself and passes this logic down as handleChangeOverride while the taxInput chose instead to use the logic above. 

I chose to follow the lead of profile and moved this logic above to the TaxInput and passed it down. 

I did this because it felt move discoverable and its a pattern we have all over profile to pass down the override instead of having it in the components itself (it also feels cleaner to have the component be more isolated) Also we needed some of this logic in the TaxInput anyways to prevent the save and register button from triggering as well.

web/src/components/tasks/TaxInput.tsx
```diff
@@ -126,11 +143,14 @@ export const TaxInput = (props: Props): ReactElement => {
               </Alert>
             ) : (
               <>
-                <OnboardingTaxId required />
+                <OnboardingTaxId
+                  required={isAuthenticated === IsAuthenticated.FALSE ? false : true}
+                  handleChangeOverride={showNeedsAccountModalForGuest()}
```
adidner, line 148: So this is so we don't have inline errors when the user tries to enter data here (I think ideally I would like to relocate this logic one level lower but for some reason that isn't working at all)

adidner: JK that doesn't work at all because its spreading props into its children so it needs to not be required to not be validated in addition to modifying the validation function itself and that doesn't feel particularly less complicated

web/src/components/tasks/TaxInput.tsx
```diff
@@ -106,6 +106,23 @@ export const TaxInput = (props: Props): ReactElement => {
     </div>
   );
 
+  const showNeedsAccountModalForGuest = (): (() => void) | undefined => {
```
aloverso, line 109: [dust/pebble] This made me do a double-take when I saw a function getting executing down on line 148, until I realized it's a function that returns a function.

That's fine, I see what we want it to be undefined or a function. But I think the naming should be revised to make it clear that this is NOT a function that "shows needs account modal for guest" - it returns a function that does that. Something like `getNeedsAccountFunction` or even `getHandleChangeOverride` would be more descriptive imo

web/src/components/tasks/TaxInput.tsx
```diff
@@ -126,11 +143,14 @@ export const TaxInput = (props: Props): ReactElement => {
...
-                <OnboardingTaxId required />
+                <OnboardingTaxId
+                  required={isAuthenticated === IsAuthenticated.TRUE}
+                  handleChangeOverride={showNeedsAccountModalForGuest()}
+                />
                 <div className="tablet:margin-top-05 tablet:margin-left-2">
                   <SecondaryButton
                     isColor="primary"
-                    onClick={onSubmit}
+                    onClick={(): void => onSave()}
```
aloverso, line 153: [pebble] why not just `onClick={onSave}`?

somabadri (approved):
good to merge after addressing Anne's comments

web/src/components/tasks/TaxInput.tsx
```diff
@@ -106,6 +106,23 @@ export const TaxInput = (props: Props): ReactElement => {
     </div>
   );
 
+  const showNeedsAccountModalForGuest = (): (() => void) | undefined => {
+    if (isAuthenticated === IsAuthenticated.FALSE) {
+      return () => {
+        return setShowNeedsAccountModal(true);
+      };
+    }
+    return undefined;
```
jphechter, line 115: [dust] This is about as nit picky as I could be, but you don't need to explicitly return `undefined`. You can just `return`.

web/src/components/tasks/TaxInput.tsx
```diff
@@ -106,6 +106,23 @@ export const TaxInput = (props: Props): ReactElement => {
...
+  };
```
jphechter, line 116: [dust] This is just nit picky. You don't need to explicitly return `undefined`. And you can also use the arrow function shorthand to make this a little smaller.
```suggestion
    if (isAuthenticated === IsAuthenticated.FALSE) {
      return () =>  setShowNeedsAccountModal(true);
    }
    return
  };
```

aloverso: for the undefined, I think I prefer the explicit undefined personally.

PR #6364, opened 2023-09-15
feat: [#185167013] enable foreign nonprofit formation

aloverso:
### Ticket

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/billing/getCost.test.ts
+++ b/web/src/components/tasks/business-formation/billing/getCost.test.ts
@@ -2,23 +2,12 @@
 
 import { getCost } from "@/components/tasks/business-formation/billing/getCost";
 import { getMergedConfig } from "@/contexts/configContext";
-import { FormationLegalType } from "@businessnjgovnavigator/shared/formationData";
+import { allFormationLegalTypes, FormationLegalType } from "@businessnjgovnavigator/shared/formationData";
 
 const Config = getMergedConfig();
 
 describe("getCost", () => {
-  const defaultCostIds: FormationLegalType[] = ["limited-liability-company", "limited-liability-partnership"];
-  const overriddenCostIds: FormationLegalType[] = ["s-corporation", "c-corporation", "limited-partnership"];
-  const allLegalTypes = [...defaultCostIds, ...overriddenCostIds];
-
-  for (const legalStructureId of allLegalTypes) {
-    describe("official formation document", () => {
-      it(`uses default cost for ${legalStructureId}`, async () => {
-        const officialFormationCost = Number.parseInt(Config.formation.fields.officialFormationDocument.cost);
-        expect(getCost("officialFormationDocument", legalStructureId)).toEqual(officialFormationCost);
-      });
-    });
-
+  for (const legalStructureId of allFormationLegalTypes) {
     describe("certified copy of formation", () => {
       it(`uses default cost for ${legalStructureId}`, async () => {
         const certifiedCopyCost = Number.parseInt(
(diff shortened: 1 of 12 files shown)


web/src/components/tasks/business-formation/billing/getCost.test.ts
```diff
@@ -29,7 +18,36 @@ describe("getCost", () => {
     });
   }
 
+  describe("official formation document", () => {
+    const overriddenCostIds: FormationLegalType[] = ["nonprofit"];
+    const defaultCostIds = allFormationLegalTypes.filter((it) => !overriddenCostIds.includes(it));
+
+    for (const legalStructureId of defaultCostIds) {
```
somabadri, line 25: [dust] Is there a reason we are using the for loop here instead of the it.each pattern? I think the for loop is bit more readable in this case but was wondering if there were other things taken into consideration. 

`it.each([...defaultCostIds])("...", (legalStructureId) => {})`

web/src/components/tasks/business-formation/review/ReviewStep.test.tsx
```diff
@@ -773,4 +773,28 @@ describe("Formation - ReviewStep", () => {
...
+
+  describe("when foreign nonprofit", () => {
+    it("displays foreign cert of good standing", async () => {
+      await renderStep({ businessPersona: "FOREIGN", legalStructureId: "nonprofit" }, {});
+      expect(screen.getByText(Config.formation.fields.foreignGoodStandingFile.label)).toBeInTheDocument();
+    });
+
+    it("displays will practice law", async () => {
+      await renderStep({ businessPersona: "FOREIGN", legalStructureId: "nonprofit" }, {});
+      expect(screen.queryByText(Config.formation.fields.willPracticeLaw.label)).not.toBeInTheDocument();
```
somabadri, line 797: [pebble] I think the `it` statement has an typo, should it be `doesn't display` based on the expect statement?

web/src/components/tasks/business-formation/review/ReviewStep.test.tsx
```diff
@@ -773,4 +773,28 @@ describe("Formation - ReviewStep", () => {
       });
     });
   });
+
+  describe("when foreign corporation", () => {
```
somabadri, line 777: [dust] we could combine these to more concise

```
describe("when foreign", () => {
it.each(["c-corp, nonprofit"])("displays foreign cert of good standing when legalStructure is %s")....

it("displays will practice law when corp", async () = {})

it("doesn't display will practice law when nonprofit", async () = {})
})
```
the last test I'm assuming a typo exists below.

web/src/components/tasks/business-formation/billing/getCost.test.ts
aloverso: I was just following the existing pattern of the file already... happy to change if we prefer it.each. Thoughts?

web/src/components/tasks/business-formation/review/ReviewStep.test.tsx
```diff
@@ -773,4 +773,28 @@ describe("Formation - ReviewStep", () => {
       });
```
somabadri, line 773: [dust] This is just a general comment, not related to your work but I'm looking through this test file and it's pretty difficult to read because there isn't a structured grouping of any sort? I think we should open a refactor ticket to organize it by section like below or whichever other grouping makes sense because it's disorganized and will grow to be more chaotic in the future? 

```
describe("business"){
    describe("llp")
    describe("llc")
    ....
}

describe("Contacts"){
    describe("llp")
    describe("llc")
    ....
}
```

web/src/components/tasks/business-formation/review/ReviewStep.test.tsx
aloverso: Hm, I'd push back. I think this is a bit harder to read in my opinion. I think we tend to think of things on a legal-type by legal-type basis, ie, phrasing the tests as "here's what happens when foreign c-corp" and "here's what happens when foreign nonprofit" is more accurate to read from a documentation perspective than having to sort through all the tests in "when foreign" to identify which ones apply when. Like, if you're reading the tests trying to understand what's different for a foreign c-corp, it's easier to have that laid out together than to have to read all 3 individual `it` statements to identify which ones might have something to do with corp. 

Plus it would require a larger refactoring of the test file because there's a lot of other tests/describes for other "when foreign" cases that would require moving into this block.

somabadri: mhm, that makes sense, thank you! I thought a larger refactoring might be necessary too, left a comment above.

somabadri (approved):

web/src/components/tasks/business-formation/billing/getCost.test.ts
somabadri: I don't have a preference but I think we should chat as a dev team on whether we have preferred choice and try to have it reflect throughout our code for uniformity sake or is having both patterns okay? I was in a similar situation for the signer and checkbox error feature where the test files had `for` loops and I refactored them to follow the `it` pattern but either would have been fine.

adidner (approved):
LGTM

PR #6396, opened 2023-09-19
feat: [#186025115] skip onboarding for industry query strings without essential questions

aloverso:
## Description

The switch to using `localUpdateQueue` and passing it around is because for the query string users, the updateQueue gets created earlier in the routing useEffect, and gets set to the global context state, but since React state updates are async, it was still undefined in state when we got down to the `completeOnboarding` function, which needs it to be defined in order to finish everything up. So we instead use the return value to reference the queue everywhere as the onboarding routing completes.

### Ticket

[#186025115](https://www.pivotaltracker.com/story/show/#186025115)

## Code author checklist

### DIFF
--- a/web/test/pages/onboarding/onboarding-shared.test.tsx
+++ b/web/test/pages/onboarding/onboarding-shared.test.tsx
@@ -1,5 +1,6 @@
 import { onboardingFlows } from "@/components/onboarding/OnboardingFlows";
 import { getMergedConfig } from "@/contexts/configContext";
+import { QUERIES, ROUTES } from "@/lib/domain-logic/routes";
 import { templateEval } from "@/lib/utils/helpers";
 import { randomElementFromArray } from "@/test/helpers/helpers-utilities";
 import * as mockRouter from "@/test/mock/mockRouter";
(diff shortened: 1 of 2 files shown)


web/test/pages/onboarding/onboarding-shared.test.tsx
```diff
@@ -88,19 +89,22 @@ describe("onboarding - shared", () => {
     expect(screen.getByTestId("step-1")).toBeInTheDocument();
   });
 
-  it("routes to industry page when industry without essential question is set by using industry query string", async () => {
+  it("routes to industry page when industry WITHOUT essential question is set by using industry query string", async () => {
```
somabadri, line 92: [pebble] I'm think I might be a little confused, I thought the point of this ticket was that if the industry is set by the query string then we skip onboarding altogether and take the user directly to the dashboard and if they have an essential question then they are taken through onboarding so they can answer it?

aloverso: you're right... I updated the test names incorrectly.
- Without essential question, it goes to dashboard
- With essential question, it stays in onboarding

The test content is right I believe but the test names got mixed up. Good catch!

somabadri (approved):

PR #6397, opened 2023-09-19
feat: [#185801876] display modal on NAICS code task for guest mode users

AnJuHyppolite:
## Description

When users are in guest mode, a modal will now appear when they click either of the NAICS code radio buttons and the task progress will remain as NOT STARTED.

### Ticket

[#185801876](https://www.pivotaltracker.com/story/show/185801876)

## Code author checklist

### DIFF
--- a/web/src/components/tasks/NaicsCodeInput.tsx
+++ b/web/src/components/tasks/NaicsCodeInput.tsx
@@ -1,6 +1,7 @@
 import { Content, ExternalLink } from "@/components/Content";
 import { GenericTextField } from "@/components/GenericTextField";
 import { SecondaryButton } from "@/components/njwds-extended/SecondaryButton";
+import { NeedsAccountContext } from "@/contexts/needsAccountContext";
 import { IsAuthenticated } from "@/lib/auth/AuthContext";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { useUpdateTaskProgress } from "@/lib/data-hooks/useUpdateTaskProgress";
@@ -10,7 +11,7 @@ import { NaicsCodeObject, Task } from "@/lib/types/types";
 import { templateEval, useMountEffectWhenDefined } from "@/lib/utils/helpers";
 import { Business, LookupIndustryById } from "@businessnjgovnavigator/shared";
 import { FormControl, FormControlLabel, Radio, RadioGroup } from "@mui/material";
-import React, { ReactElement, useMemo, useState } from "react";
+import React, { ReactElement, useContext, useMemo, useState } from "react";
 
 interface Props {
   onSave: () => void;
@@ -33,6 +34,8 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
   const [isInvalid, setIsInvalid] = useState<NaicsErrorTypes | undefined>(undefined);
   const [isLoading, setIsLoading] = useState<boolean>(false);
   const [displayInputState, setDisplayInput] = useState<boolean>(false);
+  const { isAuthenticated, setShowNeedsAccountModal } = useContext(NeedsAccountContext);
+
   const { queueUpdateTaskProgress } = useUpdateTaskProgress();
   const userDataFromHook = useUserData();
   const business = props.CMS_ONLY_fakeBusiness ?? userDataFromHook.business;
@@ -70,6 +73,11 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
   }, business);
 
   const saveNaicsCode = async (): Promise<void> => {
+    if (isAuthenticated !== IsAuthenticated.TRUE) {
+      setShowNeedsAccountModal(true);
+      return;
+    }
+
     if (!updateQueue) {
       return;
     }
@@ -101,6 +109,11 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
   };
 
   const handleChange = (value: string): void => {
+    if (isAuthenticated !== IsAuthenticated.TRUE) {
+      setShowNeedsAccountModal(true);
+      return;
+    }
+
     setNaicsCode(value);
     if (value.length === LENGTH) {
       if (getCode(value)) {
@@ -141,8 +154,12 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
               name={"naics-radio-group"}
               value={naicsCode}
               onChange={(event: React.ChangeEvent<{ name?: string; value: string }>): void => {
-                setNaicsCode(event.target.value);
                 if (displayInput) setDisplayInput(false);
+                if (isAuthenticated !== IsAuthenticated.TRUE) {
+                  setShowNeedsAccountModal(true);
+                  return;
+                }
+                setNaicsCode(event.target.value);
                 setInProgress();
               }}
               data-testid="post-onboarding-radio-btn"
@@ -178,10 +195,13 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
                 labelPlacement="end"
                 data-testid="naics-radio-input"
                 onChange={(): void => {
+                  if (isAuthenticated !== IsAuthenticated.TRUE) {
+                    setShowNeedsAccountModal(true);
+                    return;
+                  }
                   setDisplayInput(true);
                   setNaicsCode("");
                   setIsInvalid(undefined);
-                  setInProgress();
                 }}
                 value=""
                 checked={displayInput}
--- a/web/src/components/tasks/NaicsCodeTask.test.tsx
+++ b/web/src/components/tasks/NaicsCodeTask.test.tsx
@@ -24,7 +24,7 @@ import {
   TaxFilingState,
 } from "@businessnjgovnavigator/shared";
 import { createTheme, ThemeProvider } from "@mui/material";
-import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { fireEvent, render, screen, waitFor, within } from "@testing-library/react";
 
 jest.mock("@/lib/roadmap/buildUserRoadmap", () => ({ buildUserRoadmap: jest.fn() }));
 jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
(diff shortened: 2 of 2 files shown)


web/src/components/tasks/NaicsCodeInput.tsx
```diff
@@ -141,9 +144,14 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
...
               onChange={(event: React.ChangeEvent<{ name?: string; value: string }>): void => {
-                setNaicsCode(event.target.value);
                 if (displayInput) setDisplayInput(false);
-                setInProgress();
+                if (props.isAuthenticated === IsAuthenticated.TRUE) {
+                  setNaicsCode(event.target.value);
+                  setInProgress();
+                } else {
+                  setShowNeedsAccountModal(true);
+                  setDisplayInput(false);
```
aloverso, line 153: What's the purpose of this line?

web/src/components/tasks/NaicsCodeTask.test.tsx
```diff
@@ -427,5 +427,29 @@ describe("<NaicsCodeTask />", () => {
...
+      });
+      renderPage();
+
+      fireEvent.click(screen.getByTestId("naics-radio-621399"));
+
+      expect(setShowNeedsAccountModal).toHaveBeenCalledWith(true);
+      expect(setShowNeedsAccountModal).toHaveBeenCalledTimes(1);
+
+      fireEvent.click(screen.getByTestId("naics-radio-input"));
+      expect(setShowNeedsAccountModal).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
AnJuHyppolite: Done!

web/src/components/tasks/NaicsCodeInput.tsx
AnJuHyppolite: I set `setDisplayInput(false)` so that when we click the NAICS code input radio button as a guest, the input field does not display. Otherwise, it does display and even though the modal appears, the task progress changes from NOT STARTED to COMPLETED.

jphechter: Similar to what we were trying to accomplish with the task status, we want to prevent a value from being saved, not undo a change to achieve the desired outcome.

I don't know if we necessarily need `setDisplayInput(false)`. If we do need it, then we should prevent that value from getting set in the first place.

aloverso: This is part of what's causing one of two issues I'm seeing. See my main comment for details

aloverso:
I see at least 2 cases where this isn't working when I test the application:

1. As a guest user, when my industry is "generic", then the radio buttons don't exist, and the task is incorrectly allowing me to type and save a NAICS code in Guest Mode.
image

2. Due to the changes around the `displayInput`, the text input no longer disappears when I select a radio button. As a signed-in user: first, select "Search for other NAICS codes" radio button. Then, select one of the recommended radio buttons. The display does not go away, and now both radios are selected. 
image

web/src/components/tasks/NaicsCodeInput.tsx
```diff
@@ -101,15 +104,19 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
   };
 
   const handleChange = (value: string): void => {
-    setNaicsCode(value);
-    if (value.length === LENGTH) {
-      if (getCode(value)) {
-        setIsInvalid(undefined);
+    if (isAuthenticated === IsAuthenticated.FALSE && business?.profileData.industryId === "generic") {
```
jphechter, line 107:
```suggestion
    if (isAuthenticated === IsAuthenticated.FALSE {
```
[sand] We don't need to actually consider the industry type in this check. It can be limited to just checking for authentication.

Consider this, is the important part the industry selection or that the user is not registered? While this field will only display when the user's industry selection is "generic", we don't need to bake that into this check. We don't want to say "if you're _not_ logged in and your industry is _not_ generic then you _can_ enter content in this field".

web/src/components/tasks/NaicsCodeInput.tsx
jphechter: @AnJuHyppolite Can this get removed?

web/src/components/tasks/NaicsCodeInput.tsx
aloverso: +1 This is a pebble, not sand, imo

web/src/components/tasks/NaicsCodeInput.tsx
```diff
@@ -101,15 +104,19 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
...
+      setShowNeedsAccountModal(true);
+    } else {
```
aloverso, line 109: [pebble] I'd prefer to follow our our patterns and put the isAuthenticated if-check into a guard clause instead of putting the main body of the function into an else-clause.

web/src/components/tasks/NaicsCodeInput.tsx
```diff
@@ -141,9 +148,14 @@ export const NaicsCodeInput = (props: Props): ReactElement => {
               name={"naics-radio-group"}
               value={naicsCode}
               onChange={(event: React.ChangeEvent<{ name?: string; value: string }>): void => {
-                setNaicsCode(event.target.value);
                 if (displayInput) setDisplayInput(false);
-                setInProgress();
+                if (isAuthenticated === IsAuthenticated.TRUE) {
```
aloverso, line 152: I would also prefer to see this in a guard clause with the main logic of the function at the "top-level" instead of within an if statement

web/src/components/tasks/NaicsCodeTask.test.tsx
```diff
@@ -412,19 +412,59 @@ describe("<NaicsCodeTask />", () => {
...
+    it("opens Needs Account modal on NAICS Code radio button click", () => {
+      initialBusiness = generateBusiness({
+        profileData: generateProfileData({ naicsCode: "", industryId: "acupuncture" }),
+      });
       renderPage();
-      expect(screen.getByText(`Register & ${Config.determineNaicsCode.saveButtonText}`)).toBeInTheDocument();
+
+      fireEvent.click(screen.getByTestId("naics-radio-621399"));
+
+      expect(setShowNeedsAccountModal).toHaveBeenCalledWith(true);
```
aloverso, line 423: [boulder] I think it'd be valuable to also check that the radio buttons are not actually checked. (or on the input box, that there is no value typed into it)

Imagine someone is moving some code around, or say - refactoring the existing logic to use a guard clause:

```
if (isAuthenticated !== IsAuthenticated.FALSE) {
  setShowNeedsAccountModal(true);
} 

setNaicsCode(value);
      if (value.length === LENGTH) {
        ... and so on
```

What's wrong with this code? At first glance, it seems okay. And more importantly, all these tests would pass using this code.

But the code is incorrect - it's missing the `return` in the guard clause, which would mean that it would execute BOTH the open modal AND the rest of the function to save the input value.

The tests should be written in a way that describes the complete behavior so that writing it in this way would cause a failing test, indicating the problem.

This can be done by adding an additional expect statement that expects the input value to be empty, or that the radio input HTML element is not `checked`

web/src/components/tasks/NaicsCodeInput.tsx
aloverso: ^^^

web/src/components/tasks/NaicsCodeTask.test.tsx
jphechter: It looks like this has been added on `line 430` - solid add!

web/src/components/tasks/NaicsCodeTask.test.tsx
```diff
@@ -417,14 +417,53 @@ describe("<NaicsCodeTask />", () => {
...
+        profileData: generateProfileData({ naicsCode: "", industryId: "acupuncture" }),
+      });
+      renderPage();
+
+      fireEvent.click(screen.getByTestId("naics-radio-input"));
+
+      expect(setShowNeedsAccountModal).toHaveBeenCalledWith(true);
       await waitFor(() => {
-        return expect(setShowNeedsAccountModal).toHaveBeenCalledWith(true);
+        expect(screen.getByTestId("naics-radio-input")).not.toBeChecked();
```
aloverso, line 444: Same as the above comment regarding this never failing.

Here is the correct line of code that needs to be used here. Note that the waitFor is unnecessary and should be removed.
```
expect(within(screen.getByTestId("naics-radio-input")).getByRole("radio")).not.toBeChecked();
```

web/src/components/tasks/NaicsCodeTask.test.tsx
```diff
@@ -417,14 +417,53 @@ describe("<NaicsCodeTask />", () => {
...
+    describe("guest mode - generic industry", () => {
+      it("opens modal when a user types in NAICS code input field", async () => {
+        initialBusiness = generateBusiness({
+          profileData: generateProfileData({ naicsCode: "", industryId: "generic" }),
+        });
+        renderPage();
+
+        fireEvent.change(screen.getByLabelText("Save NAICS Code"), { target: { value: "123456" } });
+
+        await expect(setShowNeedsAccountModal).toHaveBeenCalledWith(true);
```
aloverso, line 466: This is missing the expectation that nothing gets typed into the radio field. Please add the following assertion as well:

```
expect(screen.getByLabelText("Save NAICS Code")).toHaveValue('');
```

Also note that the `await` is unnecessary on the line above and should be removed.

web/src/components/tasks/NaicsCodeTask.test.tsx
```diff
@@ -417,14 +417,53 @@ describe("<NaicsCodeTask />", () => {
       expect(screen.getByText(`Register & ${Config.determineNaicsCode.saveButtonText}`)).toBeInTheDocument();
     });
 
-    it("opens Needs Account modal on save button click", async () => {
```
aloverso, line 420: This test is still necessary, because I could comment out your code on lines 76-79 and all the tests would still pass. 

However, just bringing back this test as it was written wouldn't be enough because it fires a change event and THAT triggers the modal, so the test-as-written would still pass even when I remove lines 76-79. We need a test that drives out and necessitates that guard clause.

Please add the following test:

```
    it("opens Needs Account modal on save button click", () => {
      renderPage();
      fireEvent.click(screen.getByText(`Register & ${Config.determineNaicsCode.saveButtonText}`));
      expect(setShowNeedsAccountModal).toHaveBeenCalledWith(true);
    });
```

PR #6404, opened 2023-09-20
feat: [#185481806] add BRC quick action tile

aloverso:
### Ticket

[#185481806](https://www.pivotaltracker.com/story/show/#185481806)

## Code author checklist


web/src/pages/dashboard.tsx
```diff
@@ -122,6 +131,8 @@ const DashboardPage = (props: Props): ReactElement => {
     dataTestId: "fromAdditionalBusiness-alert",
   });
 
+  const operatingPhase = LookupOperatingPhaseById(business?.profileData.operatingPhase);
```
AnJuHyppolite, line 134: Nice refactor!

AnJuHyppolite (approved):

web/src/components/dashboard/QuickActionTile.tsx
```diff
@@ -0,0 +1,42 @@
...
+    router.push(`${ROUTES.quickActions}/${props.url}`);
+  };
+
+  const getPadding = (): string => {
+    if (isMobile) {
+      return "0.25rem 0 0 0";
+    }
+
+    return props.index % 2 === 0 ? "0 0.25rem 0 0" : "0 0 0 0.25rem";
+  };
```
adidner, line 27: Isn't this a bit of an anti pattern? Wouldn't it be better to do this as a matter of css

https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

image

web/src/components/dashboard/QuickActionTile.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    setMobile(false);
+    render(<QuickActionTile name="some name" url="some-url" index={1} />);
+    expect(screen.getByTestId("quick-action-1")).toHaveStyle("padding: 0px 0px 0px 0.25rem;");
+  });
+
+  it("sets top padding when mobile", () => {
+    setMobile(true);
+    render(<QuickActionTile name="some name" url="some-url" index={0} />);
+    expect(screen.getByTestId("quick-action-0")).toHaveStyle("padding: 0.25rem 0px 0px 0px;");
+  });
```
adidner, line 57: A bit strange to be doing things as styles, also having tests for them? Do we want to test styles in some cases now?

web/src/components/dashboard/QuickActionTile.tsx
adidner: Is this not possible for some reason I'm not seeing at a glance?

web/src/lib/domain-logic/routes.ts
```diff
@@ -7,6 +7,7 @@ export const ROUTES = {
   dashboard: "/dashboard",
   onboarding: "/onboarding",
   unsupported: "/unsupported",
+  quickActions: "/actions",
```
adidner, line 10: I think this should also be added to findDeadLinks.ts

web/src/lib/static/loadCertifications.ts
```diff
@@ -8,6 +7,7 @@ import { getFileNameByUrlSlug } from "./helpers";
 const certificationDir = path.join(process.cwd(), "..", "content", "src", "certifications");
 const archivedCertificationDir = path.join(process.cwd(), "..", "content", "src", "archived-certifications");
 
+type PathParams<P> = { params: P; locale?: string };
```
adidner, line 10: So we re-define this like 7 times, is there a reason we don't have a unified type that we're exporting? (Or I assume there is a reason that I don't know that would be good if I did know)

web/src/components/dashboard/QuickActionTile.test.tsx
aloverso: I tested the styles here because there's logic in when they get applied. You know, same as how we'd test an href or something.

web/src/lib/domain-logic/routes.ts
aloverso: good catch!

web/src/pages/actions/[quickActionUrlSlug].tsx
```diff
@@ -0,0 +1,89 @@
...
+interface Props {
+  quickAction: QuickAction;
+}
+
+const Config = getMergedConfig();
+export const taxFilingMethodMap: Record<TaxFilingMethod, string> = {
+  online: Config.filingDefaults.onlineTaxFilingMethod,
+  "paper-or-by-mail-only": Config.filingDefaults.paperOrMailOnlyTaxFilingMethod,
+  "online-required": Config.filingDefaults.onlineRequiredTaxFilingMethod,
+  "online-or-phone": Config.filingDefaults.onlineOrPhoneTaxFilingMethod,
```
Comment (adidner) [inline web/src/pages/actions/[quickActionUrlSlug].tsx:27]: online is not a string? but the others are?

web/src/lib/static/loadCertifications.ts
aloverso: Well, I didn't like the old pattern, where randomly one file (fundings, I think) defined it and for some reason all these other files imported from fundings? That pattern breaks domain boundaries because loading X (quick actions, certs, etc) should certainly not depend on fundings.

Putting it in a top level import would work, but since we were already redefining it in some, I just leaned into that pattern for this refactor. Incremental cleanup, but I agree it could go further to refactor out the duplication altogether

web/src/pages/actions/[quickActionUrlSlug].tsx
Comment (aloverso) [reply web/src/pages/actions/[quickActionUrlSlug].tsx:27]: oops, copy paste error! this doesn't belong here at all

Comment (aloverso) [reply web/src/pages/actions/[quickActionUrlSlug].tsx:27]: But also, to answer your question because this code IS elsewhere in the app: online is a string too, here. In javascript, all keys of an object are strings and the quotation marks are optional. They're only required in the other ones because of the hyphens.

It's like how `const a = { a: 1, b: 2 }` and `const a = { "a": 1, "b": 2 }` are the exact same piece of code in JS

web/src/components/dashboard/QuickActionTile.test.tsx
adidner: Yeah makes, sense, I saw the test first and made the comment before seeing the implementation

web/src/lib/static/loadCertifications.ts
adidner: Gotcha

web/src/pages/actions/[quickActionUrlSlug].tsx
Comment (adidner) [reply web/src/pages/actions/[quickActionUrlSlug].tsx:27]: Hmm, interesting and good to know about how it work underlying. Still probably don't want to mix the "" vs not "" in the same object though, just for it being pretty sake.

adidner (approved):
A few minor comments that you seem to already be on top of. Otherwise good

web/src/components/dashboard/QuickActionTile.tsx
aloverso: The "reason that you're not seeing" is that you are a Smartie and I am stuck in JS-brain. will do!

PR #6420, opened 2023-09-21
Add analytics event to 'Skip to main content'

jphechter:
## Description

Create a new analytics event to fire `onClick` for "Skip to main content" button.

### Ticket

[184226156](https://www.pivotaltracker.com/story/show/184226156)

### Approach

- Created an independent component
- Moved copy to config
- Added analytics event
- Added test to confirm event fires

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/config.json
+++ b/content/src/fieldConfig/config.json
@@ -21,6 +21,9 @@
     "requiredTasksSingular": "${numTasks} required task",
     "requiredTasksPlural": "${numTasks} required tasks"
   },
+  "skipToMainContent": {
+    "buttonText": "Skip to main content"
+  },
   "betaBar": {
     "betaModalButtonText": "Share your feedback!",
     "betaMainText": "This site is in beta."
--- a/web/src/components/SkipToMainContent.test.tsx
+++ b/web/src/components/SkipToMainContent.test.tsx
@@ -1,0 +1,31 @@
+import { SkipToMainContent } from "@/components/SkipToMainContent";
+import { getMergedConfig } from "@/contexts/configContext";
+import analytics from "@/lib/utils/analytics";
+import { fireEvent, render, screen } from "@testing-library/react";
+
+const Config = getMergedConfig();
+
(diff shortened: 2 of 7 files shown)


content/src/fieldConfig/config.json
```diff
@@ -21,6 +21,9 @@
     "requiredTasksSingular": "${numTasks} required task",
     "requiredTasksPlural": "${numTasks} required tasks"
   },
+  "skipToMainContent": {
```
FarazA22, line 24: [dust] We are trying to actively remove data from the `config` file. Could this live elsewhere?

content/src/fieldConfig/config.json
```diff
@@ -21,6 +21,9 @@
...
+    "buttonText": "Skip to main content"
```
adidner, line 25: I also imagined it with more Capital letters? `Skip to Main Content`

web/src/components/SkipToMainContent.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+        },
+      },
+    },
+  };
+}
+
+describe("<SkipToMainContent>", () => {
+  it("fires skip_to_main_content analytics event when link is clicked", () => {
+    render(<SkipToMainContent />);
+    fireEvent.click(screen.getByText("Skip to main content"));
```
adidner, line 25: Shouldn't this grab the string from Config? This will break if the content team changes the name right?

content/src/fieldConfig/config.json
jphechter: This is maintains the current implementation. We can ask Content, but I don't want to modify this without explicit instruction.

web/src/components/SkipToMainContent.test.tsx
jphechter: That is 100% why I added the string to config in the first place and then promptly didn't add it to the test. Thank you!

content/src/fieldConfig/config.json
jphechter: I'm open to suggestions for this one.

PR #6422, opened 2023-09-22
chore: [#185715307] refactor and add error bar to naics code input

FarazA22:
## Description

Added error bar to input and updated labels on task page
Refactored/ moved around the NaicsInput component so the code is easy to read.
Consolidated input content to profile.json and updated cms previews.

### Ticket

https://www.pivotaltracker.com/story/show/185715307

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/naics-code.json
+++ b/content/src/fieldConfig/naics-code.json
@@ -4,7 +4,6 @@
     "successMessage": "NAICS code ${code} has been saved.",
     "invalidValidationErrorText": "Your NAICS code is invalid",
     "hasSavedCodeHeader": "Your NAICS Code",
-    "inputLabel": "**Save your 6-digit NAICS code here so you can reference it later:**",
     "missingNAICSCodePlaceholder": "(You have not saved a NAICS Code yet)",
     "findCodeHeader": "Find Your NAICS Code",
     "findCodeBodyText": "Use the resource listed below to search and find your NAICS code. **This information is used for data and analysis purposes by State and Federal governments, and does not directly impact your business.**",
--- a/content/src/fieldConfig/profile.json
+++ b/content/src/fieldConfig/profile.json
@@ -207,7 +207,8 @@
           "headerContextualInfo": "naics",
           "editText": "Edit",
           "addText": "Add",
-          "notEnteredText": "Not entered"
+          "notEnteredText": "Not entered",
+          "altDescription": "Save your 6-digit NAICS code here so you can reference it later:"
         }
       },
       "isChildcareForSixOrMore": {
--- a/web/public/mgmt/config.yml
+++ b/web/public/mgmt/config.yml
@@ -2489,7 +2489,6 @@ collections:
                   hint: This field supports Markdown,
                 }
               - { label: naics.com Link, name: naicsLink, widget: string, hint: This field supports Markdown }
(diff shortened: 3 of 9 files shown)


content/src/fieldConfig/profile.json
```diff
@@ -207,7 +207,8 @@
           "headerContextualInfo": "naics",
           "editText": "Edit",
           "addText": "Add",
-          "notEnteredText": "Not entered"
+          "notEnteredText": "Not entered",
+          "altDescription": "Save your 6-digit NAICS code here so you can reference it later:"
```
AnJuHyppolite, line 211: I don't think we should move, _"Save your 6-digit NAICS code here so you can reference it later:"_ in `profile.json`. My understanding is that the copy in `profile.json` is copy that lives in the **Business Profile** section of the site.

content/src/fieldConfig/naics-code.json
```diff
@@ -4,7 +4,6 @@
     "successMessage": "NAICS code ${code} has been saved.",
     "invalidValidationErrorText": "Your NAICS code is invalid",
     "hasSavedCodeHeader": "Your NAICS Code",
-    "inputLabel": "**Save your 6-digit NAICS code here so you can reference it later:**",
```
AnJuHyppolite, line 7: I think we should leave this here. See comment re:`profile.json` update.

web/public/mgmt/config.yml
```diff
@@ -2489,7 +2489,6 @@ collections:
                   hint: This field supports Markdown,
                 }
               - { label: naics.com Link, name: naicsLink, widget: string, hint: This field supports Markdown }
-              - { label: Input Label, name: inputLabel, widget: string, hint: This field supports Markdown }
```
AnJuHyppolite, line 2492: We would add this back in if we don't make the `profile.json` change.

web/public/mgmt/config.yml
```diff
@@ -2506,6 +2505,40 @@ collections:
...
+                        collapsed: false
+                        widget: object
+                        fields:
+                          - { label: "Header (shared on profile/task)", name: "header", widget: "string" }
+                          - {
+                              label: "Alt Description Text - Renders On Task Page",
+                              name: "altDescription",
+                              widget: "string",
+                            }
+
```
AnJuHyppolite, line 2541: Not necessary if we don't make the update in `profile.json`.

web/src/pages/mgmt/cms.tsx
```diff
@@ -123,6 +122,10 @@ const CMS = dynamic(
       registerPreview(CMS, "business-formation-interim-success-page", FormationInterimSuccessPreview);
       registerPreview(CMS, "business-formation-success-page", FormationSuccessPreview);
 
+      // Naics Code
```
AnJuHyppolite, line 125: delete commented out code

web/src/styles/components/errors.scss
```diff
@@ -1,16 +1,16 @@
...
   margin-right: -17px !important;
   position: relative;
 }
 
 .input-error-bar.error {
+  border-left-width: 4px;
   border-left-color: #d54309;
+  border-left-style: solid;
+  left: -17px !important;
   margin-bottom: 1em;
```
AnJuHyppolite, line 14: We should be using `rem` in place of `px` for scalability/accessibility reasons.

PR #6440, opened 2023-09-25
Starter kit title #186099467

aloverso:
### Ticket

[#186099467](https://www.pivotaltracker.com/story/show/#186099467)

## Code author checklist

- [ x] I have not used any relative imports

### DIFF
--- a/content/src/fieldConfig/config.json
+++ b/content/src/fieldConfig/config.json
@@ -308,6 +308,8 @@
   "headerDefaults": {
     "noUserNameHeaderText": "Hello!",
     "defaultHeaderText": "Hello, ${name}",
+    "starterKitText": "Your ${industry} Starter Kit",
+    "genericStarterKitText": "Your Starter Kit",
     "newJerseyDateBodyText": "ET",
     "guestModeToProfileButtonText": "Guest Account",
     "genericToProfileButtonText": "Business Profile"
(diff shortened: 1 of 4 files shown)


content/src/fieldConfig/config.json
```diff
@@ -308,6 +308,8 @@
   "headerDefaults": {
     "noUserNameHeaderText": "Hello!",
     "defaultHeaderText": "Hello, ${name}",
+    "starterKitText": "Your ${industry} Starter Kit",
+    "genericStarterKitText": "Your Starter Kit",
```
jphechter, line 312: Question regarding implementation. At one point, the onboarding header text was something along the lines of "Tell us about your ${OoS} business" and we called a hook that conditionally rendered "Out of State" or nothing at all. Given that this is essentially the same situation, we could do the same thing here.

Any particular reason you opted for 2 separate lines as opposed to conditionally rendering `industry`?

jphechter (approved):

content/src/fieldConfig/config.json
aloverso: I don't think it's the same situation. The `${Oos}` approach was used as a global keyword specifically for the content team to be able to insert into _any_ piece of content, without additional coding effort.

Whereas here, we could certainly have one string for `"Your ${industry} Starter Kit"` and remove the `${industry}` section part altogether if they are generic, but I think that introduces 2 pieces of complexity that aren't worth solving:
- also having to conditionally remove the space so that there aren't two spaces next to each other.
- The team kinda picked the "generic" text on a whim, it wasn't originally defined in the story. Seems pretty likely it could change. And if the content team ever decided they wanted the generic text to say just "Starter Kit" or something ELSE, then they wouldn't be able to make that change and we'd have to have a whole additional story just to move FROM having 1 piece of content to having 2, so that those could independently co-exist. This approach allows them to not be coupled together, and you can change one without changing the other.

PR #6446, opened 2023-09-28
Storybook molecules buttons

adidner:
## Description

Storybook Molecules for getting Buttons up to date with the design system,
Also refactored the quaternary button a bit 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/184606443)

### Steps to Test

run storybook via the `yarn run storybook` command from the we folder and check out the button section

## Code author checklist

### DIFF
--- a/web/src/components/dashboard/HideableTasks.tsx
+++ b/web/src/components/dashboard/HideableTasks.tsx
@@ -1,5 +1,5 @@
+import { SecondaryButton } from "@/components//njwds-extended/SecondaryButton";
 import { Roadmap } from "@/components/dashboard/Roadmap";
-import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
 import { Icon } from "@/components/njwds/Icon";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { useRoadmap } from "@/lib/data-hooks/useRoadmap";
@@ -39,11 +39,7 @@ export const HideableTasks = (): ReactElement => {
       <div className={`${isTabletAndUp ? "flex flex-align-center" : ""} margin-bottom-205`}>
         <h2 className="margin-bottom-0 text-medium">{Config.dashboardDefaults.upAndRunningTaskHeader}</h2>
         <div className={`mla ${isTabletAndUp ? "" : "margin-top-2"}`}>
-          <UnStyledButton
-            style={"transparentBgColor"}
-            className={"usa-tag text-normal text-base border-1px border-base-light hide-unhide-button"}
-            onClick={handleToggleClick}
-          >
+          <SecondaryButton size={"small"} isColor={"border-base-light"} onClick={handleToggleClick}>
             <div className="fdr fac">
               <Icon>{business?.preferences.isHideableRoadmapOpen ? "visibility_off" : "visibility"}</Icon>
               <span className="margin-left-05 line-height-sans-2">
@@ -52,7 +48,7 @@ export const HideableTasks = (): ReactElement => {
                   : Config.dashboardDefaults.showTaskText}
               </span>
             </div>
-          </UnStyledButton>
+          </SecondaryButton>
         </div>
       </div>
       <hr className="margin-bottom-3 margin-top-0" aria-hidden={true} />
--- a/web/src/components/njwds-extended/PrimaryButton.tsx
+++ b/web/src/components/njwds-extended/PrimaryButton.tsx
@@ -4,10 +4,10 @@ import { forwardRef, ReactElement, Ref } from "react";
 export type PrimaryButtonColors =
   | "primary"
   | "secondary"
-  | "accent-cool-lightest"
   | "accent-cool-darker"
   | "accent-cooler"
-  | "info";
+  | "info"
+  | "success-extra-light";
 
 interface Props extends GenericButtonProps {
   isColor: PrimaryButtonColors;
@@ -16,10 +16,10 @@ interface Props extends GenericButtonProps {
 const colors = {
   primary: "usa-button",
   secondary: "usa-button usa-button--secondary",
-  "accent-cool-lightest": "usa-button btn-accent-cool-lightest",
   "accent-cool-darker": "usa-button btn-accent-cool-darker",
   "accent-cooler": "usa-button btn-accent-cooler",
   info: "usa-button btn-info",
+  "success-extra-light": "usa-button btn-success-extra-light",
 };
 
 export const PrimaryButton = forwardRef(function PrimaryButton(
--- a/web/stories/Molecules/Button/Calendar.stories.tsx
+++ b/web/stories/Molecules/Button/Calendar.stories.tsx
@@ -1,0 +1,27 @@
+import { CalendarEventItem } from "@/components/filings-calendar/CalendarEventItem";
+import { defaultDateFormat, parseDateWithFormat } from "@businessnjgovnavigator/shared";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+
+export default {
+  title: "Molecules/Button/Calendar",
+  component: CalendarEventItem,
+  decorators: [(Story) => <div style={{ width: "212px", height: "54px" }}>{Story()}</div>],
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/vAa8neaM0JJSmldck5vlBC/BFS-Design-System-(Sprint-33%2B)?node-id=2421%3A3127&t=k6Nkwsn3QXIhASwW-1",
+    },
+  },
+} as ComponentMeta<typeof CalendarEventItem>;
+
+const Template: ComponentStory<typeof CalendarEventItem> = ({ ...args }) => <CalendarEventItem {...args} />;
+export const Regular = Template.bind({});
+
+const currentDate = parseDateWithFormat(`2024-02-15`, "YYYY-MM-DD");
+
+Regular.args = {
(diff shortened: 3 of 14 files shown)


web/stories/Molecules/Button/PrimaryRegular.stories.tsx
```diff
@@ -24,25 +25,16 @@ PrimaryColor.args = {
...
-AccentCoolDarkerColor.args = {
-  isColor: "accent-cool-darker",
-  children: "button",
+IconRight.args = {
+  isColor: "primary",
+  children: (
+    <>
+      <>Button</>
+      <Icon className="usa-icon--size-3 margin-left-05">arrow_drop_down</Icon>
+    </>
```
adidner, line 36: I notice we have this pattern of icons with slightly different styles a bunch which should probably be consolidates in some fashion but for the moment I'm going to leave it as there is just a bit to much diversity to tackle it in this ticket

web/src/components/njwds-extended/PrimaryButton.tsx
```diff
@@ -3,23 +3,23 @@ import { forwardRef, ReactElement, Ref } from "react";
...
 export type PrimaryButtonColors =
   | "primary"
-  | "secondary"
-  | "accent-cool-lightest"
+  | "feedback"
   | "accent-cool-darker"
   | "accent-cooler"
-  | "info";
+  | "info"
+  | "tertiary";
```
adidner, line 10: So I'm not sure exactly what we want precent to be for this. 

We have a SecondaryButton compoennt and QuaternaryButton component

But the primary button has a bunch of sub types which are used for specific things so I'm confused as to where the line is or should be

web/src/components/BetaBar.tsx
```diff
@@ -14,7 +14,7 @@ export const BetaBar = (): ReactElement => {
     >
       <span className="margin-left-1 margin-right-1">{Config.betaBar.betaMainText}</span>
       <PrimaryButton
-        isColor="secondary"
+        isColor="feedback"
```
FarazA22, line 17: [pebble] I think this color should stay as is, when you click on an element within Figma, the color itself is identified. If we're naming the colors exactly what we are looking at in Figma it's a lot easier to figure out what prop we need to apply.
 
Screenshot 2023-09-28 at 3 25 17 PM

web/src/components/dashboard/HideableTasks.tsx
```diff
@@ -39,11 +39,7 @@ export const HideableTasks = (): ReactElement => {
       <div className={`${isTabletAndUp ? "flex flex-align-center" : ""} margin-bottom-205`}>
         <h2 className="margin-bottom-0 text-medium">{Config.dashboardDefaults.upAndRunningTaskHeader}</h2>
         <div className={`mla ${isTabletAndUp ? "" : "margin-top-2"}`}>
-          <UnStyledButton
-            style={"transparentBgColor"}
-            className={"usa-tag text-normal text-base border-1px border-base-light hide-unhide-button"}
-            onClick={handleToggleClick}
-          >
+          <QuaternaryButton size={"regular"} onClick={handleToggleClick}>
```
FarazA22, line 42: [Dust] When I remove the content from both the secondary button component and the QuaternaryButton button component, it results in a container with a border and padding. I believe both of these buttons should be categorized as secondary buttons due to their similar styling. The elements that are passed into the button component should not dictate the name of the component.

If you decide to keep the button, I suggest creating a 'buttons' folder to centralize all the various button components we use throughout the application.

secondary
Screenshot 2023-09-28 at 3 36 14 PM

quaternary
Screenshot 2023-09-28 at 3 34 08 PM

web/src/components/njwds-extended/PrimaryButton.tsx
FarazA22: [dust] When I consider the distinctions between buttons, I examine the styling applied at the parent level. For a primary button, the container itself has a solid fill, while the secondary button has a border, and an unstyled button has neither. 

The children elements of a button, including text or graphics, should not influence the decision on how many buttons to create or how to distinguish between them. For instance, in the case of the 'primary' button, the background color is determined by the 'primary button colors' property.

The available colors for the primary colors property are directly linked to the default background colors. As mentioned earlier, you can easily view them by clicking directly into the button.

web/src/components/njwds-extended/PrimaryButton.tsx
```diff
@@ -3,23 +3,23 @@ import { forwardRef, ReactElement, Ref } from "react";
...
 
 const colors = {
   primary: "usa-button",
-  secondary: "usa-button usa-button--secondary",
-  "accent-cool-lightest": "usa-button btn-accent-cool-lightest",
+  feedback: "usa-button usa-button--secondary",
   "accent-cool-darker": "usa-button btn-accent-cool-darker",
   "accent-cooler": "usa-button btn-accent-cooler",
   info: "usa-button btn-info",
+  tertiary: "usa-button btn-tertiary",
```
FarazA22, line 22: [pebble] rename based on default color

web/stories/Molecules/Button/Calendar.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import { CalendarEventItem } from "@/components/filings-calendar/CalendarEventItem";
+import { defaultDateFormat, parseDateWithFormat } from "@businessnjgovnavigator/shared";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+
+export default {
+  title: "Molecules/Button/Calendar",
+  component: CalendarEventItem,
+  decorators: [(Story) => <div className="width-mobile">{Story()}</div>],
```
FarazA22, line 8: [pebble/sand] The styling on line 19 can be applied here `<div style={{ width: "212px", height: "54px" }}>` or remove this line altogether

The decorator key allows you to apply global styling to each individual story.

web/stories/Molecules/Button/Feedback.stories.tsx
```diff
@@ -18,25 +18,10 @@ const Template: ComponentStory<typeof PrimaryButton> = ({ children, ...args }) =
...
-Regular.args = {
-  children: (
-    <>
-      <ButtonIcon svgFilename="chat-processing" />
-      <span className="text-left">Text Here</span>
-    </>
-  ),
-  isColor: "accent-cool-lightest",
-  isFullWidthOnDesktop: true,
-  isTextAlignedLeft: true,
```
FarazA22, line 32: [dust] I believe this property was exclusively used for regular feedback design. As we are removing this styling, it makes sense to delete this property from the generic button component as well.

FarazA22: I believe the same principle should also apply to the 'isFullWidthOnDesktop' property.

web/stories/Molecules/Button/PrimaryLarge.stories.tsx
```diff
@@ -25,42 +25,10 @@ PrimaryColor.args = {
   isLargeButton: true,
 };
 
-export const SecondaryColor = Template.bind({});
-
-SecondaryColor.args = {
```
FarazA22, line 30: [dust] Here's an example of the design system not being up-to-date. We're currently using some of these colors for the primary large button on the homepage. I suggest keeping these stories and adding a comment to the design team to update the design system to include these colors as well.

Screenshot 2023-09-28 at 4 12 58 PM

web/stories/Molecules/Button/PrimaryRegular.stories.tsx
FarazA22: As I mentioned earlier, the children elements of a design button can be anything, including graphics or text. I don't believe we need to address this, but I'd like to discuss it further after you've had a chance to read my thoughts.

jphechter (commented):
I think Faraz's comments make sense. I'm in favor of keeping our property names in line with the design system where we can.

Not seeing anything additional I want to call out here. Overall good work.

web/stories/Molecules/Button/PrimaryRegular.stories.tsx
adidner: So the concern here is about inconsistency over the size of icons and the padding between icons and text which is all left to the parent and is already inconsistency between various places in the code base which is bad. I was thinking maybe we could obfuscate it somehow to being more determined via props but maybe not and maybe its not worth it or a good pattern. Just a thought

adidner: I'll for sure leave it for now at least

web/src/components/BetaBar.tsx
adidner: Sure

web/stories/Molecules/Button/Feedback.stories.tsx
adidner: So it appears isFullWidthOnDesktop is used in a number of places still 
image

adidner: I removed isTextLeftAligned

web/src/components/njwds-extended/PrimaryButton.tsx
adidner: Sure

web/stories/Molecules/Button/Calendar.stories.tsx
adidner: Neto, didn't know about that key. But very cool, placed in decorator for calendar and removed the decorator key from all the other button stories.

PR #6447, opened 2023-09-28
chore: [#186126038] fix paragraph spacing

cbrunefearless:
## Description

Context: https://njcio.slack.com/archives/C04SS2KJCCA/p1695311458063769

Solution: add 1rem bottom margin to paragraph elements.

Attempted fix:
```
 .usa-prose {
  p {
     margin-bottom: 1rem !important;
     margin-top: 0 !important;
```

Problem: it increased ALL paragraphs on the page, like in the roadmap tasks, and in the alert. Need to find a better way

### Ticket 

[186126038](https://www.pivotaltracker.com/story/show/)

## Code author checklist

### DIFF
--- a/content/src/quick-actions/registry-update-brc-amendment.md
+++ b/content/src/quick-actions/registry-update-brc-amendment.md
@@ -12,8 +12,6 @@ If your business registration information has changed since you last updated you
 
 Depending on the type of changes, this may include updating your business charter or obtaining a new `Business Registration Certificate (BRC)|business-reg-certificate` .
 
----
-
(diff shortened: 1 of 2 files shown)


aloverso:
@cbrunefearless This solves the immediate problem, but when we chatted with design, it seemed like they actually *did* want to increase inter-paragraph spacing to 1rem for content blocks like this. Do you think there's a way we could do it using that approach?

jphechter (approved):
If design is good with this approach, it LGTM

PR #6423, opened 2023-09-22
refactor: [#185628613] combine nonprofit provisions and partnership rights fields

aloverso:
### Ticket

[#185628613](https://www.pivotaltracker.com/story/show/#185628613)

## Code author checklist

- [ ] I have performed a self-review of my code
- [ ] I have created and/or updated relevant documentation, if necessary
- [ ] I have not used any relative imports
- [ ] I have checked for and removed instances of unused config from CMS
- [ ] I have pruned any instances of unused code
- [ ] I have not added any markdown to labels, titles and button text in config
- [ ] If I added/updated any values in `userData` (including `profileData`, `formationData` etc), then I added a new migration file
- [ ] If I added any new collections to the CMS config, then I updated the search tool and `cmsCollections.ts`

PR #6455, opened 2023-10-02
refactor: use FormationRadio component for formation radio questions

aloverso:
Intended as a pre-factor for [#184249726](https://www.pivotaltracker.com/story/show/184249726)

- rename PracticesLaw to WillPracticeLaw to keep consistent variable naming
- WillPracticeLaw and IsVeteranNonprofit now use the shared FormationRadio component instead of redefining radio button logic
- This actually fixes a bug that was introduced by #6423 (ticket [#185628613](https://www.pivotaltracker.com/story/show/185628613) ) where the PartnershipRights radio buttons lost their special label text.


web/src/components/tasks/business-formation/business/IsVeteranNonprofit.tsx
```diff
@@ -1,72 +1,27 @@
...
-            </FormHelperText>
-          )}
-        </FormControl>
-      </FormationField>
-    </WithErrorBar>
+    <FormationRadio
+      fieldName={fieldName}
+      title={Config.formation.fields.isVeteranNonprofit.description}
+      values={["true", "false"]}
+      onChange={(e): void => {
```
jphechter, line 16: [sand] It looks like almost all implementations of `<FormationRadio>` have the same `onChange` function. Thoughts on making this the default and optionally passing a function to override?

aloverso: the exception is in `NonprofitProvisions`

Every usage uses true/false and the same onChange function, which JSON-parses the "true" and "false" strings into boolean values that get saved.

Whereas the one usage in NonprofitProvisions uses "IN_FORM" / "IN_BYLAWS" which is what necessitates passing in the values, and is also different in the onChange because it won't work if it does the JSON parse.

The thing is that if we did a default, it'd have to go hand-in-hand with "true"/"false" as a default values array, because the two go hand-in-hand.

But I would be worried about the overall potential of someone using the component, and overriding the default values with "something"/"custom" and not realizing it's ALSO necessary to override the onChange as well. I'd want some way to tie them together or something

PR #6457, opened 2023-10-03
For you section empty state #185943408

AnJuHyppolite:
## Description

Added For You empty state message and image. The message and image display when users do not have cards in their For You section.

### Ticket

[#185943408](https://www.pivotaltracker.com/story/show/185943408)

## Code author checklist


web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +63,55 @@ describe("<SidebarCardsList />", () => {
...
+          section: "below-opportunities",
+        }),
+      ];
+      render(
+        <SidebarCardsList
+          topCards={topCards}
+          bottomCards={bottomCards}
+          fundings={[]}
+          hiddenFundings={[]}
+          certifications={[]}
```
FarazA22, line 106: [Boulder] This ticket mentions this the UI is also dependent on funding and certification cards. I don't see the logic for that piece.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -133,7 +141,7 @@ export const SidebarCardsList = (props: Props): ReactElement => {
           {showEmptyState() && (
             <div className="fdc fac margin-y-3">
               <h3 className="text-normal">{Config.dashboardDefaults.emptyOpportunitiesHeader}</h3>
-              <img src={`/img/signpost.svg`} className="" alt="" />
+              <img src={`/img/for-you-section.svg`} className="" alt="" />
```
somabadri, line 144: [pebble] you should add the alt text here as well.

web/src/components/dashboard/SidebarCardsList.test.tsx
somabadri: I think this is covered in the `SidebarCardsContainer.test.tsx` under `displays empty state when all opportunities are hidden because the empty state logic for fundings and certifications already existed, but I think the test can be updated to check that the new image is being displayed.

AnJuHyppolite: @FarazA22: My understanding is that Design just wanted us to change the image. The logic already exists in our code and shouldn't change. I sent you the Slack convo.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -114,6 +114,14 @@ export const SidebarCardsList = (props: Props): ReactElement => {
     <>
       <h2 className="h1-styling margin-top-0">{Config.dashboardDefaults.sidebarHeading}</h2>
       <hr className="margin-top-2 margin-bottom-3 bg-cool-lighter" aria-hidden={true} />
+      {props.topCards.length === 0 && props.bottomCards.length === 0 && (
+        <div data-testid="empty-for-you-message" className="fdc fac">
+          <div className="text-center margin-bottom-3">
+            <Content>{Config.dashboardDefaults.emptyForYouMessage}</Content>
+          </div>
+          <img src={`img/for-you-section.svg`} alt={"For You Section"} />
```
jphechter, line 122: [boulder] I don't think "For You Section" is an accurate alt name for this image. A key finding of our accessibility audit thus far has been that we need our alt descriptions to be meaningful. If the image doesn't actually provide meaningful information it's actually better to hide this from the screen reader.

For example, if you had an image of a bar graph conveying a particular statistic you would set the alt text to be "graph illustrating covid cases over the last 6 months" (or whatever it is). In this case, providing an accurate alt description "woman pointing at vector images" doesn't add to the user experience. It's better to hide the image.

[Here's the item from Quality Logic talking about this.](https://www.pivotaltracker.com/story/show/185791759)

web/src/components/dashboard/SidebarCardsList.tsx
jphechter, line 144: [sand] I know you didn't write this originally, but this is an opportunity to clean up. We don't need an empty className attribute.
```suggestion
              <img src={`/img/for-you-section.svg`} alt="" />
```

jphechter: From what I understand, the right approach for hiding images from the screen reader is to explicitly set `alt` to be an empty string. So this is good.

[pebble] We'll also want to set `aria-hidden="true"`

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +63,55 @@ describe("<SidebarCardsList />", () => {
     fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
     expect(mockAnalytics.event.for_you_card_unhide_button.click.unhide_cards).toHaveBeenCalledTimes(1);
   });
+
+  describe("empty state messaging", () => {
```
jphechter, line 67: [pebble] We had talked about how some of the names in `SidebarCardsList` may have been slightly confusing, and because the meanings of those names would be changing with this work it would be a good opportunity to rename. One of which was the function `showEmptyState`. I see that function is still in the `SidebarCardsList` component. By calling this describe "empty state messaging" it almost seems like we're testing different functionality.

I'm not saying this name has to change, but let's make sure that this describe title is reflective of the named behavior we want to test.

I think when we originally named this section we were trying to use "empty state messaging" as an all-encompassing way to cover both of the For You messages. Let's consider that in whatever name we use for this describe or as a guide for renaming the functionality in the component.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +63,55 @@ describe("<SidebarCardsList />", () => {
...
+          fundings={[]}
+          hiddenFundings={[]}
+          certifications={[]}
+          hiddenCertifications={[]}
+          displayFundings={false}
+          displayCertifications={false}
+        />
+      );
+
+      expect(screen.getByTestId("empty-for-you-message")).toBeInTheDocument();
```
jphechter, line 82: [pebble] We can get this item using `getByText()` and referencing the `config`. We can also drop the `test-id` on the component.

web/src/components/dashboard/SidebarCardsList.test.tsx
jphechter: [boulder] I would love to see the other message behavior currently covered by the function `showEmptyState` tested as well. It's not specifically covered by this ticket, but we are making changes to this component and the tests shouldn't take much effort. Additionally, we had questions about the intended behavior of that message in relation to this new work. Those tests will serve as documentation for us and help prevent this confusion in the future.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +63,55 @@ describe("<SidebarCardsList />", () => {
...
+      const bottomCards = [
+        generateSidebarCardContent({
+          id: "some-fake-bottom-card",
+          section: "below-opportunities",
+        }),
+      ];
+      render(
+        <SidebarCardsList
+          topCards={topCards}
+          bottomCards={bottomCards}
```
aloverso, line 103: [pebble] It's generally good practice to test edge cases. In these two tests we have one test that says "when there are 0 cards, show the Empty State Message" and one test that says "when there are 2 cards, do not show the Empty State Message"

It leaves the 1 case untested. Generally it's better to test the case right on the edge of the logic instead of something in the middle (like 2). I might recommend writing this test as only 1 card instead of 2 would be more useful.

Think of it like the AQI TQ example. If you're testing that an aqi>100 is "unhealthy" and less than 100 is "good" then you could write tests like:
```
expect(getAQI(50)).toEqual("good")
expect(getAQI(125)).toEqual("unhealthy")
```

This is correct testing, but the logic could vary and the test would still pass. Any of the following lines of code would get the test passing:
```
return aqi > 100 ? "unhealthy" : "good"
return aqi >= 100 ? "unhealthy" : "good"
return aqi > 80 ? "unhealthy" : "good"
return aqi > 120 ? "unhealthy" : "good"
```

So the tests, while accurate, don't actually test that the boundary line is 100 specifically. A better test would test the specific edge cases:
```
expect(getAQI(99)).toEqual("good")
expect(getAQI(100)).toEqual("good")
expect(getAQI(101)).toEqual("unhealthy")
```

the **only** line of code that would allow this to pass is:
```
return aqi >= 100 ? "unhealthy" : "good"
```

This is why we test boundary conditions. So I think this test could be improved by testing what happens when we have _one_ card in the For You column (instead of 2)

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +63,55 @@ describe("<SidebarCardsList />", () => {
...
+        <SidebarCardsList
+          topCards={[]}
+          bottomCards={[]}
+          fundings={[]}
+          hiddenFundings={[]}
+          certifications={[]}
+          hiddenCertifications={[]}
+          displayFundings={false}
+          displayCertifications={false}
+        />
```
jphechter, line 79: [sand] Every test in this function has roughly the same render function. You could reduce each test by 10 lines by creating a global render function that takes the desired props for a given test. Then you could call something like:
```
renderComponent({displayFundings: false});
```

This would make the test more readable, and you'd have the added benefit of being able to clearly see what parameters are being tested in the specific test.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -133,7 +141,7 @@ export const SidebarCardsList = (props: Props): ReactElement => {
           {showEmptyState() && (
             <div className="fdc fac margin-y-3">
               <h3 className="text-normal">{Config.dashboardDefaults.emptyOpportunitiesHeader}</h3>
```
aloverso, line 143: [boulder] This appears to be missing the AC for "And new text styling" such as changing this h3.

See the difference in text styling between the current version in dev:
image

and the new styling in the designs from the story:
image

It doesn't just change the image; the text styling is also updating

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -114,6 +114,14 @@ export const SidebarCardsList = (props: Props): ReactElement => {
     <>
       <h2 className="h1-styling margin-top-0">{Config.dashboardDefaults.sidebarHeading}</h2>
       <hr className="margin-top-2 margin-bottom-3 bg-cool-lighter" aria-hidden={true} />
+      {props.topCards.length === 0 && props.bottomCards.length === 0 && (
```
jphechter, line 117: [boulder] I think we need to change the way we conditionally display this.

Screenshot 2023-10-03 at 2 16 00 PM

web/src/components/dashboard/SidebarCardsList.tsx
jphechter: We specifically don't want alt text here.

web/src/components/dashboard/SidebarCardsList.tsx
AnJuHyppolite: Done!

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -21,12 +23,29 @@ function setupMockAnalytics(): typeof analytics {
...
+    const sidebarCardsListProps = {
+      topCards: [],
+      bottomCards: [],
+      fundings: [],
+      hiddenFundings: [],
+      certifications: [],
+      hiddenCertifications: [],
+      displayFundings: false,
+      displayCertifications: false,
+      ...props,
```
FarazA22, line 44: pebble - consider changing the parameter name to override

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +81,56 @@ describe("<SidebarCardsList />", () => {
...
+    });
+
+    it("does not display Complete Required Tasks Message when there are top cards", () => {
+      const topCards = [
+        generateSidebarCardContent({
+          id: "some-fake-top-card",
+          section: "above-opportunities",
+        }),
+      ];
+      renderComponent({ topCards, displayFundings: true, displayCertifications: true });
```
FarazA22, line 118: pebble - we are rendering the component with `displayFundings: true` and `displayCertifications: true` but that is not reflected in the test name. We should update the title, or maybe create a separate test.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +81,56 @@ describe("<SidebarCardsList />", () => {
     fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
     expect(mockAnalytics.event.for_you_card_unhide_button.click.unhide_cards).toHaveBeenCalledTimes(1);
   });
+
+  describe("empty state messaging", () => {
+    it("displays Complete Required Tasks Message when there are no primary actions or opportunities", () => {
```
jphechter, line 86:
```suggestion
    it("displays Complete Required Tasks Message when there are no primary actions and opportunities are not displayed", () => {
```
[sand] Reading this with fresh eyes, I'm thinking about what this test name is conveying. While the message is true, the key piece of logic has to do with `displayCertifications` and `displayFundings` not specifically the opportunity cards being present.

web/src/components/dashboard/SidebarCardsList.test.tsx
FarazA22: Another option is to update the initial value of the renderComponet so `displayFundings: true` and `displayCertifications: true` by default

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -36,14 +37,22 @@ export const SidebarCardsList = (props: Props): ReactElement => {
     }
   };
 
-  const showEmptyState = (): boolean => {
+  const showEmptyOpportunitiesMsg = (): boolean => {
     return (
       props.displayCertifications &&
       props.displayFundings &&
```
jphechter, line 43:
```suggestion
      (props.displayCertifications || props.displayFundings) &&
```
This could happen here or later. It will need tests.

I was thinking about how these messages are kindof inverse cases and it made me want to check if there is a time we display certs and not fundings (or vice versa). There is, in `FORMED_AND_REGISTERED`. In that instance, if a user hides all their opportunities they will currently see nothing. This is a bug, not caused by your work, but if you wanted to fix it, it would be quick.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -36,14 +37,22 @@ export const SidebarCardsList = (props: Props): ReactElement => {
...
       props.certifications.length + props.fundings.length === 0
     );
   };
 
+  const showCompleteRequiredTasksMsg = (): boolean => {
+    return (
+      !props.displayCertifications &&
+      !props.displayFundings &&
```
FarazA22, line 51: Pebble, based on your logic here, I think the renderComponent should have certifications and fundings. I think we render the component with the data already available but choose to render the cards when certain criteria is met.

web/src/components/dashboard/SidebarCardsList.tsx
FarazA22: @jp, once a user is in `formed and registered` they see the funding nudge card and if they hide all their certs, they will still see them in the hidden section (under the accordion). I don't think this is an issue.

PR #6458, opened 2023-10-03
feat: [#185931949] For you header updates and counter

cbrunefearless:
## Description
As a Navigator user, I need to be able to differentiate the For You header in the hierarchy in order to understand which header is generally applicable to the page vs the For You section specifically, and have a consistent Opportunity counter in mobile and desktop in order to understand if my Opportunities have changed

AC
Given I am a Navigator user
When viewing my For You section in a desktop viewport
Then the "For You" header uses unbolded H2 styling
And it is followed by parenthetical counter of unhidden opportunities + cards in the for you section using "base" color
(works the same as it does in mobile currently)

Given i am user
when viewing in mobile Viewport
The the For You column Header and the header line underneath is removed

Note: Storybook typography will also need to be updated.

### Ticket

[#185931949](https://www.pivotaltracker.com/story/show/)

## Code author checklist

### DIFF
--- a/web/src/components/dashboard/SidebarCardsList.test.tsx
+++ b/web/src/components/dashboard/SidebarCardsList.test.tsx
@@ -1,8 +1,20 @@
 import { SidebarCardsList } from "@/components/dashboard/SidebarCardsList";
 import analytics from "@/lib/utils/analytics";
+import { generateCertification } from "@/test/factories";
 import { useMockBusiness } from "@/test/mock/mockUseUserData";
+import { generateBusiness, generateProfileData } from "@businessnjgovnavigator/shared/test";
+import * as materialUi from "@mui/material";
+import { useMediaQuery } from "@mui/material";
 import { fireEvent, render, screen } from "@testing-library/react";
 
+function mockMaterialUI(): typeof materialUi {
+  return {
+    ...jest.requireActual("@mui/material"),
+    useMediaQuery: jest.fn(),
+  };
+}
+
+jest.mock("@mui/material", () => mockMaterialUI());
 jest.mock("@/lib/utils/analytics", () => setupMockAnalytics());
 jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
 
@@ -24,6 +36,9 @@ function setupMockAnalytics(): typeof analytics {
 describe("<SidebarCardsList />", () => {
   beforeEach(() => {
     jest.resetAllMocks();
+    (useMediaQuery as jest.Mock).mockImplementation(() => {
+      return true;
+    });
     useMockBusiness({});
   });
 
@@ -61,5 +76,59 @@ describe("<SidebarCardsList />", () => {
     fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
     fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
     expect(mockAnalytics.event.for_you_card_unhide_button.click.unhide_cards).toHaveBeenCalledTimes(1);
+  });
+
+  describe("SidebarCardsList For You Counter", () => {
+    beforeEach(() => {
+      jest.resetAllMocks();
+      (useMediaQuery as jest.Mock).mockImplementation(() => {
+        return true;
+      });
+    });
+
+    const mockBusiness = generateBusiness({
+      profileData: generateProfileData({
+        operatingPhase: "GUEST_MODE_OWNING",
+        municipality: undefined,
+        sectorId: undefined,
+        existingEmployees: "50",
+      }),
+    });
+
+    it("render cards with a zero count in the For You header when no Fundings and no certifications are present", () => {
+      useMockBusiness(mockBusiness);
+      render(
+        <SidebarCardsList
+          topCards={[]}
+          bottomCards={[]}
+          fundings={[]}
+          hiddenFundings={[]}
+          certifications={[]}
+          hiddenCertifications={[]}
+          displayFundings={true}
+          displayCertifications={true}
+        />
+      );
+      const forYouCounter = screen.getByTestId("for-you-counter").textContent;
+      expect(forYouCounter).toEqual("(0)");
+    });
+
+    it("render cards with a one count in the For You header when one certification is present", () => {
+      useMockBusiness(mockBusiness);
+      render(
+        <SidebarCardsList
+          topCards={[]}
+          bottomCards={[]}
+          fundings={[]}
+          hiddenFundings={[]}
+          certifications={[generateCertification({})]}
+          hiddenCertifications={[]}
+          displayFundings={true}
+          displayCertifications={true}
+        />
+      );
+      const forYouCounter = screen.getByTestId("for-you-counter").textContent;
+      expect(forYouCounter).toEqual("(1)");
+    });
   });
 });
--- a/web/src/components/dashboard/SidebarCardsList.tsx
+++ b/web/src/components/dashboard/SidebarCardsList.tsx
@@ -3,10 +3,13 @@ import { SidebarCard } from "@/components/dashboard/SidebarCard";
 import { PrimaryButton } from "@/components/njwds-extended/PrimaryButton";
 import { Icon } from "@/components/njwds/Icon";
 import { useConfig } from "@/lib/data-hooks/useConfig";
+import { useUserData } from "@/lib/data-hooks/useUserData";
+import { getForYouCardCount } from "@/lib/domain-logic/getForYouCardCount";
+import { MediaQueries } from "@/lib/PageSizes";
 import { Certification, Funding, SidebarCardContent } from "@/lib/types/types";
 import analytics from "@/lib/utils/analytics";
 import { openInNewTab, templateEval } from "@/lib/utils/helpers";
-import { Accordion, AccordionDetails, AccordionSummary } from "@mui/material";
+import { Accordion, AccordionDetails, AccordionSummary, useMediaQuery } from "@mui/material";
 import { ReactElement, ReactNode, useState } from "react";
 
 interface Props {
@@ -23,6 +26,7 @@ interface Props {
 export const SidebarCardsList = (props: Props): ReactElement => {
   const [hiddenAccordionIsOpen, setHiddenAccordionIsOpen] = useState<boolean>(false);
   const { Config } = useConfig();
+  const { business } = useUserData();
 
   const hiddenOpportunitiesCount = (): number => {
     if (props.displayCertifications && props.displayFundings) {
@@ -43,6 +47,8 @@ export const SidebarCardsList = (props: Props): ReactElement => {
       props.certifications.length + props.fundings.length === 0
     );
   };
+
+  const isDesktopAndUp = useMediaQuery(MediaQueries.desktopAndUp);
 
   const hiddenCardsAccordion = (): ReactNode => {
     if (props.displayCertifications || props.displayFundings) {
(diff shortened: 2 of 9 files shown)


web/src/styles/base/global.scss
```diff
@@ -129,6 +129,10 @@ ol {
   transform: translateY(0%);
 }
 
+.text-normal-override {
```
jphechter, line 132: [sand] Can we just call this `text-normal`?

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -24,7 +35,17 @@ function setupMockAnalytics(): typeof analytics {
...
+      profileData: generateProfileData({
+        industryId: "generic",
+        businessPersona: "STARTING",
+        sectorId: undefined,
+      }),
+    });
+    useMockBusiness(business);
+    (useMediaQuery as jest.Mock).mockImplementation(() => {
+      return true;
+    }); // set large screen
```
aloverso, line 48: Let's avoid code comments. If something feels like it needs an explanatory comment, then let's write explanatory code instead. For example, a function called `setLargeScreen` here.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -24,7 +35,17 @@ function setupMockAnalytics(): typeof analytics {
 describe("<SidebarCardsList />", () => {
   beforeEach(() => {
     jest.resetAllMocks();
-    useMockBusiness({});
+    const business = generateBusiness({
+      profileData: generateProfileData({
+        industryId: "generic",
+        businessPersona: "STARTING",
+        sectorId: undefined,
```
aloverso, line 42: [pebble] Why are these values necessary as the global mocked business data? 

It seems likely unnecessary globally, because all the other tests were passing even before this change.

If it's only necessary for the new test, it should be scoped just for that test. Even so, I have questions about why some of these specific values (like sector?) are impacting the test outcome

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +83,21 @@ describe("<SidebarCardsList />", () => {
     fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
     expect(mockAnalytics.event.for_you_card_unhide_button.click.unhide_cards).toHaveBeenCalledTimes(1);
   });
+
+  it("render cards with a zero count in the For You header when no Fundings and no certifications are present", () => {
```
aloverso, line 87: [boulder] I feel like there is a missing test(s) for any number other than zero... we're not actually testing the functionality that it counts up certifications and fundings.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -112,8 +118,20 @@ export const SidebarCardsList = (props: Props): ReactElement => {
...
+            {" "}
+            {Config.dashboardDefaults.sidebarHeading}
+            <span data-testid="for-you-counter" className="margin-left-05 text-base">
+              ({getContentWithCardCount(business, props.certifications, props.fundings)})
+            </span>
+          </h2>
+          <hr className="margin-top-2 margin-bottom-3 bg-cool-lighter" aria-hidden={true} />
+        </>
+      ) : (
+        <></>
```
aloverso, line 133: [pebble] generally [we don't use the ternary style](https://njcio.slack.com/archives/C04J8MTJNS2/p1692719694434699) like `condition ? <Component/> : <></>` with an empty fragment, but prefer `&&` instead.

web/src/lib/domain-logic/getContentWithCardCount.ts
```diff
@@ -0,0 +1,30 @@
+import { filterCertifications } from "@/lib/domain-logic/filterCertifications";
```
jphechter, line 1: [pebble] I like that we split this out from `TwoTabDashboardLayout`. Given that this isn't specific to that component any more, can we pull out the associated tests as well?

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -112,8 +118,20 @@ export const SidebarCardsList = (props: Props): ReactElement => {
 
   return (
     <>
-      <h2 className="h1-styling margin-top-0">{Config.dashboardDefaults.sidebarHeading}</h2>
-      <hr className="margin-top-2 margin-bottom-3 bg-cool-lighter" aria-hidden={true} />
+      {isDesktopAndUp ? (
```
jphechter, line 121: Question about this implementation. Further up in `dashboard.tsx` we do a check for `isDesktopAndUp` and then render the dashboard either for mobile or desktop. Both contain independent implementations of `SidebarCardsContainer` which contains `SidebarCardsList`. I feel like I see the current approach pretty commonly throughout our application, where we just do the check for what we need in the component. In this case though, I wonder if it makes sense to pass a boolean down from `dashboard.tsx` where we could set this when invoking the parent.

Is there a reason we typically follow this approach? Obviously fewer props is easier to read and this isn't an intensive operation. @cbrunefearless @aloverso

web/src/styles/base/global.scss
aloverso: I left a comment on the component itself - I think we might not need/want this class at all

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -112,8 +118,20 @@ export const SidebarCardsList = (props: Props): ReactElement => {
...
+        <>
+          <h2 className="margin-top-0 text-normal-override">
```
aloverso, line 123: [pebble/dust] Instead of doing an `h2` with a text override, what if we did `<div role="heading" aria-level={2}` like we do in other places in the app? That way, it's still a level 2 header for accessibility purposes, but we don't need to override the h2 styling. What do we think?

web/src/lib/domain-logic/getContentWithCardCount.ts
```diff
@@ -0,0 +1,30 @@
...
+import { filterFundings } from "@/lib/domain-logic/filterFundings";
+import { getVisibleCertifications } from "@/lib/domain-logic/getVisibleCertifications";
+import { getVisibleFundings } from "@/lib/domain-logic/getVisibleFundings";
+import { Certification, Funding } from "@/lib/types/types";
+import { LookupOperatingPhaseById } from "@businessnjgovnavigator/shared/operatingPhase";
+import { Business } from "@businessnjgovnavigator/shared/userData";
+
+export const getContentWithCardCount = (
```
aloverso, line 9: [pebble] I'm a little confused by this function name... it's `getContent...` but it returns a number. 

I think this could be improved to describe more directly what this function does.
- `getCardCount`
- `getSidebarCardCount`
- `getForYouCardCount`
- `countCards`

web/src/lib/domain-logic/getContentWithCardCount.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (LookupOperatingPhaseById(business?.profileData.operatingPhase).displayFundings) {
+    count += getVisibleFundings(filterFundings(fundings, business as Business), business as Business).length;
+  }
+
+  if (business?.preferences.visibleSidebarCards.length) {
+    count += business?.preferences.visibleSidebarCards.length;
+  }
+  return count;
+};
```
aloverso, line 30: [boulder] I think this should have unit tests

web/src/lib/domain-logic/getContentWithCardCount.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const getContentWithCardCount = (
+  business: Business | undefined,
+  certifications: Certification[],
+  fundings: Funding[]
+): number => {
+  let count = 0;
+  if (LookupOperatingPhaseById(business?.profileData.operatingPhase).displayCertifications) {
+    count += getVisibleCertifications(
+      filterCertifications(certifications, business as Business),
```
aloverso, line 17: [boulder] All over this function, we are writing `business as Business` which is a big anti-pattern. 

The function has explicitly defined that the `business` argument has type `Business | undefined`. By using `as Business` throughout, we're basically losing any type safety this function might have had. And the type safety exists for a reason. `business` _could_ be undefined, and if it is, then what happens? Does the code break at runtime in prod with a "cannot read properties of undefined" error?

Our code should always be type-safe, and that includes handling a possible undefined gracefully and not forcing it to be ignored using `as`. There's lots of approaches here. I'd recommend a guard clause because that's the pattern we use in the rest of the code. What _should_ this return when `business` is undefined? Let's code (and even test!) for that possible use case.

web/src/styles/base/global.scss
cbrunefearless: done

web/src/components/dashboard/SidebarCardsList.tsx
cbrunefearless: totally agree..

web/src/components/dashboard/SidebarCardsList.test.tsx
cbrunefearless: done

web/src/lib/domain-logic/getContentWithCardCount.ts
cbrunefearless: getForYouCardCount

web/src/lib/domain-logic/getContentWithCardCount.ts
cbrunefearless: put a guard clause in to return 0 if its undefined.

web/src/components/dashboard/SidebarCardsList.tsx
cbrunefearless: How do you feel about saving this as a refactor?

web/src/components/dashboard/SidebarCardsList.test.tsx
cbrunefearless: taking out

web/src/lib/domain-logic/getContentWithCardCount.ts
cbrunefearless: wrote a unit test for each individual count and error case

web/src/components/dashboard/SidebarCardsList.test.tsx
cbrunefearless: added a test for certification and funding.

web/src/components/dashboard/SidebarCardsList.tsx
jphechter: Either way. I'm not too partial. Mainly curious if there was a preference on the team.

web/test/pages/dashboard.test.tsx
```diff
@@ -583,7 +583,6 @@ describe("dashboard page", () => {
       expect(userDataWasNotUpdated()).toBe(true);
 
       fireEvent.click(screen.getByTestId("for-you-tab"));
-      expect(screen.getByText(Config.dashboardDefaults.sidebarHeading)).toBeInTheDocument();
       await waitFor(() => {
```
cbrunefearless, line 586: The heading in mobile mode was taken out, that is the why this is deleted.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -62,4 +77,58 @@ describe("<SidebarCardsList />", () => {
...
+          hiddenCertifications={[]}
+          displayFundings={true}
+          displayCertifications={true}
+        />
+      );
+      const forYouCounter = screen.getByTestId("for-you-counter").textContent;
+      expect(forYouCounter).toEqual("(0)");
+    });
+
+    it("render cards with a two count in the For You header when one certification is present", () => {
```
jphechter, line 116: [pebble]
```suggestion
    it("render cards with a one count in the For You header when one certification is present", () => {
```

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -119,8 +125,18 @@ export const SidebarCardsList = (props: Props): ReactElement => {
...
+          <h2 className="h2-styling margin-top-0 font-weight-normal">
+            {" "}
+            {Config.dashboardDefaults.sidebarHeading}
+            <span data-testid="for-you-counter" className="margin-left-05 text-base">
+              ({getForYouCardCount(business, props.certifications, props.fundings)})
+            </span>
+          </h2>
+          <hr className="margin-top-2 margin-bottom-3 bg-cool-lighter" aria-hidden={true} />
+        </>
+      )}
```
jphechter, line 139: [sand] It'd be nice to check this logic with a test. I don't think jest can actually adjust the screenwidth, but we could mock `isDesktopAndUp` to return `true/false`.

PR #6466, opened 2023-10-06
Dev design

FarazA22:
### Ticket

## Code author checklist

### DIFF
--- a/web/src/components/ModalTwoButton.tsx
+++ b/web/src/components/ModalTwoButton.tsx
@@ -1,7 +1,9 @@
 import { ModalZeroButton } from "@/components/ModalZeroButton";
 import { PrimaryButton } from "@/components/njwds-extended/PrimaryButton";
 import { SecondaryButton } from "@/components/njwds-extended/SecondaryButton";
-import { Breakpoint } from "@mui/material";
+import { ReverseOrderInMobile } from "@/components/ReverseOrderInMobile";
+import { MediaQueries } from "@/lib/PageSizes";
+import { Breakpoint, useMediaQuery } from "@mui/material";
 import { ReactElement, ReactNode } from "react";
 
 interface Props {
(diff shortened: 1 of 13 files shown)


web/src/components/ModalTwoButton.tsx
```diff
@@ -18,31 +20,37 @@ interface Props {
 }
 
 export const ModalTwoButton = (props: Props): ReactElement => {
+  const isMobileAndUp = useMediaQuery(MediaQueries.mobileAndUp);
+
   const buttonNode = (
     <div
-      className="padding-x-4 padding-y-3 bg-base-lightest display-flex flex-column flex-justify-center mobile-lg:flex-row"
+      className={`padding-x-4 padding-y-3 bg-base-lightest ${
+        isMobileAndUp ? "flex flex-column flex-align-end" : ""
```
adidner, line 28: I feel like this would be better as a mixin that way we don't have to do JS media queries which I think are kinda an anti pattern

web/src/components/ReverseOrderInMobile.tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactElement } from "react";
+
+interface Props {
+  children: ReactElement;
+  className?: string;
+}
+export const ReverseOrderInMobile = (props: Props): ReactElement => (
```
adidner, line 7: Neat

web/src/components/njwds-extended/PrimaryButton.tsx
```diff
@@ -26,5 +29,14 @@ export const PrimaryButton = forwardRef(function PrimaryButton(
...
-  return <GenericButton {...props} className={colors[props.isColor]} ref={ref} />;
+  return (
+    <GenericButton
+      {...props}
+      className={`${colors[props.isColor]} 
+      ${props.isUnBolded ? "text-normal" : ""}  
+      ${props.isSmallerText ? "font-body-2xs" : ""} 
+      ${props.isLargeButton ? "usa-button--big" : ""}`}
+      ref={ref}
+    />
```
adidner, line 40: I see that you moved these classes up a level so that they only exist via the PrimaryButton rather than for all generic buttons, 

what was the reasoning like for this? Is this better somehow? I would worry about these differences have to be undone if we ever want these attributes through other things like the secondary buttons because they seem quite common and plausible to have in other buttons later

FarazA22: I'm going to send another PR. My goal is to exclude the props we don't use. I found a better solution.

web/src/components/ModalTwoButton.tsx
FarazA22: I created this chore, let's talk about it during the next grooming

PR #6467, opened 2023-10-06
chore: [#186203159] clean up generic button prop

FarazA22:
### Ticket

## Code author checklist


adidner (approved):
I like this a lot better, it seems much more clear what we're omitting. Also if we want to add any of this functionality in the future it will be a lot easier to add. Also feels cleaner to me. Thanks

PR #6474, opened 2023-10-17
feat: [#185151322] Post-onboarding Questions dropdown support

seidior:
## Description

These changes allow for post-onboarding questions to have their own CTAs, including a button-based CTA and a dropdown-based CTA, as per ticket. This flexibility is demonstrated with the construction renovation post-onboarding question, which is now configured with a dropdown with the two provided links. The functionality is also mirrored in the CMS, giving CMS users immediate feedback to see how their CTA will appear.

In order for a dropdown to appear, all five greymatter attributes must be defined:

- dropdown text (the text that should appear inside the dropdown)
- link 1 name and destination
- link 2 name and destination

If only the link 1 name and destination (or link 2 name and destination) are specified, a CTA button will appear instead. If both are provided, nothing will appear until the dropdown text is provided.

### Ticket

[#185151322](https://www.pivotaltracker.com/story/show/185151322)

### Approach

No dependencies or migration are necessary for acceptance and merger of this pull request.

### Steps to Test

1. Log in to the CMS to make changes.
2. Navigate in the CMS to the post-onboarding questions section (🟨 MISC -> Post Onboarding Content)
3. Click the `construction-renovation` post-onboarding question.
4. Observe changes to the behavior of the gray footer section of the post-onboarding question preview on the right side as the relevant items (dropdown name, link 1 name and destination, link 2 name and destination) on the left side are changed.

### Notes

Thanks to @aloverso and @FarazA22 for assistance in getting this over the line.

## Code author checklist


web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { useRouter } from "next/router";
+import React, { ReactElement } from "react";
+
+interface ButtonsArrayElement {
+  text: string;
+  destination: string;
+  onClick: () => void;
+}
+
+const TaskFooter = (props: React.ComponentPropsWithoutRef<"div">): JSX.Element => {
```
aloverso, line 16: [dust] Why pull this out into its own component if it's only ever used in this one case, in this one context? Its only purpose is to be a "cta-area".

[sand] At the very least, the `attributes` prop is unnecessary because you're using it to pass in data-testid but that's already defined on line 20. So you could simplify this by just taking `children: ReactNode` as a prop and nothing else.

[dust] I've always done `ReactElement` as the return type for a functional component, like you did on line 37. Why JSX.Element here?

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
+import { PrimaryButtonDropdown } from "@/components/njwds-extended/PrimaryButtonDropdown";
+import { TaskCTA } from "@/components/TaskCTA";
+import { useUserData } from "@/lib/data-hooks/useUserData";
+import { PostOnboarding, Task } from "@/lib/types/types";
+import { openInNewTab } from "@/lib/utils/helpers";
+import { ProfileData } from "@businessnjgovnavigator/shared/index";
+import { useRouter } from "next/router";
+import React, { ReactElement } from "react";
+
+interface ButtonsArrayElement {
```
aloverso, line 10: [sand] I can't think of a good name right now, but I feel like there's a better name out there for this. I feeling like using the term `ArrayElement` in a type name is an antipattern because it's describing _how_ the type is being used instead of _what_ it is/represents.

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+    >
+      <div className="mobile-lg:margin-bottom-0 margin-bottom-1">{children}</div>
+    </div>
+  );
+};
+
+interface Props {
+  postOnboardingQuestion: PostOnboarding | undefined;
+  task?: Task | undefined;
+  forceDisplay?: boolean | undefined;
```
aloverso, line 33: [pebble] What's the purpose of this prop? If I'm reading correctly, it seems this is always `true` in the app? Which is an indicator that it is unnecessary and should be removed. 

If it actually is necessary behavior, I'd expect it to be covered in the unit tests for this component.

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+export const TaskCtaArea = (props: Props): ReactElement => {
+  const ctaButtons: ButtonsArrayElement[] = [];
+  const { business } = useUserData();
+  const router = useRouter();
+
+  const taskCtaDestination = props.task?.callToActionLink;
+  let taskCtaText = props.task?.callToActionText ?? "";
+
+  if (taskCtaText && taskCtaText.length === 0) {
+    taskCtaText = "Start Application";
```
aloverso, line 46: [pebble] 

a. The `taskCtaText.length === 0` part of this is redundant, right?

b. What is the purpose of this statement? I don't see a test for it, and I'm surprised to see a bare string in the code - if this is desired behavior, it should be in Config. Unless this was left in by mistake?

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const text2 = props.postOnboardingQuestion?.callToActionYesText2;
+
+  const hasTaskCta = taskCtaDestination;
+  const hasPostOnboarding = !!props.postOnboardingQuestion;
+
+  const hasAnsweredYesToPostOnboarding =
+    (props.onboardingKey && business?.profileData[props.onboardingKey] === true) || props.forceDisplay;
+
+  if (hasPostOnboarding && hasAnsweredYesToPostOnboarding) {
+    if (link1 && link1.length > 0 && text1 && text1.length > 0) {
```
aloverso, line 61: [sand] Same point as above, I believe that checking length is redundant here and this could be simplified to `if (link1 && text1)` which is much more readable :)

web/src/components/TaskCtaArea.tsx
aloverso: I won't comment all over the place for the same question, but I'm also unsure about why the if/else openInNewTab controlled by this prop?

web/src/components/TaskCtaArea.tsx
aloverso: Same comment on line 74

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+        onClick: () => {
+          if (props.forceDisplay) {
+            openInNewTab(link1);
+          } else {
+            router.push(link1);
+          }
+        },
+      });
+    }
+    if (link2 && link2.length > 0 && text2 && text2.length > 0) {
```
aloverso, line 74: [sand] Same comment as above regarding length

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+        onClick: () => {
+          if (props.forceDisplay) {
+            openInNewTab(link2);
+          } else {
+            router.push(link2);
+          }
+        },
+      });
+    }
+  } else {
```
aloverso, line 87: [sand] could just be `else if` instead of an if nested in an else.

If you haven't heard the team talk about it yet, I'm a big fan of guard clauses, for the reason of keeping the "useful" logic as close to the left as possible with minimal nesting. (I imagine you also care for this, with your 80char happiness). So this comment is coming from the same place - no need for 2 layers of tab nesting when we can achieve the same with just one.

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+          } else {
+            router.push(taskCtaDestination);
+          }
+        },
+      });
+    }
+  }
+
+  if (ctaButtons.length === 1) {
+    const [buttonsMeta] = ctaButtons;
```
aloverso, line 104: [sand] This is not my favorite variable name, because it's not very descriptive. Maybe something to make it clear at quick glance that this is the one (and only) button we care about? idk, `singularButton` or something like that? Idk

web/src/components/TaskCtaArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+    return (
+      <div data-testid="cta-area">
+        <TaskCTA link={buttonsMeta.destination} text={buttonsMeta.text} />
+      </div>
+    );
+  }
+
+  if (ctaButtons.length > 1 && !!props.postOnboardingQuestion?.callToActionYesDropdownText) {
```
aloverso, line 112: [sand] I think `!!` is unnecessary here since the if-statement naturally forces the string/undefined type into truthy/falsey, so the extra is redundant.

web/src/styles/base/flex.scss
```diff
@@ -92,6 +92,10 @@
   flex-direction: column-reverse;
 }
 
+.flex-row-reverse {
```
aloverso, line 95: [sand] Am i correct in that this is unused?

web/src/components/TaskCtaArea.tsx
seidior: a. Actually, I kind of want it to be its own component. Is it better behavior to stick an export on it here or to separate it out into its own component file?

b. Do... do we not like prop spread? I've always used it for convenience when a component should behave exactly like its intrinsic HTML element.

c. Good catch. Fixed.

web/src/components/TaskCtaArea.tsx
seidior: You're right, naming is tricky, because it defines an element that can either be a button or a dropdown list item depending on how many elements in the array.

web/src/components/TaskCtaArea.tsx
seidior: a. It is not; I'm *specifically* testing for the case where the length is 0 so that I can set some fallback text to display for the CTA button.

b. You're right about the bare string. I've updated it to pull this value from `Config`.

web/src/styles/base/flex.scss
seidior: You are indeed correct; this is a fragment left over from a prior implementation that *technically* is no longer needed. However, it's a utility that I'm surprised we didn't have, given that we had its cousin for column-reversing.

Would it be better to leave this in, as someone will undoubtedly use it in the future (and maybe we should consider using something like postCSS.

web/src/components/TaskCtaArea.tsx
aloverso: `CMS_ONLY_forceDisplay` is good, that helps. For what it's worth, I don't think it's necessary at all though. Nothing in the CMS preview panel is clickable or interactable and there's no reason this one needs to be.

web/src/components/TaskCtaArea.tsx
aloverso: Maybe at a higher level - why do we want fallback text for this? I didn't think that existed previously, and it's not part of the story?

web/src/components/TaskCtaArea.tsx
seidior: Oh, I made it clickable so users can see the dropdown list, oops. Should I make it unclickable again?

web/src/components/TaskCtaArea.tsx
seidior: Not sure if this was part of the model change making this optional, but yes, this value existed previously. I pulled this from L33 of `TaskCTA.tsx`.

aloverso: But also wait... I'm still confused. I don't see how this could ever evaluate to true? 

`taskCtaText && taskCtaText.length === 0`

taskCtaText is either the text of the task, or empty string, according to the prev line. Empty string evaluates to falsey, so if it's empty, then `taskCtaText` would evaluate to false. And if it's not empty, then `taskCtaText.length === 0` would evaluate to false, so the && case can never be true. 

Am I missing something? If I am, I think a test case for this would clear up the confusion

web/src/styles/base/flex.scss
aloverso: Personally, I'm more in favor of not adding things until we have a need, but that's just me

web/src/components/TaskCtaArea.tsx
aloverso: I'm very very much in favor of one component per file, I think it can be really hard to find things when they're multiple in a file.

For prop spread.... personally I like being explicit and saying "this supports a className" or "this takes a testId" and giving someone who uses a component an explicit list of what we expect to be passed in, or not. I'd argue that otherwise it gets really complex to figure what I can or cannot pass in, and why. I've been dealing with the `GenericTextField` component for the formation refactoring and it uses prop spread and I really have a hard time working with it, and knowing what I can or can't provide; it makes it hard to find. But that's just my opinion

seidior: I've separated out this component into its own file. Given there's no logic—it's just a div with some CSS—I don't see a need to add a test file here, but if it's needed let me know and I can try to come up with something.

To that end, I've also removed the interface formerly known as `ButtonsArrayElement` (which I've renamed `CallToActionHyperlink`) and placed it in the `types.ts` file to be shared (including by the cannabis-related files) by any component working with this dropdown element.

---

For something like `GenericTextField`, for sure, I can see why that can cause a headache, given that we're passing it down to MUI's `TextField` but also have a bunch of logic in the way too. However, for components that behave just like HTML elements, we definitely do know what we can and can't pass in; in this instance, we can pass in everything that a `<div>` can take (or rather, what an `HTMLDivElement` can take), which gives us greater flexibility in the future. In fact, code completion should suggest the same attributes.

That said, I am definitely aware that what I understand to be best practices may be either or both a. wrong, and b. not the expectation on an existing codebase with existing code patterns. I'm down to eat a slice of humble pie on this one and would be okay making this less generic to fit the style of the codebase if not welcome. Let me know.

web/src/styles/base/flex.scss
jphechter: I'm with Anne. If we don't need it, I say kill it (for now).

web/src/components/TaskFooter.tsx
```diff
@@ -0,0 +1,14 @@
+import { ComponentPropsWithoutRef, ReactElement } from "react";
+
+export const TaskFooter = (props: ComponentPropsWithoutRef<"div">): ReactElement => {
+  const { children, ...attributes } = props;
+  return (
+    <div
+      style={{ marginTop: "2rem" }}
```
jphechter, line 7: [dust] You can also add the class `margin-top-4` to achieve the same thing and avoid the inline style.

seidior: Good catch, thanks!

web/src/styles/base/flex.scss
seidior: Nuked from orbit. Thanks for the feedback!

web/src/components/TaskCtaArea.tsx
aloverso: Overall: I'm good with this!

A thought: I think there's some renaming that could be done here. At first glance, `TaskCtaArea` and `TaskFooter` sound like kinda the same thing given what I know about the task page. I wouldn't know from the names which one contains the other. This was behind my initial question of - why pull it out at all if we can't descriptively name it as its own thing? But maybe just a thoughtful naming scheme could solve it too.

I'm good with your take on the HTMLDivElement stuff though!

A note: I don't really believe in the phrase "best practices" and you know lots of valuable patterns! There is no "best" imo, because practices are context-dependent and what's best for one is not for another. I prefer thinking of things as "useful patterns" because there's no one best way of doing things. Strong opinions loosely held. I think you make a good point here.

web/src/components/TaskCtaArea.tsx
aloverso: 🤷 

- it's already built
- but it adds a non-insignificant amount of complexity to the code
- but it's useful for the CMS team 
- but it's not something we do (or should plan to do) in general in the CMS, imo

Your call

PR #6479, opened 2023-10-18
feat: [#185893614] license status check migration - public movers

somabadri:
## Description

DCA Dynamics API integration for public movers. 

### Ticket

[#185893614](https://www.pivotaltracker.com/story/show/185893614)

### Approach
Built `searchLicenseStatusFactory` which returns a `SearchLicenseStatus` to use based on license type. If the license type is `Public Movers and Warehousemen` then `DynamicsLicenseStatusClient` is returned otherwise `WebserviceLicenseStatusProcessorClient` is returned.

### Steps to Test

This new integration is only for public movers so make sure your industry is Moving Company.

Here is an mover business details that is present in the Dynamics API (STAGING) only available in local env:

**Business Name**: `SIMONIK MOVING`
**Business Address Line 1**: `5E Easy Street`
**ZIP Code**: `08805`

### Notes
- Update the following environment variables in API env file
  - DCA_DYNAMICS_ORG_URL
  - DCA_DYNAMICS_CLIENT_ID
  - DCA_DYNAMICS_SECRET
  - DCA_DYNAMICS_TENANT_ID
  - FEATURE_DYNAMICS_PUBLIC_MOVERS

- There is a potential refactor to combine `WebserviceLicenseStatusProcessorClient` and `WebserviceLicenseStatusClient`.

## Code author checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,6 +201,11 @@ commands:
             echo 'export USE_BASIC_AUTH=$USE_BASIC_AUTH_DEV' >> $BASH_ENV
             echo 'export BASIC_AUTH_USERNAME=$BASIC_AUTH_USERNAME' >> $BASH_ENV
             echo 'export BASIC_AUTH_PASSWORD=$BASIC_AUTH_PASSWORD' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_ORG_URL_DEV=$DCA_DYNAMICS_ORG_URL' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_CLIENT_ID_DEV=$DCA_DYNAMICS_CLIENT_ID' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_SECRET_DEV=$DCA_DYNAMICS_SECRET' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_TENANT_ID_DEV=$DCA_DYNAMICS_TENANT_ID' >> $BASH_ENV
+            echo 'export FEATURE_DYNAMICS_PUBLIC_MOVERS_DEV=$FEATURE_DYNAMICS_PUBLIC_MOVERS' >> $BASH_ENV
 
   set-cypress-env-vars:
     steps:
@@ -226,6 +231,11 @@ commands:
             echo 'export AWS_CRYPTO_CONTEXT_PURPOSE=$AWS_CRYPTO_CONTEXT_PURPOSE' >> $BASH_ENV
             echo 'export AWS_CRYPTO_CONTEXT_ORIGIN=$AWS_CRYPTO_CONTEXT_ORIGIN' >> $BASH_ENV
             echo 'export DISABLE_GTM=true' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_ORG_URL_DEV=$DCA_DYNAMICS_ORG_URL' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_CLIENT_ID_DEV=$DCA_DYNAMICS_CLIENT_ID' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_SECRET_DEV=$DCA_DYNAMICS_SECRET' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_TENANT_ID_DEV=$DCA_DYNAMICS_TENANT_ID' >> $BASH_ENV
+            echo 'export FEATURE_DYNAMICS_PUBLIC_MOVERS_DEV=$FEATURE_DYNAMICS_PUBLIC_MOVERS' >> $BASH_ENV
 
   set-staging-env-vars:
     steps:
@@ -262,6 +272,11 @@ commands:
             echo 'export USE_BASIC_AUTH=$USE_BASIC_AUTH_STAGING' >> $BASH_ENV
             echo 'export BASIC_AUTH_USERNAME=$BASIC_AUTH_USERNAME' >> $BASH_ENV
             echo 'export BASIC_AUTH_PASSWORD=$BASIC_AUTH_PASSWORD' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_ORG_URL_STAGING=$DCA_DYNAMICS_ORG_URL' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_CLIENT_ID_STAGING=$DCA_DYNAMICS_CLIENT_ID' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_SECRET_STAGING=$DCA_DYNAMICS_SECRET' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_TENANT_ID_STAGING=$DCA_DYNAMICS_TENANT_ID' >> $BASH_ENV
+            echo 'export FEATURE_DYNAMICS_PUBLIC_MOVERS_STAGING=$FEATURE_DYNAMICS_PUBLIC_MOVERS' >> $BASH_ENV
 
   set-prod-env-vars:
     steps:
@@ -296,6 +311,11 @@ commands:
             echo 'export AWS_CRYPTO_CONTEXT_PURPOSE=$AWS_CRYPTO_CONTEXT_PURPOSE' >> $BASH_ENV
             echo 'export AWS_CRYPTO_CONTEXT_ORIGIN=$AWS_CRYPTO_CONTEXT_ORIGIN' >> $BASH_ENV
             echo 'export OUTAGE_ALERT_CONFIG_URL=$OUTAGE_ALERT_CONFIG_URL' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_ORG_URL_PROD=$DCA_DYNAMICS_ORG_URL' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_CLIENT_ID_PROD=$DCA_DYNAMICS_CLIENT_ID' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_SECRET_PROD=$DCA_DYNAMICS_SECRET' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_TENANT_ID_PROD=$DCA_DYNAMICS_TENANT_ID' >> $BASH_ENV
+            echo 'export FEATURE_DYNAMICS_PUBLIC_MOVERS_PROD=$FEATURE_DYNAMICS_PUBLIC_MOVERS' >> $BASH_ENV
 
   set-content-env-vars:
     steps:
--- a/api/src/client/dynamics/DynamicsBusinessAddressClient.test.ts
+++ b/api/src/client/dynamics/DynamicsBusinessAddressClient.test.ts
@@ -1,0 +1,184 @@
+/* eslint-disable unicorn/no-null */
+import { LogWriter, LogWriterType } from "@libs/logWriter";
+import { generateLicenseSearchAddress } from "@shared/test";
+import axios from "axios";
+import { DynamicsBusinessAddressClient } from "./DynamicsBusinessAddressClient";
+import { BusinessAddressClient } from "./types";
+
+jest.mock("axios");
+jest.mock("winston");
+const mockAxios = axios as jest.Mocked<typeof axios>;
+
+describe("DynamicsBusinessAddressClient", () => {
+  let client: BusinessAddressClient;
+  let logger: LogWriterType;
+
+  const ORG_URL = "www.test-org-url.com";
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    logger = LogWriter("NavigatorWebService", "ApiLogs", "us-test-1");
+
+    client = DynamicsBusinessAddressClient(logger, ORG_URL);
+  });
+
+  const mockAccessToken = "access-granted";
+  const mockBusinessIds = ["id2", "id3"];
+
+  it("makes a get request to dynamics api with business address query and auth token in header", async () => {
+    const businessAddressMockResponse = {
+      value: [
+        {
+          createdon: "2023-05-31T10:31:51Z",
+          rgb_city: "Washington",
+          rgb_county: null,
+          rgb_name: "Apt GFG 3900A Watson Pl NW",
+          rgb_state: "DC",
+          rgb_street1: "3900A Watson Pl NW",
+          rgb_street2: "Apt GFG",
+          rgb_typecode: 100000000,
+          rgb_zip: "20016-6739",
+          statecode: 0,
+          rgb_addressid: "ff166755-9eff-ed11-a81b-001dd80703e0",
+        },
+        {
+          createdon: "2023-05-31T10:32:04Z",
+          rgb_city: "Chesapeake",
+          rgb_county: null,
+          rgb_name: "Trlr T 322 Ballahack Rd",
+          rgb_state: "VA",
+          rgb_street1: "Trlr T 322 Ballahack Rd",
+          rgb_street2: null,
+          rgb_typecode: 100000001,
+          rgb_zip: "23322-2479",
+          statecode: 0,
+          rgb_addressid: "4da7455c-9eff-ed11-a81b-001dd80703e0",
+        },
+      ],
+    };
+
+    mockAxios.get.mockResolvedValue({ data: businessAddressMockResponse });
+    expect(await client.getBusinessIdsAndLicenseSearchAddresses(mockAccessToken, mockBusinessIds)).toEqual([
+      {
+        businessId: "id2",
+        addresses: [
+          generateLicenseSearchAddress({
+            addressLine1: "3900A Watson Pl NW",
+            addressLine2: "Apt GFG",
+            zipCode: "20016",
+          }),
+          generateLicenseSearchAddress({
+            addressLine1: "Trlr T 322 Ballahack Rd",
+            addressLine2: "",
+            zipCode: "23322",
+          }),
+        ],
+      },
+      {
+        businessId: "id3",
+        addresses: [
+          generateLicenseSearchAddress({
+            addressLine1: "3900A Watson Pl NW",
+            addressLine2: "Apt GFG",
+            zipCode: "20016",
+          }),
+          generateLicenseSearchAddress({
+            addressLine1: "Trlr T 322 Ballahack Rd",
+            addressLine2: "",
+            zipCode: "23322",
+          }),
+        ],
+      },
+    ]);
+    expect(mockAxios.get).toHaveBeenNthCalledWith(
+      1,
+      `${ORG_URL}/api/data/v9.2/rgb_addresses?$select=createdon,rgb_city,rgb_county,rgb_name,rgb_state,rgb_street1,rgb_street2,rgb_typecode,rgb_zip,statecode&$filter=(_rgb_businessid_value eq ${mockBusinessIds[0]})&$top=50`,
+      {
+        headers: {
+          Authorization: `Bearer ${mockAccessToken}`,
+        },
+      }
+    );
+
+    expect(mockAxios.get).toHaveBeenNthCalledWith(
+      2,
+      `${ORG_URL}/api/data/v9.2/rgb_addresses?$select=createdon,rgb_city,rgb_county,rgb_name,rgb_state,rgb_street1,rgb_street2,rgb_typecode,rgb_zip,statecode&$filter=(_rgb_businessid_value eq ${mockBusinessIds[1]})&$top=50`,
+      {
+        headers: {
+          Authorization: `Bearer ${mockAccessToken}`,
+        },
+      }
+    );
+  });
+
+  it("filters out addresses whose statecode is not 0", async () => {
+    const businessAddressMockResponse = {
+      value: [
+        {
+          createdon: "2023-05-31T10:31:51Z",
+          rgb_city: "Washington",
+          rgb_county: null,
+          rgb_name: "Apt GFG 3900A Watson Pl NW",
+          rgb_state: "DC",
+          rgb_street1: "Apt GFG 3900A Watson Pl NW",
+          rgb_street2: null,
+          rgb_typecode: 100000000,
+          rgb_zip: "20016-6739",
+          statecode: 0,
+          rgb_addressid: "ff166755-9eff-ed11-a81b-001dd80703e0",
+        },
+        {
+          createdon: "2023-05-31T10:32:04Z",
+          rgb_city: "Chesapeake",
+          rgb_county: null,
+          rgb_name: "Trlr T 322 Ballahack Rd",
+          rgb_state: "VA",
+          rgb_street1: "Trlr T 322 Ballahack Rd",
+          rgb_street2: null,
+          rgb_typecode: 100000001,
+          rgb_zip: "23322-2479",
+          statecode: 1,
+          rgb_addressid: "4da7455c-9eff-ed11-a81b-001dd80703e0",
+        },
+      ],
+    };
+
+    mockAxios.get.mockResolvedValue({ data: businessAddressMockResponse });
+    expect(await client.getBusinessIdsAndLicenseSearchAddresses(mockAccessToken, mockBusinessIds)).toEqual([
+      {
+        businessId: "id2",
+        addresses: [
+          generateLicenseSearchAddress({
+            addressLine1: "Apt GFG 3900A Watson Pl NW",
+            addressLine2: "",
+            zipCode: "20016",
+          }),
+        ],
+      },
+      {
+        businessId: "id3",
+        addresses: [
+          generateLicenseSearchAddress({
+            addressLine1: "Apt GFG 3900A Watson Pl NW",
+            addressLine2: "",
+            zipCode: "20016",
+          }),
+        ],
+      },
+    ]);
+  });
+
+  it("returns empty array if received data value is empty", async () => {
+    mockAxios.get.mockResolvedValue({ data: { value: [] } });
+    expect(await client.getBusinessIdsAndLicenseSearchAddresses(mockAccessToken, mockBusinessIds)).toEqual([
+      {
+        businessId: "id2",
+        addresses: [],
+      },
+      {
+        businessId: "id3",
+        addresses: [],
+      },
+    ]);
+  });
+});
--- a/api/src/client/dynamics/DynamicsLicenseApplicationIdClient.test.ts
+++ b/api/src/client/dynamics/DynamicsLicenseApplicationIdClient.test.ts
@@ -1,0 +1,236 @@
+/* eslint-disable unicorn/no-null */
+import { MULTIPLE_MAIN_APPS_ERROR, NO_MAIN_APPS_ERROR } from "@domain/types";
+import { LogWriter, LogWriterType } from "@libs/logWriter";
+import axios from "axios";
+import { DynamicsLicenseApplicationIdClient } from "./DynamicsLicenseApplicationIdClient";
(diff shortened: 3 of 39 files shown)


api/src/client/DynamicsAccessTokenClient.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      clientId: process.env.DCA_DYNAMICS_CLIENT_ID || "",
+      clientSecret: process.env.DCA_DYNAMICS_SECRET || "",
+    });
+  });
+
+  afterEach(() => {
+    process.env.DCA_DYNAMICS_ORG_URL = ORIGINAL_ORG_URL;
+    process.env.DCA_DYNAMICS_TENANT_ID = ORIGINAL_TENANT_ID;
+    process.env.DCA_DYNAMICS_CLIENT_ID = ORIGINAL_CLIENT_ID;
+    process.env.DCA_DYNAMICS_SECRET = ORIGINAL_CLIENT_SECRET;
```
aloverso, line 54: [pebble] I think this whole process of setting the env vars in `process.env` is unnecessary, because the code under test isn't pulling them out of the environment, they're getting passed in as config vars to the Client. And that's good! It means that the client doesn't depend on the node environment. But the tests here can be significantly simpler because you just need to pass the string into the client, you shouldn't need anything to do anything with `process.env` at all.

aloverso: Same comment in the other tests also.

api/src/client/DynamicsBusinessAddressClient.ts
```diff
@@ -0,0 +1,81 @@
...
+      .catch((error: AxiosError) => {
+        logWriter.LogError(`Dynamics Business Address Client - Id:${logId} - Error:`, error);
+        throw error.response?.status;
+      });
+  };
+
+  const getMappedBusinessAddressesByIds = async (
+    accessToken: string,
+    businessIds: string[]
+  ): Promise<Awaited<BusinessIdToAddressesMapping>[]> => {
```
aloverso, line 44: [pebble] I don't think the `Awaited` type is doing anything here... this is just returning a type `Promise<BusinessIdToAddressesMapping[]>` and that still compiles

api/src/client/DynamicsBusinessAddressClient.ts
```diff
@@ -0,0 +1,81 @@
...
+    const getBusinessAddressesById = async (id: string): Promise<BusinessIdToAddressesMapping> => {
+      return { [id]: await getBusinessAddresses(accessToken, id) };
+    };
+
+    return await Promise.all(
+      businessIds.map(
+        async (id: string): Promise<BusinessIdToAddressesMapping> => await getBusinessAddressesById(id)
+      )
```
aloverso, line 52: [sand/dust] If you like, this could be written as 
```
return await Promise.all(businessIds.map(getBusinessAddressesById));
```

which might be easier to read

api/src/client/DynamicsBusinessAddressClient.test.ts
```diff
@@ -0,0 +1,182 @@
...
+          rgb_street2: null,
+          rgb_typecode: 100000001,
+          rgb_zip: "23322-2479",
+          statecode: 0,
+          rgb_addressid: "4da7455c-9eff-ed11-a81b-001dd80703e0",
+        },
+      ],
+    };
+
+    mockAxios.get.mockResolvedValue({ data: { ...businessAddressMockResponse } });
```
aloverso, line 67: [sand] could be written simpler as `{ data: businessAddressMockResponse }` I think

api/src/client/DynamicsBusinessAddressClient.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    ]);
+    expect(mockAxios.get).toHaveBeenCalledWith(
+      `${ORG_URL}/api/data/v9.2/rgb_addresses?$select=createdon,rgb_city,rgb_county,rgb_name,rgb_state,rgb_street1,rgb_street2,rgb_typecode,rgb_zip,statecode&$filter=(_rgb_businessid_value eq ${mockBusinessIds[0]})&$top=50`,
+      {
+        headers: {
+          Authorization: `Bearer ${mockAccessToken}`,
+        },
+      }
+    );
+    expect(mockAxios.get).toHaveBeenCalledWith(
```
aloverso, line 106: [dust] If it helps, there's a matcher called `toHaveBeenNthCalledWith` where you can assert which one was called first and second. I don't know if you care the order. But it also helps readability for me to be like, "oh yeah, both of these were supposed to have been called" instead of thinking "wait did we just do the same assertion twice" and then realizing

api/src/client/DynamicsBusinessIdsClient.ts
```diff
@@ -0,0 +1,36 @@
+import { LogWriterType } from "@libs/logWriter";
+import axios, { AxiosError } from "axios";
+import { BusinessIdClient, BusinessIdResponse } from "../domain/types";
```
aloverso, line 3: [pebble] I think the `BusinessIdResponse` type should be defined in this file, not in domain... it defines the data structure for a Dynamics response type, right? That would be client-specific code, not our domain.

api/src/client/DynamicsBusinessIdsClient.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const DynamicsBusinessIdsClient = (logWriter: LogWriterType, orgUrl: string): BusinessIdClient => {
+  const getBusinessId = async (accessToken: string, nameToSearch: string): Promise<string[]> => {
```
aloverso, line 6: [sand] I might rename this `getBusinessIds` because it returns an array of strings. Or something even more descriptive, because it's based on a name search? `getMatchingBusinessIds` or something?

api/src/client/DynamicsBusinessIdsClient.test.ts
```diff
@@ -0,0 +1,74 @@
...
+          name: mockBusinessName2,
+          accountid: mockBusinessId2,
+        },
+      ],
+    };
+
+    mockAxios.get.mockResolvedValue({ data: { ...businessIdMockResponse } });
+    expect(await client.getBusinessId(mockAccessToken, mockNametoSearch)).toEqual([
+      businessIdMockResponse.value[0].accountid,
+      businessIdMockResponse.value[1].accountid,
```
aloverso, line 57: [sand/dust] Since you're using variables so heavily, I think this might be more readable if you set up your expectation using the variable names instead of extracting them from the structure of the mock object.

Just something like:

```
 expect(await client.getBusinessId(mockAccessToken, mockNametoSearch)).toEqual([
      mockBusinessId1,
      mockBusinessId2
]
```

api/src/client/DynamicsAccessTokenClient.test.ts
somabadri: I originally wasn't passing them in so this made sense but I should have changed it to your point after the refactor to pass them in. Good call out! I changed them all expect for `searchLicenseStatusFactory` because that actually depends on the feature flag to determine whether to use Dynamics or not, so I left that one the same.

api/src/client/DynamicsLicenseApplicationIdClient.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      ],
+    };
+
+    mockAxios.get.mockResolvedValue({ data: { ...mockLicenseApplicationIdResponse } });
+    expect(
+      await client.getLicenseApplicationId(mockAccessToken, mockBusinessId, "Public Movers and Warehousemen")
+    ).toEqual({
+      expirationDate: "",
+      applicationId: "license-id-1357",
+      licenseStatus: "APPROVED",
```
aloverso, line 157: [pebble] I think in order for this test to really give confidence, the Inactive record should come first in the mocked value.

Right now, we're just selecting and returning the `[0]` array element. If you removed the logic that filters out inactive applications, this test would still pass because it'd just return the first element from the array.

If you switch the order in the mock response here, then the test would only pass if the filtering really is happening.

api/src/client/DynamicsLicenseStatusClient.ts
```diff
@@ -0,0 +1,64 @@
...
+      expirationISO: applicationIdResponse.expirationDate,
+      checklistItems: [...checkListItems],
+    };
+
+    const logId = logWriter.GetId();
+    logWriter.LogInfo(
+      `Dynamics License Status Client - Id:${logId} - Response:  ${JSON.stringify(response)}`
+    );
+
+    return response;
```
aloverso, line 62: [boulder] This is good but I think it's missing error handling - both in the code and tests. What happens if any of these awaited calls throw an error?

There's no need to get specific with different types of errors, but I think there should be a test for what this returns if any one of the many calls goes wrong.

api/src/domain/license-status/searchBusinessAddressesForMatch.ts
```diff
@@ -0,0 +1,38 @@
...
+): string => {
+  const result: string[] = [];
+
+  const isAddressMatch = (
+    address: LicenseSearchAddress,
+    userInputtedAddress: LicenseSearchAddress
+  ): boolean => {
+    return (
+      userInputtedAddress.addressLine1 === address.addressLine1 &&
+      userInputtedAddress.addressLine2 === address.addressLine2 &&
```
aloverso, line 16: [pebble] Our existing logic does not match on addressLine2 at all (it is ignored completely) - we only match on addressLine1 and zipCode. We should keep this same behavior.

api/src/domain/license-status/searchBusinessAddressesForMatch.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  const randomBusinessWithRandomAddresses2 = {
+    businessId2345: [generateLicenseSearchAddress({}), generateLicenseSearchAddress({})],
+  };
+  const businessWithNoAddresses = {
+    businessId3456: [],
+  };
+
+  describe("single business ids", () => {
+    it("returns correct business id when it has two matching addresses", () => {
+      const businessId1 = { matchingId: [matchedAddress1, matchedAddress2] };
```
aloverso, line 32: [sand] It might be easier to read if this variable was named something else, because this object is not a "businessId", it's a "businessIdToAddressesMapping" and that's important.

Also, this is tiny, but I think it's useful to put `"matchingId"` in quotes here in this object, which helps trigger the brain that this is a test string and not an actual key in a defined type (if that makes sense)

api/src/domain/license-status/searchBusinessAddressesForMatch.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      const result = searchBusinessAddressesForMatch(
+        [businessIdWithMatchingAddress, randomBusinessWithRandomAddresses1],
+        userInputtedAddress
+      );
+      expect(result).toEqual("matchingId");
+    });
+
+    it("returns correct business id when it has one matching addresses and other has no address", () => {
+      const businessIdWithMatchingAddress = {
+        matchingId: [matchedAddress1, generateLicenseSearchAddress({})],
```
aloverso, line 73: [dust/sand] This occurred to me as I'm reading through the test and code here. Is it possible to ever have one of these `BusinessIdToAddressesMapping` types with more than one ID key? For example:
```
  const businessIdWithMatchingAddress = {
        "matchingId": [matchedAddress1, generateLicenseSearchAddress({})],
        "some-other-id" [adddress1]
  }
```

If so: I think we need a test for that case.

If not: Then I don't think a record is the appropriate type to represent this. If there's only ever ONE id with addresses, then why have a record (that you do Object.entries on)? A more appropriate type might be:
```
type X = {
  businessId: string;
  addresses: LicenseSearchAddress[]
}
```

api/src/domain/license-status/searchLicenseStatusFactory.ts
```diff
@@ -1,84 +1,18 @@
...
-      return cleanAddress(it.addressLine1).startsWith(cleanAddress(nameAndAddress.addressLine1));
-    });
-
-    if (allMatchingAddressesArray.length === 0) {
-      throw new Error("NO_MATCH");
+import { SearchLicenseStatus } from "../types";
+
+export const searchLicenseStatusFactory = (
+  webServiceSearchLicenseStatus: SearchLicenseStatus,
+  dynamicsSearchLicenseStatus: SearchLicenseStatus
```
aloverso, line 5: [sand] Small & picky but for pure DI, I'd really want to avoid referencing any specific implementations by name anywhere in our domain code. Like, this approach is injecting them properly (that's DI) but just the variable naming is still making this code "verbally depend" on the concept of dynamics specifically.

We could call them ("legacy" and "new") or even ("1" and "2"). Idk. This is probably the one place in the whole code you'll ever hear me arguing AGAINST more specific, explicit, readable variable naming, but I feel like it's not right to go to these lengths to have our domain code decoupled from specific implementations, and then have the variable name undo that by referencing the specific implementations directly.

api/src/domain/types.ts
```diff
@@ -61,6 +68,45 @@ export type EncryptTaxId = (userData: UserData) => Promise<UserData>;
 export type AddNewsletter = (userData: UserData) => Promise<UserData>;
 export type AddToUserTesting = (userData: UserData) => Promise<UserData>;
 
+export interface AccessTokenClient {
+  getAccessToken: () => Promise<string>;
```
aloverso, line 72: [dust/sand] This is an idea for a follow-up refactor for sure, but hear me out:

I think almost all of these type definitions (for AccessTokenClient, BusinessAddressClient, etc) do not belong here in domain. I think they should live in a `types` file within the `client` folder specifically. Because they should be scoped to that folder and that folder alone. The app at large does NOT have a concept of an AccessTokenClient, and I don't think it should. It's a sub-concept of how the DynamicsLicenseStatusClient works - it's not its own thing in isolation.

From there, you could actually _cement_ that idea in the architecture of the code. You could do this by putting these types and all the "sub-clients" for Dynamics in their own subfolder inside of `/client`. And then you can use the DependencyChecker to explicitly prevent any imports from anywhere else in the code other than the `client` folder and the `app.ts` setup.

What this would do is concretely define that these are NOT types that are available to the application interface. By having them in the domain here, we're basically saying that any domain code could in theory make a call to a `getAccessToken` function on a client if it's provided a client of that type. That isn't really accurate because these should not be functions that are "available" to anything in the domain - they should only be available to the stuff in Client, because that's their purpose of existing.

I might be off-base on this idea, but I'm happy to talk more about it if you want.

api/src/domain/types.ts
```diff
@@ -61,6 +68,45 @@ export type EncryptTaxId = (userData: UserData) => Promise<UserData>;
...
+}
+export interface BusinessIdClient {
+  getBusinessId: (accessToken: string, nameToSearch: string) => Promise<string[]>;
+}
+
+export interface BusinessAddressClient {
+  getMappedBusinessAddressesByIds: (
+    accessToken: string,
+    businessIds: string[]
+  ) => Promise<Awaited<BusinessIdToAddressesMapping>[]>;
```
aloverso, line 87: same comment as before about not needing `Awaited`

web/src/components/tasks/LicenseStatusReceipt.tsx
```diff
@@ -61,16 +61,47 @@ const LicenseStatusLookup: Record<LicenseStatus, string> = {
...
+  REVOKED: Config.licenseSearchTask.revokedStatusText,
 };
 
 export const LicenseStatusReceipt = (props: Props): ReactElement => {
   const [theme, setTheme] = useState<PermitTheme>(pendingPermitTheme);
   const { business } = useUserData();
 
+  const isPending = (licenseStatus: LicenseStatus): boolean => {
+    const pendingStatus: LicenseStatus[] = [
+      "PENDING",
```
aloverso, line 86: [pebble] Can we add a test for this? To test for something other than "PENDING"?

api/src/client/DynamicsLicenseStatusClient.ts
```diff
@@ -0,0 +1,64 @@
...
+    const accessToken = await config.tokenFactory.getAccessToken();
+
+    const businessIds = await config.businessIdFactory.getBusinessId(accessToken, nameAndAddress.name);
+
+    const mappedBusinessAddressesByIds = await config.businessAddressFactory.getMappedBusinessAddressesByIds(
+      accessToken,
+      businessIds
+    );
+
+    const matchedBusinessId = searchBusinessAddressesForMatch(mappedBusinessAddressesByIds, nameAndAddress);
```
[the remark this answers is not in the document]
aloverso: update on this: maybe not internal, because it seems like it should definitely continue to have its own unit test. Maybe scoped to the same folder?

.circleci/config.yml
```diff
@@ -222,6 +222,12 @@ commands:
             echo 'export BASIC_AUTH_USERNAME=$BASIC_AUTH_USERNAME' >> $BASH_ENV
             echo 'export BASIC_AUTH_PASSWORD=$BASIC_AUTH_PASSWORD' >> $BASH_ENV
 
+            echo 'export DCA_DYNAMICS_ORG_URL_DEV=$DCA_DYNAMICS_ORG_URL' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_CLIENT_ID_DEV=$DCA_DYNAMICS_CLIENT_ID' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_SECRET_DEV=$DCA_DYNAMICS_SECRET' >> $BASH_ENV
+            echo 'export DCA_DYNAMICS_TENANT_ID_DEV=$DCA_DYNAMICS_TENANT_ID' >> $BASH_ENV
+            echo 'export FEATURE_DYNAMICS_PUBLIC_MOVERS_DEV=$FEATURE_DYNAMICS_PUBLIC_MOVERS' >> $BASH_ENV
+
```
jphechter, line 230: [pebble] Even though we don't know what these values will be, should we set these for `set-prod-env-vars`?

jphechter: Just thinking so that when the time comes to officially release this it's just a change in CircleCI.

api/src/client/DynamicsBusinessAddressClient.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  });
+
+  const mockAccessToken = "access-granted";
+  const mockBusinessIds = ["id2", "id3"];
+
+  it("makes a get request to dynamics api with business address query and auth token in header", async () => {
+    const businessAddressMockResponse = {
+      value: [
+        {
+          createdo: "2023-05-31T10:31:51Z",
```
adidner, line 32: createdon?

api/src/client/DynamicsBusinessAddressClient.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        },
+      }
+    );
+  });
+
+  it("only returns addresses that are active based on statecode", async () => {
+    const businessAddressMockResponse = {
+      value: [
+        {
+          createdo: "2023-05-31T10:31:51Z",
```
adidner, line 113: createdon

somabadri: This is the response directly from Dynamics API so we have no control over how the key is named, same reason for comment above as well.

.circleci/config.yml
somabadri: yeah sure, I don't see why not, I will add them!

api/src/client/DynamicsBusinessIdsClient.ts
somabadri: I changed it to `getBusinessIds` because `getMatchingBusinessIds` might be confusing as we use the `matching` prefix later on in the file to describe the business id after processing.

api/src/domain/license-status/searchLicenseStatusFactory.ts
somabadri: I understand the decoupling of the name and I thought about that as I was building this as well for the same reason. I was hesitant as the test names felt not specific enough? Do we change it so it just says`returns new license status client when env var is true`, is that okay?

api/src/domain/types.ts
somabadri: @FarazA22 and I talked about this and we want to bring it up to the team at large because there are tradeoffs to the different solutions we came up with and there is value to getting everyones input.

api/src/client/DynamicsLicenseStatusClient.ts
somabadri: added both instances of errors (status code error and NO_MATCH)

api/src/client/dynamics/DynamicsLicenseStatusClient.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        },
+        {
+          title: "Registration Fee",
+          status: "ACTIVE",
+        },
+      ],
+    });
+  });
+
+  it("throws a NO_MATCH error when a sub client client throws a N0_MATCH", async () => {
```
aloverso, line 100: typo? `N0_MATCH`

api/src/client/dynamics/DynamicsBusinessAddressClient.test.ts
```diff
@@ -0,0 +1,184 @@
...
+  });
+
+  const mockAccessToken = "access-granted";
+  const mockBusinessIds = ["id2", "id3"];
+
+  it("makes a get request to dynamics api with business address query and auth token in header", async () => {
+    const businessAddressMockResponse = {
+      value: [
+        {
+          createdo: "2023-05-31T10:31:51Z",
```
jphechter, line 32:
```suggestion
          createdon: "2023-05-31T10:31:51Z",
```

api/src/client/dynamics/searchBusinessAddressesForMatch.ts
```diff
@@ -0,0 +1,31 @@
...
+      userInputtedAddress.addressLine1 === address.addressLine1 &&
+      address.zipCode === userInputtedAddress.zipCode
+    );
+  };
+
+  for (const { businessId, addresses } of businessIdsAndLicenseAddresses) {
+    for (const address of addresses) {
+      if (isAddressMatch(address, nameAndAddress)) {
+        result.push(businessId);
+        break;
```
jphechter, line 24: Should this break be here? If we want to make sure that we're only getting one match, wouldn't we want to iterate over the whole list? This would cause us to exit the for loop on the first match, right?

api/src/client/dynamics/DynamicsLicenseStatusClient.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const checkListItems = await config.checkListItemFactory.getChecklistItems(
+      accessToken,
+      applicationIdResponse.applicationId
+    );
+
+    const response = {
+      status: applicationIdResponse.licenseStatus,
+      expirationISO: applicationIdResponse.expirationDate,
+      checklistItems: [...checkListItems],
```
jphechter, line 61:
```js
      checklistItems: checkListItems,
```
or
```js
      checklistItems,
```

api/src/client/dynamics/searchBusinessAddressesForMatch.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    it("throws error when there are no addresses", () => {
+      expect(() => searchBusinessAddressesForMatch([businessWithNoAddresses], userInputtedAddress)).toThrow(
+        "NO_MATCH"
+      );
+    });
+  });
+
+  describe("multiple business ids", () => {
+    it("returns correct business id when it has one matching addresses", () => {
```
jphechter, line 57:
```suggestion
    it("returns correct business id when it has one matching address", () => {
```

api/src/client/dynamics/DynamicsLicenseApplicationIdClient.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    expect(
+      await client.getLicenseApplicationId(mockAccessToken, mockBusinessId, "Public Movers and Warehousemen")
+    ).toEqual({
+      expirationDate: "2023-09-30T00:00:00Z",
+      applicationId: "7b5c178c-2805-ee11-a81c-001dd805ef6c",
+      licenseStatus: "ACTIVE",
+    });
+  });
+
+  it("throws NO_MAIN_APPS error if there are no matching main applications", async () => {
```
aloverso, line 169: [sand] I think it'd be helpful for this test name (and the following one) to state/clarify what makes something a main application (it seems like ending with "00"?)

Like, just from reading this test setup, I can't immediately tell  _why_ this particular setup has "no main apps" unless you already know.

api/src/client/dynamics/DynamicsLicenseApplicationIdClient.ts
```diff
@@ -0,0 +1,109 @@
...
+      .then((response) => {
+        logWriter.LogInfo(
+          `Dynamics License Application Id Client - Id:${logId} - Response: ${JSON.stringify(response.data)}`
+        );
+
+        const mainActiveApplications = response.data.value
+          .filter((applicationDetails: LicenseApplicationIdApiResponse) => applicationDetails.statecode === 0)
+          .filter(
+            (applicationDetails: LicenseApplicationIdApiResponse) =>
+              applicationDetails.rgb_number.slice(-2) === "00"
```
aloverso, line 36: [sand] The magic string `"00"` could be stored as a variable with a descriptive name. That would help with clarity

aloverso: Thinking about it now, even the `0` for statecode is a magic number and could be stored as an `ACTIVE_STATECODE` variable or something

api/src/client/dynamics/DynamicsLicenseStatusClient.ts
```diff
@@ -0,0 +1,71 @@
...
+  ChecklistItemsClient,
+  LicenseApplicationIdClient,
+} from "./types";
+
+type Config = {
+  tokenFactory: AccessTokenClient;
+  businessIdFactory: BusinessIdClient;
+  businessAddressFactory: BusinessAddressClient;
+  applicationIdFactory: LicenseApplicationIdClient;
+  checkListItemFactory: ChecklistItemsClient;
```
aloverso, line 18: [sand] This is a little nitpicky, but I wouldn't name these variables with the word `Factory`. (and when you _do_ that, they should reflect what is actually being returned)

In my opinion, and in the patterns we use, a "ThingFactory" is a function that returns a Thing. 

So a `checklistItemsFactory` by naming pattern would mean "a function that returns to me checklist items" - which doesn't make sense, because traditionally a "factory" returns an object, not a piece of data.

Maybe the following naming / type matchup will help explain:
```
checklistItemsFactory: () => ChecklistItems[] // as stated above, doesn't really make sense
checklistItemsClientFactory: () => ChecklistItemsClient // this is an expected use of a factory
checklistItemsClient: ChecklistItemsClient // when you just have an object of that type, it's not a factory, so the variable name should reflect that
```

api/src/domain/user/updateLicenseStatusFactory.ts
```diff
@@ -61,7 +64,11 @@ export const updateLicenseStatusFactory = (searchLicenseStatus: SearchLicenseSta
         });
       })
       .catch(async (error: Error) => {
-        if (error.message === "NO_MATCH") {
+        if (
+          error.message === "NO_MATCH" ||
+          error.message === "NO_MAIN_APPS" ||
+          error.message === "MULTIPLE_MAIN_APPS"
```
aloverso, line 70: [sand] You could store these hardcoded strings in the domain/types file so that we're not relying on typing the strings out.

```
export const NO_MATCH_ERROR = "NO_MATCH";
export const NO_MAIN_APPS_ERROR = "NO_MAIN_APPS";
// and so on
```

PR #6480, opened 2023-10-19
feat: [#185681031] added note alert when businesses formed outside th…

adidner:
…e Navigator in profile

## Description

We wanted businesses that formed outside the Navigator to get a special message in profile

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/185681031)

## Code author checklist

### DIFF
--- a/web/src/components/profile/ProfileNoteForBusinessesFormedOutsideNavigator.tsx
+++ b/web/src/components/profile/ProfileNoteForBusinessesFormedOutsideNavigator.tsx
@@ -1,0 +1,23 @@
+import { Alert } from "@/components/njwds-extended/Alert";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { isFormedOutsideNavigator } from "@/lib/domain-logic/isFormedOutsideNavigator";
+import { Business } from "@businessnjgovnavigator/shared/userData";
+import { ReactElement } from "react";
+
+interface Props {
+  business: Business | undefined;
+}
+
+export const ProfileNoteForBusinessFormedOutsideNavigator = (props: Props): ReactElement => {
+  const { Config } = useConfig();
+
+  if (!isFormedOutsideNavigator(props.business)) {
+    return <></>;
+  }
+
+  return (
+    <Alert variant="note" borderSmall borderRight>
+      {Config.profileDefaults.default.noteForBusinessesFormedOutsideNavigator}
+    </Alert>
+  );
+};
--- a/web/test/pages/profile/profile-shared.test.tsx
+++ b/web/test/pages/profile/profile-shared.test.tsx
@@ -28,7 +28,11 @@ import {
   OperatingPhases,
   randomInt,
 } from "@businessnjgovnavigator/shared";
-import { generateTaxFilingData, randomFilteredIndustry } from "@businessnjgovnavigator/shared/test";
+import {
+  generateFormationData,
+  generateTaxFilingData,
+  randomFilteredIndustry,
+} from "@businessnjgovnavigator/shared/test";
 
 import {
   chooseTab,
@@ -98,7 +102,7 @@ jest.mock("@/lib/data-hooks/useRoadmap", () => ({ useRoadmap: jest.fn() }));
 jest.mock("@/lib/utils/analytics", () => setupMockAnalytics());
 
 const mockAnalytics = analytics as jest.Mocked<typeof analytics>;
-const personas: BusinessPersona[] = ["STARTING", "FOREIGN"];
+const nonOwningPersonas: BusinessPersona[] = ["STARTING", "FOREIGN"];
 
 describe("profile - shared", () => {
   beforeEach(() => {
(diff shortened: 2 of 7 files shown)


web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -386,6 +390,77 @@ describe("profile - shared", () => {
...
+          businessPersona: "OWNING",
+        }),
+      });
+      renderPage({ business });
+      expect(
+        screen.getByText(Config.profileDefaults.default.noteForBusinessesFormedOutsideNavigator)
+      ).toBeInTheDocument();
+    });
+
+    it("does NOT show Note Alert for businesses that payed via the Navigator", () => {
```
AnJuHyppolite, line 429: Updated `payed` to `paid`

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -386,6 +390,77 @@ describe("profile - shared", () => {
     });
   });
 
+  describe("Special Note Alert for Businesses Formed outside the Navigator", () => {
```
AnJuHyppolite, line 393: Can we remove `Special` from the test description or add it to the others below, so it doesn't seem like it's two different notes (e.g. a special one and a not-so-special one)?

web/src/lib/utils/helpers.ts
```diff
@@ -239,6 +240,17 @@ export const getConfigFieldByLegalStructure = (
   }
 };
 
+export const isFormedOutsideNavigator = (business: Business | undefined): boolean => {
```
FarazA22, line 243: This function has a good name, easy to understand

web/src/components/profile/ProfileNoteForBusinessesFormedOutsideNavigator.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { ReactElement } from "react";
+
+interface Props {
+  business: Business | undefined;
+}
+
+export const ProfileNoteForBusinessFormedOutsideNavigator = (props: Props): ReactElement => {
+  const { Config } = useConfig();
+
+  return isFormedOutsideNavigator(props.business) ? (
```
FarazA22, line 14: Since All we are doing here is returning an alert, is this component `ProfileNoteForBusinessFormedOutsideNavigator` necessary?

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -386,6 +390,77 @@ describe("profile - shared", () => {
...
+    it.each(["FOREIGN", "STARTING"])(
```
FarazA22, line 394: Since your using `.each` each test is run independently. There when the Foreign test is run, the `title - for FOREIGN and STARTING businesses` is not accurate. You can reference the the value of each array element in the title to resolve this. 

Link that was helpful for me - https://elfi-y.medium.com/reduce-boilerplate-test-code-with-jest-it-each-30a0eec9776d

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -386,6 +390,77 @@ describe("profile - shared", () => {
...
+        expect(
+          screen.getByText(Config.profileDefaults.default.noteForBusinessesFormedOutsideNavigator)
+        ).toBeInTheDocument();
+      }
+    );
+
+    it("shows the Note Alert for OWNING businesses always", () => {
+      const business = generateBusinessForProfile({
+        formationData: generateFormationData({
+          completedFilingPayment: false,
```
FarazA22, line 416: Dust - to fully bulletproof the `isFormedOutsideNavigator` function, should we update this to true? Wondering this because, let's say in the future we remove `business.profileData.dateOfFormation)` from the condition, the function would be failing because the first half of the conditional is true. 

  if (
    (business.formationData.completedFilingPayment === false && business.profileData.dateOfFormation) ||
    business.profileData.businessPersona === "OWNING"
  ) {

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -386,6 +390,77 @@ describe("profile - shared", () => {
...
+        }),
+      });
+      renderPage({ business });
+      expect(
+        screen.queryByText(Config.profileDefaults.default.noteForBusinessesFormedOutsideNavigator)
+      ).not.toBeInTheDocument();
+    });
+
+    it.each(["FOREIGN", "STARTING"])(
+      "does NOT show Note Alert for businesses (FOREIGN and STARTING) that have not payed and not formed",
```
FarazA22, line 445: pebble/sand - here is another instance where the test and titles don't match

web/src/components/profile/ProfileNoteForBusinessesFormedOutsideNavigator.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <Alert variant="note" borderSmall borderRight>
+      {Config.profileDefaults.default.noteForBusinessesFormedOutsideNavigator}
+    </Alert>
+  ) : (
+    <></>
```
aloverso, line 19: [pebble] We try to avoid ternaries in JSX, the nesting makes it hard to read. I think this would do a guard clause here.

web/src/lib/utils/helpers.ts
aloverso, line 243: [pebble] I think this would be better to live in its own file in `domain-logic` folder. And have its own little unit test.

web/src/components/profile/ProfileNoteForBusinessesFormedOutsideNavigator.tsx
adidner: It has a little bit of logic to see if it exists or not plus I kinda like the obfuscation with a name that says whats going on.

web/test/pages/profile/profile-shared.test.tsx
adidner: Thanks for the tip, thats something I totally forget to do more often than I should

web/src/lib/utils/helpers.ts
adidner: Does it kinda felt redundant given that I'm basically testing all the same logic one level higher when I check if the Note Alert exists?

I'm already doing the tests and plan on finishing but I'm curious if you feel like there is a point where it becomes a poor idea

web/test/pages/profile/profile-shared.test.tsx
adidner: Sure, that feels a bit more robust.

PR #6481, opened 2023-10-20
feat: [185932083] view hidden items accordion scrolls to top when opened

AnJuHyppolite:
## Description

The `View Hidden Items` accordion in the `For You` section scrolls to the top below the nav bar when the accordion opens up.

### Ticket

[#185932083](https://www.pivotaltracker.com/story/show/185932083)

## Code author checklist

### DIFF
--- a/web/src/components/dashboard/SidebarCardsList.test.tsx
+++ b/web/src/components/dashboard/SidebarCardsList.test.tsx
@@ -1,6 +1,7 @@
 import { SidebarCardsList, SidebarCardsListProps } from "@/components/dashboard/SidebarCardsList";
 import { getMergedConfig } from "@/contexts/configContext";
 import analytics from "@/lib/utils/analytics";
+import * as helpers from "@/lib/utils/helpers";
 import { generateCertification, generateFunding, generateSidebarCardContent } from "@/test/factories";
 import { useMockBusiness } from "@/test/mock/mockUseUserData";
 import { generateBusiness, generateProfileData } from "@businessnjgovnavigator/shared/test";
@@ -18,6 +19,13 @@ function mockMaterialUI(): typeof materialUi {
 jest.mock("@mui/material", () => mockMaterialUI());
 jest.mock("@/lib/utils/analytics", () => setupMockAnalytics());
 jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
+jest.mock("@/lib/utils/helpers", () => {
+  return {
+    ...jest.requireActual("@/lib/utils/helpers"),
+    scrollToTopOfElement: jest.fn(),
+  };
+});
+const mockHelpers = helpers as jest.Mocked<typeof helpers>;
 
 const mockAnalytics = analytics as jest.Mocked<typeof analytics>;
 function setupMockAnalytics(): typeof analytics {
@@ -96,7 +104,7 @@ describe("<SidebarCardsList />", () => {
     expect(mockAnalytics.event.for_you_card_unhide_button.click.unhide_cards).toHaveBeenCalledTimes(1);
   });
 
-  describe("empty state messaging", () => {
+  describe("Empty State Messaging", () => {
     it("displays Complete Required Tasks Message when there are no top or bottom cards and display Fundings and Certifications are false", () => {
       renderComponent({});
       expect(screen.getByTestId("complete-required-tasks-msg")).toBeInTheDocument();
@@ -190,6 +198,21 @@ describe("<SidebarCardsList />", () => {
       );
       const forYouCounter = screen.getByTestId("for-you-counter").textContent;
       expect(forYouCounter).toEqual("(1)");
+    });
+  });
+
+  describe("View Hidden Items Accordion", () => {
+    it("calls scrollToTopOfElement when the accordion opens", async () => {
+      renderComponent({ displayFundings: true });
+      fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
+      expect(mockHelpers.scrollToTopOfElement).toHaveBeenCalledTimes(1);
+    });
+
+    it("does not call scrollToTopOfElement when the accordion closes", async () => {
+      renderComponent({ displayCertifications: true });
+      fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
+      fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
+      expect(mockHelpers.scrollToTopOfElement).toHaveBeenCalledTimes(1);
     });
   });
 });
--- a/web/src/components/dashboard/SidebarCardsList.tsx
+++ b/web/src/components/dashboard/SidebarCardsList.tsx
@@ -9,9 +9,9 @@ import { getForYouCardCount } from "@/lib/domain-logic/getForYouCardCount";
 import { MediaQueries } from "@/lib/PageSizes";
 import { Certification, Funding, SidebarCardContent } from "@/lib/types/types";
 import analytics from "@/lib/utils/analytics";
-import { openInNewTab, templateEval } from "@/lib/utils/helpers";
+import { openInNewTab, scrollToTopOfElement, templateEval } from "@/lib/utils/helpers";
 import { Accordion, AccordionDetails, AccordionSummary, useMediaQuery } from "@mui/material";
-import { ReactElement, ReactNode, useState } from "react";
+import { ReactElement, ReactNode, useRef, useState } from "react";
 
 export interface SidebarCardsListProps {
   topCards: SidebarCardContent[];
@@ -28,6 +28,7 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement => 
   const [hiddenAccordionIsOpen, setHiddenAccordionIsOpen] = useState<boolean>(false);
   const { Config } = useConfig();
   const { business } = useUserData();
+  const accordionRef = useRef(null);
 
   const hiddenOpportunitiesCount = (): number => {
     if (props.displayCertifications && props.displayFundings) {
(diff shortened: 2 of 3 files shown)


web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -96,7 +96,7 @@ describe("<SidebarCardsList />", () => {
     expect(mockAnalytics.event.for_you_card_unhide_button.click.unhide_cards).toHaveBeenCalledTimes(1);
   });
 
-  describe("empty state messaging", () => {
+  describe("Empty State Messaging", () => {
```
AnJuHyppolite, line 99: Not part of the story, but the lowercase bothered my eyes. Since I worked on this, I figured it'd be ok to update here.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -58,13 +59,22 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
     );
   };
 
+  const scrollAccordionToTop = (element: HTMLElement | null, { isDesktop }: { isDesktop: boolean }): void => {
```
jphechter, line 62:
```suggestion
  const scrollAccordionToTop = (element: HTMLElement | null, isDesktop: boolean): void => {
```
[dust] My personal preference for this would be to make this just a boolean, not an object. It would still get passed to `scrollToTopOfElement` the same way, but in the context of this function I just think this makes more sense.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -73,6 +83,7 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
                 setHiddenAccordionIsOpen((prevAccordionStatus) => {
                   return !prevAccordionStatus;
                 });
+                scrollAccordionToTop(accordionRef.current, { isDesktop: isDesktopAndUp });
```
jphechter, line 86: [sand] I think it might make more sense to have this only fire when the menu opens. Right now it's also firing on close.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -58,13 +59,22 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
     );
   };
 
+  const scrollAccordionToTop = (element: HTMLElement | null, { isDesktop }: { isDesktop: boolean }): void => {
+    if (element) {
+      setTimeout(() => {
+        scrollToTopOfElement(element.parentElement as HTMLDivElement, { isDesktop });
+      }, 200);
```
jphechter, line 66: [sand] I know this 200ms timeout is here to account for the time it takes for the menu to open. Currently, this hard-coded value lacks context. It might seem a little overkill, but I think it might be valuable to initialize a variable
```js
const timeForAccordionToOpen = 200  // or some other better name
```
and use that in your `setTimeout` to give a reason for the delay.

web/src/components/dashboard/SidebarCardsList.tsx
aloverso: Strong disagree. I think the team has discussed this before, but I really believe that passing raw booleans to any function ever is an anti-pattern, and they should always be parameterized/labelled like AnJu has here. 

If you change this, then when you call the function, it would look like: `scrollToTopOfElement(element, true)` and there is no way of knowing what the true/false boolean is controlling.

web/src/components/dashboard/SidebarCardsList.tsx
aloverso: Agree, but I think this is a pebble. I don't think we should merge the code until this is addressed, because right now it scrolls to an empty part of the page on close, which is antithetical to the intentions set out by the user story.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -58,13 +59,22 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
     );
   };
 
+  const scrollAccordionToTop = (element: HTMLElement | null, { isDesktop }: { isDesktop: boolean }): void => {
+    if (element) {
```
aloverso, line 63: I'd vote for a guard clause instead of nesting the functionality.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -26,8 +26,10 @@ export interface SidebarCardsListProps {
 
 export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement => {
   const [hiddenAccordionIsOpen, setHiddenAccordionIsOpen] = useState<boolean>(false);
+  const [accordionPosition, setAccordionPosition] = useState<number>(0);
```
AnJuHyppolite, line 29: I added the accordion position so we can manage it so when the accordion closes, there isn't that empty space.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -65,10 +67,15 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
           <hr className="margin-top-3 bg-cool-lighter" aria-hidden={true} />
           <div className="desktop:margin-bottom-1">
             <Accordion
+              ref={accordionRef}
               expanded={hiddenAccordionIsOpen}
               onChange={(): void => {
-                if (!hiddenAccordionIsOpen) {
+                if (hiddenAccordionIsOpen) {
```
AnJuHyppolite, line 73: Updated this so when adding the `else`, we don't have a negated condition.

web/src/lib/utils/helpers.ts
```diff
@@ -112,6 +112,18 @@ export const scrollToTopOfElement = (
   }, 100);
 };
 
+export const scrollAccordionToTop = (
```
AnJuHyppolite, line 115: I moved `scrollAccordionToTop` to `helpers.ts` to live with the other utility functions (to be more in line with the function's function (see what I did there...lol)).

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -192,4 +198,18 @@ describe("<SidebarCardsList />", () => {
...
+      renderComponent({ displayFundings: true });
+      fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
+      expect(scrollAccordionToTop).toHaveBeenCalledTimes(1);
+    });
+
+    it("calls scrollAccordionToTop when the accordion opens with certifications", () => {
+      renderComponent({ displayCertifications: true });
+      fireEvent.click(screen.getByTestId("hidden-opportunity-header"));
+      expect(scrollAccordionToTop).toHaveBeenCalledTimes(1);
+    });
```
aloverso, line 213: I think these tests are a good start but I have two notes here.

First, I'm not sure both these tests are necessary. From a TDD perspective, we write one test that fails and then get it passing. So if we write the first test (calls when the accordion opens with fundings) and then we get it passing. Based on the code and the component that we're testing, I don't see a reason that it work work any differently for certifications. So when we write the second test, it would pass immediately. In my opinion, it doesn't add any additional confidence because the accordion doesn't work any differently for certs vs fundings. If there's a specific reason for both these tests to exist that I'm missing, please let me know!

Second, I think there's a missing test here. What happens when we _close_ the accordion? There's specific behavior that was added: that the `scrollAccordionToTop` should NOT get called when the accordion is closed. And there's logic in the code for that (an if/else statement) and right now the tests are only covering one of those 2 paths. So I think we need a test here for the "when accordion is closed" case.

web/src/lib/utils/helpers.ts
aloverso: In my personal opinion, I don't think this belongs in `helpers`. This function is called `scrollAccordionToTop` and includes a variable called `timeForAccordionToOpen`. Basically, everything in here is specific to the accordion. So I don't foresee any other components in the code calling the "scroll accordion" function. And since it's not actually shared code, I think it should stay in the same place with the accordion itself

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -65,10 +67,15 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
...
+                  window.scroll(0, accordionPosition);
```
aloverso, line 74: Was this behavior discussed with the product/design team? It feels a little jarring to have the page jump when I close the accordion.

If this was just the interpretation of the comment from myself and JP - I think the intention is just that _no_ scroll behavior should happen on close. This would also make the code a lot simpler because then you would not need to track the accordionPosition variable at all.

If you did discuss this with product/design and this is desired behavior, then you can ignore this comment.

AnJuHyppolite: I did this based on the comments from you and JP (and what I saw happening in the app).

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -65,10 +66,19 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
...
               onChange={(): void => {
                 if (!hiddenAccordionIsOpen) {
                   analytics.event.for_you_card_unhide_button.click.unhide_cards();
+                  const element = accordionRef.current as unknown as HTMLElement;
+                  if (element) {
+                    scrollToTopOfElement(
+                      element.parentElement as HTMLDivElement,
+                      { isDesktop: isDesktopAndUp },
+                      200
+                    );
```
jphechter, line 80: [pebble] I liked what you had before, where this was contained in it's own function. I think there was a comment about keeping the logic in this file and not in a helper file, but that doesn't mean we have to just get rid of the function.

The other thing I really liked before was explicitly [giving a name to the value 200](https://github.com/newjersey/navigator.business.nj.gov/pull/6481#discussion_r1370315240). This gives additional context and purpose to the value.

web/src/lib/utils/helpers.ts
```diff
@@ -97,7 +97,8 @@ export const scrollToTop = (props?: { smooth?: boolean }): void => {
 
 export const scrollToTopOfElement = (
   element: HTMLDivElement | null,
-  { isDesktop }: { isDesktop: boolean }
+  { isDesktop }: { isDesktop: boolean },
+  waitTime: number | undefined = 100
```
jphechter, line 101: [sand] Cool idea about passing the `waitTime` as a param here! You could take this a step further and add it to the object we have defined on line 100.

jphechter: It could look something like this:
```js
export const scrollToTopOfElement = (
  element: HTMLDivElement | null,
  { isDesktop, waitTime = 100 }: { isDesktop: boolean; waitTime: number | undefined }
): void => {
```

jphechter (approved):

PR #6489, opened 2023-10-24
chore: fix failing FilingCalendar tests

AnJuHyppolite:
## Description

- Updated two failing tests in `FilingCalendar.tests.tsx`:
1. `shows 5 more events when the view more button is clicked`
2. `resets view more button when year is changed`

**Issue**: The way our tests were set up, we get the current date and go two months into the future and populate `x` number of events (for the current year). In the case of the failing tests, there are 10 events. 

On Friday, October 20th, these tests passed, because two months out brings us you to December 20th, and 10 events would be December 30th. However, on Monday, October 23rd, the tests failed because two months out brings us to December 23rd, and 10 events brings us to January 1st (the following year).

**Resolution**: Set up a `currentDateMock` with April as the current date (versus using the actual current date) so there would be no issues with tests with future dates or past dates (since we have one test that calls `.subtract`).

### Ticket

[N/A](https://www.pivotaltracker.com/story/show/)

## Code author checklist

### DIFF
--- a/web/src/components/filings-calendar/FilingsCalendar.test.tsx
+++ b/web/src/components/filings-calendar/FilingsCalendar.test.tsx
@@ -21,12 +21,12 @@ import {
   generateBusiness,
   generateTaxFilingCalendarEvent,
   generateUserDataForBusiness,
-  getCurrentDate,
   LookupIndustryById,
   OperatingPhases,
   randomInt,
   TaxFilingCalendarEvent,
 } from "@businessnjgovnavigator/shared";
+import * as getCurrentDateModule from "@businessnjgovnavigator/shared/dateHelpers";
 import {
   generateLicenseData,
   generatePreferences,
@@ -37,8 +37,7 @@ import {
 import * as materialUi from "@mui/material";
 import { createTheme, ThemeProvider, useMediaQuery } from "@mui/material";
 import { fireEvent, render, screen, within } from "@testing-library/react";
-import dayjs from "dayjs";
-
+import dayjs, { Dayjs } from "dayjs";
 function mockMaterialUI(): typeof materialUi {
   return {
     ...jest.requireActual("@mui/material"),
@@ -65,7 +64,17 @@ jest.mock("@/lib/data-hooks/useUserData", () => ({ useUserData: jest.fn() }));
 jest.mock("@/lib/data-hooks/useRoadmap", () => ({ useRoadmap: jest.fn() }));
 jest.mock("next/router", () => ({ useRouter: jest.fn() }));
 jest.mock("@/lib/utils/analytics", () => setupMockAnalytics());
+jest.mock("@businessnjgovnavigator/shared/dateHelpers", () => {
+  return {
+    ...jest.requireActual("@businessnjgovnavigator/shared/dateHelpers"),
+    getCurrentDate: jest.fn(),
+  };
+});
+const currentDateMock = (getCurrentDateModule as jest.Mocked<typeof getCurrentDateModule>).getCurrentDate;
 
+const getAprilDateOfThisYear = (): Dayjs => {
+  return dayjs().month(3);
+};
 const mockAnalytics = analytics as jest.Mocked<typeof analytics>;
 const Config = getMergedConfig();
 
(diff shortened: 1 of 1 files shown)


web/src/components/filings-calendar/FilingsCalendar.test.tsx
```diff
@@ -37,8 +37,7 @@ import {
 import * as materialUi from "@mui/material";
 import { createTheme, ThemeProvider, useMediaQuery } from "@mui/material";
 import { fireEvent, render, screen, within } from "@testing-library/react";
-import dayjs from "dayjs";
-
+import dayjs, { Dayjs } from "dayjs";
```
seidior, line 40: [dust] We don't use it very much across the codebase (really I only found two instances, `serverless.ts` and `basicauth.ts`), but I'd like for us to get in the habit of using `import type` when importing types. So this line would look like:

```typescript
import dayjs, { type Dayjs } from "dayjs";
```

As a side note, in newer codebases, you might see something like this, but I'd advise against this pattern because we don't use it anywhere else in our code:

```typescript
import { default as dayjs, type Dayjs } from "dayjs";
```

web/src/components/filings-calendar/FilingsCalendar.test.tsx
```diff
@@ -131,12 +141,12 @@ describe("<FilingsCalendar />", () => {
...
 
   it("displays filings only within a year in list view", () => {
-    const farDueDate = getCurrentDate().add(2, "years");
+    const farDueDate = getAprilDateOfThisYear().add(2, "years");
     const farReport = generateTaxFilingCalendarEvent({
       identifier: "whatever",
       dueDate: farDueDate.format(defaultDateFormat),
     });
-    const recentDueDate = getCurrentDate().add(2, "months");
+    const recentDueDate = getAprilDateOfThisYear().add(2, "months");
```
seidior, line 149: [dust] As I understand it, we added two months to get some amount of time in the future, but now that we're specifically setting it to April, any reason we're still adding two months? Just curious. Perhaps a refactor to consider for another day.

web/src/components/filings-calendar/FilingsCalendar.test.tsx
```diff
@@ -719,13 +729,13 @@ describe("<FilingsCalendar />", () => {
     });
 
     it("doesn't display past months in list view", () => {
-      const pastDueDate = getCurrentDate().subtract(1, "months");
+      const pastDueDate = getAprilDateOfThisYear().subtract(1, "months");
```
seidior, line 732: [sand] Ideally, I'd love if this also covered the edge case of the prior year, which is why I was interested when you and @adidner were talking about setting the month to January initially. It's not super important though, and I think this is sufficient to cover our bases for now, but something worth considering if we come back to this test file again.

seidior (approved):
Thank you for turning this around! Great work thinking through the changes that needed to be made to make this test pass again. This was a fun puzzle.

For testing, I made sure to test the changes manually and in CircleCI. My manual test included advancing my computer's date to December 31, a date in January, and a date in April to see if any of those variables affected the test. All three of those times, the test passed, so hopefully we don't run into this situation again with this test file.

I'm approving this for expedited merger into the codebase as these changes are critical to ensuring CircleCI returns to normal for everyone.

web/src/components/filings-calendar/FilingsCalendar.test.tsx
adidner: No idea, wasn't something we wanted to investigate given the blocking nature. Totally agree it's a good candidate for understanding and potential refactor

web/src/components/filings-calendar/FilingsCalendar.test.tsx
adidner: Interesting, didn't know about this. Will do in the future

web/src/components/filings-calendar/FilingsCalendar.test.tsx
adidner: I'm not sure I understand what you mean here. Do you mean adding another test for looking at the prior year? 

This works better with April because if you did January and then subtracted a month you would still be cut off but by the fact you were in the prior calendar year rather than being in the current calendar year but already past that due date 

So the yes would pass but you wouldn't be testing what you wanted.

PR #6511, opened 2023-10-30
chore: [#186365366] reduce relative import use in api folder

FarazA22:
### Ticket
https://www.pivotaltracker.com/story/show/186365366

### Approach

I updated `tsconfig.paths.json ` and `jest.config.ts` in API to allow path names. All backend folders are covered, which should prevent relative imports from being used when importing files from an existing folder..

## Code author checklist


jphechter (commented):
This is a great add! Thank you!

I noticed 3 files that I think we can address.
`formationRouter.test.ts`
`taxFiliingsInterfaceFactory.test.ts`
`fetchMunicipalityByName.ts`

The 2 test files use relative imports in their mocks. I think we can use the new `@domain` path in both of these instances.

I'm not positive about the other, but I think we can use `@shared` there.

If these don't work, that's fine, I'm happy with this change. I didn't check every single instance of the imports. Assuming that the app wouldn't build and tests wouldn't pass if these were wrong.

FarazA22:
> This is a great add! Thank you!
> 
> I noticed 3 files that I think we can address. `formationRouter.test.ts` `taxFiliingsInterfaceFactory.test.ts` `fetchMunicipalityByName.ts`
> 
> The 2 test files use relative imports in their mocks. I think we can use the new `@domain` path in both of these instances.
> 
> I'm not positive about the other, but I think we can use `@shared` there.
> 
> If these don't work, that's fine, I'm happy with this change. I didn't check every single instance of the imports. Assuming that the app wouldn't build and tests wouldn't pass if these were wrong.

I'll create a ticket for this

PR #6518, opened 2023-10-31
fix: [#186366473] added Analytics for formation task help button

adidner:
## Description

We need analytics on the formation help buttons

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186366473)

### Steps to Test

Look at formation pages and hit help button and look at network tab

Also DBA version

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
+++ b/web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
@@ -1,6 +1,6 @@
 import { Content } from "@/components/Content";
 import { ModalTwoButton } from "@/components/ModalTwoButton";
-import { HelpButton } from "@/components/njwds-extended/HelpButton";
+import { FormationHelpButton } from "@/components/njwds-extended/FormationHelpButton";
 import { HorizontalStepper } from "@/components/njwds-extended/HorizontalStepper";
 import { PrimaryButton } from "@/components/njwds-extended/PrimaryButton";
 import { SecondaryButton } from "@/components/njwds-extended/SecondaryButton";
@@ -171,14 +171,14 @@ export const DbaFormationPaginator = (): ReactElement => {
     if (state.stepIndex === 0 && nameIsAvailable) {
       return (
         <ButtonWrapper>
-          <HelpButton />
+          <FormationHelpButton />
           <ForwardButton />
         </ButtonWrapper>
       );
     } else if (state.stepIndex === 1) {
       return (
         <ButtonWrapper>
-          <HelpButton />
+          <FormationHelpButton />
           <BackButton />
           <ForwardButton />
         </ButtonWrapper>
@@ -190,7 +190,7 @@ export const DbaFormationPaginator = (): ReactElement => {
           text={state.dbaContent.Authorize.callToActionText}
           onClick={(): void => setShowCtaModal(true)}
         >
-          <HelpButton />
+          <FormationHelpButton />
           <BackButton />
         </TaskCTA>
       );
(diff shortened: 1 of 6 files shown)


web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
```diff
@@ -190,7 +198,11 @@ export const DbaFormationPaginator = (): ReactElement => {
           text={state.dbaContent.Authorize.callToActionText}
           onClick={(): void => setShowCtaModal(true)}
         >
-          <HelpButton />
+          <HelpButton
+            onClick={(): void => {
+              analytics.event.business_formation_help_button.click.open_live_chat();
+            }}
+          />
```
adidner, line 205: So the DBA formation paginator should also have tests associated probably for analytics I think.

The issue here being I started doing this then realized that there are a bunch of missing tests and for this file and that because there is no file at the moment there a bunch of missing helper functions and its going to be a bit of work to get it done

So I made a follow up refactor friday ticket so I can deal with it in agregate. https://www.pivotaltracker.com/story/show/186377291

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -161,6 +166,39 @@ describe("<BusinessFormationPaginator />", () => {
       await page.stepperClickToBillingStep();
       expect(screen.getByTestId("help-button")).toBeInTheDocument();
     });
+
+    it("the help button fires analytics on every page", async () => {
```
FarazA22, line 170: pebble - in other tests we've typically separated each of the assertions into their own test. You can find this example in the onboarding test files. There are instances where we do integration tests, I believe there are large tests in formation. But I think smaller unit test that test one thing is preferred.

it fires on tab 1
it fires on tab 2, etc

web/src/components/njwds-extended/HelpButton.tsx
```diff
@@ -3,7 +3,11 @@ import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { ReactElement } from "react";
 
-export const HelpButton = (): ReactElement => {
+interface Props {
+  onClick?: () => void;
```
FarazA22, line 7: dust - since this button is only used in formation, have you considered defining the onclick within the component itself? The same set of analytics is fired for this button.

If you prefer the props, I strongly recommend making the props required and not optional. I would make it required, because, if we use this button elsewhere, and AC is forgotten, we'll technically be blocked. Or, if we are using this button else where, we could easily forget it.

web/src/components/tasks/business-formation/DbaFormationPaginator.tsx
FarazA22: The ticket is a great idea,

If the analytics is defined inside the component, you would only need to test firing inside the component.

web/src/components/njwds-extended/HelpButton.tsx
adidner: I did but I feel like the odds that we want to reuse this component going forwards is high, given the generic style and use and I think we would want to possibly reuse it in the future. 

Having hard coded analytics wouldn't work if we want to reuse it and be generic which would be better in my opinion. 

I'm okay making this required I suppose, but it kinda assumes that we will always have analytics associated with all intercom things, which is probably fine.

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
adidner: Yeah, thats fine. Historically I've not had super hard preferences but in hind sight granularity is probably not a bad thing. 

I will fix

web/src/components/njwds-extended/HelpButton.tsx
jphechter: Agreed with Faraz's recommendation. In favor of renaming the component to something like `FormationHelpButton` at least for now to ensure the analytics event doesn't fire if the component is used elsewhere.

PR #6537, opened 2023-11-03
feat: [#186093180] HTML Page Titles should reference page content

adidner:
## Description

We wanted page titles for accessibility for some of the pages that were missing. Largely home, profile, and dashboard

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186093180)

### Steps to Test

Turn a screen reader on and navigate to the pages. 

## Code author checklist


web/public/mgmt/config.yml
```diff
@@ -3074,6 +3074,15 @@ collections:
...
               - { label: Beta Modal Button Text, name: betaModalButtonText, widget: string }
+          - label: SEO and Accessibility Page Titles
+            name: seoAndAccessibilityPageTitles
+            collapsed: true
+            widget: object
+            summary: "{{seoAndAccessibilityPageTitles}}"
+            fields:
+              - { label: Home Page Title, name: homePageTitle, widget: string }
+              - { label: Dashboard Page Title, name: dashboardPageTitle, widget: string }
+              - { label: Business Profile Page Title, name: profilePageTitle, widget: string }
```
adidner, line 3085: Yes we probably need all of these. The word dashboard and home don't appear on their pages. 

And the Business Profile does have the words Business Profile but it has markdown header hashtags in front of it and I don't think it wou
ld be good to remove those and I'm not sure it woudl be worth the refactor 
image

content/src/fieldConfig/config.json
```diff
@@ -319,6 +319,11 @@
     "savingText": "Saving",
     "savedText": "Saved 1 minute ago"
   },
+  "seoAndAccessibilityPageTitles": {
+    "homePageTitle": "Home",
+    "dashboardPageTitle": "Dashboard",
+    "profilePageTitle": "Business Profile"
```
seidior, line 325: `seoAndAccessibilityPageTitles.profilePageTitles` seems unnecessarily redundant. I would either put these under their appropriate pages, like so:

- `pagesMetadata.home.title`
- `pagesMetadata.dashboard.title`
- `pagesMetadata.profile.title`

Or if you think the better thing to do would be to keep the key `seoAndAccessibilityPageTitles`, I'd recommend this approach:

- `seoAndAccessibilityPageTitles.home`
- `seoAndAccessibilityPageTitles.dashboard`
- `seoAndAccessibilityPageTitles.profile`

I prefer the first approach, though, since it gives us room to expand page metadata in the future.

web/jest.config.ts
```diff
@@ -8,6 +8,8 @@ export default async (): Promise<Config.InitialOptions> => {
     setupFilesAfterEnv: ["./setupTests.js"],
     testEnvironment: "jsdom",
     testPathIgnorePatterns: ["<rootDir>/.next/", "<rootDir>/node_modules/", "<rootDir>/cypress/"],
+    rootDir: "./",
+    moduleDirectories: ["node_modules", "<rootDir>"],
```
seidior, line 12: If these two additions are necessary, I'd separate them out into a separate chore.

web/public/mgmt/config.yml
seidior: I tend to agree with you here. I'd like to have config-style documents as purely markup-free strings and other primitives. I'd create a separate chore and make that change, assuming your story requires Design's review anyway.

web/src/pages/index.tsx
```diff
@@ -147,95 +149,100 @@ const Home = (props: Props): ReactElement => {
...
-                  {landingPageConfig.section4FirstIconText}
-                </div>
-              </div>
+    <>
+      <NextSeo title={`${seoTitle.title} - ${Config.seoAndAccessibilityPageTitles.homePageTitle}`} />
+      <PageSkeleton landingPage={true}>
+        <NavBar landingPage={true} />
+        <main data-testid="main">
+          <Hero isWelcomePage={props.isWelcomePage} />
+          <section ref={sectionHowItWorks} aria-label="How it works">
```
seidior, line 158: I know this was pre-existing, but I would recommend moving this label to config as well.

seidior (commented):
Thorough work here; well done, Aaron. I appreciate that you thought about best practices at every step here, both for configurability and long-term maintainability. I have just a couple of notes!

web/public/mgmt/config.yml
```diff
@@ -3074,6 +3074,15 @@ collections:
             fields:
               - { label: Beta Main Text, name: betaMainText, widget: string }
               - { label: Beta Modal Button Text, name: betaModalButtonText, widget: string }
+          - label: SEO and Accessibility Page Titles
```
jphechter, line 3077: [pebble] How do you feel about moving the contents of `next-seo.config.js` to be defined here? We could still import `config` and set the values in that file.

This would give the content team the ability to modify the site name and description if necessary. Further gives them control to change the name of "The Navigator" when the time comes.

web/jest.config.ts
jphechter, line 12: What is this changing for us?

web/src/components/navbar/NavBarLandingDesktop.tsx
```diff
@@ -18,11 +19,7 @@ export const NavBarLandingDesktop = (): ReactElement => {
           "grid-container-widescreen desktop:padding-x-7 height-8 flex flex-justify flex-align-center flex-wrap"
         }
       >
-        <img
-          className="height-4 margin-y-2"
-          src="/img/Navigator-logo@2x.png"
-          alt="Business.NJ.Gov Navigator"
-        />
+        <img className="height-4 margin-y-2" src="/img/Navigator-logo@2x.png" alt={`${seoTitle.title}`} />
```
jphechter, line 22: [sand]
```suggestion
        <img className="height-4 margin-y-2" src="/img/Navigator-logo@2x.png" alt={seoTitle.title} />
```

web/src/components/navbar/NavigatorLogo.tsx
```diff
@@ -13,7 +14,7 @@ export const NavigatorLogo = (props: Props): ReactElement => {
         <img
           className={props.useSmall ? "height-3" : "height-4"}
           src="/img/Navigator-logo@2x.png"
-          alt="Business.NJ.Gov Navigator"
+          alt={`${seoTitle.title}`}
```
jphechter, line 17:
```suggestion
          alt={seoTitle.title}
```

web/jest.config.ts
adidner: I imagine you mean separate commits not separate pivotal tickets?

This is necessary, because if we're using the seo.config.js which is the the root directory our jest tests can't actually see that file unless the root directory is a moduleDirectory

the rootDir declaration is just a good habit and the node_modules are necessary because I believe it implicitly grabs them unless you declare this moduleDirectories field in which case you have to be explicit.

web/public/mgmt/config.yml
adidner: I'mma look into this after lunch, I think that would be fine. Idk how easy it would be to setup given its a file in a different place that needs to be named as such to work with our SEO stuff but I'll explore a bit

web/jest.config.ts
adidner: This is necessary, because if we're using the seo.config.js to grab the pre fix of `Navigator.nj.gov` the file is in the root directory and our jest tests can't actually see that file unless the root directory is a moduleDirectory

the rootDir declaration is just a good habit and the node_modules are necessary because I believe it implicitly grabs them unless you declare this moduleDirectories field in which case you have to be explicit.

web/src/components/navbar/NavBarLandingDesktop.tsx
adidner: Totally, will fix after lunch (bad habit of mine)

web/src/components/navbar/NavigatorLogo.tsx
adidner: Totally, will fix after lunch (bad habit of mine)

web/src/pages/index.tsx
adidner: Where in config were you imagining? I can look around for a good spot but I'm wondering if you had a place in mind

content/src/fieldConfig/config.json
adidner: Will modify accordingly after lunch

web/public/mgmt/config.yml
adidner: I'll write a follow up ticket after lunch for getting rid of the whole H3 in a json file thing and finding similar issues as a refactor ticket

web/src/pages/index.tsx
seidior: I did not have a specific place in mind; I trust wherever you place it.

web/public/mgmt/config.yml
adidner: https://www.pivotaltracker.com/story/show/186435464

web/public/mgmt/config.yml
adidner: I did this just FYI

PR #6544, opened 2023-11-09
feat: [#186519052]: DCA Dynamics Fire Safety and Housing Setup

SirSamuelJoseph:
## Description

Enable API access to Fire Safety and Housing DCA Dynamics Environment

### Ticket

[186519052](https://www.pivotaltracker.com/story/show/)

### Approach
This follows similar dynamics integrations

### Steps to Test
Hit the API endpoints at /api/fire-safety and /api/housing

### Notes

We are still waiting for decisions from product for what we are doing with these APIs so this is just to enable the connections

I still have to add the values to the Bitbucket and will be doing so today before a merge. I will also add them to CircleCI 

## Code author checklist


api/src/client/dynamics/fire-safety/DynamicsFireSafetyInspectionClient.ts
```diff
@@ -0,0 +1,60 @@
...
+  getFireSafetyInspectionsByAddress: (
+    accessToken: string,
+    address: string
+  ) => Promise<FireSafetyInspection[]>;
+}
+export const DynamicsFireSafetyInspectionClient = (
+  logWriter: LogWriterType,
+  orgUrl: string
+): FireSafetyInspectionClient => {
+  const getFireSafetyInspectionsByAddress = async (
```
somabadri, line 15: [pebble/sand] getFireSafetyInspection instead of getFireSafetyInspectionByAddress because address isn't the only parameter being passed in?

api/src/client/dynamics/housing/DynamicsHousingPropertyInterestClient.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+      )
+      .then((response) => {
+        logWriter.LogInfo(
+          `Dynamics Housing Property Interest Client - Id:${logId} - Response: ${JSON.stringify(
+            response.data
+          )}`
+        );
+        return response.data.value.map((response: DynamicsHousingPropertyInterestResponse) => {
+          return {
```
somabadri, line 32: [pebble] thoughts on pulling the processing of the response into it's own function below just to reduce clutter within this function

api/src/api/fireSafetyRouter.ts
```diff
@@ -0,0 +1,21 @@
+import { FireSafetyInspectionStatus } from "@domain/types";
+
+import { FireSafetyInspectionResult } from "@shared/fireSafety";
+import { Router } from "express";
+
+export const fireSafetyRouterFactory = (fireSafetyInspection: FireSafetyInspectionStatus): Router => {
+  const router = Router();
+
+  router.get("/fire-safety/:address", async (req, res) => {
+    const address = req.params.address as string;
```
somabadri, line 10: [sand] I know we don't have a ui designed yet, but will the address be a single string? Will it not have like a street address 1, zipcode, city? If so we might have rethink how this is parameterized but if not then nevermind.

shared/src/housing.ts
```diff
@@ -0,0 +1,8 @@
+export type HousingPropertyInterestDetails = {
+  createdOn: string;
+  isFireSafety: boolean;
+  isBHIRegistered: boolean;
+  address: string;
+  BHINextInspectionDueDate: string;
+  stateCode: number;
```
somabadri, line 7: [pebble] Is there a specific reason to why we are planning on sending the `stateCode` to the frontend? I'm not sure what it represents but for the license work it represented if a data object was active or inactive and if that's the case here as well I think we should be filtering out the inactive objects before sending it to the frontend and the details object shouldn't have this field.

api/src/client/dynamics/fire-safety/DynamicsFireSafetyInspectionClient.ts
```diff
@@ -0,0 +1,60 @@
...
+            Authorization: `Bearer ${accessToken}`,
+          },
+        }
+      )
+      .then((response) => {
+        logWriter.LogInfo(
+          `Dynamics Fire Safety Client - Id:${logId} - Response: ${JSON.stringify(response.data)}`
+        );
+        return response.data.value.map((response: DynamicsFireSafetyInspectionResponse) => {
+          return {
```
somabadri, line 36: [pebble] same as the other client file, can we pull the processing portion into it's own function so there is less going on in the function to improve readability

shared/src/housing.ts
SirSamuelJoseph: No, I still have to confirm that that's the behavior of statecode on our APIs. But depending on the definition of "active", we might still want it. If we're pulling a history of inspections, a non-active inspection could still be valuable. Unless I'm misinterpreting the definition of an active data object here.

api/src/client/dynamics/housing/DynamicsHousingPropertyInterestClient.ts
SirSamuelJoseph: That makes sense as it might evolve to be more complicated, will do

shared/src/housing.ts
somabadri: I should have specified further, but in the license status world it meant like an old address or an application that is not relevant anymore but since you aren't sure yet what it represents, I think it's okay to leave as is and then refactor later when there is more clarity?

SirSamuelJoseph: Yeah that was my plan!

api/src/api/fireSafetyRouter.ts
SirSamuelJoseph: Right now we have just been doing address but this is a great point, I can definitely expand this with named parameters if we expand like that

PR #6562, opened 2023-11-15
bug: fix landing test

cbrunefearless:
### Ticket

## Code author checklist

### DIFF
--- a/web/cypress/support/page_objects/landingPage.ts
+++ b/web/cypress/support/page_objects/landingPage.ts
@@ -1,6 +1,10 @@
+import { useConfig } from "@/lib/data-hooks/useConfig";
+
+const { Config } = useConfig();
+
 class LandingPage {
   getHowItWorksSection = () => {
-    return cy.get('[aria-label="How it works"]');
+    return cy.get(`[aria-label="${Config.landingPage.section4HeaderText}"]`);
   };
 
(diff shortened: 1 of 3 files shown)


web/cypress/support/page_objects/landingPage.ts
```diff
@@ -1,14 +1,19 @@
+import { useConfig } from "@/lib/data-hooks/useConfig";
+
+const { Config } = useConfig();
+let landingPageConfig = Config.landingPage;
```
FarazA22, line 4: @cbrunefearless is the variable necessary?

cbrunefearless: no, just want it to pass first than change :)

web/cypress/support/page_objects/landingPage.ts
```diff
@@ -1,14 +1,19 @@
...
+
 class LandingPage {
   getHowItWorksSection = () => {
-    return cy.get('[aria-label="How it works"]');
+    return cy.get('[aria-label="' + `${landingPageConfig.section4HeaderText}` + '"]');
```
FarazA22, line 8: @cbrunefearless why not use a backtick instead of using single quotes and adding them together, I. think it would be more readable

PR #6567, opened 2023-11-15
Footer

FarazA22:
### Ticket

The purpose of this ticket is to update the footer styling based off of the design system
https://www.pivotaltracker.com/story/show/186384501

### Approach
I cleaned up the styling file for the footer and ended up redoing a lot of the HTML because they were using unnecessary CSS

## Code author checklist


web/src/components/Footer.tsx
seidior, line 1: [pebble] If we're going to rename this file, my personal preference would probably be something like `PageFooter`, as "footer" can mean different things in different contexts.

web/src/components/Footer.tsx
```diff
@@ -0,0 +1,87 @@
+import { Icon } from "@/components/njwds/Icon";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { MediaQueries } from "@/lib/PageSizes";
+import { useMediaQuery } from "@mui/material";
+import { ReactElement } from "react";
+
+export const Footer = (): ReactElement => {
+  const isLargeScreen = useMediaQuery(MediaQueries.desktopAndUp);
```
seidior, line 8: [sand] Is there a way to ditch this media query hook in a follow-up refactor? It seems weird to me that we can't draw the footer without knowing the user's viewport width instead of just letting CSS reflow the content.

web/src/components/Footer.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  return (
+    <div
+      className={`${
+        isLargeScreen ? "flex flex-align-center fjb" : ""
+      } margin-x-auto padding-y-4 grid-container-widescreen desktop:padding-x-7 bg-base-darkest text-white`}
+    >
+      <section aria-label="New Jersey Office of Innovation information and services">
+        <div className={`${isLargeScreen ? "" : "flex flex-column flex-align-center"}`}>
+          <a
```
seidior, line 19: [pebble] In order to better match the Figma, it appears you'll need something like this declaration for links in the footer:

```css
text-underline-offset: 0.25rem;
```

(Not sure if that's something that needs to be set for all links across the site or just these links in the footer, but that makes the underline's spacing better match the design.)

jphechter:
I think everything looks good here.

seidior (commented):
Great turnaround time on this, Faraz. Couple of pebbles and a tangential comment for you!

web/src/components/Footer.tsx
FarazA22: i'll bring this up in design qa
https://www.pivotaltracker.com/story/show/186490214

web/src/components/Footer.tsx
FarazA22: I've noticed they don't work well with Flex / and when your manually changing the size of the view port

FarazA22: let's chat in thread

PR #6573, opened 2023-11-16
feat: [#186363219] Note alert disclaimer in profile for guest mode

adidner:
## Description

We wanted out note alert to appear in guest mode before the user logins

### Ticket

## Code author checklist


cspell.json
```diff
@@ -2,158 +2,164 @@
...
   "words": [
+    "aesthetician",
     "Altius",
-    "BOC-3",
     "Balloonport",
     "Balloonsport",
+    "BOC-3",
+    "brownfields",
     "Brownfields",
+    "businessnjgovnavigator",
```
adidner, line 12: So my IDEA auto alphabetized this. or maybe prettier did. Anyways I just added a few things here like " and "mynj"businessnjgovnavigator"

web/src/lib/domain-logic/shouldShowDisclaimerForProfileNotSubmittingData.ts
```diff
@@ -0,0 +1,24 @@
+import { IsAuthenticated } from "@/lib/auth/AuthContext";
+import { Business } from "@businessnjgovnavigator/shared/userData";
+
+const isFormedOutsideNavigator = (business: Business): boolean => {
```
adidner, line 4: So I thought about making this its own function in its own file with its own tests. But I think that might a refactor if we ever need it somewhere else rather than something to do pre-emptively.

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -459,12 +457,31 @@ describe("profile - shared", () => {
...
         ).not.toBeInTheDocument();
       }
     );
+
+    it.each(businessPersonas)(
+      "shows Note alert for %s business that set DateOfFormation but did NOT pay",
+      (persona) => {
+        const business = generateBusinessForProfile({
+          formationData: generateFormationData({
+            completedFilingPayment: false,
```
FarazA22, line 472: pebble: is it necessary to test `OWNING` / would iterating over`nonOwningPersonas` make more sense? They would never be asked to pay for formation. What is the scenario your trying to test for Owning, maybe another test is needed?

web/src/lib/domain-logic/shouldShowDisclaimerForProfileNotSubmittingData.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    expect(shouldShowDisclaimerForProfileNotSubmittingData(undefined, IsAuthenticated.FALSE)).toBe(true);
+  });
+
+  it("returns true if isAuthenticated is UNKNOWN", () => {
+    const business = generateBusiness({
+      formationData: generateFormationData({
+        completedFilingPayment: false,
+      }),
+      profileData: generateProfileData({
+        dateOfFormation: undefined,
```
FarazA22, line 25: Pebble: based on how your conditional is written, `isFormedOutsideNavigator(business)` is causing the conditional to fail. 

I'd update the `dateOfFormation` and `businessPersona` so the first two conditionals are true (see below)
```
    isFormedOutsideNavigator(business) ||
    business.profileData.businessPersona === "OWNING" ||
    isAuthenticated === IsAuthenticated.FALSE
```

web/src/lib/domain-logic/shouldShowDisclaimerForProfileNotSubmittingData.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        businessPersona: "STARTING",
+      }),
+    });
+    expect(shouldShowDisclaimerForProfileNotSubmittingData(business, IsAuthenticated.UNKNOWN)).toBe(true);
+    expect(shouldShowDisclaimerForProfileNotSubmittingData(business, IsAuthenticated.UNKNOWN)).toBe(true);
+  });
+
+  it("returns true if persona is OWNING and not Authenticated", () => {
```
FarazA22, line 33: pebble - title says ` and not Authenticated` but your passing in `IsAuthenticated.TRUE`.

FarazA22, line 33: Boulder: see comment in Eng Channel, we can chat more about this with the broader team on Monday. It would be great if you can follow the `red, green, refactor` approach to see what tests need to be staged correctly.

adidner: Yeah, in my first iteration I thought we would test via operating phase to know whether a user had logged in (GUEST mode operating phases), guess I missed this test when refactoring to Authentication. My bad

web/src/lib/domain-logic/shouldShowDisclaimerForProfileNotSubmittingData.ts
FarazA22: I don't think it necessary to extract it out, if we ever do, at that point it would make sense to create test. I see it as a helper function that only lives in this file

PR #6588, opened 2023-11-18
Create REMOTE_SELLER_WORKER operating phase

jphechter:
## Description

Creates a dedicated operating phase for remote sellers and workers. This use case is often overlooked, but will likely have it's own flows in the near term. Those flows will likely differ from what most Dakotas encounter. This specific operating phase allows us to preserve existing functionality and may be a good jumping off point for us to work on remote seller and worker behavior in the future.

### Ticket

[#185757536](https://www.pivotaltracker.com/story/show/185757536)

## Code author checklist


web/src/lib/utils/helpers.ts
```diff
@@ -246,3 +246,32 @@ export const getConfigFieldByLegalStructure = (
...
+      // Remove ordered list numbers
+      .replaceAll(/^\s*\d+\.\s/gm, "")
+      // Remove horizontal rules
+      .replaceAll(/^-{3,}\s*$/gm, "")
+      // Remove extra lines
+      .replaceAll(/\n{2,}/g, " ")
+      // Remove all remaining line breaks
+      .replaceAll(/(\r\n|\n|\r)/gm, "")
+  );
+};
```
jphechter, line 277: Most of this is unused, but the idea is that it could be easily reference to comprehensively strip formatting in the future.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -68,82 +69,112 @@ describe("<SidebarCardsList />", () => {
...
 
-    it("does not display Complete Required Tasks Message when displayCertifications is true", () => {
-      renderComponent({ displayCertifications: true });
-      expect(screen.queryByTestId("complete-required-tasks-msg")).not.toBeInTheDocument();
-    });
+        const emptyMessage = screen.getByTestId("empty-for-you-message");
+        const sanitizedString = removeMarkdownFormatting(Config.dashboardDefaults.emptyOpportunitiesHeader);
+        const combinedMessage = `${sanitizedString}${Config.dashboardDefaults.emptyOpportunitiesText}`;
+        expect(emptyMessage).toHaveTextContent(combinedMessage);
+      });
```
jphechter, line 114: Instead of referencing a test id, this will sanitize any markdown formatting and allow us to get the string from the page. It still utilizes the config and is more in-line with the way we've been testing for an element's presence on the page.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -161,44 +218,66 @@ describe("<SidebarCardsList />", () => {
...
+    });
+
+    it("render cards with a one count in the For You header when one funding is present", () => {
+      useMockBusiness(mockBusiness);
+      const genericFunding = generateFunding({
+        isNonprofitOnly: false,
+        county: [],
+        sector: [],
+        employeesRequired: undefined,
+      });
```
jphechter, line 260: This test and the one previous were somewhat flaky. The generator functions were occasionally creating fundings and certifications that would get filtered out before being displayed in the UI. Tried to make a generic funding and certification that would never get filtered.

api/src/db/migrations/v127_create_remote_worker_seller_phase.ts
```diff
@@ -0,0 +1,494 @@
...
+  if (ids.includes("companyOperatedVehiclesInNJ")) return "NEXUS";
+  if (ids.includes("employeesInNJ")) return "REMOTE_WORKER";
+  if (ids.includes("revenueInNJ")) return "REMOTE_SELLER";
+  if (ids.includes("transactionsInNJ")) return "REMOTE_SELLER";
+  return undefined;
+};
+
+const migrate_v126Business_to_v127Business = (business: v126Business): v127Business => {
+  // no data is changing - casting this type was missed in a previous migration
+  const ids = business.profileData.foreignBusinessTypeIds as v127ForeignBusinessTypeId[];
```
FarazA22, line 53: Nice job in catching this.

api/src/db/migrations/v127_create_remote_worker_seller_phase.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  it("does not modify operatingPhase for other Foreign businesses", () => {
+    const foreignBusinessTypeIds = [
+      "none",
+      "employeeOrContractorInNJ",
+      "officeInNJ",
+      "propertyInNJ",
+      "companyOperatedVehiclesInNJ",
+    ];
+    const randomIndex = Math.floor(Math.random() * foreignBusinessTypeIds.length);
```
FarazA22, line 51: pebble - we've moved away from single test that randomly selects a value, reason being is that when a test fails and then passes when re-running, it's now clear why it's failing. We typically test for ever case.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -154,20 +189,13 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
...
-          <img src={`img/for-you-section.svg`} aria-hidden="true" alt={""} />
-        </div>
-      )}
       <div>
         <div data-testid="top-cards">
           {props.topCards.map((card) => {
             return <SidebarCard card={card} key={card.id} />;
           })}
         </div>
+        {emptyForYouMessage()}
```
FarazA22, line 198: I really like how you consolidated all the error messages

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -68,82 +70,137 @@ describe("<SidebarCardsList />", () => {
...
-      expect(screen.queryByText(Config.dashboardDefaults.emptyOpportunitiesText)).not.toBeInTheDocument();
-    });
+  describe("Empty State Messages", () => {
+    describe("Remote Seller/Worker Message", () => {
+      const foreignBusinessTypeIds: ForeignBusinessTypeId[] = [
+        "employeesInNJ",
+        "revenueInNJ",
+        "transactionsInNJ",
+      ];
+      const randomIndex = Math.floor(Math.random() * foreignBusinessTypeIds.length);
```
FarazA22, line 94: pebble - as mentioned earlier I think it's better to test against all use cases

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -141,6 +127,55 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
...
+      if (showEmptyOpportunitiesMsg) return Config.dashboardDefaults.emptyOpportunitiesText;
+      return "";
+    };
+
+    return (
+      <>
+        {showEmptyForYouMessage && (
+          <div className="fdc fac margin-y-3" data-testid="empty-for-you-message">
+            <div className="text-center margin-bottom-3">
+              <Content>{getTopText()}</Content>
```
FarazA22, line 169: Pebble - since this is a child of `Content`, we should change the 3 widget types in the YAML file from string to markdown. The `getBottonText` fn would stay as a string.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -141,6 +127,55 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
...
+    };
+
+    const getBottomText = (): string => {
+      if (showEmptyOpportunitiesMsg) return Config.dashboardDefaults.emptyOpportunitiesText;
+      return "";
+    };
+
+    return (
+      <>
+        {showEmptyForYouMessage && (
```
FarazA22, line 166: dust - would it make sense to add a guard clause for `!showEmptyForYouMessage`

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -141,6 +127,55 @@ export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement =>
...
+        return foreignBusinessType === "REMOTE_SELLER" || foreignBusinessType === "REMOTE_WORKER";
+      }
+      return false;
+    };
+
+    const showEmptyForYouMessage =
+      showEmptyOpportunitiesMsg || showCompleteRequiredTasksMsg || isRemoteSellerWorker();
+
+    const getTopText = (): string => {
+      if (isRemoteSellerWorker()) return Config.dashboardDefaults.emptyOpportunitiesRemoteSellerWorkerText;
```
FarazA22, line 153: pebble - from a content/cms perspective, I think we may need go update the preview so it renders the component

api/src/db/migrations/v127_create_remote_worker_seller_phase.test.ts
jphechter: Great point

web/src/components/dashboard/SidebarCardsList.test.tsx
jphechter: Included

web/src/components/dashboard/SidebarCardsList.tsx
jphechter: Another good callout

web/src/components/dashboard/SidebarCardsList.tsx
jphechter: I thought we landed on this `&&` pattern, didn't we? It's kindof behaving like a guard clause and the alternative would be 2 return statements, one with an empty fragment and the other being the actual content.

FarazA22: Based on how this logic is written, if `showEmptyForYouMessage` is false a react fragment is returned, but if `showEmptyForYouMessage` is true, the same react fragment is returned with additional children elements.

The guard clause would separate the logic on line 165 from 166
```
if (!showEmptyForYouMessage) return null
return <div>...
```

jphechter: Right, I thought in this instance we opted for the && pattern for instances like this. I can go either way.

FarazA22: ok, I'm good either way as well

PR #6701, opened 2023-12-01
chore: [#186598228] updated license calendar test to only use valid m…

somabadri:
…onths

## Description

The logic we have in place indicates that we only show calendar events for that year and the test had to be updated because it wasn't restricted to month values that reflected that. It started to fail because the current month is December, and both the expiration date and renewal date were being dynamically setup using currentDate. The test was expecting a renewal date 30 days later which would be in January of 2024 but that breaks our initial logic that we only ever return events that are in the same year, so as a solution we restricted the test to use only the months of January to November for the expiration date so the renewal date would be in the same year. 

**Note**: We explicitly test the December case in a later test, `returns an array containing only license events within the year`.

### Ticket

[#186598228](https://www.pivotaltracker.com/story/show/186598228)

## Code author checklist


cbrunefearless (approved):
lgtm

seidior (approved):
@somabadri walked me through any concerns I had in person. Approved!

PR #6702, opened 2023-12-01
fix: [#185757752] fix scrolling behavior on alert with href routing

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/185757752

## Code author checklist

### DIFF
--- a/web/src/components/FieldEntryAlert.tsx
+++ b/web/src/components/FieldEntryAlert.tsx
@@ -1,5 +1,6 @@
 import { Content } from "@/components/Content";
 import { Alert, AlertProps } from "@/components/njwds-extended/Alert";
+import { useRouter } from "next/router";
 import { ReactElement, ReactNode } from "react";
 
 interface Props {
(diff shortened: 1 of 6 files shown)


web/src/styles/base/global.scss
```diff
@@ -347,3 +347,7 @@ ol {
 .text-no-underline-override {
   text-decoration: none;
 }
+
+.add-spacing-on-ele-scroll {
+  scroll-margin-top: 100px;
+}
```
jphechter, line 353: [dust] This might be a good time to also set `scroll-behavior: smooth`. It's not part of the scope of work, but it might be a nice thing for us to add given it's behavior that we're incorporating in other parts of the app (ex: hidden opportunities in the For You column).

https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior

jphechter: You may need to add this to a higher level container element.

web/src/components/FieldEntryAlert.tsx
```diff
@@ -13,6 +14,14 @@ interface Props {
 }
 
 export const FieldEntryAlert = (props: Props): ReactElement => {
+  const router = useRouter();
+
+  const onAnchorClick = (): void => {
+    setTimeout(() => {
+      router.push(router.asPath, undefined, { shallow: true });
+    });
+  };
```
jphechter, line 23: I'm assuming this was added to handle some issues with form data being modified or getting wiped, is that correct?

FarazA22: we spoke about this

PR #6727, opened 2023-12-05
Create Webflow industry starter kit dropdown

jphechter:
## Description

Create a dropdown component that can be used in Webflow. The dropdown hosts a list of industries supported by the navigator. Selecting one will drive the user to the Navigator dashboard with that industry pre-selected.

### Ticket

[#185698591](https://www.pivotaltracker.com/story/show/185698591)

### Approach

- Create a new sync task that will update a Webflow collection with Navigator industry information
- Create the dropdown that will populate using the data from Webflow
- Implement the dropdown in the Webflow environment

### Steps to Test

You can run the file locally in your web browser, or visit the Webflow beta site [here](https://beta.business.nj.gov/starter-kits).

### Notes

At the time of writing this PR, I do not have tests around this. I'm having issues getting the JS to fire correctly in Jest. Will try to follow up by adding tests for this functionality.

## Code author checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -757,6 +757,21 @@ jobs:
             yarn workspace @businessnjgovnavigator/content build
             yarn workspace @businessnjgovnavigator/web webflow:licenses-ci-sync
 
+  sync-industries:
+    docker:
+      - image: $AWS_ACCOUNT_ID_DEV.dkr.ecr.$AWS_REGION.amazonaws.com/navigator_builder:bfs_navigator_builder-18_18_2
+        aws_auth:
+          aws_access_key_id: $AWS_ACCESS_KEY_ID_DEV
+          aws_secret_access_key: $AWS_SECRET_ACCESS_KEY_DEV
+    steps:
+      - bootstrap-code
+      - run:
+          name: Node Industry Sync
+          command: |
+            echo 'export WEBFLOW_API_TOKEN=$WEBFLOW_API_TOKEN' >> $BASH_ENV
+            yarn workspace @businessnjgovnavigator/content build
+            yarn workspace @businessnjgovnavigator/web webflow:industries-ci-sync
+
 parameters:
   run_workflow_builder:
     default: false
@@ -797,6 +812,10 @@ workflows:
           name: webflow-licenses-sync
           requires:
             - production-deploy
+      - sync-industries:
+          name: webflow-industries-sync
+          requires:
+            - production-deploy
 
   webflow-fundings-sync:
     jobs:
@@ -813,6 +832,14 @@ workflows:
             branches:
               only:
                 - /^webflow-licenses-sync-.*/
+
+  webflow-industries-sync:
+    jobs:
+      - sync-industries:
+          filters:
+            branches:
+              only:
+                - /^webflow-industries-sync-.*/
 
   build-and-test:
     jobs:
@@ -836,6 +863,7 @@ workflows:
                 - /^cms/.*/
                 - /^webflow-fundings-sync-.*/
                 - /^webflow-licenses-sync-.*/
+                - /^webflow-industries-sync-.*/
                 - /^new-docker-image$/
       - lint:
           name: Lint
@@ -847,6 +875,7 @@ workflows:
                 - /^cms/.*/
                 - /^webflow-fundings-sync-.*/
                 - /^webflow-licenses-sync-.*/
+                - /^webflow-industries-sync-.*/
           requires:
             - build
       - test:
@@ -859,6 +888,7 @@ workflows:
                 - /^cms/.*/
                 - /^webflow-fundings-sync-.*/
                 - /^webflow-licenses-sync-.*/
+                - /^webflow-industries-sync-.*/
           requires:
             - build
       - integration-tests:
@@ -874,6 +904,7 @@ workflows:
                 - /^cms/.*/
                 - /^webflow-fundings-sync-.*/
                 - /^webflow-licenses-sync-.*/
+                - /^webflow-industries-sync-.*/
           requires:
             - build
       - integration-tests:
@@ -889,6 +920,7 @@ workflows:
                 - /^cms/.*/
                 - /^webflow-fundings-sync-.*/
                 - /^webflow-licenses-sync-.*/
+                - /^webflow-industries-sync-.*/
           requires:
             - build
       - integration-tests:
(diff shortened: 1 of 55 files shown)


content/src/roadmaps/industries/acupuncture.json
```diff
@@ -57,5 +57,6 @@
   "isEnabled": true,
   "additionalSearchTerms": "acupuncturist, spa, alternative medicine",
   "id": "acupuncture",
-  "description": "Traditional Chinese medicine involving the insertion of needles into the body"
-}
\ No newline at end of file
+  "description": "Traditional Chinese medicine involving the insertion of needles into the body",
+  "webflowId": "656a136266a0b2c046a17462"
```
jphechter, line 61: Running the sync job created a `webflowId` for all industries.

.circleci/config.yml
```diff
@@ -757,6 +757,21 @@ jobs:
...
+          aws_access_key_id: $AWS_ACCESS_KEY_ID_DEV
+          aws_secret_access_key: $AWS_SECRET_ACCESS_KEY_DEV
+    steps:
+      - bootstrap-code
+      - run:
+          name: Node Industry Sync
+          command: |
+            echo 'export WEBFLOW_API_TOKEN=$WEBFLOW_API_TOKEN' >> $BASH_ENV
+            yarn workspace @businessnjgovnavigator/content build
+            yarn workspace @businessnjgovnavigator/web webflow:industries-ci-sync
```
jphechter, line 773: I accidentally added this command to `package.json` awhile ago which is why that change isn't here in the diff.

There are 2 options `webflow:industries-sync` and `webflow:industries-ci-sync`.

web/src/scripts/webflow/industrySync.mjs
```diff
@@ -0,0 +1,171 @@
...
+const syncIndustries = async (params) => {
+  console.log("updating industry names");
+  const industriesAlreadyInWebflow = await getIndustryNamesAlreadyInWebflow();
+
+  await updateWebflowIndustryNames(industriesAlreadyInWebflow);
+  console.log("creating new industry names");
+
+  if (params.create) {
+    await createNewWebflowIndustryNames();
+    console.log("Complete industry sync!");
```
adidner, line 148: [dust]: You're using console.logs in some areas and console.info in others. Feels inconsistent to me? But perhaps there is a reason I'm missing?

web/src/webflow/dropdown.html
```diff
@@ -0,0 +1,864 @@
...
+  }
+
+  @media screen and (max-width: 500px) {
+    #skf-combo-box-button {
+      margin: 20px 0 0;
+    }
+  }
+</style>
+<div id="skf-container" onclick="detectClickLocation(event)">
+  <h3>Find Other Starter Kits</h3>
```
adidner, line 213: [dust]: Does this work with the page hierarchy? IE is there an H2 before it on the page where it will be used?

web/src/webflow/dropdown.html
```diff
@@ -0,0 +1,864 @@
...
+  }
+
+  .skf-industry-description {
+    color: #5c5c5c;
+    font-weight: 400;
+    font-size: 13px;
+    margin: 0;
+  }
+
+  #skf-combo-box-button {
```
adidner, line 169: [dust]: I'm not sure how I feel about selecting by ID, I don't think its an anti-pattern just not something I see used very often

web/src/webflow/dropdown.html
```diff
@@ -0,0 +1,864 @@
...
+    });
+  };
+
+  const sanitizeString = (string) => {
+    return string.replace("'", "’");
+  };
+
+  const populateDropdown = (industries) => {
+    const dropdown = document.getElementById("skf-dropdown");
+    const input = document.getElementById("skf-input");
```
adidner, line 301: [dust] I feel like my instinct here is to always use ID's to grab elements and always use classes for CSS to select and style things. I feel this division is more robust and more easily changeable.

adidner (approved):
Have a few question and potential changes based on opinion stuff but nothing major. Good job

web/src/scripts/webflow/industrySync.mjs
```diff
@@ -0,0 +1,171 @@
+/* eslint-disable unicorn/no-process-exit */
+/* eslint-disable no-undef */
+
+import fs from "fs";
+import path from "path";
+import { fileURLToPath } from "url";
+import industryJson from "../../../../content/lib/industry.json" assert { type: "json" };
```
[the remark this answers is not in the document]
jphechter: There was an issue with this in CI on the `licenseSync`. For some reason it acted wonky with the absolute path and fine with the relative path. It works both ways locally. I could try and figure this out further, but it just didn't seem like it was worth the effort.

web/src/webflow/dropdown.html
jphechter: I believe so, but I'll double check. Great accessibility thinking.

web/src/webflow/dropdown.html
jphechter: In most instances here I'm using `getElementById` to reference these same elements so I just applied style rules on the ids. I can change that, but given this won't scale I don't know if it's critical.

web/src/webflow/dropdown.html
seidior, line 1: [dust] I'd like to see us use `rem` instead of `px` in the future. Most browsers intelligently scale with increased text size, but it's a good habit to get into and a best practice.

web/src/webflow/dropdown.html
```diff
@@ -0,0 +1,864 @@
...
+    }
+  }
+
+  @media screen and (max-width: 500px) {
+    #skf-combo-box-button {
+      margin: 20px 0 0;
+    }
+  }
+</style>
+<div id="skf-container" onclick="detectClickLocation(event)">
```
seidior, line 212: [dust] If we're adding an `onclick` handler to a non-interactive component, there may be accessibility-related attributes we need to add like `role` so that it gets read out to a screen reader as interactive.

seidior (approved):
Spectacular work as I would expect from you, @jphechter. I have only minor style comments here, no showstoppers.

I've also added a follow-up refactoring chore in the icebox for us to revisit this and convert it to TypeScript rather than ESM. Hopefully that can cut down on the amount of code you have to write and will be easier for us to maintain this in the future.

web/src/scripts/webflow/industrySync.mjs
jphechter: There's no reason. There's some copy/paste going on here. I'll make these all `console.log()`. I can see where someone might argue one or the other, but I don't think it matters for our purposes.

PR #6781, opened 2023-12-11
chore: [#186659137] remove style prop from unstyled button

FarazA22:
### Ticket

## Code author checklist

### DIFF
--- a/web/src/components/PageFooter.tsx
+++ b/web/src/components/PageFooter.tsx
@@ -1,8 +1,6 @@
-import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
 import { Icon } from "@/components/njwds/Icon";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { MediaQueries } from "@/lib/PageSizes";
-import { openInNewTab } from "@/lib/utils/helpers";
 import { useMediaQuery } from "@mui/material";
 import { ReactElement } from "react";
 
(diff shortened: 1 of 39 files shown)


web/src/components/PageFooter.tsx
```diff
@@ -48,17 +46,16 @@ export const PageFooter = (): ReactElement => {
           <hr className="margin-y-2" />
           <div className={`${isLargeScreen ? "" : "flex flex-column flex-align-center"}`}>
             <span className="padding-right-1 text-center">
-              <UnStyledButton
-                onClick={(): void => openInNewTab(Config.footer.officeLink)}
-                className="text-white text-center"
-                ariaLabel="Made with love by the Office of Innovation"
-                style="footerLink"
-                isUnderline={true}
+              <a
```
AnJuHyppolite, line 49: [boulder] Can you provide context as to why we're removing this? This recent update was done to make "Made with ❤️..." accessible.

web/src/components/njwds-extended/UnStyledButton.tsx
```diff
@@ -1,34 +1,24 @@
 import React, { forwardRef, ReactElement, useEffect, useRef, useState } from "react";
 
 interface Props {
-  style: "default" | "transparentBgColor" | "footerLink";
   className?: string;
```
AnJuHyppolite, line 4: [boulder] How are we accounting for the footer?

web/src/components/njwds-extended/UnStyledButton.tsx
```diff
@@ -42,14 +32,12 @@ export const UnStyledButton = forwardRef(
       }
     }, [height, width, disabledClass]);
 
-    const noRightMargin =
```
AnJuHyppolite, line 45: [boulder] This change affects the the Primary and Secondary buttons. Are the style changes affecting these other two components in a way that we want them displayed?

AnJuHyppolite (commented):
@FarazA22 Can you update the Tracker link so it takes us to the story? Also, would love some context around why we're removing the style prop in the Tracker ticket itself and here in the description.

web/src/components/njwds-extended/UnStyledButton.tsx
FarazA22: Based on how the logic was written, the no right margin was applied to every button except the "help" formation button. I've corrected that padding.

web/src/components/njwds-extended/UnStyledButton.tsx
FarazA22: it's converted to an anchor link. Defining styles by location of the app was a motivating factor behind the refactor; I don't believe footerLink accurately describes what the button is doing if selected.

web/src/components/PageFooter.tsx
FarazA22: an anchor link should be used since it's an external link

PR #6834, opened 2023-12-22
chore: add script to extract front matter key from markdown files

seidior:
## Description

Adding in a script to extract a given key from Markdown frontmatter, then added additional guidance for Python unit testing in the Readme.

### Approach

You'll need to set up a virtual environment for Python. Follow the steps in the revised Readme.

### Steps to Test

To test, run `yarn test:python`.

## Code author checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ for CI/CD.
 You will need Node.js (with Yarn installed via `npm` or `corepack`) installed
 for primary development. Additionally, for running the server in local
 development mode, you will need a Java runtime (for `serverless-dynamodb`) and
-Python (for the AWS CLI) installed (details below).
+Python (for the AWS CLI and some of our scripts) installed (details below).
 
 We recommend using WSL2 if developing on Windows.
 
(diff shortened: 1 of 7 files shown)


content/scripts/test_get_frontmatter_key.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
+
```
adidner, line 2: [dust] I get that this is the expected output, but I still found it a bit confusing before looking at the tests, any way we could try and make this a tad clearer?
image

adidner (approved):
Seems good, minor question and comment

AnJuHyppolite (commented):
Looks good save for Aaron's comments!

README.md
```diff
@@ -80,10 +80,19 @@ Before you can run locally, you will need to:
   `./web/.env-template` file.
 - create a `./api/.env` that includes all the values laid out in the
   `./api/.env-template` file.
+- create a `.venv` virtual environment and install requirements if working on
+  Python:
+
+```shell
+python3 -m venv .venv
+source .venv/bin/activate
+pip install -r requirements.txt
```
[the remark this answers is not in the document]
seidior: You would need to `source .venv/bin/activate` each time, as you would need to do that _before_ running `yarn test:python` unfortunately, as it relies on bash/zsh functionality in the same process.

I've added VSCode functionality that automatically detects the virtual environment and sources it when you open a new terminal. Is there a way I can do the same for your editor as well?

content/scripts/test_get_frontmatter_key.py
seidior: Yeah... I didn't love that either, much as I don't love our `console.log()` outputs when we run Jest. I've addressed this in a revised commit that buffers (`-b`) the output, so it shouldn't display anymore. Thanks for the feedback :)

PR #6841, opened 2023-12-26
feat: [#186657856] [#186741011] add new routes for health checks

seidior:
## Description

The first of two health check-related pull requests. (See "Notes" later for more information.) This adds or modifies the following routes to the API for health checks:

| route | description |
| ----- | ------------ |
| `/health` | [Existing] Self-check to ensure the API is up and running. |
| `/health/webservice/license-status` | Establishes a connection to the Webservice server (for Hub License Status) to ensure the server is online and that we can successfully retrieve data. |
| `/health/dynamics/elevator` | Establishes a connection to the Dynamics server (for Muni Elevator Safety) to ensure the server is online and that we can successfully retrieve data. |
| `/health/dynamics/fire-safety` | Establishes a connection to the Dynamics server (for Muni Fire Safety) to ensure the server is online and that we can successfully retrieve data. |
| `/health/dynamics/housing` | Establishes a connection to the Dynamics server (for Muni Housing) to ensure the server is online and that we can successfully retrieve data. |
| `/health/dynamics/license-status` | Establishes a connection to the Dynamics server (for Hub License Status) to ensure the server is online and that we can successfully retrieve data. |

For external services, these are the valid states:

| HTTP Code | Description | Example API Response |
| ----------- | ----------- | --- |
| HTTP 200 OK | Server is online and returns expected result. | `{timestamp: 123, success: true, data: { message: "Alive" } }` |
| HTTP 502 Bad Gateway | Server is online but returned an unexpected result. | `{timestamp: 123, success: false, error: { message: "Unexpected result.", timeout: false, serverResponseCode: 404, serverResponseBody: "Page not found." } }` |
| HTTP 504 Gateway Timeout | Server is unreachable. | `{timestamp: 123, success: false, error: { message: "Timeout of 5000ms reached.", timeout: true } }` |

### Ticket

[#186657856](https://www.pivotaltracker.com/story/show/186657856)

### Steps to Test

1. Start the dev server locally.
2. In another Terminal window, check the HTTP status code:

   ```shell
   curl -s -w "%{http_code}\n" -o /dev/null "http://localhost:5002/local/health"
   ```

3. If you would like to see the body of the request, parsed and pretty-printed:

   ```shell
   curl -s -o - "http://localhost:5002/local/health" | python -m json.tool
   ```

### Notes

Thanks to @SirSamuelJoseph for contributing the health checks for the Muni team's work, to @somabadri for Dynamics license status, and to @jphechter for the webservice guidance.

## Code author checklist

### DIFF
--- a/api/serverless.ts
+++ b/api/serverless.ts
@@ -111,7 +111,7 @@ const serverlessConfiguration: AWS = {
   ],
   provider: {
     name: "aws",
-    runtime: "nodejs16.x",
+    runtime: "nodejs18.x",
     stage: stage,
     region: region,
     iam: {
--- a/api/src/api/healthCheckRouter.test.ts
+++ b/api/src/api/healthCheckRouter.test.ts
@@ -1,0 +1,100 @@
+import { healthCheckRouterFactory } from "@api/healthCheckRouter";
+import type { HealthCheckMethod } from "@domain/types";
+import { setupExpress } from "@libs/express";
+import type { Express } from "express";
+import { ReasonPhrases, StatusCodes } from "http-status-codes";
+import request from "supertest";
+
+const mockOKHealthCheckMethod: HealthCheckMethod = async () => {
+  return new Promise((resolve) => {
+    resolve({
+      success: true,
+      data: {
+        message: ReasonPhrases.OK,
+      },
+    });
+  });
+};
+
+const mockErrorHealthCheckMethod: HealthCheckMethod = async () => {
+  return new Promise((resolve) => {
+    resolve({
+      success: false,
+      error: {
+        message: ReasonPhrases.BAD_GATEWAY,
+        serverResponseBody: "",
+        serverResponseCode: 404,
+        timeout: false,
+      },
+    });
+  });
+};
+
+const mockTimeoutHealthCheckMethod: HealthCheckMethod = async () => {
+  return new Promise((resolve) => {
+    resolve({
+      success: false,
+      error: {
+        message: ReasonPhrases.GATEWAY_TIMEOUT,
+        timeout: true,
+      },
+    });
+  });
+};
+
+describe("healthCheckRouter", () => {
+  let app: Express;
+
+  beforeEach(async () => {
+    app = setupExpress(false);
+    app.use(
+      healthCheckRouterFactory(new Map<string, HealthCheckMethod>([["mockOK", mockOKHealthCheckMethod]])),
+      healthCheckRouterFactory(new Map<string, HealthCheckMethod>([["mockErr", mockErrorHealthCheckMethod]])),
+      healthCheckRouterFactory(
+        new Map<string, HealthCheckMethod>([["mockTimeout", mockTimeoutHealthCheckMethod]])
+      )
+    );
+  });
+
+  afterAll(async () => {
+    await new Promise((resolve) => {
+      return setTimeout(resolve, 500);
+    });
+  });
+
+  describe("GET /", () => {
+    it("should return a successful response for the server's self check", async () => {
+      const response = await request(app).get("/");
+      expect(response.status).toBe(StatusCodes.OK);
+      expect(response.body.success).toBe(true);
+      expect(response.body.data.message).toEqual("OK");
+    });
+  });
+
+  describe("GET /mockOK", () => {
+    it("should return a successful response for a mocked health check", async () => {
+      const response = await request(app).get("/mockOK");
+      expect(response.status).toBe(StatusCodes.OK);
+      expect(response.body.success).toBe(true);
+      expect(response.body.data.message).toEqual("OK");
+    });
+  });
+
+  describe("GET /mockErr", () => {
+    it("should return an unsuccessful response for a errored health check", async () => {
+      const response = await request(app).get("/mockErr");
+      expect(response.status).toBe(StatusCodes.BAD_GATEWAY);
+      expect(response.body.success).toBe(false);
+      expect(response.body.error.message).toEqual(ReasonPhrases.BAD_GATEWAY);
+    });
+  });
+
+  describe("GET /mockTimeout", () => {
+    it("should return a successful response for a mocked health check", async () => {
+      const response = await request(app).get("/mockTimeout");
+      expect(response.status).toBe(StatusCodes.GATEWAY_TIMEOUT);
+      expect(response.body.success).toBe(false);
+      expect(response.body.error.message).toEqual(ReasonPhrases.GATEWAY_TIMEOUT);
+    });
+  });
+});
--- a/api/src/api/healthCheckRouter.ts
+++ b/api/src/api/healthCheckRouter.ts
@@ -1,0 +1,47 @@
+import type { HealthCheckMetadata, HealthCheckMethod, SuccessfulHealthCheckResponse } from "@domain/types";
+import { Router } from "express";
+import { ReasonPhrases, StatusCodes } from "http-status-codes";
+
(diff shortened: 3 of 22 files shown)


api/src/api/healthCheckRouter.ts
```diff
@@ -0,0 +1,61 @@
...
+    message: string;
+  };
+}
+
+interface UnsuccessfulHealthCheckResponse {
+  success: false;
+  error: {
+    timeout: boolean;
+    serverResponseCode?: number;
+    serverResponseBody?: string;
```
SirSamuelJoseph, line 23: [pebble] I like having `message` as a field in the Successful response because it lets individual handlers add context as needed- can we add that here as well?

api/src/api/healthCheckRouter.ts
```diff
@@ -0,0 +1,61 @@
+import { Router } from "express";
+
+/* TODO: Building out this feature:
+ *
+ * - Ensure continued access to the /health endpoint for the entire app.
```
SirSamuelJoseph, line 5: What more needs to be done to ensure this? I would think you've already finished this

api/src/api/healthCheckRouter.ts
seidior: Yep! I'll add it. I'd love a way to indicate to others that it should be a safe return message that contains no PII, but would love some input on an appropriate name here. Don't want to accidentally leak info from our status checks somehow.

api/src/api/healthCheckRouter.ts
seidior: Just testing, which I haven't done yet.

api/src/api/healthCheckRouter.ts
SirSamuelJoseph: Yeah that's a great point. I think `data` is the field name that would most likely tempt me to put userData in it. Maybe `responseContext`? 

Or if we want to be even stricter, we could have a type for possible responses (maybe `HealthCheckResponseMessage`). It would be slightly more annoying to add possible responses to the type, but you'd be looking at every previous example to see that its not user data. Plus, we'd likely have one response for 99% of healthy messages ("Alive"), so it would only be a few special cases where additional messages would be needed.

seidior: Error -37. Haha. Yeah we can definitely do that. I'll tack onto any of the ones you think of here; and I'm definitely open to changing "Alive", as that was the pre-existing message but I have no loyalty to it.

api/serverless.ts
```diff
@@ -111,7 +111,7 @@ const serverlessConfiguration: AWS = {
   ],
   provider: {
     name: "aws",
-    runtime: "nodejs16.x",
+    runtime: "nodejs18.x",
```
jphechter, line 114: We've updated other parts of our app to run Node `20.10.0`. It looks like there's an [AWS Lambda runtime for nodejs20.x](https://aws.amazon.com/blogs/compute/node-js-20-x-runtime-now-available-in-aws-lambda/). Is there a reason you picked 18?

I'm not opposed. Just curious.

seidior: I knew about this and tried it, but our version of `serverless` doesn't support it, and upgrading all our `serverless`-related dependencies was well out of scope of this ticket.

I'll add this to the release process doc though, and whoever gets a `serverless`-related Renovate should try making `nodejs20.x` work.

api/src/api/healthCheckRouter.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    await new Promise((resolve) => {
+      return setTimeout(resolve, 500);
+    });
+  });
+
+  describe("GET /", () => {
+    it("should return a successful response for the server's self check", async () => {
+      const response = await request(app).get("/");
+      expect(response.status).toBe(200);
+      expect(response.body.success).toBeTruthy();
```
somabadri, line 36: [pebble] I think we should be using `toBe(true)` instead of` toBeTruthy()` because we explicitly want to check for a `true` and `toBeTruthy()` evaluates to true when there is any value that is not `false`, `0`, `''`, `null`, `undefined`, and `NaN` and that is not what we are testing in this case?

api/src/api/healthCheckRouter.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      expect(response.body.data.message).toEqual("Alive");
+    });
+  });
+
+  describe("GET /mock", () => {
+    it("should return a successful response for a mocked health check", async () => {
+      const response = await request(app).get("/mock");
+      expect(response.status).toBe(200);
+      expect(response.body.success).toBeTruthy();
```
somabadri, line 45: [pebble] same comment as above

api/src/api/healthCheckRouter.ts
```diff
@@ -0,0 +1,46 @@
...
+  const router = Router();
+
+  router.get("/", (_req, res) => {
+    res.json({
+      timestamp: requestTimestamp(),
+      success: true,
+      data: {
+        message: "Alive",
+      },
+    } as SuccessfulHealthCheckResponse);
```
somabadri, line 29: question for my understanding, doesn't this always take the response and transform it into a successful health response even when it is not alive or does that error out in a different way?

api/src/domain/types.ts
```diff
@@ -118,6 +119,40 @@ export type ElevatorSafetyInspectionStatus = (
...
+
+export type UnsuccessfulHealthCheckResponse = {
+  timestamp: number;
+} & UnsuccessfulHealthCheckMetadata;
+
+export type HealthCheckResponse = SuccessfulHealthCheckResponse | UnsuccessfulHealthCheckResponse;
+
+export type HealthCheckMethod = () => Promise<HealthCheckMetadata>;
+
+// TODO: Make this more explicit?
```
somabadri, line 153: [pebble] is this something we are still planning on doing?

somabadri (commented):
Left a few comments and additionally why are there no test files for any of the HealthCheckClients? Isn't the lack of unit tests an anti-pattern?

api/src/client/dynamics/license-status/DynamicsLicenseHealthCheckClient.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { HealthCheckMetadata, HealthCheckMethod } from "@domain/types";
+import type { LogWriterType } from "@libs/logWriter";
+import axios, { type AxiosError } from "axios";
+
+type Config = {
+  accessTokenClient: AccessTokenClient;
+  orgUrl: string;
+};
+
+export const DynamicsLicenseHealthCheckClient = (
```
jphechter, line 11: I'm noticing that each API has a corresponding health check client. Is there a reason we opted for this approach instead of just utilizing the existing resources we built?

seidior: Dependency inversion, and because each has unique needs that aren't applicable to the others.

api/src/client/dynamics/fire-safety/DynamicsFireSafetyHealthCheckClient.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      error: {
+        message: "Bad Gateway",
+        serverResponseBody: "",
+        serverResponseCode: 404,
+        timeout: false,
+      },
+    });
+  });
+
+  it("returns a passing health check if axios request times out", async () => {
```
somabadri, line 47: [pebble] Is a timeout still considered to be a passing health check? The question applies for other test files as well. Let me know if I'm not understanding this correctly.

PR #6847, opened 2023-12-27
feat: [#186704006] Formation Review Page Heading Structure for Will P…

adidner:
…ractice law fixed

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186704006)

## Code author checklist


web/src/components/tasks/business-formation/review/ReviewBusinessSuffixAndStartDate.tsx
```diff
@@ -22,6 +23,15 @@ export const ReviewBusinessSuffixAndStartDate = (): ReactElement => {
     return `${name} ${suffix}`;
   };
 
+  const formattLawResponse = (willPracticeLaw: boolean | undefined): "Yes" | "No" | undefined => {
+    if (willPracticeLaw === true) {
+      return "Yes";
```
FarazA22, line 28: Pebble - this content as well as the one on line 30 should be editable by the content team/live in the CMS

web/src/components/tasks/business-formation/review/ReviewBusinessSuffixAndStartDate.tsx
```diff
@@ -22,6 +23,15 @@ export const ReviewBusinessSuffixAndStartDate = (): ReactElement => {
...
+    } else if (willPracticeLaw === false) {
```
FarazA22, line 29: pebble - you don't need the else if statement here; the reason for this is that an undefined value will result in a false boolean value, thus, you would never hit line 32. You can return "NO".

web/src/components/tasks/business-formation/review/section/ReviewLineItem.tsx
```diff
@@ -12,6 +12,7 @@ interface Props {
   dataTestId?: string;
   marginOverride?: string;
   formatter?: (value: string) => string;
+  noColenAfterLabel?: boolean;
```
FarazA22, line 15: dust - why is this being added, it feels weird to not have a Colon for 1 item on the review page. I noticed there are no designs in the ticket, I'd check with the design team about this.

web/src/components/tasks/business-formation/review/ReviewWillPracticeLaw.tsx
```diff
@@ -1,25 +0,0 @@
...
-interface Props {
-  willPracticeLaw: boolean | undefined;
-}
-
-export const ReviewWillPracticeLaw = (props: Props): ReactElement => {
-  const { Config } = useConfig();
-
-  const displayResponse = (): ReactElement => {
-    if (props.willPracticeLaw === undefined) {
-      return <ReviewNotEntered />;
```
FarazA22, line 15: Pebble - why did we remove the logic to render? `ReviewNotEntered`

web/src/components/tasks/business-formation/review/ReviewBusinessSuffixAndStartDate.tsx
FarazA22: NVM, I think we want to return a different value for when it's not selected by the user

web/src/components/tasks/business-formation/review/section/ReviewLineItem.tsx
adidner: We explicitly asked content about it and this is what they wanted.

web/src/components/tasks/business-formation/review/ReviewWillPracticeLaw.tsx
adidner: ReviewLineItem has logic inside it that renders a ReivewNotEntered when applicable.

web/src/components/tasks/business-formation/review/ReviewBusinessSuffixAndStartDate.tsx
adidner: Yep, thats a nice catch. Definielty copy pasted from the deleted ReviewWillPracticeLaw component but it should not have been that way in the first place.

web/src/components/tasks/business-formation/review/section/ReviewLineItem.tsx
FarazA22: I'd check with design. One of their principles is consistency. I'd be surprised if they agree with Content. 

Or at a Min, add design to review the ticket and write a comment to design letting them know the formatting is different.

web/src/components/tasks/business-formation/review/ReviewBusinessSuffixAndStartDate.tsx
adidner: So the way the ReviewLineItem component is written, when a value is passed if that value is undefined it will display the <ReviewNotEntered /> component, but if the content is defined it will display that content. 

So the text will come back as that text and the undefined will render the ReviewNotEntered associated text

web/src/components/tasks/business-formation/review/ReviewBusinessSuffixAndStartDate.tsx
```diff
@@ -64,6 +74,14 @@ export const ReviewBusinessSuffixAndStartDate = (): ReactElement => {
           dataTestId={"review-total-business-stock"}
         />
       )}
+      {isForeignCorporation(state.formationFormData.legalType) && (
+        <ReviewLineItem
+          label={Config.formation.fields.willPracticeLaw.label}
+          value={formattLawResponse(state.formationFormData.willPracticeLaw)}
+          dataTestId={"review-will-practice-law"}
+          noColenAfterLabel={true}
```
AnJuHyppolite, line 82: [pebble] We should change `noColen...`to `noColon...`

web/src/components/tasks/business-formation/review/section/ReviewLineItem.tsx
AnJuHyppolite: [pebble] in the meantime, while it's still part of the code, we should change `noColen...` to `noColon...`

AnJuHyppolite (commented):
Looks good, save for the typo and circling with UX on Faraz's comment.

PR #6848, opened 2023-12-27
fix: [#186682632] update certifications snackbar to display after formation

AnJuHyppolite:
## Description

**Undesired Behavior:**
When the user manually marks the Tax Registration task complete, the certification snackbar is displayed regardless of the user's formation status. If the user hasn't completed the Formation task, the For You section doesn't display certifications, but the snackbar still states certifications have been added.

**Current Behavior:**
When the user has not formed and marks the Tax Registration task complete, they are routed to the dashboard and a snackbar stating certifications have been added is displayed even though they not been added since the user hasn't yet formed.

**Expected Behavior:**
When the user has not formed and marks the Tax Registration task complete, they are routed to the dashboard

### Ticket

[#186682632](https://www.pivotaltracker.com/story/show/#186682632)

## Code author checklist

### DIFF
--- a/web/src/pages/dashboard.tsx
+++ b/web/src/pages/dashboard.tsx
@@ -89,7 +89,7 @@ const DashboardPage = (props: Props): ReactElement => {
   });
 
   const CertificationsAlert = useQueryControlledAlert({
-    queryKey: QUERIES.fromTaxRegistration,
+    queryKey: QUERIES.fromForming,
     pagePath: ROUTES.dashboard,
     headerText: Config.dashboardDefaults.certificationsSnackbarHeading,
     bodyText: Config.dashboardDefaults.certificationsSnackbarBody,
@@ -104,6 +104,7 @@ const DashboardPage = (props: Props): ReactElement => {
     bodyText: Config.dashboardDefaults.opportunitiesAndDeadlinesBody,
     variant: "success",
     dataTestId: "deadlines-opportunities-alert",
+    delayInMilliseconds: Config.dashboardDefaults.opportunitiesAndDeadlinesTimeDelay,
   });
 
   const FundingAlert = useQueryControlledAlert({
@@ -144,6 +145,7 @@ const DashboardPage = (props: Props): ReactElement => {
   });
 
   const operatingPhase = LookupOperatingPhaseById(business?.profileData.operatingPhase);
+  console.log({ operatingPhase });
 
   useMountEffectWhenDefined(() => {
     (async (): Promise<void> => {
--- a/web/test/pages/dashboard.test.tsx
+++ b/web/test/pages/dashboard.test.tsx
@@ -269,16 +269,16 @@ describe("dashboard page", () => {
     expect(screen.getByTestId("snackbar-alert-calendar")).toBeInTheDocument();
   });
 
-  it("renders certification snackbar when fromTaxRegistration query parameter is provided", () => {
-    useMockRouter({ isReady: true, query: { [QUERIES.fromTaxRegistration]: "true" } });
+  it("renders certifications snackbar when fromForming query parameter is provided", () => {
+    useMockRouter({ isReady: true, query: { [QUERIES.fromForming]: "true" } });
     renderDashboardPage({});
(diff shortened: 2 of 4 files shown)


web/src/pages/dashboard.tsx
```diff
@@ -89,7 +89,7 @@ const DashboardPage = (props: Props): ReactElement => {
   });
 
   const CertificationsAlert = useQueryControlledAlert({
-    queryKey: QUERIES.fromTaxRegistration,
```
FarazA22, line 92: `<TaskProgressCheckbox />` references `QUERIES.fromTaxRegistration`. I don't see it being used else where. Can we delete the key? Are changer required for `<TaskProgressCheckbox />`?

web/src/pages/dashboard.tsx
```diff
@@ -104,6 +104,7 @@ const DashboardPage = (props: Props): ReactElement => {
     bodyText: Config.dashboardDefaults.opportunitiesAndDeadlinesBody,
     variant: "success",
     dataTestId: "deadlines-opportunities-alert",
+    delayInMilliseconds: 6000,
```
FarazA22, line 107: dust - why are we adding a delay?

web/src/pages/dashboard.tsx
```diff
@@ -144,6 +145,7 @@ const DashboardPage = (props: Props): ReactElement => {
   });
 
   const operatingPhase = LookupOperatingPhaseById(business?.profileData.operatingPhase);
+  console.log({ operatingPhase });
```
FarazA22, line 148: remove

web/src/pages/dashboard.tsx
```diff
@@ -89,7 +89,7 @@ const DashboardPage = (props: Props): ReactElement => {
   });
 
   const CertificationsAlert = useQueryControlledAlert({
-    queryKey: QUERIES.fromTaxRegistration,
+    queryKey: QUERIES.fromForming,
```
FarazA22, line 92: pebble - this change creates an entirely different issue. A user no longer sees the snack bar when they complete the formation task. You could view this by navigating to the formation task, and then clicking on a checkbox at the top to manually change the task to complete it. When the user is routed back the snackbar does not render

FarazA22: Separately, I'd expect the snack bar to render when a user forms via the API, and they successfully processed the payment. I think when they return, they are also routed to the dashboard. I might be a mistaken here, I think it's worth looking into. If they are routed to the dashboard, then I expect to see the Snack bar here as well

web/test/pages/dashboard.test.tsx
```diff
@@ -269,16 +269,16 @@ describe("dashboard page", () => {
...
+    useMockRouter({ isReady: true, query: { [QUERIES.fromForming]: "true" } });
     renderDashboardPage({});
     expect(screen.getByTestId("certification-alert")).toBeInTheDocument();
   });
 
   it("renders deadlines and opportunities snackbar when fromForming query parameter is provided", () => {
     useMockRouter({ isReady: true, query: { [QUERIES.fromForming]: "true" } });
-
     renderDashboardPage({});
+    act(() => jest.runAllTimers());
```
FarazA22, line 281: you'll need to update the test based on how you solve the bug

FarazA22, line 281: you'll need to update the test based on how you solve the bug

web/src/pages/dashboard.tsx
seidior, line 107: [sand] I like the custom delay, but I wonder if this is a sensible value to put into config so that Content can change it later if need be.

seidior (approved):
No comments besides Faraz's. Thank you for taking this on!

web/src/pages/dashboard.tsx
AnJuHyppolite: re: snackbar issue - brought to product's attention with an ask to resolve in another story
re: API - User is not re-routed to dashboard once formation is complete/successful.

PR #6851, opened 2023-12-27
feat: [#185681159] add business name field and note

FarazA22:
to remote seller remote worker in profile

## Description

Add Business Name field to Profile for Dakota Remote Sellers and Dakota Remote Workers  

### Ticket

https://www.pivotaltracker.com/story/show/185681159

## Code author checklist

### DIFF
--- a/web/src/lib/domain-logic/getNavBarBusinessTitle.ts
+++ b/web/src/lib/domain-logic/getNavBarBusinessTitle.ts
@@ -1,11 +1,7 @@
 import { getMergedConfig } from "@/contexts/configContext";
 import { IsAuthenticated } from "@/lib/auth/AuthContext";
-import {
-  Business,
-  determineForeignBusinessType,
-  LookupIndustryById,
-  LookupLegalStructureById,
-} from "@businessnjgovnavigator/shared";
+import { isRemoteWorkerOrSeller } from "@/lib/domain-logic/isRemoteWorkerOrSeller";
+import { Business, LookupIndustryById, LookupLegalStructureById } from "@businessnjgovnavigator/shared";
 
 export const getNavBarBusinessTitle = (
   business: Business | undefined,
@@ -24,15 +20,7 @@ export const getNavBarBusinessTitle = (
     businessPersona,
     needsNexusDbaName,
     nexusDbaName,
-    foreignBusinessTypeIds,
   } = business.profileData;
-
-  const isRemoteWorkerOrSeller = (): boolean => {
-    return (
-      determineForeignBusinessType(foreignBusinessTypeIds) === "REMOTE_SELLER" ||
-      determineForeignBusinessType(foreignBusinessTypeIds) === "REMOTE_WORKER"
-    );
-  };
 
   const determineName = (): string => {
     if (needsNexusDbaName) {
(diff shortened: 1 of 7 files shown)


web/src/lib/domain-logic/isRemoteWorkerOrSeller.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  it("returns true when revenueInNJ", () => {
+    expect(isRemoteWorkerOrSeller(["revenueInNJ"])).toBe(true);
+  });
+
+  it("returns true when transactionsInNJ", () => {
+    expect(isRemoteWorkerOrSeller(["transactionsInNJ"])).toBe(true);
+  });
+
+  it("returns true when empty", () => {
```
adidner, line 16: [pebble] Description should be returns false (not true)

web/src/lib/domain-logic/isRemoteWorkerOrSeller.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(isRemoteWorkerOrSeller([])).toBe(false);
+  });
+
+  it("returns true when nexus", () => {
```
adidner, line 20: [pebble] Description should be returns false (not true)

web/src/lib/domain-logic/getNavBarBusinessTitle.ts
```diff
@@ -27,13 +23,6 @@ export const getNavBarBusinessTitle = (
     foreignBusinessTypeIds,
   } = business.profileData;
 
-  const isRemoteWorkerOrSeller = (): boolean => {
```
adidner, line 30: [sand] So I'm totally down for making this domain logic, but we have like six other places where we locally declare an isRemoteWorkerOrSeller function which should also be consolidated in this PR or as a follow up

web/src/lib/domain-logic/isRemoteWorkerOrSeller.ts
```diff
@@ -0,0 +1,8 @@
+import { ForeignBusinessTypeId } from "@businessnjgovnavigator/shared/";
+import { determineForeignBusinessType } from "@businessnjgovnavigator/shared/domain-logic/determineForeignBusinessType";
+
+export const isRemoteWorkerOrSeller = (foreignBusinessTypeIds: ForeignBusinessTypeId[]): boolean => {
```
adidner, line 4: [pebble] I think this would be better if you passed a business instead of an ID, that way you can check for that it's foreign inside rather than outside.

adidner: [dust] I also think it might be good to check for undefined state as a guard clause. (And then test that as well) just for robustness sake

adidner (commented):
Seems good. A few minor description of test changes and some minor ideas for improving robustness and further consideration of domain logic

seidior (approved):
Aside from the items Aaron raised, everything else looks good code-wise.

web/src/lib/domain-logic/getNavBarBusinessTitle.ts
FarazA22: https://www.pivotaltracker.com/story/show/186749997

PR #6869, opened 2024-01-03
feat: [#186569119] Test no certification alert for foreign remote sel…

adidner:
…ler/worker

## Description

Added missing test for removal of certification snackbar for Dakota Remote Seller / Worker

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186569119)

## Code author checklist


cbrunefearless (approved):
lgtm

web/src/components/TaskProgressCheckbox.test.tsx
```diff
@@ -234,6 +234,48 @@ describe("<TaskProgressCheckbox />", () => {
...
+      renderTaskCheckbox(taxTaskId, generateBusiness({ profileData: foreignRemoteWorker }));
+      await selectCompleted();
+      expect(currentBusiness().taskProgress[taxTaskId]).toEqual("COMPLETED");
+      expect(mockPush).toHaveBeenCalledWith({
+        pathname: ROUTES.dashboard,
+        query: { fromFormBusinessEntity: "false", fromTaxRegistration: "false" },
+      });
+    });
+
+    it("redirects with fromTaxRegistration as true when checking TaskProgressCheckbox as dakota nexus", async () => {
```
AnJuHyppolite, line 265: [pebble] This test doesn't cover the logic we implemented with the new feature. We can delete this.

AnJuHyppolite (commented):
Save for my one comment, looks good, Aaron. Thank you for revisiting and adding the tests.

web/src/components/TaskProgressCheckbox.test.tsx
adidner: I think it would be good to have anyways. 

Before this we didn't have any test covering the "default case" of showing the certification snackbar from Tax Redirect. 

It's also great to have a positive case here to make sure its possible (if the alert was for some reason always disabled in our code we would want our tests to cover that)

adidner: We could try and make this more generic? Perhaps as the snackbar is suppose to show for all other cases coming from the tax registration page. Or more robust by doing all the various persona but I didn't think it was super necessary here.

PR #6873, opened 2024-01-05
chore: [#186303948] refactor drop shadows scss to use mixins

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/186303948

### Approach
Refactored drop shadows to use mixin feature so box shadow can be included in existing classes.

## Code author checklist

### DIFF
--- a/web/src/styles/components/buttons.scss
+++ b/web/src/styles/components/buttons.scss
@@ -128,7 +128,12 @@
 }
 
 .btn-white {
+  &:hover {
+    @include addDropShadow($drop-shadow-md);
+  }
+
   background-color: $white !important;
+
   &:active {
     background-color: $base-lighter !important;
   }
@@ -194,7 +199,9 @@
   border: 0.125rem solid $secondary-lighter;
   border-radius: 0.625rem;
   cursor: pointer;
-  @extend .hover\:drop-shadow-md;
+  &:hover {
+    @include addDropShadow($drop-shadow-md);
+  }
   &:active {
     background-color: $secondary-lighter;
   }
--- a/web/src/components/LandingPageActionTile.tsx
+++ b/web/src/components/LandingPageActionTile.tsx
@@ -33,7 +33,9 @@ export const LandingPageActionTile = (props: Props): ReactElement => {
   return (
     <div className={props.className} ref={props.reference}>
       <button
-        className={`${props.isPrimary ? "landing-page-tile" : "landing-page-secondary-tile"} margin-1`}
+        className={`${
+          props.isPrimary ? "landing-page-tile" : "landing-page-secondary-tile"
+        } margin-1 margin-bottom-3`}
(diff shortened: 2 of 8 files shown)


web/src/styles/_sass-files/variables.scss
```diff
@@ -5,6 +5,13 @@ $sm: 40em; //tablet + up (640px)
 $md: 50em; //big tablet (800px)
```
adidner, line 5: [dust] curious why the underscore for this folder name

web/src/styles/_sass-files/variables.scss
```diff
@@ -5,6 +5,13 @@ $sm: 40em; //tablet + up (640px)
...
 $lg: 64em; //desktop and up (1024px)
 
+$drop-shadow-xs: 0px, 4px, 8px, 0.1;
+$drop-shadow-xs-darker: 4px, 4px, 8px, 0.25;
```
adidner, line 9: [dust] So it kinda seems like Alph here is its own attribute,

so you could do have a second param to the mixin and have like 0.1 and 0.25 as dark and darker respectively as passed as such 

(I'm not saying that this is better or that anything has to change, just kinda thinking out loud)

web/src/styles/components/buttons.scss
```diff
@@ -128,7 +128,12 @@
 }
 
 .btn-white {
+  &:hover {
+    @include addDropShadow($drop-shadow-md);
+  }
```
adidner, line 133: [dust] interesting that you chose to include this code here rather than
something like `white: "usa-button text-info-darkest btn-white drop-shadow-md-on-hover",` (I think that this would also work right?)

Did you have a specific reason for favoring this approach?

adidner (approved):
Looks good to me. Curious about a few different approach things and if there was any thought behind those decisions because there was more than 1 way to do some of these and I personally don't have strong opinions about which is better but I'm curious if you do.

seidior (approved):
Feedback given async.

web/src/styles/_sass-files/variables.scss
FarazA22: I'm doing to remove it.

web/src/styles/_sass-files/variables.scss
FarazA22: This variable is a list with 4 items. I think it would be more straightforward to define all the attributes within the list than passing in a value like dark/ darker and having the value or those variables defined in the function. Side note - I'm also planning to speak to design about using USDWS shadows so the darker might go away.

web/src/styles/components/buttons.scss
FarazA22: I thought that by incorporating it into the btn-white or landing-page-tile classes, we wouldn't have to be concerned about including the drop-shadow class when using them in other sections of the application.

PR #6900, opened 2024-01-09
feat: [#186638569] Quick Action Icon Options Added

adidner:
## Description

We wanted content to be able to select icons for quick actions (more different icons).

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186638569)

## Code author checklist


web/public/mgmt/config.yml
```diff
@@ -3027,7 +3035,15 @@ collections:
...
+        options:
+          - { label: "Update", value: "update-purple.svg" }
+          - { label: "Government", value: "govt-purple.svg" }
+          - { label: "Add", value: "add-purple.svg" }
+          - { label: "Change", value: "change-purple.svg" }
+          - { label: "Sell", value: "sell-purple.svg" }
+          - { label: "Close", value: "close-purple.svg" }
+          - { label: "Edit", value: "edit-purple.svg" }
+          - { label: "File", value: "file-purple.svg" }
       - { label: "External Link", name: "externalRoute", widget: "nospace", required: true }
```
adidner, line 3047: image
This was how I was thinking about things initially, I think this is fine and not wrong (espeicially if it only happens twice) but I think a slightly more centeralized and robust solution is probably a bit better.

content/src/mappings/quickActionIcons.json
```diff
@@ -0,0 +1,12 @@
+{
+  "arrayOfQuickActionIcons": [
+    { "name": "Update", "id": "update-purple.svg" },
+    { "name": "Government", "id": "govt-purple.svg" },
+    { "name": "Add", "id": "add-purple.svg" },
+    { "name": "Change", "id": "change-purple.svg" },
+    { "name": "Sell", "id": "sell-purple.svg" },
+    { "name": "Close", "id": "close-purple.svg" },
+    { "name": "Edit", "id": "edit-purple.svg" },
+    { "name": "File", "id": "file-purple.svg" }
```
jphechter, line 10: [sand] should `id` be `filename`?

web/stories/Molecules/QuickActions.stories.tsx
```diff
@@ -16,8 +16,44 @@ const meta: Meta<typeof QuickActionTaskTile> = {
 export default meta;
 type Story = StoryObj<typeof QuickActionTaskTile>;
 
-export const QuickAction: Story = {
+export const QuickActionGovernment: Story = {
```
FarazA22, line 19: dust - why are we calling this `QuickActionGovernment`?

web/public/mgmt/config.yml
```diff
@@ -3060,6 +3066,22 @@ collections:
...
+        fields:
+          - label: Quick Action Icons
+            name: arrayOfQuickActionIcons
+            widget: list
+            allow_add: false
+            summary: "{{name}}"
+            fields:
+              - label: Name
+                name: name
+                widget: string
```
FarazA22, line 3081: dust - I don't think this should be editable by Content, if a change is made, the icons won't render.

web/public/mgmt/config.yml
```diff
@@ -3060,6 +3066,22 @@ collections:
       preview: false
     delete: false
     files:
+      - label: "Quick Action Icons"
+        name: "arrayOfQuickActionIcons"
+        file: "content/src/mappings/quickActionIcons.json"
+        fields:
+          - label: Quick Action Icons
+            name: arrayOfQuickActionIcons
+            widget: list
```
FarazA22, line 3075: dust -  maybe make this widget hidden and update the label so it informs content this collection is not editable

web/public/mgmt/config.yml
FarazA22: dust - It might be a good idea to create an uneditable mapping collection, the industry mapping ticket I'm working on also prevents content from editing the field

web/public/mgmt/config.yml
```diff
@@ -2997,10 +2997,13 @@ collections:
       - { label: "Name/Title", name: "name", widget: "string" }
       - label: "Icon"
         name: icon
-        widget: "select"
+        widget: "relation"
         multiple: false
```
FarazA22, line 3001: dust - I think the default is false, so you could remove `multiple: false`

web/public/mgmt/config.yml
FarazA22: nvm, I see there is a name and id/filename field. If a change is made, I think the files with the previous value would not render. Is content responsible for making updates. I kind of lean on making it hidden

adidner: I think it would be fine actually, the name can be edited so long as the ID stays the same, as the files only have ID's and that's what is being used for reference I believe. It just allows content to change the name that corresponds to the value of the file. 

Like the ID is the only thing stored in the files and the name is just for display in the CMS so the name can be changed without ruining anything I believe.

web/public/mgmt/config.yml
adidner: Good callout

web/stories/Molecules/QuickActions.stories.tsx
```diff
@@ -5,7 +5,7 @@ import { Meta, StoryObj } from "@storybook/react";
 const meta: Meta<typeof QuickActionTaskTile> = {
   title: "Molecules/QuickAction",
   component: QuickActionTaskTile,
-  decorators: [(Story) => <div className="width-mobile">{Story()}</div>],
+  decorators: [(Story) => <div>{Story()}</div>],
```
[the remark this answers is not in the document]
adidner: Makes sense, i will add it back in

content/src/mappings/quickActionIcons.json
adidner: Yeah, thats a good point

PR #7083, opened 2024-01-26
chore: [#186749997] refactored isRemoteSellerRemoteWorker to be used …

adidner:
…in more places

we wanted to further consolidate into the function isRemoteWorkerOrSeller

### Ticket

## Code author checklist


shared/src/domain-logic/businessPersonaHelpers.test.ts
```diff
@@ -36,7 +36,7 @@ const generateForeignBusinessWithForeignBusinessTypeIds = (
 describe("businessPersonaHelpers", () => {
   describe("isOwningBusiness", () => {
     it("returns false when business is undefined", () => {
-      expect(isOwningBusiness(undefined)).toBe(false);
+      expect(isOwningBusiness()).toBe(false);
```
FarazA22, line 39: I didn't think about passing 0 arguments when I wrote these tests... It's cleaner. 🥇

somabadri (approved):
lgtm, good job Aaron!

web/src/pages/onboarding.tsx
```diff
@@ -257,10 +258,10 @@ const OnboardingPage = (props: Props): ReactElement => {
       analytics.event.onboarding_last_step.submit.finish_onboarding();
     }
 
-    const isRemoteSellerWorker =
-      newProfileData.businessPersona === "FOREIGN" &&
-      (determineForeignBusinessType(newProfileData.foreignBusinessTypeIds) === "REMOTE_SELLER" ||
-        determineForeignBusinessType(newProfileData.foreignBusinessTypeIds) === "REMOTE_WORKER");
+    const isRemoteSellerWorker = isRemoteWorkerOrSellerBusiness({
+      ...updateQueue.currentBusiness(),
```
FarazA22, line 262: This is a dope solution. You could also use `determineForeignBusinessType`. However, I have some thoughts on the function. See below.

In the ticket, I made a note to consider where the parameter for the `businessPersonalHelpers function` should be the ProfileData object. I noted this because we are using `determineForeignBusinessType` in the app and I was wondering if it's possible to kill that function and use one of the ones in the `businessPersonalHelpers` file. In areas like the profile data text field, we only have access to profile data.

What do you think? If we keep it, I think `determineForeignBusinessType` function should also be move to the `businessPersonalHelpers` file. It would make it easier to find.

shared/src/domain-logic/businessPersonaHelpers.test.ts
```diff
@@ -14,6 +8,12 @@ import {
   generateBusiness,
   generateProfileData,
 } from "@businessnjgovnavigator/shared/";
+import {
+  isNexusBusiness,
+  isOwningBusiness,
+  isRemoteWorkerOrSellerBusiness,
+  isStartingBusiness,
+} from "./businessPersonaHelpers";
```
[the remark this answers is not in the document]
seidior: (fwiw I just fixed this for VSCode [here](https://github.com/newjersey/navigator.business.nj.gov/commit/be367cc5ca4747862fd8794d0129df70937866a8). if there's an equivalent for other editors I'm totally open to adding that configuration.)

PR #7090, opened 2024-01-29
feat: [#186444137] [#186444230] navigation bar title updates

somabadri:
## Description
Updated the navbar to include the new logo that routes to business.nj.gov or dashboard based on which portion of the logo the user clicks. 

### Ticket

[#186444137](https://www.pivotaltracker.com/story/show/186444137)
[#186444230](https://www.pivotaltracker.com/story/show/186444230)

## Code author checklist


jphechter:
Home and Welcome still seem to be using the Navigator logo
Screenshot 2024-01-29 at 4 45 45 PM

web/src/components/navbar/NavigatorLogo.tsx
```diff
@@ -10,14 +9,16 @@ interface Props {
 export const NavigatorLogo = (props: Props): ReactElement => {
```
jphechter, line 9: [sand] In an effort to rename, I think we should change the name of this component as well. Maybe `MyAccountLogo` or even just `Logo`? I dunno. Follow your heart.

AnJuHyppolite:
> Home and Welcome still seem to be using the Navigator logo Screenshot 2024-01-29 at 4 45 45 PM

Where are you looking? Which URL?

web/src/components/navbar/NavBarDesktop.tsx
```diff
@@ -73,7 +75,16 @@ export const NavBarDesktop = (): ReactElement => {
     <div className="position-sticky top-0 z-500 bg-white">
       <nav aria-label="Primary" className="grid-container-widescreen desktop:padding-x-7">
         <div className="display-flex flex-row flex-justify flex-align-center height-8">
-          <NavigatorLogo />
+          <div className="display-flex flex-row flex-align-center">
```
jphechter, line 78: [sand] Kind of nit-picky - the focus box doesn't actually surround the image on click. It'd be nice to fix this.
Screenshot 2024-01-29 at 4 59 29 PM

jphechter: Not specifically part of the AC, but it might be nice to have this use the blue color if it's not a PITA.

web/src/components/navbar/NavBarVerticalLine.tsx
```diff
@@ -0,0 +1,14 @@
...
+  <div
+    className="height-4"
+    style={{
+      width: "2px",
+      background:
+        "linear-gradient(0, rgba(46,65,83,0) 0%, rgba(46,65,83,0) 25%, rgba(46,65,83,1) 25%, rgba(46,65,83,1) 75%, rgba(46,65,83,0) 75%, rgba(46,65,83,0) 100%)",
+    }}
+  >
+    &nbsp;
+  </div>
```
jphechter, line 13: Is there a reason we didn't just add this to a pseudo class? It feels like overkill to create a component for this.

web/src/components/navbar/NavBarDesktop.tsx
```diff
@@ -73,7 +75,16 @@ export const NavBarDesktop = (): ReactElement => {
...
+            <NavigatorLogo />
```
AnJuHyppolite, line 79: I agree with JP's comment. Since we're rebranding, let's just do away with the `Navigator` prefix. I think `MyAccountLogo` is suitable.

web/public/mgmt/config.yml
```diff
@@ -3671,6 +3671,8 @@ collections:
               - { label: Needs Account Modal Button, name: needsAccountModalButtonText, widget: string }
               - { label: Needs Account Modal Sub Button text, name: needsAccountModalSubText, widget: string }
               - { label: Add Business Button, name: addBusinessButton, widget: string }
+              - { label: My Account Text, name: navBarMyAccountText, widget: string }
+              - { label: Business.nj.gov Link, name: navBarBusinessNJGovHomePageLink, widget: string }
```
AnJuHyppolite, line 3675: [dust] I don't think we need to specify the home page. This could help in places where the text is a bit verbose. Consider changing the `name` to `navBarBusinessNJGovLink`.

web/src/components/navbar/NavBarDashboardLink.tsx
```diff
@@ -0,0 +1,28 @@
+import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
+import { ROUTES } from "@/lib/domain-logic/routes";
+import { useRouter } from "next/router";
+import { ReactNode } from "react";
+
+interface Props {
+  businessName: string;
+  stringToDisplay: string;
```
jphechter, line 8: [sand]
```suggestion
  linkText: string;
```

web/src/components/navbar/NavBarDashboardLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+        }}
+        ariaLabel={`My Account for ${props.businessName} Home`}
+      >
+        <span className={`text-primary text-semibold font-body-sm ${props.className ?? ""}`}>
+          {props.stringToDisplay}
+        </span>
+      </UnStyledButton>
+    </div>
+  );
+};
```
jphechter, line 28: Looking at this after it's been implemented, it's kindof goofy that we have a link to the dashboard on the dashboard. Might have made more sense for us to only make this a link on tasks/details/profile.

web/src/components/navbar/NavBarDashboardLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const NavBarDashboardLink = (props: Props): ReactNode => {
+  const router = useRouter();
+  return (
+    <div>
+      <UnStyledButton
+        onClick={(): void => {
+          router.push(ROUTES.dashboard);
+        }}
+        ariaLabel={`My Account for ${props.businessName} Home`}
```
jphechter, line 20: [sand] `businessName` is allowed to be undefined. It will be until the user supplies a business name, which may be a while. During that time the ariaLabel will be `My Account for  Home`. What if we change this to say either `My Account Home` or `My Account for ABC Business Home`?

web/src/components/navbar/NavigatorLogo.tsx
```diff
@@ -10,14 +9,16 @@ interface Props {
...
   const { Config } = useConfig();
   return (
-    <Link href={ROUTES.dashboard} passHref>
-      <a href={ROUTES.dashboard}>
-        <img
-          className={props.useSmall ? "height-3" : "height-4"}
-          src="/img/Navigator-logo@2x.png"
```
jphechter, line 17: [sand] Delete this file after we remove it from the home and welcome pages.

jphechter:
[This ticket](https://www.pivotaltracker.com/n/projects/2573455/stories/186444137) has 2 analytics events listed. I know Alex mentioned at one point that he might try to add events through the GA interface. Not sure if he's ready for that yet. Do we need to include those events in this work?

jphechter:
https://github.com/newjersey/navigator.business.nj.gov/pull/7090#issuecomment-1915647531

@AnJuHyppolite I pulled the code and ran locally

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -102,6 +102,7 @@ describe("<NavBar />", () => {
   beforeEach(() => {
     jest.resetAllMocks();
     useMockRouter({});
+    useMockUserData({});
```
adidner, line 105: dust - I'm confused about why this is needed? Given how no other modifications were made and I assume things were passing here before hand?

web/src/components/navbar/NavigatorLogo.tsx
```diff
@@ -10,14 +9,16 @@ interface Props {
...
-          alt={Config.pagesMetadata.titlePrefix}
-        />
-      </a>
+    <Link href={Config.navigationDefaults.navBarBusinessNJGovHomePageLink} passHref>
+      <div className={"height-4"}>
+        <a href={Config.navigationDefaults.navBarBusinessNJGovHomePageLink}>
+          <img
+            className={props.useSmall ? "height-3" : "height-4"}
+            src="/img/business.NJ.gov-logo.svg"
+            aria-label={"Business.nj.gov Home"}
```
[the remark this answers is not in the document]
AnJuHyppolite: My understanding is that `aria-label` is used for interactive elements. Since we're making the logo clickable to route to the dashboard, we should have it on the image.

web/src/components/navbar/NavBarDashboardLink.tsx
adidner: I'm not sure I agree, I feel like its a well established pattern to just always hame a home icon in the top right regardless of whether or not you're on the home page.

AnJuHyppolite:
> [#7090 (comment)](https://github.com/newjersey/navigator.business.nj.gov/pull/7090#issuecomment-1915647531)
> 
> @AnJuHyppolite I pulled the code and ran locally

I did the same. It may be a timing thing, b/c it was "fixed" when I looked at it.

web/src/components/navbar/NavBarVerticalLine.tsx
somabadri: @seidior had created this component and I just kept it. When you say psuedo class do you mean just make it styling within a the navbar component?

web/src/components/navbar/NavigatorLogo.tsx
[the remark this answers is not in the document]
somabadri: I can make it part of the CMS

web/src/components/navbar/NavBarDashboardLink.tsx
somabadri: This wouldn't be the home icon, but rather the my account or business name portion which takes you to the dashboard.

web/src/components/navbar/NavBar.test.tsx
somabadri: It's because the sub component now uses userData to get the businessName

web/src/components/navbar/NavBarDashboardLink.tsx
somabadri: I made it already that you can't click it to navigate to the dashboard while you are in onboarding so we could make it so it doesn't work on the dashboard page with some more logic if we wanted?

web/src/components/navbar/NavBarVerticalLine.tsx
jphechter: Yeah, like just add a `::after` to the CSS rule for the logo image or something.

seidior: Nah it's not valid for ::after since it needs to be a conditional render, if I recall correctly. And I'm totally okay having this be a component as it improves code readability.

OTOH, there's probably more to do here to clean this up, like moving more of this into USWDS that can be moved in.

jphechter: I don't think I understand why `::after` wouldn't work. We could still manage this with CSS rules, but even still adding this to a rule applying to the logo component would suffice. This is only relevant any time the logo is on the page. The conditional rendering for the logo would be enough.

I'm not super torn, I just think in this instance there's a lot of component boiler plate being added here for what is essentially a CSS change.

web/src/components/navbar/NavBarDashboardLink.tsx
jphechter: [sand] @somabadri If I'm in the "add another business" onboarding flow, and I click the "My Account" link text, I get redirected to the onboarding flow and no longer have the option to return to the dashboard. Can we make sure that essentially has the same behavior as the "Return to ${business_name}" button in that use case?
Screenshot 2024-01-31 at 3 27 01 PM

web/src/components/navbar/NavBar.test.tsx
```diff
@@ -102,6 +102,7 @@ describe("<NavBar />", () => {
...
   });
```
FarazA22, line 106: @somabadri The `NavBar` component uses dependency injection/ passes business as a Prop, still not clear as to why useMockUserData({}) is needed. Shouldn't we be using the generator fn to create the business and pass it in as an argument?

web/src/components/navbar/NavBarDashboardLink.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  it("fires the dashboard analytics event when link is clicked by the user", async () => {
+    useMockRouter({});
+    renderComponent({});
+    fireEvent.click(screen.getByText("text-to-display"));
+    expect(mockAnalytics.event.my_account.click.my_account).toHaveBeenCalled();
+  });
+
+  it("returns to dashboard and updates userData when link is clicked by the user and there is a previous business prop", async () => {
```
FarazA22, line 79: dust - should the word `and` be replaced with `when`?

web/src/components/navbar/NavBarDashboardLink.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const renderComponent = ({
+  userData,
+  previousBusiness,
+}: {
+  userData?: UserData | undefined;
+  previousBusiness?: Business | undefined;
+}): void => {
+  render(
+    <WithStatefulUserData
```
FarazA22, line 47: dust - why do we need `WithStatefulUserData`?

web/src/components/navbar/NavBar.test.tsx
jphechter: I think Soma answers this in the "resolved" comment.

content/src/fieldConfig/config.json
```diff
@@ -68,7 +68,11 @@
     "navBarUnnamedBusinessText": "Unnamed",
     "taskPageNavBarHeading": "Business Roadmap",
     "needsAccountSnackbarTitleExistingAccount": "Link Your myNewJersey Account to Save Your Progress",
-    "needsAccountModalBody": "To use this feature, you need to complete your registration with the State of New Jersey's secure login system, MyNewJersey. This will ensure your privacy, and allow you to come back any time."
+    "needsAccountModalBody": "To use this feature, you need to complete your registration with the State of New Jersey's secure login system, MyNewJersey. This will ensure your privacy, and allow you to come back any time.",
+    "navBarMyAccountText": "My Account",
```
FarazA22, line 72: dust - we talked about not adding to this file, instead we discussed carving out the areas of this confid into their own collection / with another collection.

web/src/components/navbar/NavBarDashboardLink.test.tsx
somabadri: this is needed for the third test to check to see if we properly update the userData in order to return to the previous business

content/src/fieldConfig/config.json
somabadri: I can do that in a follow up ticket!

web/src/components/navbar/NavBarDashboardLink.test.tsx
somabadri: I initially had it as `when` but changed it to and because it read as `when link is clicked by the user when there is a previous business prop` and it didn't sound well written.

web/src/components/navbar/NavBar.tsx
```diff
@@ -13,6 +14,7 @@ type Props = {
   showSidebar?: boolean;
   hideMiniRoadmap?: boolean;
   logoOnly?: "NAVIGATOR_LOGO" | "NAVIGATOR_MYNJ_LOGO" | undefined;
+  previousBusiness?: Business | undefined;
```
FarazA22, line 17: dust - i get why `useMockUserData({});` was added to `NavBar.test.tsx` - that being said, it feels weird passing in the the `business` object while also using the userData hook to grab the `business` within individual components.

I think it might be better to have a helper function that takes a business as an argument and returns the `id` and this prop takes that id. 
`<Navbar prevBusId={getBusinessId(business)} />`

web/src/components/navbar/NavBarDashboardLink.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useUserData } from "@/lib/data-hooks/useUserData";
+import { removeBusiness } from "@/lib/domain-logic/removeBusiness";
+import { ROUTES } from "@/lib/domain-logic/routes";
+import analytics from "@/lib/utils/analytics";
+import { Business } from "@businessnjgovnavigator/shared/userData";
+import { useRouter } from "next/router";
+import { ReactNode } from "react";
+
+interface Props {
+  businessName: string;
```
FarazA22, line 11: pebble - do we need this as a prop, can we pull it from the `useUserData` hook? Same question with linkText? It's feel a bit confustion why we are using dependency injection here

web/src/components/navbar/NavBarDashboardLink.tsx
FarazA22: Did design confirm the link should be disabled. I agree with Aaron, there should be a consistent pattern, Conditionally disable/enable a link is a jarring experience.

web/src/components/navbar/NavBar.tsx
somabadri: I think this is a valid point. I was just following the pattern established in `ReturnToPreviousBusinessBar.tsx` but we only require an id in this case so I can make that change.

web/src/components/navbar/NavBarMobile.tsx
```diff
@@ -75,13 +81,26 @@ export const NavBarMobile = (props: Props): ReactElement => {
...
-            <div className="text-bold font-body-sm truncate-long-business-names_NavBarMobile">
-              {getNavBarBusinessTitle(business, state.isAuthenticated)}
+          <div className={"display-flex flex-align-center"}>
+            <div className={props.showSidebar ? "width-15" : ""}>
+              <BusinessNjGovLogo />
             </div>
-          ) : (
-            <NavigatorLogo />
-          )}
+            <div className="margin-x-105">
```
FarazA22, line 88: pebble - keeping things dry, this could be a component

somabadri: which portion are you referring to?

web/src/styles/base/typography.scss
```diff
@@ -1,3 +1,14 @@
+@font-face {
```
FarazA22, line 1: dust - we have some font limitations, what exactly are we introducing here? Are we using a different font weigh, if so the design system needs to be updated and the typography section of storybook. And this new font style needs to be globally availabile via a CSS name, note sure if it is

somabadri: @seidior implemented this portion of it, can you shed some light on what was your intention here?

web/src/components/navbar/NavBarDashboardLink.tsx
somabadri: yes we can, I initially wasn't using the `useUserData` hook but I had to in order for the return to previous business to work, but I didn't catch this refactor

web/src/components/navbar/NavBarMobile.tsx
FarazA22: Yeah, aren't we using it elsewhere?

somabadri: yeah, we do. I was thinking of a way to make a component out of it, but each case is slightly different enough where  the final component wouldn't be readable

web/src/components/navbar/NavBarDashboardLink.tsx
somabadri: @FarazA22 I asked about this in slack and the consensus was that it should always navigate to dashboard

web/src/components/navbar/BusinessNjGovLogo.tsx
```diff
@@ -0,0 +1,31 @@
+import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import analytics from "@/lib/utils/analytics";
+import { useRouter } from "next/router";
+import { ReactElement } from "react";
+interface Props {
+  useSmall?: boolean;
+}
+
+export const BusinessNjGovLogo = (props: Props): ReactElement => {
```
adidner, line 10: pebble/dust - So is this only for the nav bar? It kinda seems like it based on how the onClicks are handled. Maybe should be named as such

adidner (approved):
Left a tiny comment, but otherwise looks good!

web/src/styles/base/typography.scss
seidior: You're correct @FarazA22, this introduces the semibold weight, which is missing due to NJWDS's implementation of USWDS, which doesn't include semibold by default.

web/src/components/navbar/BusinessNjGovLogo.tsx
somabadri: named it `NavbarBusinessNjGovLogo.tsx` to match the style of the other navbar components

PR #7157, opened 2024-02-06
feat: [##184764052] change business name error from alert to inline

AnJuHyppolite:
## Description

The business name error should no longer render as an `<Alert>`.

1. Remove the `<Alert>` that renders on the top in the Business Name Step.
2. Change the `<Alert>` that renders under the business name field in the Business Name Step to an inline error. 

### Ticket

[#184764052](https://www.pivotaltracker.com/story/show/184764052)

### Steps to Test

1. Leave the business name empty and move on to other steps in the Business Paginator and the business name no longer renders above (until you're on the Review Step).
2. Enter a business name that already exists and watch the error message render as an inline error.

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
+++ b/web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
@@ -518,15 +518,12 @@ describe("<BusinessFormationPaginator />", () => {
         expect(screen.getByText(Config.formation.errorBanner.incompleteStepsError)).toBeInTheDocument();
       });
 
-      it("shows error field states for each step with error", async () => {
+      it("shows error field states for each step with error except for businessName field", async () => {
         const page = preparePage({ business, displayContent });
         await page.stepperClickToBillingStep();
         page.completeRequiredBillingFields();
         await page.stepperClickToReviewStep();
         await page.clickSubmit();
-
-        await page.stepperClickToBusinessNameStep();
-        expect(screen.getByText(Config.formation.errorBanner.errorOnStep)).toBeInTheDocument();
         await page.stepperClickToBusinessStep();
         expect(screen.getByText(Config.formation.errorBanner.errorOnStep)).toBeInTheDocument();
         await page.stepperClickToContactsStep();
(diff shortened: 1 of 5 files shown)


web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
```diff
@@ -691,70 +688,6 @@ describe("<BusinessFormationPaginator />", () => {
             expect(page.getStepStateInStepper(LookupStepIndexByName(formationStepName))).toEqual("ERROR");
             expect(screen.getByText(Config.formation.errorBanner.incompleteStepsError)).toBeInTheDocument();
           });
-
-          it("shows API error message on step for businessName API error", async () => {
```
adidner, line 695: dust - these tests were for the error above the section right? The one the story wants to remove, ergo we don't need them anymore? (Just want to double check)

web/src/components/tasks/business-formation/name/BusinessNameStep.tsx
```diff
@@ -86,6 +86,31 @@ export const BusinessNameStep = (): ReactElement => {
                     onBlurNameField(event.target.value);
                   }}
                 />
+                {state.businessNameAvailability?.status === "UNAVAILABLE" && (
```
adidner, line 89: I was looking around for coverage of this functionality because I was surprised you didn't have to change any test to look for this data on the page

and I don't think we have any?

I was looking in BusinessNameStep I think its a missed test that probably deserves coverage (but can definielty be a post ticket refactor) (also you should look around to make sure I didn't miss it being covered somewhere else in the tree)

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
```diff
@@ -109,7 +109,7 @@ describe("getErrorStateForField", () => {
           formationFormData,
           businessNameAvailability,
         }).label
-      ).toEqual(Config.formation.fields.businessName.errorInlineUnavailable);
+      ).toEqual(undefined);
```
adidner, line 112: dust - So this is undefined, because we need to render something custom for a multi-line inline error right?

adidner (approved):
A couple of questions I had to confirm my understanding is correct. 

And a potential sand follow up test addition for ideal coverage. Looks and works great. Nice job.

web/src/components/tasks/business-formation/getErrorStateForField.ts
```diff
@@ -152,12 +152,13 @@ export const getErrorStateForField = (inputParams: {
     const isAvailable = businessNameAvailability?.status === "AVAILABLE";
     const isValid = exists && isAvailable;
     let label = errorState.label;
+
     if (!exists) {
       label = Config.formation.fields.businessName.errorInlineEmpty;
     } else if (businessNameAvailability?.status === undefined) {
       label = Config.formation.fields.businessName.errorInlineNeedsToSearch;
     } else if (businessNameAvailability?.status !== "AVAILABLE") {
-      label = Config.formation.fields.businessName.errorInlineUnavailable;
```
somabadri, line 160: [pebble] if we are not using this anymore, we should remove it from the json and yaml config files

somabadri (approved):
Great work AnJu! I wish there was a better way to implement this other than returning an undefined label but given the structure of the getErrorStateForFields, but I think this is the best solution, so great job on identifying it! You should bring this up with the team at large to see if we should be implementing a different pattern for the errors as a whole and refactor or if the one off case is okay.

web/src/components/tasks/business-formation/BusinessFormationPaginator.test.tsx
AnJuHyppolite: yes, that's correct.

web/src/components/tasks/business-formation/getErrorStateForField.ts
AnJuHyppolite: Yes!

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
AnJuHyppolite: Yes, I'll bring this up to the team at-large, per Soma's suggestion.

web/src/components/tasks/business-formation/name/BusinessNameStep.tsx
AnJuHyppolite: Great observation, @adidner! We have coverage in `BusinessFormationPaginator` and `BusinessNameStep`, where four  tests failed due to __`TestingLibraryElementError: Unable to find an element by: [data-testid="unavailable-text"]`__. The `data-testid` was on the now deleted `<Alert>` so I added to the `<div>`.

PR #7206, opened 2024-02-09
feat: [#186775240] Add health checks for Formation

cbrunefearless:
## Description

Add health checks for Formation API
### Ticket

[#186775240](https://www.pivotaltracker.com/story/show/)

### Approach

As an engineer supporting the site, I want advance notice of changes in availability or format of the NICUSA Formation API to ensure optimal availability of the Formation tasks.

AC
Given a change is made to the availability of the NICUSA Formation API
When the change is identified in the Dev environment
Then a notification will be sent to the engineers

Given the NICUSA Formation API does not respond to the correct Content-Type header
When identified in the Dev environment
Then a notification will be sent to the engineers

Given a change is made to the availability of the NICUSA Formation API
When the change is identified in the Prod environment
Then a notification will be sent to the engineers

Given the NICUSA Formation API does not respond to the correct Content-Type header
When identified in the Prod environment
Then a notification will be sent to the engineers

### Steps to Test
yarn start:dev the application in one console and open another console and  'curl -I "http://localhost:5002/local/health/self"

## Code author checklist


api/src/client/ApiFormationHelpers.ts
```diff
@@ -0,0 +1,546 @@
...
+  | "ForeignLimitedLiabilityCompany"
+  | "ForeignLimitedPartnership"
+  | "ForeignLimitedLiabilityPartnership";
+
+export type FormationFields = {
+  businessType: BusinessType;
+  shortDescription: "LLC" | "LLP" | "LP" | "DP" | "NP" | "NV" | "NF" | "LF" | "LFC" | "FLC" | "FLP" | "FR";
+};
+
+export const BusinessTypeMap: Record<FormationLegalType | "veteran-nonprofit", FormationFields> = {
```
jphechter, line 464: [sand] Some of these functions and variables are only being used in this file. I'm wondering if they got exported automatically by the IDE when you restructured the files.
```suggestion
const BusinessTypeMap: Record<FormationLegalType | "veteran-nonprofit", FormationFields> = {
```

api/src/client/ApiFormationClient.ts
```diff
@@ -161,534 +156,83 @@ export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): Fo
...
-    };
+      formationFormData
+    );
 
-    // eslint-disable-next-line unicorn/consistent-function-scoping
-    const isInFormOrBylaw = (input: "IN_BYLAWS" | "IN_FORM" | undefined): "Herein" | "Bylaw" => {
-      return input === "IN_FORM" ? "Herein" : "Bylaw";
-    };
+    const returnUrl: string = "returnUrl";
+    const foreignGoodStandingFile: InputFile | undefined = undefined;
```
jphechter, line 186: [sand] If you're just setting this to undefined there's no reason to declare it. You can just leave the positional arg out of the call to `makePostBody`.

api/src/client/ApiFormationClient.ts
```diff
@@ -1,31 +1,26 @@
...
+} from "@client/ApiFormationHelpers";
 import { splitErrorField } from "@client/splitErrorField";
-import { getCurrentBusiness } from "@shared/domain-logic/getCurrentBusiness";
-
-type ApiConfig = {
-  account: string;
-  key: string;
-  baseUrl: string;
-};
+import { generateFormationFormData, generateFormationSigner } from "@shared/test";
```
FarazA22, line 21: boulder - we may want to chat with the rest of the team, if feels weird to use generator functions in our logic

api/src/client/ApiFormationClient.ts
```diff
@@ -161,534 +156,83 @@ export const ApiFormationClient = (config: ApiConfig, logger: LogWriterType): Fo
...
-              OtherProvisions: additionalProvisions,
-            },
-          };
-        }
-        case "nonprofit":
-        case "veteran-nonprofit": {
+    return axios
+      .post(`${config.baseUrl}/PrepareFiling`, postBody, {
+        headers: {
+          "Content-Type": "application/json",
```
jphechter, line 199: [boulder] Currently, the Formation API can support application/json and text/plain, but long term they're going to exclusively use text/plain. Our goal is to only use text/plain. Can we account for this here?

FarazA22 (changes requested):

api/src/client/ApiFormationClient.test.ts
```diff
@@ -1606,4 +1601,35 @@ describe("ApiFormationClient", () => {
...
+      mockAxios.post.mockRejectedValue({});
+      expect(await client.health()).toEqual({
+        success: false,
+        error: {
+          message: "Gateway Timeout",
+          timeout: true,
+        },
+      });
+    });
+  });
```
jphechter, line 1634: [sand] The behavior we want to capture is that there's a notification thrown if anything other than 200 is thrown. I don't know what all values we could expect from the API, but it might be nice to have a wildcard test in here.

api/src/client/ApiFormationHelpers.ts
cbrunefearless: done

api/src/client/ApiFormationClient.ts
cbrunefearless: done changed to  text/plain

api/src/client/ApiFormationClient.ts
cbrunefearless: changed

api/src/client/ApiFormationClient.ts
jphechter: I see what you're saying. Pulling test helpers into production code does feel a little icky to me, but I do see the thought behind this.

Chris and I talked about this yesterday, and at the time I didn't really think anything of it. For a couple reasons, I'm more inclined now to say that it would be worth specifically building data for us to test with. A key reason is the formation API does use some conditional logic to validate a payload. If we're just generating stuff randomly we could also get some false negatives in this health check.

Additionally in considering this now after Faraz called it out, I do also think it's a mistake to use a test generator in our production code. This could end up biting us.

My rec here would be to create a separate file that could be used to load in predefined data so that we could test against specific scenarios. This becomes more relevant if we do find the need to check other validation rules, such as business effective date.

api/src/client/ApiFormationClient.test.ts
cbrunefearless: gotcha, checking to see if the response.data.success is true now, if return success false

api/src/client/ApiFormationClient.ts
cbrunefearless: Made it its own file

PR #7238, opened 2024-02-16
refactor: [#186333254] update sidebar files

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/186333254

### Approach
All sidebar cards appear on top of the opportunities cards and only render one at a time. The weights and above and below criteria are no longes required.

I refactored the props in SidebarCardsList, the display props can be invoked in the child component so I moved them there.

I also created helper functions so domain logic can live outside SidebarCardsList.

I made some updates to the preview component in the CMS.

## Code author checklist


jphechter:
Given this is removing functionality, should it be QA'd by one of the other teams? If so, should we also be calling this a feature instead of a chore?

adidner (approved):
Looks good from what I see, probably do wanna ping content or design for a review just in case we missed something and this is required for some weird reason (or incase this makes them think of something else they want done) like JP said.

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -12,31 +12,45 @@ import { Certification, Funding, SidebarCardContent } from "@/lib/types/types";
...
+  CMS_ONLY_DISPLAY_CERTIFICATIONS?: boolean;
+  CMS_ONLY_IS_REMOTE_SELLER_WORKER?: boolean;
 }
 
 export const SidebarCardsList = (props: SidebarCardsListProps): ReactElement => {
   const [hiddenAccordionIsOpen, setHiddenAccordionIsOpen] = useState<boolean>(false);
   const { Config } = useConfig();
   const { business } = useUserData();
   const accordionRef = useRef(null);
+  const isDesktopAndUp = useMediaQuery(MediaQueries.desktopAndUp);
```
seidior, line 35: [dust] Do we have a follow-up refactor ticket to remove this?

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -12,31 +12,45 @@ import { Certification, Funding, SidebarCardContent } from "@/lib/types/types";
...
+  const displayCertificationsCards = (): boolean => {
+    return (
+      props.CMS_ONLY_DISPLAY_CERTIFICATIONS ||
+      LookupOperatingPhaseById(business?.profileData.operatingPhase).displayCertifications
+    );
+  };
 
   const hiddenOpportunitiesCount = (): number => {
-    if (props.displayCertifications && props.displayFundings) {
+    if (displayCertificationsCards() && displayFundingCards()) {
```
seidior, line 51: [sand] Given how many times you call these functions, perhaps we should cache these in a constant rather than call the function each time?

web/src/lib/domain-logic/filterCertifications.ts
```diff
@@ -3,9 +3,13 @@ import { Certification } from "@/lib/types/types";
 import { Business } from "@businessnjgovnavigator/shared";
 
 export const filterCertifications = (
-  certifications: Certification[],
-  business: Business
+  certifications: Certification[] | undefined,
```
seidior, line 6: [sand] These should probably be designated as optionals (`certifications?`, for example) if they are `| undefined` types.

seidior (approved):
Massive review, so it's a bit of a challenge for me to feel like I've reviewed it properly. Couple of notes, but an overall approval. Thank you for going through all this!

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -46,6 +47,30 @@ function setupMockAnalytics(): typeof analytics {
...
+  (op): boolean => {
+    return !op.displayFundings && op.displayCertifications;
+  }
+);
+
+const operatingPhaseWithDisplayFundingAndDisplayCertificationAreFalse = OperatingPhases.find(
+  (op): boolean => {
+    return !op.displayFundings && !op.displayCertifications;
+  }
+);
```
jphechter, line 72: I think I understand why you did this. Because it's actually getting the Operating Phases that will produce these cases?

Given the component doesn't require complete user data, isn't this a little overkill? I feel like the other approach was simpler to read and only defined what was absolutely necessary to run the test.

web/src/components/dashboard/SidebarCardsList.test.tsx
```diff
@@ -147,65 +191,80 @@ describe("<SidebarCardsList />", () => {
...
           generateSidebarCardContent({
             id: "some-fake-bottom-card",
-            section: "below-opportunities",
           }),
         ];
-        renderComponent({ bottomCards });
+        renderComponent({ sideBarCards });
         expect(screen.queryByTestId("empty-for-you-message")).not.toBeInTheDocument();
       });
     });
```
jphechter, line 270: [sand] Aren't these 2 the same test?

web/src/components/dashboard/SidebarCardsList.tsx
```diff
@@ -12,31 +12,45 @@ import { Certification, Funding, SidebarCardContent } from "@/lib/types/types";
...
+      props.CMS_ONLY_DISPLAY_FUNDINGS ||
+      LookupOperatingPhaseById(business?.profileData.operatingPhase).displayFundings
+    );
+  };
+  const displayCertificationsCards = (): boolean => {
+    return (
+      props.CMS_ONLY_DISPLAY_CERTIFICATIONS ||
+      LookupOperatingPhaseById(business?.profileData.operatingPhase).displayCertifications
+    );
+  };
```
jphechter, line 48: [dust] super nit picky of me, but I probably would have just calculated the value instead of making arrow functions.

```suggestion
  const displayFundingCards = props.CMS_ONLY_DISPLAY_FUNDINGS || LookupOperatingPhaseById(business?.profileData.operatingPhase).displayFundings;
  
  const displayCertificationsCards = props.CMS_ONLY_DISPLAY_CERTIFICATIONS || LookupOperatingPhaseById(business?.profileData.operatingPhase).displayCertifications;
```

web/src/lib/domain-logic/getVisibleSideBarCards.ts
```diff
@@ -0,0 +1,13 @@
+import { SidebarCardContent } from "@/lib/types/types";
```
jphechter, line 1: I don't know if I agree with some of these functions being added to `/domain-logic` this still feels pretty component specific to me.

web/src/components/dashboard/SidebarCardsList.tsx
FarazA22: love it

web/src/components/dashboard/SidebarCardsList.test.tsx
FarazA22: Yes, I'll delete the duplicate and update the test name

web/src/components/dashboard/SidebarCardsList.tsx
FarazA22: yeah, I missed this, I will update, Spencer brought it up as well.

web/src/components/dashboard/SidebarCardsList.tsx
FarazA22: https://www.pivotaltracker.com/story/show/187105184

web/src/components/dashboard/SidebarCardsList.test.tsx
FarazA22: discussed

web/src/lib/domain-logic/getVisibleSideBarCards.ts
FarazA22: moved related helpers to a single function, and created this ticket - https://www.pivotaltracker.com/story/show/187105804

jphechter (approved):
I think this all looks good. Thanks for making the changes! 👍

PR #7244, opened 2024-02-19
Set initial business effective date relative to NJ local date

jphechter:
## Description

Users in time zones outside ET run the risk of getting a business effective date set to something that breaks existing validation rules. In many places throughout our application we use `dayJs` to check the current date, but this ends up being relative to the user, not the state of New Jersey. This is a small fix that attempts to fix the most common area where this causes issues: the business effective date field in the business formation task.

### Ticket

[#185761403](https://www.pivotaltracker.com/story/show/185761403)

### Approach

Utilize the timezone extension to check dates relative to the US Eastern time zone where New Jersey falls.

### Steps to Test

- Change machine timezone manually to something that will have a date different than US Eastern time. (I used Manila, Philippines as my location, but anything around the international date line should work)
- Start a new business
- Open formation task
- Check that the initial value for the business start date is the current date in NJ
- If the legal structure is one that can select a different date (ex: foreign non-profit)
    - check that the first selectable date is today's date in NJ

### Notes

There are many, many places where we are getting the current date, not necessarily relative to NJ. We should audit these. None of these have come up in user reported issues that we know of, so they may not be as pressing. Additionally, none of them should prevent a user from performing an important task, but for consistency sake we should check.

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/business-formation.json
+++ b/content/src/fieldConfig/business-formation.json
@@ -265,7 +265,7 @@
       "businessStartDate": {
         "label": "NJ Business Effective Date",
         "labelContextualInfo": "effective-date",
-        "labelSecondaryText": "(mm/dd/yyyy)",
+        "labelSecondaryText": "(mm/dd/yyyy) Eastern Time",
         "errorFuture": "The NJ business effective date must be today or a future date in format MM/DD/YYYY.",
         "errorToday": "The NJ business effective date must be today and in format MM/DD/YYYY.",
         "error30Days": "The NJ business effective date must be today or a future date within 30 days in format MM/DD/YYYY.",
--- a/shared/src/dateHelpers.ts
+++ b/shared/src/dateHelpers.ts
@@ -21,12 +21,12 @@ export const getCurrentDateInNewJersey = (): Dayjs => {
   return dayjs().tz("America/New_York");
 };
 
-export const getCurrentDateInNewJerseyISOFormatted = (): string => {
-  return getCurrentDateInNewJersey().toISOString();
+export const getCurrentDateFormatted = (format: string): string => {
+  return getCurrentDate().format(format);
 };
 
-export const getCurrentDateFormatted = (format: string): string => {
-  return dayjs().format(format);
+export const getCurrentDateInNewJerseyFormatted = (format: string): string => {
+  return getCurrentDateInNewJersey().format(format);
 };
 
 export const getDateInCurrentYear = (date: string): Dayjs => {
@@ -39,6 +39,10 @@ export const isDateAfterCurrentDate = (date: string): boolean => {
 
 export const getCurrentDateISOString = (): string => {
   return dayjs().toISOString();
+};
+
+export const getCurrentDateInNewJerseyISOString = (): string => {
+  return getCurrentDateInNewJersey().toISOString();
 };
 
 export const parseDate = (date: string | number | undefined): Dayjs => {
--- a/web/src/components/tasks/business-formation/BusinessFormation.test.tsx
+++ b/web/src/components/tasks/business-formation/BusinessFormation.test.tsx
@@ -20,6 +20,7 @@ import {
   generateMunicipality,
   getCurrentBusiness,
   getCurrentDate,
+  getCurrentDateInNewJerseyFormatted,
 } from "@businessnjgovnavigator/shared";
 import {
   generateFormationData,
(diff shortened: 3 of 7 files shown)


shared/src/dateHelpers.ts
```diff
@@ -41,6 +41,10 @@ export const getCurrentDateISOString = (): string => {
   return dayjs().toISOString();
 };
 
+export const getCurrentDateInNewJerseyISOString = (): string => {
+  return getCurrentDateInNewJersey().toISOString();
+};
+
```
jphechter, line 47: Changed to match established pattern.

web/src/components/tasks/business-formation/business/FormationDate.tsx
```diff
@@ -73,30 +75,36 @@ export const FormationDate = (props: Props): ReactElement => {
     });
   };
 
+  const getMinDate = (): dayjs.Dayjs | undefined => {
+    const startDate = dayjs(getCurrentDateInNewJersey().format(dateFormat));
+    return props.fieldName === "businessStartDate" ? startDate : undefined;
+  };
+
```
jphechter, line 82: There should be a way to set a relative locale via the `LocalizationProvider`, but everything I tried wasn't working. What did work was getting the current date in NJ, creating a string, and then creating a new date object to be used in setting the date minimum. I don't love this, but it works.

shared/src/dateHelpers.ts
```diff
@@ -21,12 +21,12 @@ export const getCurrentDateInNewJersey = (): Dayjs => {
   return dayjs().tz("America/New_York");
 };
 
-export const getCurrentDateInNewJerseyISOFormatted = (): string => {
-  return getCurrentDateInNewJersey().toISOString();
+export const getCurrentDateFormatted = (format: string): string => {
+  return getCurrentDate().format(format);
```
AnJuHyppolite, line 25: Nice catch!

shared/src/dateHelpers.ts
```diff
@@ -21,12 +21,12 @@ export const getCurrentDateInNewJersey = (): Dayjs => {
...
 };
 
-export const getCurrentDateFormatted = (format: string): string => {
-  return dayjs().format(format);
+export const getCurrentDateInNewJerseyFormatted = (format: string): string => {
```
AnJuHyppolite, line 28: [pebble] Should write a test for this function? We have one for `getCurrentDateFormatted`.

web/src/components/tasks/business-formation/business/FormationDate.tsx
```diff
@@ -73,30 +75,36 @@ export const FormationDate = (props: Props): ReactElement => {
     });
   };
 
+  const getMinDate = (): dayjs.Dayjs | undefined => {
```
AnJuHyppolite, line 78: [pebble] Should we write a test?

shared/src/dateHelpers.ts
seidior: [dust] Question for @jphechter: is there ever a situation where we want to `getCurrentDate` without getting the current date in New Jersey?

seidior (approved):
Already some great comments from @AnJuHyppolite, so just a couple minor notes from me.

shared/src/dateHelpers.ts
jphechter: I'm sure, for instance something that has to do with handling data server-side. I don't see a reason that we'd want to calculate for NJ specifically, even if they do happen to be the same. I didn't want to completely replace for that reason.

web/src/components/tasks/business-formation/business/FormationDate.tsx
jphechter: probably 🙃

web/src/components/tasks/business-formation/BusinessFormation.test.tsx
```diff
@@ -1191,4 +1192,77 @@ describe("<BusinessFormation />", () => {
     expect(formationFormData.nonprofitTrusteesMethodSpecified).toEqual(undefined);
     expect(formationFormData.nonprofitAssetDistributionSpecified).toEqual(undefined);
   }, 60000);
+
```
jphechter, line 1195: @AnJuHyppolite added tests

content/src/fieldConfig/business-formation.json
```diff
@@ -265,7 +265,7 @@
       "businessStartDate": {
         "label": "NJ Business Effective Date",
         "labelContextualInfo": "effective-date",
-        "labelSecondaryText": "(mm/dd/yyyy)",
+        "labelSecondaryText": "(mm/dd/yyyy) Eastern Time",
```
FarazA22, line 268: eastern date?

web/src/components/tasks/business-formation/BusinessFormation.test.tsx
```diff
@@ -1191,4 +1192,70 @@ describe("<BusinessFormation />", () => {
...
+          }),
+          formationData: generateEmptyFormationData(),
+        },
+        displayContent,
+      });
+      await page.fillAndSubmitBusinessNameStep("Pizza Joint");
+      expect(screen.getByTestId("date-businessStartDate")).toHaveValue(expectedDateString);
+    });
+
+    it("when user is in Phillipine Timezone, initial value is current date in NJ", async () => {
```
seidior, line 1219: [pebble] The time zone's name is Philippine Standard Time.

content/src/fieldConfig/business-formation.json
jphechter: This is what Content and Design asked for here.

web/src/components/tasks/business-formation/BusinessFormation.test.tsx
```diff
@@ -1191,4 +1192,70 @@ describe("<BusinessFormation />", () => {
...
+          }),
+          formationData: generateEmptyFormationData(),
+        },
+        displayContent,
+      });
+      await page.fillAndSubmitBusinessNameStep("Pizza Joint");
+      expect(screen.getByTestId("date-businessStartDate")).toHaveValue(expectedDateString);
+    });
+
+    it("when user is in US Pacific Timezone, initial value is current date in NJ", async () => {
```
seidior, line 1240: [dust] I'd rather see this be in the Hawaii–Aleutian Time Zone, especially since you're not testing DST. Also US/Pacific isn't -08:00 all year round, most especially not on 4/13/2020 😉

content/src/fieldConfig/business-formation.json
seidior: "Eastern Time" would actually be the most acceptable name here for the timezone, even though you're only specifying the date.

web/src/components/tasks/business-formation/BusinessFormation.test.tsx
FarazA22: I think the test title should be something along the lines of... when the system date is 1 day before the date in NJ, the NJ date rendered. The current title makes me think we have some special logic around timezones

PR #7249, opened 2024-02-22
feat: [#186781958] remove step 4 from dakota nexus onboarding

somabadri:
## Description

Removed step 4 from dakota nexus onboarding and now utilizing the office in NJ checkbox to get the same information from the user.

### Ticket

[#186781958](https://www.pivotaltracker.com/story/show/186781958)

## Code author checklist


shared/src/domain-logic/nexusLocationInNewJersey.ts
```diff
@@ -0,0 +1,13 @@
+import { ProfileData } from "../profileData";
+import { determineForeignBusinessType } from "./businessPersonaHelpers";
```
AnJuHyppolite, line 2: [pebble] Can we change out the relative imports?

shared/src/domain-logic/opportunityFields.ts
```diff
@@ -1,6 +1,7 @@
 import { LookupLegalStructureById } from "../legalStructure";
 import { ProfileData } from "../profileData";
 import { determineForeignBusinessType } from "./businessPersonaHelpers";
+import { nexusLocationInNewJersey } from "./nexusLocationInNewJersey";
```
AnJuHyppolite, line 4: [pebble] Let's update the import so it's not relative.

web/src/components/data-fields/IndustryDropdown.tsx
```diff
@@ -55,9 +56,8 @@ export const IndustryDropdown = (props: Props): ReactElement => {
     let homeBasedBusiness: boolean | undefined = undefined;
     let cannabisLicenseType = undefined;
 
-    if (!isHomeBasedBusinessApplicable(industryId)) {
-      homeBasedBusiness = false;
-    }
+    homeBasedBusiness = isHomeBasedBusinessApplicable(industryId) ? homeBasedBusiness : false;
+    homeBasedBusiness = nexusLocationInNewJersey(state.profileData) ? false : homeBasedBusiness;
```
AnJuHyppolite, line 60: [dust] Can we combine the logic for this variable? Would something like `homeBasedBusiness = isHomeBasedBusinessApplicable(industryId) && !nexusLocationInNewJersey(state.profileData)` work?

shared/src/domain-logic/nexusLocationInNewJersey.ts
somabadri: we don't have relative imports set up for the shared directory

shared/src/domain-logic/opportunityFields.ts
somabadri: we don't have relative imports set up for the shared directory

web/src/components/data-fields/IndustryDropdown.tsx
```diff
@@ -55,9 +56,10 @@ export const IndustryDropdown = (props: Props): ReactElement => {
     let homeBasedBusiness: boolean | undefined = undefined;
     let cannabisLicenseType = undefined;
 
-    if (!isHomeBasedBusinessApplicable(industryId)) {
-      homeBasedBusiness = false;
-    }
+    homeBasedBusiness =
+      !isHomeBasedBusinessApplicable(industryId) || nexusLocationInNewJersey(state.profileData)
+        ? false
+        : homeBasedBusiness;
```
AnJuHyppolite, line 62: [dust] Would we consider `homeBasedBusiness =
      isHomeBasedBusinessApplicable(industryId) || !nexusLocationInNewJersey(state.profileData)
        ? homeBasedBusiness : false` for readability?

web/src/components/data-fields/IndustryDropdown.tsx
```diff
@@ -55,9 +56,10 @@ export const IndustryDropdown = (props: Props): ReactElement => {
     let homeBasedBusiness: boolean | undefined = undefined;
     let cannabisLicenseType = undefined;
 
-    if (!isHomeBasedBusinessApplicable(industryId)) {
-      homeBasedBusiness = false;
-    }
+    homeBasedBusiness =
+      isHomeBasedBusinessApplicable(industryId) || !nexusLocationInNewJersey(state.profileData)
+        ? homeBasedBusiness
+        : false;
```
adidner, line 62: dust - I think this could be more readable?

`homeBasedBusiness =
      isHomeBasedBusinessApplicable(industryId) || !nexusLocationInNewJersey(state.profileData)
        ? undefined
        : false;`

maybe? Because it homeBaseBusiness will always be undefined at this point?

Alternatively you could do an if because the if case isn't actually changing the value of homeBasedBusiness. 

Super nit picky on my end. Feel free to ignore

adidner: Also this is a bit of change in logic isn't it? I didn't see something about this in the ticket initially, but I assume you talked to someone and they wanted it right?

adidner (approved):
LGMT few nit picky comments and questions

api/src/db/migrations/v130_remove_nexus_location_in_new_jersey.ts
```diff
@@ -0,0 +1,475 @@
...
+const migrate_v129Business_to_v130Business = (business: v129Business): v130Business => {
+  if (
+    business.profileData.nexusLocationInNewJersey === true &&
+    !business.profileData.foreignBusinessTypeIds.includes("officeInNJ")
+  ) {
+    business.profileData.foreignBusinessTypeIds.push("officeInNJ");
+  } else if (business.profileData.nexusLocationInNewJersey === false) {
+    business.profileData.foreignBusinessTypeIds = business.profileData.foreignBusinessTypeIds.filter(
+      (id) => id !== "officeInNJ"
+    );
```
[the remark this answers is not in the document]
somabadri: I will bring it up standup tomorrow because I was thinking the same thing but wasn't sure what the right decision was.

web/src/components/data-fields/IndustryDropdown.tsx
somabadri: I refactored it back to what I had initially/you suggested because the logic is different when written this way.

PR #7291, opened 2024-02-29
feat: [#186811252] consistent navigation

adidner:
## Description

This was a monster of a ticket (but only because I wanted it to be)

I basically ended up refactoring the entire navigation system. 

So now its first a division from mobile vs desktop then a division across loading, landing, onboarding and authed states. 

So there are 10 total states, their divisions should be obvious as well as the composition for each state super obvious. Which was my goal

I also ended up refactoring our testing situation because of this as well, trying to make the smallest composite part (individual menu items) an easy thing to test their actual ability to redirect to other locations and have correct text. 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186811252)

### Approach

The prior version of this was really confusing because the lines of division were all weird, we were dividing the states of the nav bar up at different levels of the application, we were also being really unclear as to what things showed in the nav bar during what states because we were using useContext hooks really deep into the stack at the level of the dropdown / drawers to change content. 

I removed this by making the various states more obvious and readable and making them be more explicit about what they were passing into their submenus. 

### Steps to Test

Hop in the app, look at the various states 

### Notes

I think this was good and a nice step. 

I think with this new paradigm we could potentially also refactor the getBusinessNavTitle function to be a little better as the states are now pretty explicit so we could pass things in a bit more clearly. 

I also think there is a weird naming convention here that I want to resolve in a follow up becasue its odd. 

## Code author checklist


web/src/components/navbar/desktop/NavBarDesktop.tsx
```diff
@@ -0,0 +1,185 @@
...
+  }, [open]);
+
+  const textColor = props.isAuthenticated ? "primary" : "base";
+  const navBarBusinessTitle = getNavBarBusinessTitle(business, props.isAuthenticated);
+  const currentIndex =
+    userData && business ? orderBusinessIdsByDateCreated(userData).indexOf(business.id) : 0;
+
+  if (props.logoOnlyType) {
+    // loading/redirect/ethan onboarding
+    return <NavBarLogoOnly logoType={props.logoOnlyType} />;
```
adidner, line 65: sand - So this component could and probably should be refactored into the same pattern as below (I will write a ticket to do so)

web/src/components/navbar/desktop/NavBarDesktop.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const prevOpen = useRef(open);
+  useEffect(() => {
+    if (null !== anchorRef.current && prevOpen.current && !open) {
+      anchorRef.current.focus();
+    }
+    prevOpen.current = open;
+  }, [open]);
+
+  const textColor = props.isAuthenticated ? "primary" : "base";
+  const navBarBusinessTitle = getNavBarBusinessTitle(business, props.isAuthenticated);
```
adidner, line 59: sand - so this is a function I was thinking about refactoring as well considering our cases are pretty distinct now below

web/src/components/navbar/shared-submenu-components.tsx
```diff
@@ -0,0 +1,232 @@
...
+      analytics.event.landing_page_navbar_log_in.click.go_to_myNJ_login();
+      triggerSignIn();
+    },
+    icon: <ButtonIcon svgFilename="login" sizePx="25px" />,
+    itemText: Config.navigationDefaults.logInButton,
+    key: "loginMenuItem",
+  });
+};
+
+export const LogoutMenuItem = (props: { handleClose: () => void }): ReactElement => {
```
adidner, line 32: dust - so I thought about breaking this out into many files and it just seemed kinda like pointless file bloat given these are all pretty simple in terms of implementation and testing (expcept profile which is more complicated)

web/src/components/onboarding/ReturnToPreviousBusinessBar.tsx
```diff
@@ -20,7 +21,10 @@ export const ReturnToPreviousBusinessBar = (props: Props): ReactElement | null =
   const { updateQueue, userData } = useUserData();
   const { state } = useContext(AuthContext);
   const router = useRouter();
-  const businessName = getNavBarBusinessTitle(props.previousBusiness, state.isAuthenticated);
+  const businessName = getNavBarBusinessTitle(
+    props.previousBusiness,
+    state.isAuthenticated === IsAuthenticated.TRUE
```
adidner, line 26: This was just a much needed simplification 

I still think this could be refactored to be simpler as many of the use cases are better divided now because of how I chose to do our cases for, onboarding, landing, guest and authed nav,  but its actuallly just complicated enough to probably warrent a follow up instead

web/src/components/navbar/NavBar.tsx
```diff
@@ -20,6 +21,13 @@ export const NavBar = (props: Props): ReactElement => {
   const isLargeScreen = useMediaQuery(MediaQueries.desktopAndUp);
```
seidior, line 21: [boulder] I was hoping your refactor would cover the removal of `useMediaQuery()`.

seidior (changes requested):

web/src/components/navbar/NavBar.tsx
adidner: Yeah, alright. I can do that. 

I was imagining like a css style display none for depending on a css media query. Is that what you were thinking?

seidior: Yep that sounds good!

adidner: I used inline for one of the 4 states becasue 
image
this was overriding my block, and I don't want to touch that because who knows what depends on that existing.

adidner: image
is fixed now though. Will push up in a second

seidior (commented):
Thanks for addressing my feedback! Switching to a comment approval so I'm not in your way from merging this in once your PR reviewers approve.

seidior (approved):
That didn't work, apparently. Guess I have to expressly switch to approve.

content/src/fieldConfig/config.json
```diff
@@ -46,33 +46,50 @@
   },
   "navigationDefaults": {
     "navBarUnnamedDbaBusinessText": "Unnamed DBA",
+    "navigationQuickLinks": {
```
FarazA22, line 49: pebble / sand: given the complexity of the NAV bar, we should extract the content in the config.Json to another file. We are actively trying to not add to this file, because the end goal is to delete it completely. Given the amount of time you have taken on this large ticket, perhaps it makes sense to create a chore to re-factor this. I also think given the different variations of the Nav Bar, we should probably add a preview so content can view what the changes look like in the CMS.

web/src/components/onboarding/ReturnToPreviousBusinessBar.tsx
```diff
@@ -20,7 +21,10 @@ export const ReturnToPreviousBusinessBar = (props: Props): ReactElement | null =
   const { updateQueue, userData } = useUserData();
   const { state } = useContext(AuthContext);
   const router = useRouter();
-  const businessName = getNavBarBusinessTitle(props.previousBusiness, state.isAuthenticated);
+  const businessName = getNavBarBusinessTitle(
```
FarazA22, line 24: sand - this component would be a good candidate for a refactor where we inject domain logic via props and the would render as a dummy component. Thoughts?

web/src/styles/sections/nav-bar.scss
```diff
@@ -20,6 +20,13 @@ $icon-offset: 45px;
   align-items: center;
 }
 
+.reduced-width {
```
FarazA22, line 23: pebble - can we update the variable name so it's clear it's more specific to the Nac bar

web/src/components/navbar/NavBar.tsx
```diff
@@ -36,25 +41,38 @@ export const NavBar = (props: Props): ReactElement => {
...
-  } else if (isLargeScreen) {
-    return <NavBarDesktop previousBusinessId={props.previousBusinessId} />;
-  } else {
-    return (
-      <>
+  const currentlyOnboarding = (): boolean => {
+    if (!router) {
+      return false;
+    }
+    return router.pathname === ROUTES.onboarding;
```
FarazA22, line 48: sand - can we use `router?.pathname` and remove the if statement all together?

web/src/components/navbar/desktop/NavBarDesktop.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const currentIndex =
+    userData && business ? orderBusinessIdsByDateCreated(userData).indexOf(business.id) : 0;
+
+  if (props.logoOnlyType) {
+    // loading/redirect/ethan onboarding
+    return <NavBarLogoOnly logoType={props.logoOnlyType} />;
+  } else if (props.isLanding) {
+    // landing
```
FarazA22, line 67: pebble - to avoid using these comments you could extract each navbar component into it's own component that lives within the page 

`const LandingNavBar = ({ ...props }) => {/* ... */};`
`const OnboardingNavBar = ({ ...props }) => {/* ... */};`

web/src/components/navbar/desktop/DesktopLoginButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const DesktopLoginButton = (): ReactElement => {
+  const { Config } = useConfig();
+  return (
+    <span className="margin-right-2">
+      <UnStyledButton
+        dataTestid="login-button"
+        onClick={(): void => {
+          triggerSignIn();
+          analytics.event.landing_page_navbar_log_in.click.go_to_myNJ_login();
```
FarazA22, line 15: pebble - are we testing the analytics firing or the trigger sign in function being called?

web/src/components/navbar/desktop/NavBarDesktop.tsx
```diff
@@ -0,0 +1,185 @@
...
+interface Props {
+  previousBusinessId?: string | undefined;
+  isLanding: boolean | undefined;
+  logoOnlyType?: "NAVIGATOR_LOGO" | "NAVIGATOR_MYNJ_LOGO" | undefined;
+  currentlyOnboarding: boolean | undefined;
+  isAuthenticated: boolean;
+}
+
+export const NavBarDesktop = (props: Props): ReactElement => {
+  const { business, userData } = useUserData();
```
FarazA22, line 34: dust - if we pass in userdata as a prop, this component can be a dummy UI component

web/src/components/navbar/desktop/NavBarDesktopDropDown.tsx
```diff
@@ -0,0 +1,98 @@
...
+  setOpen: React.Dispatch<React.SetStateAction<boolean>>;
+  handleClose: () => void;
+  icon?: ReactElement;
+  subMenuElement: ReactElement[];
+}
+
+export const NavBarDesktopDropDown = (props: Props): ReactElement => {
+  const toggleDropdown = (): void => {
+    if (!open) {
+      analytics.event.account_name.click.expand_account_menu();
```
FarazA22, line 22: pebble - are we testing analytics firing? Should we move it to the relevant test file?

web/src/components/navbar/desktop/NavBarDesktopHomeLogo.tsx
```diff
@@ -0,0 +1,26 @@
+import { NavBarDashboardLink } from "@/components/navbar/NavBarDashboardLink";
+import { NavBarVerticalLine } from "@/components/navbar/NavBarVerticalLine";
+import { NavbarBusinessNjGovLogo } from "@/components/navbar/NavbarBusinessNjGovLogo";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { ReactElement } from "react";
+
+interface Props {
+  previousBusinessId: string | undefined;
```
FarazA22, line 8: pebble - spencer had a great reason for using `?` instead of explicitly using undefined as a type. I forgot why, but we can simply say `previousBusinessId?: string`

web/src/components/navbar/desktop/NavBarDesktopQuickLinks.test.tsx
```diff
@@ -0,0 +1,84 @@
+import { NavBarDesktopQuickLinks } from "@/components/navbar/desktop/NavBarDesktopQuickLinks";
+import { getMergedConfig } from "@/contexts/configContext";
+import { fireEvent, render, screen } from "@testing-library/react";
+
+describe("<NavBarDesktopQuickLinks />", () => {
+  const Config = getMergedConfig();
+
+  window.open = jest.fn();
```
FarazA22, line 8: dust - we've turned `open` into a jest function. Any reason we are not resetting mocks after each test. Any reason why we are not reverting the `open` fn back to what it previously was?

web/src/components/navbar/desktop/NavBarDesktopQuickLinks.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+      "_blank",
+      "noopener noreferrer"
+    );
+  });
+
+  it("renders the operate quick link and redirects correctly", async () => {
+    render(<NavBarDesktopQuickLinks />);
+    fireEvent.click(screen.getByText(Config.navigationDefaults.navigationQuickLinks.navBarOperateText));
+
+    expect(window.open).toHaveBeenCalledWith(
```
FarazA22, line 45: pebble - because your not resetting the mocks this result can be a carry over from a previous test.

web/src/components/navbar/mobile/NavBarMobile.tsx
```diff
@@ -0,0 +1,166 @@
...
+  showSidebar?: boolean;
+  isLanding?: boolean;
+  previousBusinessId?: string | undefined;
+  logoOnlyType?: "NAVIGATOR_LOGO" | "NAVIGATOR_MYNJ_LOGO" | undefined;
+  currentlyOnboarding: boolean;
+  isAuthenticated: boolean;
+}
+
+export const NavBarMobile = (props: Props): ReactElement => {
+  const { business } = useUserData();
```
FarazA22, line 35: dust - injecting business as a prop should turn this component into a dummy UI component

web/src/components/navbar/mobile/NavBarMobile.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const openSidebar = (): void => {
+    setIsSidebarOpen(true);
+  };
+
+  const closeSideBar = (): void => {
+    setIsSidebarOpen(false);
+  };
+
+  // logo only / loading/redirect
```
FarazA22, line 50: pasting same comment I wrote on the desktop component

pebble - to avoid using these comments you could extract each navbar component into it's own component that lives within the page
const LandingNavBar = ({ ...props }) => {/* ... */}; const OnboardingNavBar = ({ ...props }) => {/* ... */};

web/src/components/navbar/mobile/NavBarMobileAccountSlideOutMenu.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export const NavBarMobileAccountSlideOutMenu = (props: Props): ReactElement => {
+  return (
+    <>
+      <button
+        className="right-nav-menu-button radius-0"
+        data-testid="nav-menu-mobile-account-open"
+        aria-label="open menu"
+        onClick={(): void => {
+          analytics.event.mobile_hamburger_icon.click.open_mobile_menu();
```
FarazA22, line 29: dust - we should be testing the 3 analytics events that fire

web/src/components/navbar/mobile/NavBarMobileAccountSlideOutMenu.test.tsx
```diff
@@ -0,0 +1,67 @@
+import { NavBarMobileAccountSlideOutMenu } from "@/components/navbar/mobile/NavBarMobileAccountSlideOutMenu";
+import { MenuItem } from "@mui/material";
+import { fireEvent, render, screen } from "@testing-library/react";
+
+describe("<NavBarMobileAccountSlideOutMenu />", () => {
+  window.open = jest.fn();
```
FarazA22, line 6: posting same comment as desktop 

pebble - we've turned open into a jest function. Any reason we are not resetting mocks after each test. Any reason why we are not reverting the open fn back to what it previously was?

web/src/components/navbar/mobile/NavBarMobileQuickLinksSlideOutMenu.test.tsx
```diff
@@ -0,0 +1,173 @@
+import { NavBarMobileQuickLinksSlideOutMenu } from "@/components/navbar/mobile/NavBarMobileQuickLinksSlideOutMenu";
+import { getMergedConfig } from "@/contexts/configContext";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+
+describe("<NavBarMobileQuickLinksSlideOutMenu />", () => {
+  const Config = getMergedConfig();
+
+  window.open = jest.fn();
```
FarazA22, line 8: posting same comment as desktop
pebble - we've turned open into a jest function. Any reason we are not resetting mocks after each test. Any reason why we are not reverting the open fn back to what it previously was?

web/src/components/navbar/mobile/NavBarMobileQuickLinksSlideOutMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+          </MenuList>
+        </nav>
+      </FocusTrappedSidebar>
+
+      <div
+        className={`right-nav-overlay ${sidebarIsOpen ? "is-visible" : ""}`}
+        data-testid="nav-menu-mobile-quick-link-close"
+        aria-hidden="true"
+        onClick={(): void => {
+          analytics.event.mobile_menu.click_outside.close_mobile_menu();
```
FarazA22, line 88: dust - we should have test for the two anaytics events that fire

web/src/components/navbar/NavBar.tsx
```diff
@@ -17,9 +16,15 @@ type Props = {
 };
 
 export const NavBar = (props: Props): ReactElement => {
-  const isLargeScreen = useMediaQuery(MediaQueries.desktopAndUp);
   const [scrolled, setScrolled] = useState(false);
 
+  const { state } = useContext(AuthContext);
+  const isAuthenticated = useMemo(() => {
```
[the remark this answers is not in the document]
adidner: So we don't want it to be re-caculated every time `state` changes just every time `state.isAuthenticated` changes. (I believe thats what this code is doing, I copied it from somewhere during the refactor)

content/src/fieldConfig/config.json
adidner: That makes sense, I would say given the amount of time on the ticket and the amount of work in doing the preview as well it would probably make more sense as a refactor ticket and I can try and pick it up on Friday

web/src/components/navbar/NavBar.tsx
adidner: ya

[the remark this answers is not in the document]
adidner: I belive you're asking me to reverse the order of my statements?

What I have works in app because the so there is no precedence issue, I'm down to change it for style reasons though.

web/src/components/navbar/desktop/DesktopLoginButton.tsx
adidner: Nope, I ripped this out of another function and thought we had tests but I double checked and we do not. So I will add, nice catch

web/src/components/navbar/desktop/NavBarDesktop.tsx
adidner: I thought about that and elected not to. 

The reason is because I think that any change you might make its probably relevant or good even to see how the other states are structure to know the differences in their composition. 

If you divide it up I'm worried it will kinda obfuscate / making some details harder to find (searching across files tends to be more annoying than in a file, and looking at things below one another is easier than across 4 or 5 files typically)

web/src/components/navbar/desktop/NavBarDesktopDropDown.tsx
adidner: ya

web/src/components/navbar/desktop/NavBarDesktopHomeLogo.tsx
adidner: Sure

web/src/components/navbar/desktop/NavBarDesktopQuickLinks.test.tsx
adidner: I forgot is the reason

web/src/components/navbar/mobile/NavBarMobile.tsx
adidner: see my rebuttle

web/src/styles/sections/nav-bar.scss
adidner: sure

web/src/components/onboarding/ReturnToPreviousBusinessBar.tsx
adidner: Sure? I think we could pass in all the data we're getting from useX calls here. I'll write a ticket for it for the future.

web/src/components/navbar/desktop/NavBarDesktop.tsx
adidner: Also userData?

PR #7333, opened 2024-03-07
chore: [#187004657] employment agency task id required

cbrunefearless:
## Description
Steps to Reproduce:

Onboard as an employment agency industry and complete the business structure task
In step 4 of the roadmap, go to the "register your" task details page
Current Behavior:

In the roadmap, there are multiple Active/Current tasks instead of one. Also, the tasks are displaying the same content.

Expected Behavior:
One active task in the mini roadmap, and each task routes/renders the correct information.

Dev Note
Since the impact of this bug is currently not clear, it would be helpful to first ensure that triage is completed to help understand whether or not this bug has significant impact (i.e. can impact any/all users) or if the impact is limited (i.e. certain set of circumstances or narrow use cases). If the impact is limited and/or if the fix requires significant effort then a discussion could occur to see if the fix could occur at a later time.

### Ticket

[#187004657](https://www.pivotaltracker.com/story/show/)

## Code author checklist


web/public/mgmt/config.yml
```diff
@@ -2110,7 +2110,7 @@ collections:
       - { label: "Filename", name: "filename", widget: "write-once-read-only" }
       - { label: "Display Name", name: "displayname", widget: "nospace" }
       - { label: "Name", name: "name", widget: "string" }
-      - { label: "Id", name: "id", widget: "write-once-read-only" }
+      - { label: "Id", name: "id", widget: "write-once-read-only", required: true }
```
FarazA22, line 2113: The required flag is fine here.. the issue is that this field doesn't exist in another collection... I believe this is the right collection - License Tasks (Navigator with Webflow mappings)

web/public/mgmt/config.yml
```diff
@@ -2166,6 +2166,7 @@ collections:
       preview: true
     fields:
       - { label: "Internal Notes", name: "notesMd", widget: "markdown", required: false }
+      - { label: "Id", name: "id", widget: "string", required: true, hint: "This is the id of the file." }
```
FarazA22, line 2169: Can we add the hint text to Task - All, and maybe update it to say, it should be the same as filename. Also, I think this should also use the write once widget. thoughts?

cbrunefearless: i took it out of the task-all bc i dunno if that functionality of required would do anything to the current code, should i add it back?

PR #7406, opened 2024-03-20
chore: [#186665645] add healthCheck lambda function

somabadri:
## Description

Added a lambda function that runs a healthCheck on all of the endpoints.

### Ticket

[#186665645](https://www.pivotaltracker.com/story/show/186665645)

## Code author checklist

### DIFF
--- a/api/src/libs/healthCheck.ts
+++ b/api/src/libs/healthCheck.ts
@@ -1,0 +1,46 @@
+import { LogWriterType } from "@libs/logWriter";
+import axios, { AxiosError, AxiosResponse } from "axios";
+
+type Status = "PASS" | "FAIL" | "ERROR";
+type StatusResult = Record<string, Status>;
+
+const healthCheckEndPoints: Record<string, string> = {
+  self: "self",
+  elevator: "dynamics/elevator",
+  fireSafety: "dynamics/fire-safety",
+  housing: "dynamics/housing",
+  dynamicsLicenseStatus: "dynamics/license-status",
+  webserviceLicenseStatus: "webservice/license-status",
+  webserviceFormation: "webservice/formation",
+};
+
+const url =
+  process.env.STAGE === "prod"
+    ? "https://api.navigator.business.nj.gov"
+    : "https://dev.api.navigator.business.nj.gov";
+
+const healthCheck = async (type: string, logger: LogWriterType): Promise<Status> => {
+  return axios
+    .get(`${url}/health/${type}`)
+    .then((response: AxiosResponse) => {
+      if (response.data.success === true) {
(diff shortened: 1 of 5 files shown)


api/src/libs/healthCheck.ts
```diff
@@ -0,0 +1,46 @@
...
+  housing: "dynamics/housing",
+  dynamicsLicenseStatus: "dynamics/license-status",
+  webserviceLicenseStatus: "webservice/license-status",
+  webserviceFormation: "webservice/formation",
+};
+
+const url =
+  process.env.STAGE === "prod"
+    ? "https://api.navigator.business.nj.gov"
+    : "https://dev.api.navigator.business.nj.gov";
```
seidior, line 20: [sand] Let's create environment variables for these, if we don't already have them. That way it's easier for us to change them in the future.

api/src/libs/healthCheck.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const healthCheck = async (type: string, logger: LogWriterType): Promise<Status> => {
+  return axios
```
seidior, line 23: [pebble] It's 2024; let's ditch axios for `fetch()` in all new development and slowly phase it out of our code base.

api/src/libs/healthCheck.ts
seidior: Also, since we're not appending to this string, it should be called `domain` or `urlBase`, since it won't be the final full resource locator.

api/src/libs/healthCheck.ts
```diff
@@ -0,0 +1,46 @@
+import { LogWriterType } from "@libs/logWriter";
+import axios, { AxiosError, AxiosResponse } from "axios";
+
+type Status = "PASS" | "FAIL" | "ERROR";
+type StatusResult = Record<string, Status>;
```
seidior, line 5: [sand/boulder] I don't love the result of these being string operations. Ideally you'd have one of these two different return types:

- 0 (C's `EXIT_SUCCESS`) for success, non-0 (C's `EXIT_FAILURE`) for failure, so that your exit code can just be the sum of the exit codes for your operation.
- the HTTP status of the operation.

The first makes more sense to me, as the result of your lambda shouldn't be a string, object, or array but rather an exit code. If you wanted to return anything from the lambda, it would in theory be the resultant objects from each of the health checks as a structured object. But you'd still set the exit code.

Can you tell me more about your approach here and why a `Record<string, Status>` is the ideal return from the lambda?

seidior: [sand] Also, opportunity for refactor: I don't love the names `Status` and `StatusResult` here.

seidior (commented):
Great first pass so far, but some opportunities for Refactor Friday.

PR #7457, opened 2024-03-29
Dashboard page structure2

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/186004286

### Steps to Test
I ended up doing a pretty large reactor to split out the components that render in desktop view versus mobile view for the dashboard. I also duplicated all of the tests so that they are there for both desktop and mobile views, what was nice about this was that I was able to extract the majority of the test from the page itself and move the test to the individual components.

There is a refactor on the circular component that is used for displaying loading state. I noticed that we were using different padding and felt that we could consolidate.

## Code author checklist


web/src/components/dashboard/DashboardOnDesktop.tsx
```diff
@@ -0,0 +1,128 @@
...
+      if (isDesktopAndUp && business?.preferences.phaseNewlyChanged) {
+        if (!updateQueue || business?.onboardingFormProgress !== "COMPLETED") {
+          return;
+        }
+        await updateQueue.queuePreferences({ phaseNewlyChanged: false }).update();
+      }
+    })();
+  }, updateQueue);
+
+  const displayHomedBaseBusinessQuestion = (): boolean => {
```
seidior, line 61: [sand] Did we mean for this to be called `displayHomeBasedBusinessQuestion()` instead?

web/src/components/dashboard/DashboardOnDesktop.tsx
```diff
@@ -0,0 +1,128 @@
...
+  operateReferences: Record<string, OperateReference>;
+  fundings: Funding[];
+  certifications: Certification[];
+}
+
+export const DashboardOnDesktop = (props: Props): ReactElement => {
+  const { business, updateQueue } = useUserData();
+  const router = useRouter();
+  const operatingPhase = LookupOperatingPhaseById(business?.profileData.operatingPhase);
+  const isDesktopAndUp = useMediaQuery(MediaQueries.desktopAndUp);
```
seidior, line 47: [boulder] If it's `DashboardOnDesktop`, why do we still need a `useMediaQuery()` call?

web/src/components/dashboard/DashboardOnDesktop.tsx
```diff
@@ -0,0 +1,128 @@
...
+  return (
+    <div className="display-none desktop:display-block" data-testid="desktop">
+      <RightSidebarPageLayout
+        mainContent={
+          <div className="margin-top-0 desktop:margin-top-0">
+            <UserDataErrorAlert />
+            <DashboardHeader />
+            <div className="margin-top-3">
+              {displayHomedBaseBusinessQuestion() && (
+                <div className="margin-bottom-4">
```
seidior, line 79: [dust] I'd probably extract everything displayed as a result of the true conditional into its own component. This mirrors the behavior we see later with the `<QuickActionsContainer>`.

web/src/styles/sections/dashboard.scss
```diff
@@ -1,3 +1,5 @@
 .blueRightGutter {
-  background: linear-gradient(to right, $white 70%, $cool-extra-light 70% 100%);
+  @media (min-width: 64em) {
```
seidior, line 2: [sand] Media query should go outside the selector rather than inside.

web/src/components/PageCircularIndicator.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface Props {
+  displayText?: string;
+}
+export const PageCircularIndicator = (props?: Props): ReactElement => {
+  return (
+    <div className="margin-top-3 desktop:margin-top-0 padding-top-0 desktop:padding-top-6 padding-bottom-15">
+      <div className="flex flex-justify-center flex-align-center" data-testid={"loading"}>
+        <CircularProgress aria-label="loading indicator" aria-busy={true} />
+        <Heading level={0} styleVariant="h3" style={{ marginBottom: 0, marginLeft: "1rem" }}>
```
seidior, line 13: [dust] Nice use of `level={0}` here.

seidior (commented):
Overall, great refactor! I think it makes things like `config` a lot more straightforward.

web/src/pages/export.tsx
```diff
@@ -22,11 +22,11 @@ const ExportPage = (): ReactElement => {
     if (!FEATURE_EXPORT_PDF) router.push(ROUTES.dashboard);
   }, [router, FEATURE_EXPORT_PDF]);
 
-  if (!FEATURE_EXPORT_PDF || !roadmap || !business) return <CircularIndicator />;
+  if (!FEATURE_EXPORT_PDF || !roadmap || !business) return <PageCircularIndicator />;
```
cbrunefearless, line 25: nice

cbrunefearless (approved):
lgtm!!! good work

web/src/components/PageCircularIndicator.tsx
FarazA22: changing to a div

web/src/components/dashboard/DashboardOnDesktop.tsx
FarazA22: There is a function inside of a use effect hook that should only be run if the screen size is desktop. The useMediaQuery is required. See <DashboardOnDesktop />

FarazA22: moving to parent container

PR #7473, opened 2024-04-04
Callouts

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/187200130
https://www.pivotaltracker.com/story/show/187200069
https://www.pivotaltracker.com/story/show/187200176
https://www.pivotaltracker.com/story/show/187200152

### Approach
Create a callout widget and remove all instances of the greenbox widget. There will be a follow-up PR to remove other ways the greenbox is rendered (i.e. blockquotes and use of greenbox css).

I manually checked all the MD changed in the CMS (local), the CMS did it's thing and made some additional changes to the markdown files.

## Code author checklist


web/src/components/Callout.tsx
```diff
@@ -0,0 +1,77 @@
...
+  headingDefaultText: string;
+  bodyStyling: string;
+  iconStyling: string;
+}
+
+export const calloutComponentMapping: Record<string, CalloutMappingObj> = {
+  informational: {
+    containerStyling: "bg-accent-cool-lightest",
+    headingStyling: "text-accent-cool-more-dark",
+    headingDefaultText: "Keep in mind:",
```
jphechter, line 24: [sand] If content intends to use the default texts, for all variants, widely as a default we may want it to be available in the CMS.

web/src/components/Callout.tsx
```diff
@@ -0,0 +1,77 @@
...
+    headingStyling: "text-accent-warm-darker",
+    headingDefaultText: "Be advised:",
+    bodyStyling: "text-primary-darker",
+    iconStyling: "callout-warning-icon",
+  },
+  note: {
+    containerStyling: "bg-base-lightest",
+    headingStyling: "text-base-darkest",
+    headingDefaultText: "Note:",
+    bodyStyling: "text-primary-darker",
```
jphechter, line 46: [sand] All of these instances of `bodyStyling` are the same. Should we just add `text-primary-darker` directly to the element?

web/src/components/Callout.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <div className="flex">
+        {props.icon === "true" && (
+          <div
+            data-testid={calloutComponentMapping[props.calloutType].iconStyling}
+            className={calloutComponentMapping[props.calloutType].iconStyling}
+            aria-hidden="true"
+          />
+        )}
+        <span className={`text-bold ${calloutComponentMapping[props.calloutType].headingStyling}`}>
+          {props.header && props.header?.length > 0
```
jphechter, line 67: [pebble] This will do the same thing
```suggestion
          {!!props.header
```

web/src/components/Content.test.tsx
```diff
@@ -44,13 +44,6 @@ describe("<Content />", () => {
     });
   });
 
-  it("renders greenBox content", () => {
-    const mdString = ":::greenBox\n" + "body text\n" + ":::";
-
-    render(<Content>{mdString}</Content>);
-    expect(screen.getByText("body text")).toBeInTheDocument();
-  });
-
```
jphechter, line 53: [sand] Should this get replaced with a test that checks the `<Callout />` gets substituted?

web/src/components/Content.tsx
```diff
@@ -54,12 +55,17 @@ export const Content = (props: ContentProps): ReactElement => {
       return <HorizontalLine />;
     },
     blockquote: GreenBox,
```
jphechter, line 57: Not your fault, but this tripped me up that there's a component declared in this file called `GreenBox` that's different from the `greenBox` we've been talking about.

web/src/components/Callout.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const Callout = (props: Props): ReactElement => {
+  return (
+    <div
+      className={`padding-205 radius-md margin-bottom-2 ${
+        calloutComponentMapping[props.calloutType].containerStyling
+      }`}
+    >
+      <div className="flex">
+        {props.icon === "true" && (
```
jphechter, line 59: [pebble] I don't like that we doing a string comparison here. Let's turn this into a boolean. Then the check can just be:
```suggestion
        {props.icon && (
```

web/src/components/Callout.tsx
```diff
@@ -0,0 +1,77 @@
+import { ReactElement } from "react";
+
+export type CalloutTypes = "informational" | "conditional" | "warning" | "note";
+
+interface Props {
+  calloutType: CalloutTypes;
+  header?: string;
+  children: string;
+  icon?: "true" | "false";
```
jphechter, line 9:
```suggestion
  icon: boolean;
```

web/src/components/Content.tsx
```diff
@@ -54,12 +55,17 @@ export const Content = (props: ContentProps): ReactElement => {
...
-    greenBox: (props: any): ReactElement => {
-      return <GreenBox>{props.children}</GreenBox>;
-    },
     note: (props: any): ReactElement => {
       return <Alert variant="note">{props.children}</Alert>;
     },
+    callout: (props: any): ReactElement => {
+      return (
+        <Callout header={props.header} icon={props.icon} calloutType={props.calloutType}>
```
jphechter, line 63: [sand] Let's turn `icon` into a boolean. You could do something like: 
```suggestion
        <Callout header={props.header} icon={props.icon === "true"} calloutType={props.calloutType}>
```

web/src/components/Content.test.tsx
FarazA22: I don't think so, the green box syntax is getting deleted all together

web/src/components/Content.tsx
FarazA22: it will get deleted once the clean up work is done

PR #7478, opened 2024-04-08
feat: [#187354337] use cards for elevator registration task

SirSamuelJoseph:
## Description
Update display of elevator registration status returns to use carded designs

### Ticket

[187354337](https://www.pivotaltracker.com/story/show/187354337)

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/elevator-registration.json
+++ b/content/src/fieldConfig/elevator-registration.json
@@ -4,9 +4,9 @@
     "errorTextNoPropertyInterestFound": "We cannot find your registration. Make sure the building address you entered matches what is in your application.",
     "errorTextNoElevatorRegistrations": "We cannot find your registration. Make sure you submit a registration application for your elevator device(s).",
     "errorTextSearchFailed": "Sorry, something went wrong. Please try again later.",
-    "registrationRequestDateText": "Application requested: ",
+    "registrationRequestDateText": "Application Requested on ",
     "registrationStatusText": "Status: ",
-    "registrationDeviceCountText": "Total devices in this application: ",
+    "registrationDeviceCountText": "Elevator devices in this application: ",
     "registrationApplicationFoundText": "Great, we've found your application(s)! Please check the details below.",
     "registrationFoundInReviewText": "Check back later to make sure your elevator device(s) has been successfully registered",
     "registrationFoundIncompleteReturnedText": "Review your incomplete or returned application(s) for any errors or missing information. If your application was returned, you can resubmit it using the same application link in the DCA Service Portal.",
@@ -18,9 +18,13 @@
     "registrationRejectedOrReturnedMessage": "Review your incomplete or returned application(s) for any errors or missing information. If your application was returned, you can resubmit it using the same application link in the DCA Service Portal.",
     "registrationSearchPrompt": "Enter your building address exactly as it appears on your application to get automatic updates about the registration status of your elevator device(s) below.",
     "address1Label": "Building Street Address Line 1",
-    "address2Label": "Business Address Line 2 (Optional)",
+    "address2Label": "Building Address Line 2 (Optional)",
     "zipCodeLabel": "ZIP Code",
     "stateLabel": "State",
-    "submitText": "Submit"
+    "submitText": "Submit",
+    "informationalInReview": "Check back later to make sure your elevator device(s) are successfully registered.",
+    "informationalReturned": "Review your returned application for any errors or missing information. You can resubmit it using the same application link in the DCA Service Portal.",
+    "informationalRejected": "Review your rejected application for any errors or missing information. You can resubmit it using the same application link in the DCA Service Portal.",
+    "informationalIncomplete": "Review your incomplete application. Make sure you complete all sections and submit it."
   }
 }
--- a/web/public/mgmt/config.yml
+++ b/web/public/mgmt/config.yml
@@ -2169,6 +2169,10 @@ collections:
                 }
               - { label: Error - Missing Required Fields, name: errorTextFieldsRequired, widget: string }
               - { label: Error - Search Failed, name: errorTextSearchFailed, widget: string }
+              - { label: Informational Message - In Review, name: informationalInReview, widget: string }
+              - { label: Informational Message - Returned, name: informationalReturned, widget: string }
+              - { label: Informational Message - Rejected, name: informationalRejected, widget: string }
+              - { label: Informational Message - Incomplete, name: informationalIncomplete, widget: string }
 
   - name: "tasks"
     label: "Tasks - All"
--- a/web/src/components/tasks/ElevatorRegistrationStatusSummary.tsx
+++ b/web/src/components/tasks/ElevatorRegistrationStatusSummary.tsx
@@ -1,9 +1,11 @@
 import { HorizontalLine } from "@/components/HorizontalLine";
 import { UserDataErrorAlert } from "@/components/UserDataErrorAlert";
 import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
+import { Icon } from "@/components/njwds/Icon";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { useUserData } from "@/lib/data-hooks/useUserData";
 import { Task } from "@/lib/types/types";
+import { parseDate } from "@businessnjgovnavigator/shared/dateHelpers";
 import {
   ElevatorSafetyAddress,
   ElevatorSafetyRegistrationSummary,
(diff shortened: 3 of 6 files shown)


web/src/components/tasks/ElevatorRegistrationStatusSummary.tsx
```diff
@@ -23,49 +25,86 @@ export const ElevatorRegistrationStatusSummary = (props: Props): ReactElement =>
...
-    for (const reg of props.summary.registrations) {
-      statuses.push(reg.status);
+  type CardDetails = {
+    headerColor: string
+    bodyColor: string;
+    icon: string;
+    iconTextColor: string;
+  }
+
+  const getDetailsForRegistrationCard = (status: string): CardDetails => {
```
SirSamuelJoseph, line 35: I know multiple switches over the same bits of information is a bit of an anti-pattern, but I couldn't find a better way of doing it.

The first switch statement covers general formatting. The second switch statement handles messages only relevant to certain statuses. The third switch statement only covers a unique case with a single status and is related to a special formatting case. 

If anyone has ideas on a better way to organize this, I'm all ears

web/src/lib/async-content-fetchers/fetchTaskByFilename.ts
```diff
@@ -48,7 +48,12 @@ const fetchTaskFile = async (filename: string): Promise<string> => {
     try {
       file = await import(`@businessnjgovnavigator/content/roadmaps/tasks/${filename}.md`);
     } catch {
-      file = await import(`@businessnjgovnavigator/content/roadmaps/license-tasks/${filename}.md`);
+      try {
+        file = await import(`@businessnjgovnavigator/content/roadmaps/license-tasks/${filename}.md`);
+      } catch {
+        file = await import(`@businessnjgovnavigator/content/roadmaps/municipal-tasks/${filename}.md`);
```
SirSamuelJoseph, line 54: This was missed in adding the new municipal tasks directory

content/src/fieldConfig/elevator-registration.json
```diff
@@ -4,9 +4,9 @@
     "errorTextNoPropertyInterestFound": "We cannot find your registration. Make sure the building address you entered matches what is in your application.",
     "errorTextNoElevatorRegistrations": "We cannot find your registration. Make sure you submit a registration application for your elevator device(s).",
     "errorTextSearchFailed": "Sorry, something went wrong. Please try again later.",
-    "registrationRequestDateText": "Application requested: ",
+    "registrationRequestDateText": "Application Requested on ",
```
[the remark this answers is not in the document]
SirSamuelJoseph: Not for this one! The designs don't have it, it's sentence case for this title now

content/src/fieldConfig/elevator-registration.json
```diff
@@ -21,6 +21,10 @@
     "address2Label": "Business Address Line 2 (Optional)",
     "zipCodeLabel": "ZIP Code",
     "stateLabel": "State",
-    "submitText": "Submit"
+    "submitText": "Submit",
+    "informationalInReview": "Check back later to make sure your elevator device(s) are successfully registered",
+    "informationalReturned": "Review your returned application for any errors or missing information. You can resubmit it using the same application link in the DCA Service Portal",
+    "informationalRejected": "Review your rejected application for any errors or missing information. You can resubmit it using the same application link in the DCA Service Portal",
```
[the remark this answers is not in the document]
SirSamuelJoseph: I'll add these, good catch!

web/src/components/tasks/ElevatorRegistrationStatusSummary.tsx
SirSamuelJoseph: I considered that (the `CardDetails` object was sort of starting this) but I think that would be harder to read/maintain. In that case, we would have to lay out every possible status instead of being able to batch them and make good use of the `default` case. 

But I'm not opposed to it if that's what the consensus is!

AnJuHyppolite: > I considered that (the `CardDetails` object was sort of starting this) but I think that would be harder to read/maintain. In that case, we would have to lay out every possible status instead of being able to batch them and make good use of the `default` case.
> 
> But I'm not opposed to it if that's what the consensus is!

For sure, which is why I understand going the `switch` route. I think I would have done the same thing. I was just thinking about your comment and thought of a map, because it seems to be the approach when someone doesn't want to use a switch. 

I'm excited about this convo! 😄

web/public/mgmt/config.yml
```diff
@@ -2133,6 +2133,10 @@ collections:
               - { label: Error - No Elevator Registrations Found, name: errorTextNoElevatorRegistrations, widget: string }
               - { label: Error - Missing Required Fields, name: errorTextFieldsRequired, widget: string }
               - { label: Error - Search Failed, name: errorTextSearchFailed, widget: string }
+              - { label: Informational Message - In Review, name: informationalInReview, widget: string }
```
FarazA22, line 2136: Have you added the following fields to the search.tsx file>

web/src/components/tasks/ElevatorRegistrationStatusSummary.tsx
```diff
@@ -23,49 +25,110 @@ export const ElevatorRegistrationStatusSummary = (props: Props): ReactElement =>
 
   const { business } = useUserData();
 
-  const getInformationalMessageText = (): string => {
-    const statuses: string[] = [];
-    for (const reg of props.summary.registrations) {
-      statuses.push(reg.status);
-    }
+  type CardDetails = {
```
FarazA22, line 28: [dust] There is a types file in the lib folder in web, we should store this type in that file and be more specific with the name. 

On the other hand, I'm now wondering if we need the types file at all. What do you think?

web/src/components/tasks/ElevatorRegistrationTask.test.tsx
```diff
@@ -121,6 +121,35 @@ describe("<ElevatorRegistrationTask />", () => {
       expect(screen.getByTestId("registration-2")).toBeInTheDocument();
       expect(screen.queryByTestId("registration-3")).not.toBeInTheDocument();
     });
+
+    it("shows informational messages for only relevant statuses", async () => {
```
FarazA22, line 125: [pebble] - we've typically tried to separate test so that one test will assert one thing. When we are making changes to the application, it'll be a lot easier to figure out which test are failing and confirm which one should be deleted, or updated to make them passing. What are your thoughts on separating them out?

web/src/styles/base/margin-padding.scss
```diff
@@ -27,3 +27,8 @@
 .mla {
   margin-left: auto !important;
 }
+
+.bottom-radius-lg {
```
FarazA22, line 31: dust - it's kind of odd that the border at the bottom has a radius, but the border at the top does not. From the other designs that I've seen, there is a radius applied across the entire element. Separately, we do try to use USWDS for radiuses can we use one of those instead? i.e. radius-bottom-lg

https://designsystem.digital.gov/utilities/border/#border-radius

web/src/components/tasks/ElevatorRegistrationStatusSummary.tsx
FarazA22: If you're going to keep it inside the file, perhaps it should live outside of the function?

content/src/fieldConfig/elevator-registration.json
```diff
@@ -21,6 +21,10 @@
     "address2Label": "Business Address Line 2 (Optional)",
     "zipCodeLabel": "ZIP Code",
     "stateLabel": "State",
-    "submitText": "Submit"
+    "submitText": "Submit",
+    "informationalInReview": "Check back later to make sure your elevator device(s) are successfully registered",
```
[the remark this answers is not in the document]
SirSamuelJoseph: Right now this is the content I was given but I'll bring it up with them!

web/src/styles/base/margin-padding.scss
SirSamuelJoseph: I can pull from USWDS, thanks for the callout! As far as only the bottom having padding, there are two separate parts, and the border between them does not have a border-radius.

web/src/components/tasks/ElevatorRegistrationStatusSummary.tsx
SirSamuelJoseph: This type is just a convenience to better organize what card display settings  for what status, I don't think it needs to exist outside of this file. I will rename it to be more clear though

web/public/mgmt/config.yml
SirSamuelJoseph: I've added a refactor ticket to do this and add the preview

PR #7489, opened 2024-04-09
feat: [#186003964] home page support card headers are 1 header line

adidner:
## Description

We wanted to change the home page support cards (at the bottom of the page). The headers on these cards were previously 2 headers actually to get them to look a certain way. We wanted to maintain the look while getting everything under one header for screen reader accessibility purposes. 

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186003964)

## Code author checklist


web/src/components/UserSupportActionCard.tsx
```diff
@@ -27,12 +26,14 @@ export const UserSupportActionCard = (props: Props): ReactElement => {
...
-          </Heading>
-          <Heading level={3} className="margin-bottom-3">
-            {props.headerLine2}
+        <div className="text-center flex flex-column fac">
+          <Heading
+            level={3}
+            className={`margin-0 ${
+              props.header.length < 25 ? "width-card" : "width-card-lg"
+            } support-action-card`}
+          >
```
adidner, line 35: [dust] - So this is a little brittle, but it was the only way I found to get the same behavior with all the cards and titles where the 3rd and longest card is only on 2 lines and the other shorter cards also have their titles spread across 2 lines. 

Also tried 

```
<Heading>
{title part 1}
<br/>
{title part 2}
</Heading>
```
but this is read poorly by the screen readr

web/public/mgmt/config.yml
```diff
@@ -3754,18 +3799,15 @@ collections:
...
-              - { label: Card 2 - Header Line 2, name: card2HeaderLine1, widget: string }
-              - { label: Card 2 - Header Line 2, name: card2HeaderLine2, widget: string }
+              - { label: Card 2 - Header Line 2, name: card2Header, widget: string }
               - { label: Card 2 - Supporting Text, name: card2SupportingText, widget: text }
               - { label: Card 2 - Button Text, name: card2Button, widget: string }
               - { label: Card 2 - Button Link, name: card2ButtonLink, widget: string }
-              - { label: Card 3 - Header Line 3, name: card3HeaderLine1, widget: string }
-              - { label: Card 3 - Header Line 3, name: card3HeaderLine2, widget: string }
+              - { label: Card 3 - Header Line 3, name: card3Header, widget: string }
               - { label: Card 3 - Supporting Text, name: card3SupportingText, widget: text }
```
adidner, line 3811: [dust] - here is the only real changes in the file, all the rest is formatting which happened without my meddling

content/src/fieldConfig/config.json
```diff
@@ -75,14 +75,11 @@
     "landingPageTileRunBizTile": "I'm running a NJ business",
     "card1ButtonLink": "https://business.nj.gov/",
     "section2SupportingText": "See the forms and licenses specific to starting your business and avoiding fines. Be prepared to tell us your business structure, industry, and business location.",
-    "card3HeaderLine1": "Chat with a New Jersey",
-    "card3HeaderLine2": "Representative",
+    "card3Header": "Chat with a New Jersey Representative",
```
somabadri, line 78: [dust] thoughts on renaming the card[NUMBER]Header fields to something more descriptive?

somabadri (approved):

web/src/components/UserSupportActionCard.tsx
jphechter: @adidner I'm pretty sure design and content said they don't need this to be _exactly_ the same. Might be worth checking, but if memory serves we can just let the browser wrap it however it wants based on div width.

web/public/mgmt/config.yml
```diff
@@ -2114,23 +2114,59 @@ collections:
               - { label: Registration Request Date Text, name: registrationRequestDateText, widget: string }
               - { label: Registration Status Text, name: registrationStatusText, widget: string }
               - { label: Registration Device Count Text, name: registrationDeviceCountText, widget: string }
-              - { label: Registration Application Found Text, name: registrationApplicationFoundText, widget: string }
-              - { label: Registration Found In Review Text, name: registrationFoundInReviewText, widget: string }
-              - { label: Registration Found Incomplete/Returned Ext, name: registrationFoundIncompleteReturnedText, widget: string }
-              - { label: Registration Call to Action Primary, name: registrationCallToActionPrimaryText, widget: string }
-              - { label: Registration Call to Action Secondary, name: registrationCallToActionSecondaryText, widget: string }
+              - {
+                  label: Registration Application Found Text,
```
FarazA22, line 2118: dust - consider having these updates in a separate commit?

web/src/styles/components/card.scss
```diff
@@ -28,3 +28,11 @@
   font-size: 0.88rem;
   font-weight: 600;
 }
+
+.usa-prose > h3,
+.usa-prose > .h3-styling,
+.h3-styling,
+h3,
+.support-action-card {
```
FarazA22, line 36: pebble - all of the typography has been set. I'd confirm with Design about the line height change. I don't see any updates in the design system. I think it might be an oversight. If there is a global change to the H3 styling, I'd update the line height in typography.scss. Right now you are overriding the line height for all H3s, and it's noted in card.scss

adidner: So I merged, without re-freshing my page so I didn't see these comments until after. 

but you're totally right that this applies to wayyy more than I wanted it to. (I got my CSS selectors mixed up).

That being said its not really a change to design actually becuase now we're using line height to make something happen that was previously 2 elements with padding. Lmk if that doesn't sense and I can clarify what I mean better. 

image
vs
image

PR #7492, opened 2024-04-11
chore: [#186665645] make api fail

cbrunefearless:
## Description
Make the health check api fail in dev and prod

### Ticket

[#186665645](https://www.pivotaltracker.com/story/show/)

## Code author checklist

### DIFF
--- a/api/src/libs/healthCheck.ts
+++ b/api/src/libs/healthCheck.ts
@@ -5,13 +5,13 @@ type Status = "PASS" | "FAIL" | "ERROR";
 type StatusResult = Record<string, Status>;
 
 const healthCheckEndPoints: Record<string, string> = {
-  self: "self2",
-  elevator: "dynamics/elevator",
-  fireSafety: "dynamics/fire-safety",
-  housing: "dynamics/housing",
-  dynamicsLicenseStatus: "dynamics/license-status",
-  webserviceLicenseStatus: "webservice/license-status",
-  webserviceFormation: "webservice/formation",
+  self: "selfError",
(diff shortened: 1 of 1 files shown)


api/src/libs/healthCheck.ts
```diff
@@ -5,13 +5,13 @@ type Status = "PASS" | "FAIL" | "ERROR";
...
 
 const healthCheckEndPoints: Record<string, string> = {
-  self: "self2",
-  elevator: "dynamics/elevator",
-  fireSafety: "dynamics/fire-safety",
-  housing: "dynamics/housing",
-  dynamicsLicenseStatus: "dynamics/license-status",
-  webserviceLicenseStatus: "webservice/license-status",
-  webserviceFormation: "webservice/formation",
+  self: "self2Error",
```
seidior, line 8: [pebble] This is not needed. `self2` was already the change that broke the self-check, since the self-check path is `self`, not `self2`.

seidior (approved):
Two questions:

- Is this to check the fail state or the error state?
- Is it necessary to make all of them error, or just one?

api/src/libs/healthCheck.ts
cbrunefearless: done

jphechter (approved):
This looks good to me.

Given we'll have another API we're consuming soon, can we add documentation for how to add resources to the health checks?

cbrunefearless:
> This looks good to me.
> 
> Given we'll have another API we're consuming soon, can we add documentation for how to add resources to the health checks?

yes sir, it will be easy to add

PR #7496, opened 2024-04-11
feat: [#186316063] revisions to account for the case where we want a snackbar after page redirect

adidner:
## Description

We had a weird case where the TaskProgress Checkbox triggers a redirect and thus we needed to use our redirect alerts instead of our page alerts. 

### Ticket

## Code author checklist

### DIFF
--- a/web/public/mgmt/config.yml
+++ b/web/public/mgmt/config.yml
@@ -3120,8 +3120,8 @@ collections:
                   widget: text,
                 }
               - {
-                  label: Opportunities and Deadlines Snackbar Time Delay,
-                  name: opportunitiesAndDeadlinesTimeDelay,
+                  label: Time Delay Between Multiple Snackbars (Time In Milliseconds),
+                  name: timeDelayBetweenMultipleSnackbars,
                   widget: text,
                 }
       - label: "Dashboard Calendar"
@@ -4168,7 +4168,8 @@ collections:
               - { label: Page Title, name: pageTitle, widget: string }
               - { label: Back Button Text, name: backToRoadmapText, widget: string }
               - { label: Default Call to Action Test, name: defaultCallToActionText, widget: string }
-              - { label: Task Progress - Prefix, name: taskProgressSnackbarPrefix, widget: string }
+              - { label: Task Progress Alert - Body Prefix, name: taskProgressSnackbarPrefix, widget: string }
+              - { label: Task Progress Alert - Heading, name: taskProgressSnackbarHeading, widget: string }
               - { label: Next Task Button Text, name: nextTaskButtonText, widget: string }
               - { label: Previous Task Button Text, name: previousTaskButtonText, widget: string }
               - { label: Unlocked By - Singular, name: unlockedBySingular, widget: string }
--- a/web/src/components/TaskProgressCheckbox.tsx
+++ b/web/src/components/TaskProgressCheckbox.tsx
@@ -94,7 +94,7 @@ export const TaskProgressCheckbox = (props: Props): ReactElement => {
     updateQueue
       .update()
       .then(() => {
-        setSuccessSnackbarIsOpen(true);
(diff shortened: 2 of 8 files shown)


web/src/components/TaskProgressCheckbox.tsx
```diff
@@ -94,7 +94,7 @@ export const TaskProgressCheckbox = (props: Props): ReactElement => {
     updateQueue
       .update()
       .then(() => {
-        setSuccessSnackbarIsOpen(true);
+        if (!(isFormationTask(props.taskId) && nextStatus === "COMPLETED")) setSuccessSnackbarIsOpen(true);
```
adidner, line 97: So we need this because when formationCheckbox is set to complete we re-direct to the dashboad and the dashboard has its own code to make the same snackbar appear. So this would just flash the snackbar before the page redirect

Also it would queue the text in the screen reader and the snackbar would be read twice.

web/public/mgmt/config.yml
```diff
@@ -3120,8 +3120,8 @@ collections:
                   widget: text,
                 }
               - {
-                  label: Opportunities and Deadlines Snackbar Time Delay,
-                  name: opportunitiesAndDeadlinesTimeDelay,
+                  label: Time Delay Between Multiple Snackbars,
+                  name: timeDelayBetweenMultipleSnackbars,
```
jphechter, line 3124: [sand] Add a descriptor here to indicate what this value represents.

Might be worth adding a note as well. If we're surfacing this in the CMS then the content team should know what this is.
```suggestion
                  label: Time Delay Between Multiple Snackbars (time in milliseconds),
                  name: timeDelayBetweenMultipleSnackbars,
```

jphechter: I'd check with them for specific messaging for the note.

web/src/components/TaskProgressCheckbox.tsx
jphechter: Should we add a test to ensure this behavior?

somabadri (approved):

web/src/components/dashboard/DashboardAlerts.tsx
```diff
@@ -23,6 +24,16 @@ export const DashboardAlerts = (): ReactElement => {
...
   });
 
+  const CheckboxStatusUpdated = useQueryControlledAlert({
+    queryKey: QUERIES.fromFormBusinessEntity,
+    pagePath: ROUTES.dashboard,
+    headerText: Config.taskDefaults.taskProgressSnackbarHeading,
+    bodyText: getTaskStatusUpdatedMessage("COMPLETED"),
+    variant: "success",
+    dataTestId: "checkbox-status-updated-snackbar-alert",
+    delayInMilliseconds: Config.dashboardDefaults.timeDelayBetweenMultipleSnackbars,
```
jphechter, line 34: [dust] I might set this at the top and reference it below.

```typescript
const delayBetweenMultipleSnackbars = Config.dashboardDefaults.timeDelayBetweenMultipleSnackbars;
 

const CheckboxStatusUpdated = useQueryControlledAlert({
  ...
  delayInMilliseconds: delayBetweenMultipleSnackbars;
});


const DeadlinesAndOpportunitiesAlert = useQueryControlledAlert({
  ...
  delayInMilliseconds: delayBetweenMultipleSnackbars;
});
```

adidner: That makes sense

web/src/components/TaskProgressCheckbox.tsx
SirSamuelJoseph: I believe that's what the added test is testing for, unless I'm reading it wrong. It's testing to make sure that the snackbar still appears on the dashboard if you are coming from the formation task. 

Unless you're saying a test to make sure it doesn't appear twice?

adidner: No JP is right, so we have a test to test for what you described @SirSamuelJoseph but we don't have a test to make sure we don't display the snackbar on the task page before you move to the dashboard.

PR #7511, opened 2024-04-15
Return first matching license from Dynamics license API

jphechter:
## Description

We display license data to users using multiple APIs as a source. Regardless of the data source, the logic should be as close to the same as possible. MLO currently returns the first matching license, but Dynamics will return an error if more than once license is found. This work removes that error and the associated check which results in a single license being returned.

### Ticket

[#187423161](https://www.pivotaltracker.com/story/show/187423161)

### Approach

Simply removes the error, the associated check for multiple licenses, and updates the associated tests.

### Steps to Test

License data coming from the Dynamics API should still work. Possibly more license lookups will also result in data being displayed on the frontend.

### Notes

This is not an exact way for us to display license data, but it is incrementally better than what currently exists and is consistent between data sources.

## Code author checklist

### DIFF
--- a/api/src/client/dynamics/license-status/DynamicsLicenseApplicationIdClient.test.ts
+++ b/api/src/client/dynamics/license-status/DynamicsLicenseApplicationIdClient.test.ts
@@ -1,5 +1,5 @@
 /* eslint-disable unicorn/no-null */
-import { MULTIPLE_MAIN_APPS_ERROR, NO_MAIN_APPS_ERROR } from "@domain/types";
+import { NO_MAIN_APPS_ERROR } from "@domain/types";
 import { LogWriter, LogWriterType } from "@libs/logWriter";
 import axios from "axios";
 import { DynamicsLicenseApplicationIdClient } from "./DynamicsLicenseApplicationIdClient";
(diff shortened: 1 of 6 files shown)


seidior (approved):
- cannot find dangling instances of `MULTIPLE_MAIN_APPS_ERROR`
- ran pre-commit hooks
- tested locally without test failures

I'd love in the future if you can add specific steps to test in the "Steps to test" section of your PR, especially for those less experienced with Wiremock.

Further, do we need any documentation updates for this?

jphechter:
@seidior 

> I'd love in the future if you can add specific steps to test in the "Steps to test" section of your PR, especially for those less experienced with Wiremock.

Are you thinking specifically how to configure this using the Dynamics API information or do you mean explaining how to get to a task that allows the user to check their license?

> Further, do we need any documentation updates for this?

I don't think so. Not really sure what documentation would come from this. If you have something in mind though, I'm happy to update.

seidior:
@jphechter 

> Are you thinking specifically how to configure this using the Dynamics API information or do you mean explaining how to get to a task that allows the user to check their license?

Specifically, the minimum number of steps needed to verify the changes in the PR are working as expected, whether via Postman or running the app or whatever. If the team split is going to work, we'll need to share more context in the PR, especially given the abbreviated AC in the story.

PR #7548, opened 2024-04-22
feat: [#187476824] profile side navigation doesn't shrink to wrap

adidner:
### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/187476824)

### Approach

I decided to do this using flex. 

The idea is we flex-shrink 0 on the nav section meaning it can't shrink below the size of its content. We set a flex grow 1 on this section so it will grow, and the right nav section we set a flex grow of 10 so it will grow much more and take more of the space. 

## Code author checklist

### DIFF
--- a/web/src/components/njwds-layout/SidebarPageLayout.tsx
+++ b/web/src/components/njwds-layout/SidebarPageLayout.tsx
@@ -12,6 +12,7 @@ export interface SidebarPageLayoutProps {
   outlineBox?: boolean;
   belowBoxComponent?: React.ReactNode;
   titleOverColumns?: React.ReactNode;
+  nonWrappingLeftColumn?: boolean;
 }
 
 export const SidebarPageLayout = ({
@@ -22,6 +23,7 @@ export const SidebarPageLayout = ({
   stackNav,
   belowBoxComponent,
   titleOverColumns,
+  nonWrappingLeftColumn,
 }: SidebarPageLayoutProps): ReactElement => {
   const isLargeScreen = useMediaQuery(MediaQueries.desktopAndUp);
   const { Config } = useConfig();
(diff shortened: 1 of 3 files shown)


web/src/components/njwds-layout/SidebarPageLayout.tsx
```diff
@@ -53,14 +53,15 @@ export const SidebarPageLayout = ({
               </div>
               {belowBoxComponent}
             </div>
-            <div className="desktop:grid-col-3 order-first">
-              {isLargeScreen && (
+
+            {isLargeScreen && (
+              <div className="fs0 fg1 order-first">
```
jphechter, line 58: [dust] Do we need this `div`? Could these classes just be applied directly to the `nav` element?

web/src/styles/base/flex.scss
```diff
@@ -88,6 +88,14 @@
   flex-grow: 1;
 }
 
+.fg10 {
+  flex-grow: 10;
```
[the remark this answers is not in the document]
jphechter: If there is an option for us to use `.flex-10` I'd be for it, but it does look like there's a class `fg1` directly above this. Leads me to think it's not an option.

jphechter (approved):

web/src/components/njwds-layout/SidebarPageLayout.tsx
```diff
@@ -22,6 +23,7 @@ export const SidebarPageLayout = ({
   stackNav,
   belowBoxComponent,
   titleOverColumns,
+  nonWrappingLeftColumn,
 }: SidebarPageLayoutProps): ReactElement => {
   const isLargeScreen = useMediaQuery(MediaQueries.desktopAndUp);
```
seidior, line 28: [dust] How much of a lift would it be to remove this `useMediaQuery()` as part of this ticket?

PR #7566, opened 2024-04-23
feat: [#187485886] save and retrieve community affairs address on ele…

SirSamuelJoseph:
…vator registration task

### Ticket

## Code author checklist

- [ ] If I added any new collections to the CMS config, then I updated the search tool and `cmsCollections.ts` (see [CMS Additions in Engineering Reference/FAQ](https://docs.google.com/document/d/1X4iWSGmBZdHYZ0jGqkwTR3_yyyqI_TiiJptfc8pi9Po/edit#heading=h.fu5jdsrcqxbh))


api/src/db/migrations/v132_add_community_affairs_address.ts
```diff
@@ -0,0 +1,483 @@
...
+  sectorId: string | undefined;
+  naicsCode: string;
+  foreignBusinessTypeIds: ForeignBusinessTypeId[];
+  nexusDbaName: string;
+  needsNexusDbaName: boolean;
+  operatingPhase: OperatingPhaseId;
+  isNonprofitOnboardingRadio: boolean;
+  nonEssentialRadioAnswers: Record<string, boolean | undefined>;
+  elevatorOwningBusiness: boolean | undefined;
+  communityAffairsAddress?: CommunityAffairsAddress;
```
SirSamuelJoseph, line 34: This is the new field added

shared/src/housing.ts
```diff
@@ -7,8 +9,8 @@ export type HousingPropertyInterestDetails = {
   stateCode: number;
 };
 
-export type HousingMunicipality = {
-  name: string;
-  id: string;
-  county: string;
+export type CommunityAffairsAddress = {
```
SirSamuelJoseph, line 12: The previous municipality type was not being used, so it was deleted. The diff makes it looks like a drop-in replacement but `CommunityAffairsAddress` is actually a new type

web/src/components/tasks/CheckElevatorRegistrationStatus.tsx
```diff
@@ -38,8 +40,38 @@ export const CheckElevatorRegistrationStatus = (props: Props): ReactElement => {
     } as Municipality;
   });
 
+  useEffect(() => {
+    if (business?.profileData.communityAffairsAddress?.streetAddress1 && formValues.address1 === "") {
+      formValues.address1 = business.profileData.communityAffairsAddress.streetAddress1;
+    }
+    if (business?.profileData.communityAffairsAddress?.streetAddress2 && formValues.address2 === "") {
+      formValues.address2 = business.profileData.communityAffairsAddress.streetAddress2;
```
adidner, line 48: pebble - Shouldn't you use setFormValues here? Isn't it bad practice to set values in a useState variable directly?

adidner: For both this if statement and the one above it

web/src/components/tasks/CheckElevatorRegistrationStatus.tsx
```diff
@@ -38,8 +40,38 @@ export const CheckElevatorRegistrationStatus = (props: Props): ReactElement => {
...
+    if (business?.profileData.communityAffairsAddress?.municipality && !selectedMunicipality) {
+      setSelectedMunicipality(business.profileData.communityAffairsAddress.municipality);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
   const onSubmit = (event: FormEvent<HTMLFormElement>): void => {
     event.preventDefault();
+
+    if (business?.profileData && formValues.address1 && selectedMunicipality) {
```
adidner, line 59: Should we also check address2?

web/src/components/tasks/CheckElevatorRegistrationStatus.tsx
```diff
@@ -38,8 +40,38 @@ export const CheckElevatorRegistrationStatus = (props: Props): ReactElement => {
...
+          streetAddress2: formValues.address2,
+          municipality: {
+            id: selectedMunicipality?.id,
+            name: selectedMunicipality?.name,
+            displayName: selectedMunicipality?.name,
+            county: selectedMunicipality?.county,
+          },
+        },
+      };
+      updateQueue?.queueProfileData(profileData).update();
```
adidner, line 73: So based on the way this is written I assume the intention is to only save if you have both address1 and selectedMunicipality selected as opposed to any one field filled out and then saved in isolation. 

I just wanted to double check that this is desired functionality.

adidner (approved):
Approved with a question and a pebble

web/src/components/tasks/CheckElevatorRegistrationStatus.tsx
```diff
@@ -38,8 +40,38 @@ export const CheckElevatorRegistrationStatus = (props: Props): ReactElement => {
...
+    if (business?.profileData.communityAffairsAddress?.streetAddress1 && formValues.address1 === "") {
+      formValues.address1 = business.profileData.communityAffairsAddress.streetAddress1;
+    }
+    if (business?.profileData.communityAffairsAddress?.streetAddress2 && formValues.address2 === "") {
+      formValues.address2 = business.profileData.communityAffairsAddress.streetAddress2;
+    }
+    if (business?.profileData.communityAffairsAddress?.municipality && !selectedMunicipality) {
+      setSelectedMunicipality(business.profileData.communityAffairsAddress.municipality);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
seidior, line 53: [dust] I know this is our pattern, but we should look into fixing this rather than disabling the eslint rule. It might be a good idea for a lunch and learn.

web/src/components/tasks/CheckElevatorRegistrationStatus.tsx
SirSamuelJoseph: You're totally right, I should not be directly touching these values! Update to use the setter function

web/src/components/tasks/CheckElevatorRegistrationStatus.tsx
SirSamuelJoseph: Would love to discuss this further in a LnL!

PR #7590, opened 2024-05-01
feat: [#186897265]change registered agent office from dropdown to field text

AnJuHyppolite:
## Description

As a user completing the registrations process with a manually entered Registered Agent I want to be able to manually enter the "Office City" within the address section so that I do not have to select my city from the options which are available in the dropdown menu.

### Ticket

[#186897265](https://www.pivotaltracker.com/story/show/#186897265)

### Steps to Test

In the Contacts Step, select the `Manually Identify Your Registered Agent` radio button. You can enter the agent's name, email, Office Address Line 1 as usual.

The Office City field will render an error under these conditions:
- It is empty.
- It contains an input of 30 characters or more. 

The error disappears once there is input AND the input is 30 characters or less. 

## Code author checklist


shared/src/formationData.ts
```diff
@@ -139,7 +139,7 @@ export interface FormationFormData extends FormationAddress {
   readonly agentEmail: string;
   readonly agentOfficeAddressLine1: string;
   readonly agentOfficeAddressLine2: string;
-  readonly agentOfficeAddressMunicipality: Municipality | undefined;
+  readonly agentOfficeAddressMunicipality: string;
   readonly agentOfficeAddressCity: string | undefined;
```
jphechter, line 143: What's the difference between these 2 fields?

web/src/components/tasks/business-formation/getErrorStateForField.ts
```diff
@@ -129,6 +126,8 @@ export const getErrorStateForField = (inputParams: {
     "nonprofitBoardMemberRightsSpecified",
     "nonprofitTrusteesMethodSpecified",
     "nonprofitAssetDistributionSpecified",
+    "paymentType",
+    "businessSuffix",
```
jphechter, line 130: This is a great example of leaving things better than you found it. 🙌

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
```diff
@@ -1119,7 +1129,7 @@ describe("getErrorStateForField", () => {
   });
 
   describe("fields that have error when undefined", () => {
-    const hasErrorIfUndefined: FormationFields[] = [
+    const onlyHasErrorIfUndefined: FormationFields[] = [
```
jphechter, line 1132: [pebble] In `getErrorStateForField.ts` we added `paymentType` and `businessSuffix`. Those need to be in this list.

Instead of trying to keep `onlyHasErrorIfUndefined` and `onlyHasErrorIfEmpty` up to date in 2 places, could we export from `getErrorStateForField.ts` and use a single implementation in both places?

api/src/db/migrations/v133_change_agent_office_address_municipality.ts
```diff
@@ -0,0 +1,488 @@
...
+};
+
+const migrate_v132Business_to_v133Business = (business: v132Business): v133Business => {
+  return {
+    ...business,
+    formationData: {
+      ...business.formationData,
+      formationFormData: {
+        ...business.formationData.formationFormData,
+        agentOfficeAddressMunicipality: "",
```
FarazA22, line 57: boulder - why are we clearing the field for all existing users?

content/src/fieldConfig/business-formation.json
```diff
@@ -250,7 +250,7 @@
       },
       "agentOfficeAddressMunicipality": {
         "label": "Office City",
-        "error": "Select a city."
+        "error": "Enter a city."
```
FarazA22, line 253: Pebble - All other error states for registered agent contain the term agent, this does not. Designs should have been provided. I'd check with Content on what the content should be.

shared/src/formationData.ts
FarazA22: Municipality is an object, we are not replacing it with a string

web/src/components/tasks/business-formation/contacts/RegisteredAgent.tsx
```diff
@@ -40,7 +36,7 @@ export const RegisteredAgent = (): ReactElement => {
         agentUseBusinessAddress &&
         (agentOfficeAddressLine1 !== addressLine1 ||
           agentOfficeAddressLine2 !== addressLine2 ||
-          agentOfficeAddressMunicipality?.name !== addressMunicipality?.name ||
+          agentOfficeAddressMunicipality !== addressMunicipality?.displayName ||
```
FarazA22, line 39: good catch

web/src/components/tasks/business-formation/contacts/RegisteredAgent.tsx
```diff
@@ -266,27 +262,14 @@ export const RegisteredAgent = (): ReactElement => {
...
                       <FormationField fieldName="agentOfficeAddressMunicipality">
-                        <MunicipalityDropdown
-                          municipalities={municipalities}
+                        <BusinessFormationTextField
+                          label={Config.formation.fields.agentOfficeAddressMunicipality.label}
                           fieldName="agentOfficeAddressMunicipality"
-                          error={doesFieldHaveError("agentOfficeAddressMunicipality")}
+                          required={true}
+                          validationText={getFieldErrorLabel("agentOfficeAddressMunicipality")} // renders Office City
                           disabled={shouldBeDisabled("agentOfficeAddressMunicipality", "ADDRESS")}
```
FarazA22, line 271: pebble - why did you decide to keep this prop?

web/src/components/tasks/business-formation/contacts/RegisteredAgent.tsx
```diff
@@ -266,27 +262,14 @@ export const RegisteredAgent = (): ReactElement => {
...
-                      </strong>
                       <FormationField fieldName="agentOfficeAddressMunicipality">
-                        <MunicipalityDropdown
-                          municipalities={municipalities}
+                        <BusinessFormationTextField
+                          label={Config.formation.fields.agentOfficeAddressMunicipality.label}
                           fieldName="agentOfficeAddressMunicipality"
-                          error={doesFieldHaveError("agentOfficeAddressMunicipality")}
+                          required={true}
+                          validationText={getFieldErrorLabel("agentOfficeAddressMunicipality")} // renders Office City
```
FarazA22, line 270: pebble - remove comment

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
FarazA22: I agree there should be only list.

FarazA22 (changes requested):
I've reviewed the PR

api/src/db/migrations/v133_change_agent_office_address_municipality.ts
FarazA22: Can you share your though process/ reasoning?

api/src/client/ApiFormationHealth.ts
```diff
@@ -134,12 +134,7 @@ export const ApiFormationHealth: UserData = {
           agentOfficeAddressLine1: "agent-office-addr-1-94810479",
           agentOfficeAddressLine2: "agent-office-addr-2-93634915",
           agentOfficeAddressCity: "agent-office-addr-city-31370791",
-          agentOfficeAddressMunicipality: {
-            displayName: "some-display-name-19185415",
-            name: "some-name-32477899",
-            county: "some-county-35537788",
-            id: "some-id-51671099",
-          },
+          agentOfficeAddressMunicipality: "municipality-123",
```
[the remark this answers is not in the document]
FarazA22: updates should be made across the board

web/src/components/tasks/business-formation/contacts/RegisteredAgent.tsx
FarazA22: referring do the disabled

AnJuHyppolite: Oversight. Thanks for catching it!

content/src/fieldConfig/business-formation.json
AnJuHyppolite: Changed per Content's confirmation.

shared/src/formationData.ts
AnJuHyppolite: @Faraz Akhtar, just verifying that you meant **now** _NOT_ **not** in this comment.

api/src/client/ApiFormationHealth.ts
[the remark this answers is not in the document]
FarazA22: update should be made in migration file as well

api/src/db/migrations/v133_change_agent_office_address_municipality.ts
FarazA22: this will need to be re-reviewed

web/src/components/tasks/business-formation/contacts/RegisteredAgent.tsx
FarazA22: needs to be confirmed with product.. missed A/C

shared/src/formationData.ts
FarazA22: I deleted it, JP's question is valid

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
```diff
@@ -1119,23 +1133,8 @@ describe("getErrorStateForField", () => {
...
-      "addressCountry",
-      "addressState",
-      "willPracticeLaw",
-      "isVeteranNonprofit",
-      "hasNonprofitBoardMembers",
-      "nonprofitBoardMemberQualificationsSpecified",
-      "nonprofitBoardMemberRightsSpecified",
-      "nonprofitTrusteesMethodSpecified",
-      "nonprofitAssetDistributionSpecified",
-    ];
```
jphechter, line 1135: Looking at this with fresh eyes got me thinking...

Originally I suggested exporting the arrays `onlyHasErrorIfUndefined / Empty` from `getErrorStateForField` to handle for possible inconsistencies. There were missing fields in the test file that we weren't checking for and doing this ensured that we wouldn't miss anything.

The problem with this approach is that the test no longer validates a rule for a given field. If the value of the array is mistakenly changed in the `getErrorStateForField.ts` file, there is nothing in the test alerting that the change was wrong. EX: I accidentally delete `addressCountry`, all tests will still be green.

The value of having these defined separately in the test is the safeguard to hopefully prevent changes.

_That said,_ simply redefining these in the test file brings us back to the original problem where the values could be mismatched and doesn't account for potentially missing fields. What probably should have been there from the beginning is a test to validate that the arrays in both files match. (`onlyHasErrorIfUndefined` in the function and `onlyHasErrorIfUndefined` in the test are equal).

jphechter (approved):
1 note you can choose to include, but overall this works.

api/src/db/migrations/v133_change_agent_office_address_municipality.ts
```diff
@@ -0,0 +1,489 @@
...
+const migrate_v132Business_to_v133Business = (business: v132Business): v133Business => {
+  const agentOfficeAddressCityAlreadyExists =
+    business.formationData.formationFormData.agentOfficeAddressMunicipality?.displayName;
+  return {
+    ...business,
+    formationData: {
+      ...business.formationData,
+      formationFormData: {
+        ...business.formationData.formationFormData,
+        agentOfficeAddressCity: agentOfficeAddressCityAlreadyExists || "",
```
FarazA22, line 59: pebble - this migration is not correct. `business.formationData.formationFormData` still contains `agentOfficeAddressMunicipality` and you're adding it back by using the spread operator

web/cypress/e2e/business-formation.spec.ts
```diff
@@ -254,12 +254,12 @@ const selectAndTypeRegisteredAgent = ({
         .should("contain", agentOfficeAddressLine2);
     }
 
-    if (agentOfficeAddressMunicipality) {
-      onBusinessFormationPage.selectRegisteredAgentMunicipality(agentOfficeAddressMunicipality.displayName);
+    if (agentOfficeAddressCity) {
+      onBusinessFormationPage.typeRegisteredAgentMunicipality(agentOfficeAddressCity);
```
FarazA22, line 258: pebble - update the function name - `typeRegisteredAgentMunicipality` is not correct b/c we removed Municipality

web/cypress/support/page_objects/businessFormationPage.ts
```diff
@@ -57,7 +57,7 @@ class BusinessFormationPage {
     return cy.get('input[name="agentOfficeAddressLine2"]');
   }
   getRegisteredAgentMunicipality() {
```
FarazA22, line 59: pebble - update to `getRegisteredAgentCity`?

web/src/components/tasks/business-formation/BusinessFormation.test.tsx
```diff
@@ -296,7 +296,7 @@ describe("<BusinessFormation />", () => {
     page.fillText("Address line1", "1234 main street");
     page.fillText("Address line2", "Suite 304");
     page.fillText("Address zip code", "08001");
-    page.selectByText("Address municipality", "Newark");
+    page.fillText("Address municipality", "Newark");
```
FarazA22, line 299: boulder - Why is this method being changed? Isn't it a part of th main business address?

web/src/components/tasks/business-formation/BusinessFormation.test.tsx
```diff
@@ -521,7 +521,7 @@ describe("<BusinessFormation />", () => {
...
```
FarazA22, line 524: same question here... I thought we are only updating the logic for registered agent.. does this ticket now include the other city field?

web/src/components/tasks/business-formation/contacts/RegisteredAgent.test.tsx
```diff
@@ -45,6 +45,7 @@ describe("Formation - Registered Agent Field", () => {
   });
 
   it("auto-fills fields when agent is manual entry", async () => {
+    const randomAgentOfficeAddressCity = `agent-city-${randomInt()}`;
```
FarazA22, line 48: pebble - in the other staget data example, we are not using the `randomInt` func, why use it? If your using it, why should we selectively use it?

FarazA22: it seems like this pattern is used in other test.. we should be consistent.

web/cypress/e2e/business-formation.spec.ts
AnJuHyppolite: I made this change. Perhaps it changed back when I rebased.

api/src/db/migrations/v133_change_agent_office_address_municipality.ts
jphechter: I didn't notice that. Good catch.

Other similar migrations have a test around this kind of logic. v130 is a good example of this.

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
AnJuHyppolite: 100%!!! I added two tests.

api/src/db/migrations/v133_change_agent_office_address_municipality.ts
AnJuHyppolite: Updated!

api/src/db/migrations/v133_change_agent_office_address_city_field.ts
```diff
@@ -0,0 +1,494 @@
...
+  const agentOfficeAddressCityAlreadyExists =
+    business.formationData.formationFormData.agentOfficeAddressMunicipality?.displayName;
+
+  const v133BusinessObj = {
+    ...business,
+    formationData: {
+      ...business.formationData,
+      formationFormData: {
+        ...business.formationData.formationFormData,
+        agentOfficeAddressCity: agentOfficeAddressCityAlreadyExists || "",
```
FarazA22, line 60: pebble - this just hit me during the re-review, the `agentOfficeAddressCity` key previously existed in user data. I think we need to consider the value that is stored within the key. Should the existing value be stored during the migration, it looks like we are overwriting it with the value stored in `agentOfficeAddressMunicipality` or an empty string. I don't think it was previously considered

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
```diff
@@ -23,6 +27,39 @@ import {
 const Config = getMergedConfig();
 
 describe("getErrorStateForField", () => {
+  const onlyHasErrorIfUndefinedTest: FormationFields[] = [
```
FarazA22, line 30: pebble - if we are exporting the two array's from `getErrorStateForField`, why are we creating them here again? It doesn't make sense to have duplicate logic

FarazA22: Im some of the test you are referencing the imported arrays, in other tests, you are referencing the arrays you've created here

web/src/components/tasks/business-formation/getErrorStateForField.test.ts
```diff
@@ -1253,6 +1264,16 @@ describe("getErrorStateForField", () => {
       }
     };
 
-    runTests(hasErrorIfEmpty);
+    runTests(onlyHasErrorIfEmptyTest);
+  });
+
+  describe("confirm field names in test and getErrorStateForField component match", () => {
+    it("onlyHasErrorIfEmpty", () => {
+      expect(onlyHasErrorIfEmpty).toEqual(onlyHasErrorIfEmptyTest);
```
FarazA22, line 1272: pebble - if your importing the arrays, why not reference them and the individual test as opposed to creating an entirely different array, using them in the test, and then having another test to verify that they match the imported arrays. What's the value of the current approach?

api/src/db/migrations/v133_change_agent_office_address_city_field.ts
FarazA22: It would be helpful to get an understanding of the findings, perpahs a quick call once you have an idea on the best decision forword.

FarazA22 (approved):

PR #7595, opened 2024-05-01
chore: [#186948602] updated webflow + netlify sync scripts to handle …

adidner:
…rate limit

We were running into issues where running the scripts for webflow netlify sync were hitting the webflow API rate limit. 

This PR resolves that by resolving promises individually and waiting for their responses before making more calls and checking the rate limit remaining header to make sure we don't have to wait (and waiting if we do)

### Ticket

[Replace with ticket_id](https://www.pivotaltracker.com/story/show/186948602)

### Steps to Test

You can manually run the scripts (they're located in package.json web)

yarn workspace @businessnjgovnavigator/web webflow:licenses-sync

## Code author checklist

### DIFF
--- a/web/src/scripts/webflow/fundingSync.mjs
+++ b/web/src/scripts/webflow/fundingSync.mjs
@@ -1,7 +1,14 @@
 /* eslint-disable unicorn/no-process-exit */
 /* eslint-disable no-undef */
 import { loadAllFundings } from "../fundingExport.mjs";
-import { argsInclude, contentToStrings, getHtml, wait } from "./helpers.mjs";
+import {
+  argsInclude,
+  catchRateLimitErrorAndRetry,
+  contentToStrings,
+  getHtml,
+  resolveApiPromises,
+} from "./helpers.mjs";
+import { wait } from "./helpers2.mjs";
 import { createItem, deleteItem, getAllItems, getCollection, modifyItem } from "./methods.mjs";
 import { allIndustryId, getCurrentSectors, syncSectors } from "./sectorSync.mjs";
 import { certificationCollectionId, fundingCollectionId } from "./webflowIds.mjs";
@@ -251,15 +258,14 @@ const deleteFundings = async () => {
     try {
       return await deleteItem(funding.id, fundingCollectionId);
     } catch (error) {
-      console.error(error.response.data);
-      throw error;
+      await catchRateLimitErrorAndRetry(error, deleteItem, funding.id, fundingCollectionId);
     }
   };
-  await Promise.all(
-    fundings.map(async (item) => {
-      return await deleteFunding(item);
-    })
-  );
+
+  const fundingPromises = fundings.map((item) => {
+    return () => deleteFunding(item);
+  });
+  await resolveApiPromises(fundingPromises);
 };
 
 const updateFundings = async () => {
@@ -278,16 +284,15 @@ const updateFundings = async () => {
     try {
       return await modifyItem(item.id, fundingCollectionId, funding);
     } catch (error) {
-      console.error(error);
-      throw error;
+      await catchRateLimitErrorAndRetry(error, modifyItem, item.id, fundingCollectionId, funding);
     }
   };
 
-  return Promise.all(
-    overlappingFundings.map(async (item) => {
-      return await modify(item);
-    })
-  );
+  const fundingsPromises = overlappingFundings.map((item) => {
+    return () => modify(item);
+  });
+
+  await resolveApiPromises(fundingsPromises);
 };
 const createNewFundings = async () => {
   const newFundings = await getNewFundings();
@@ -296,25 +301,23 @@ const createNewFundings = async () => {
     try {
       return await createItem(funding, fundingCollectionId, false);
     } catch (error) {
-      console.error(error.response.data);
-      throw error;
+      await catchRateLimitErrorAndRetry(error, createItem, funding, fundingCollectionId, false);
     }
   };
-  return await Promise.all(
-    newFundings.map(async (item) => {
-      return await create(item);
-    })
-  );
+
+  const fundingPromises = newFundings.map((item) => {
+    return () => create(item);
+  });
+
+  await resolveApiPromises(fundingPromises);
 };
 
 const syncFundings = async () => {
   console.log("deleting unused fundings");
   await deleteFundings();
   console.log("updating fundings");
-  await wait();
   await updateFundings();
   console.log("creating new fundings");
-  await wait();
   await createNewFundings();
   console.log("Complete Funding Sync!");
 };
--- a/web/src/scripts/webflow/helpers.mjs
+++ b/web/src/scripts/webflow/helpers.mjs
@@ -6,17 +6,51 @@ import remarkGfm from "remark-gfm";
 import remarkParse from "remark-parse";
 import remarkRehype from "remark-rehype";
 import { unified } from "unified";
+import { wait } from "./helpers2.mjs";
 
-export const wait = (milliseconds = 10000) => {
-  return new Promise((resolve) => {
-    return setTimeout(resolve, milliseconds);
-  });
-};
+const RATE_LIMIT_WAIT_SECONDS = 20;
 
 export const argsInclude = (query) => {
   return process.argv.some((i) => {
     return i.includes(query);
   });
+};
+
+export const checkRateLimitAndWait = async (response) => {
+  if (
+    response &&
+    response.headers &&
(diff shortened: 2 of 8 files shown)


web/src/scripts/webflow/fundingSync.mjs
```diff
@@ -279,15 +286,20 @@ const updateFundings = async () => {
       return await modifyItem(item.id, fundingCollectionId, funding);
     } catch (error) {
       console.error(error);
+      if (error.response.status === 429){
+        console.debug("Catching rate limit error hit")
+        await wait(65000);
+        return await modifyItem(item.id, fundingCollectionId, funding);
+      }
```
adidner, line 293: I ended up handling the errors inline everywhere. 

I realized that because we want to retry and recall this function we woudl still need to pass the function (in this case modifyItem and all its params)

so we could have done something like this 

```
async function handleRateLimitError(func, ...params) {
   if (error.response && error.response.status === 429) {
      console.debug("Catching rate limit error hit");
      await wait(65000);
      return await func(...params);
    }
}


try {
  return await modify( item.id, fundingCollectionId, funding);
} catch (error) {
  console.debug({ status: error.response.status, type: typeof error.response.status });
  await handleRateLimitError(modify, item.id, fundingCollectionid, funding);
  throw error;
}
```

adidner: I think this would work but something about it kinda bothers me. I can do it if people feel strongly.

web/src/scripts/webflow/fundingSync.mjs
```diff
@@ -252,14 +253,20 @@ const deleteFundings = async () => {
       return await deleteItem(funding.id, fundingCollectionId);
     } catch (error) {
       console.error(error.response.data);
+      if (error.response.status === 429){
+        console.debug("Catching rate limit error hit")
+        await wait(65000);
```
jphechter, line 258: Why 65000? Buffer?

web/src/scripts/webflow/fundingSync.mjs
jphechter: Given how many times you're repeating roughly the same logic, this would be a good opportunity to make a function. It's not a common pattern for us, but I don't think there's anything wrong with it. I'm in the camp of make-it-a-function.

jphechter: At the very least, you should make `65000` a global const.

web/src/scripts/webflow/helpers.mjs
```diff
@@ -7,6 +7,8 @@ import remarkParse from "remark-parse";
 import remarkRehype from "remark-rehype";
 import { unified } from "unified";
 
+const WAIT_AMOUNT = 20000;
```
jphechter, line 10: Why is this one 20000?

jphechter (commented):
Overall, this looks like it should work fine. I'd like to see some of these changes consolidated in a way that we don't have to repeat logic in so many places.

web/src/scripts/webflow/fundingSync.mjs
seidior: Buffer. But yes, we should probably call this out as a global const.

web/src/scripts/webflow/helpers.mjs
seidior: It depends on when the counter resets. In theory, we wouldn't have to wait a full 60 seconds to send additional requests because we don't necessarily need to send 120 more requests. So, to save on our CircleCI dollars, it's probably best we keep this arbitrarily short, but an amount longer than 60 / 4 = 15 seconds. (since we only have, theoretically 4-5 requests to spare before triggering the rate limit.)

web/src/scripts/webflow/helpers.mjs
```diff
@@ -19,6 +21,28 @@ export const argsInclude = (query) => {
...
+export const checkRateLimitAndWait = async (response) => {
+  if (response && response.headers && response.headers["x-ratelimit-remaining"]) {
+
+    if (response && response.headers && response.headers["x-ratelimit-remaining"] <= 5){
+      console.debug({"retry-after": response.headers["retry-after"]});
+      console.debug(`Rate limit close; waiting ${WAIT_AMOUNT / 1000} seconds...`);
+      await wait(WAIT_AMOUNT);
+    }
+  } else {
+    console.debug("Couldn't find x-ratelimit-remaining header. Here are the headers I've found.", response.headers);
```
seidior, line 34: [pebble] I don't think we need this anymore.

web/src/scripts/webflow/fundingSync.mjs
```diff
@@ -296,15 +308,21 @@ const createNewFundings = async () => {
     try {
       return await createItem(funding, fundingCollectionId, false);
     } catch (error) {
+      if (error.response.status === 429){
```
seidior, line 311: [pebble] This should have been caught by automatic code formatting: there should be a space between `)` and `{`. Are you formatting the code before commit?

web/src/scripts/webflow/fundingSync.mjs
seidior: [boulder] Yes, please factor out. Also, you'll need to pass the error, or at least the response, to `handleRateLimitError()`. Also, as discussed, I'd prefer if the `throw error` occurred in the `handleRateLimitError()` instead, on the logic branch where `error.response.status` was not 429.

seidior (commented):
It's a great start! I'll take a more fine-tooth comb on re-review.

web/src/scripts/webflow/helpers.mjs
seidior: Specifically this `console.debug()` line on line 34 above.

web/src/scripts/webflow/fundingSync.mjs
adidner: It should do that automatically right? Our commit script should run prettier I believe. Maybe mine isn't working for some reason? Or maybe it doesn't run on the .mjs files?

web/src/scripts/webflow/helpers2.mjs
```diff
@@ -0,0 +1,5 @@
+export const wait = (milliseconds = 10000) => {
```
jphechter, line 1: [sand] Can we call this file something other than `helpers2`? Even just `wait.mjs` I think would make more sense.

seidior (approved):
I'm not a huge fan of the name "helpers2.mjs", but we can always rename it later. Thanks for this! I'm so excited to try it out tomorrow.

web/src/scripts/webflow/helpers.mjs
```diff
@@ -6,19 +6,53 @@ import remarkGfm from "remark-gfm";
...
+    await checkRateLimitAndWait(response);
+  }
+};
+
+export const catchRateLimitErrorAndRetry = async (error, retryFunc, ...params) => {
+  console.debug("in Catch Rate and retry");
+  if (error.response.status === 429) {
+    console.debug("Catching rate limit error hit");
+    await wait(65000);
+    // eslint-disable-next-line no-useless-catch
```
jphechter, line 43: [dust] I don't think this `eslint` rule is necessary given we're actually throwing the error.

web/src/scripts/webflow/webflowSync.test.js
```diff
@@ -261,6 +263,12 @@ jest.mock("fs", () => {
   };
 });
 
+jest.mock("./helpers2.mjs", () => {
+  return {
+    wait: jest.fn(),
+  };
+});
+
 describe("webflow syncing", () => {
```
jphechter, line 272: It would probably be helpful to add tests that confirm that `catchRateLimitErrorAndRetry` is called if the API does respond with something other than `200`. You can take or leave this now, but in a re-write this should for sure be covered.

adidner: Isn't that the second test I wrote? "should retry function after waiting if rate limit error occurs" or did you mean something else?

adidner: Either way I'm going to deal with it during a Friday as a sand

web/src/scripts/webflow/helpers2.mjs
adidner: Yeah, I"ll deal with it in a friday Sand ticket.

web/src/scripts/webflow/helpers.mjs
adidner: I'm pretty sure it is. 

the reason is becasue it thinks that this try catch is useless becasause it can't recognize the retryFunc has an API call inside of it.

PR #7977, opened 2024-06-03
feat: [#187420711]: Set up both testing and content apis

aadedejifearless:
## Description
As part of our solution to migrate `testing.navigator` to `testing.account` and both `content.navigator` to `content.account` , we are updating our the testing and content environment variables in `config.yml` to ensure that the api associated with the testing resource  API endpoints in the newly created. 
- `businessnjgov-api-testing`
- `businessnjgov-api-content` 
get created. This change would be verified in deployment. 
- the `Serverless.ts`  is also updated to have reference the parameters created for both content and testing in the ssm store. 
-

### Ticket

This pull request resolves [#187420711](https://www.pivotaltracker.com/story/show/187420711).

### Approach
- Updates the `config.yml` by updating the `set-accessibility-testing-env-vars` section of the config.yml

### Steps to Test

- This change was verified in the `Accessiblity-testing-repo` testing branch's CI job. This can be referenced[ here ](https://app.circleci.com/pipelines/github/newjersey/navigator.business.nj.gov?branch=accessibility-testing-repo).
- 

## Code author checklist


api/serverless.ts
```diff
@@ -237,13 +237,22 @@ const serverlessConfiguration: AWS = {
...
+          infrastructure: "${ssm:/config/infrastructure}",
+        },
+      }
+    : {
+        ...serverlessConfiguration.custom,
+        config: {
+          application: "${ssm:/${self:provider.stage}/config/application}",
+          infrastructure: "${ssm:/config/infrastructure}",
+        },
+      };
```
jphechter, line 255: We were adding this as a short-term fix to get the testing site up. Do we actually want this included going forward?

.circleci/config.yml
```diff
@@ -381,20 +381,22 @@ commands:
...
+            echo 'export REDIRECT_URL=$REDIRECT_URL_CONTENT' >> $BASH_ENV
 
   set-accessibility-testing-env-vars:
     steps:
       - run:
           name: Set Accessibility Testing Environment Variables
           command: |
+            echo 'export COGNITO_WEB_CLIENT_ID=$COGNITO_WEB_CLIENT_ID_ACCESSIBILITY_TESTING' >> $BASH_ENV
+            echo 'export COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_ACCESSIBILITY_TESTING' >> $BASH_ENV
+            echo 'export COGNITO_IDENTITY_POOL_ID=$COGNITO_IDENTITY_POOL_ID_ACCESSIBILITY_TESTING' >> $BASH_ENV
```
seidior, line 395: [dust] This is more a question for @jphechter, I suppose, but I don't imagine that we would ever want a different `COGNITO_IDENTITY_POOL_ID` (and other Cognitos) for testing and content that is distinct from dev, no?

.circleci/config.yml
```diff
@@ -381,20 +381,22 @@ commands:
...
             echo 'export REDIRECT_URL=$REDIRECT_URL_ACCESSIBILITY_TESTING' >> $BASH_ENV
 
   deploy-content:
     steps:
-      - configure-and-build
-      - setup_remote_docker:
-          version: edge
-          docker_layer_caching: true
```
seidior, line 397: [dust] Do we no longer need to set up Docker before running `package-webapp-docker` later in the script?

seidior (commented):
Closer and closer! Just a couple comments for you in addition to JP's comment. Thanks for fighting through all this; it's taking an extraordinary amount of patience.

.circleci/config.yml
jphechter: I'd need to sign into AWS to look where the lines of the service end... but don't we need this if we're creating a new connection to myNJ? That'd be the only reason. I think right now they're actually all the same. I'm cool with this if it's a preemptive step.

.circleci/config.yml
aadedejifearless: @seidior So we are still setting up Docker. However because we run it in the deploy job prior to `deploy-content` and `deploy-accessibility-testing`  which is redundant and we get an error thrown as well in the build. See Error below: 
 ```More than one setup_remote_docker is not valid, please adjust in job , , accessibility-testing-deploy,```

api/serverless.ts
aadedejifearless: We can discuss this sometime tomorrow. But i think there are ways we can go about this. We would have to update the current config/application across the 3 environments to reference the stages so that the change here would be minimal. (Meaning we update the /config/application in dev to be /dev/config/application). This would also apply to staging and prod

.circleci/config.yml
aadedejifearless: https://app.circleci.com/pipelines/github/newjersey/navigator.business.nj.gov/27281

adidner (commented):
This seems reasonable though it should be noted that I am very much out of my depth on this.

jphechter (approved):

PR #8016, opened 2024-06-12
chore: rename reverseInMobile prop on action bar layout component

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/187780095

### Approach
I renamed the reverseInMobile prop to disableReverseOrderInMobile.
I reversed the rendering logic for this prop in the component so the reversing is ignored when optional prop is true
When used if original value was reverseInMobile=TRUE, I added the prop. If it was false I did not add the prop.

## Code author checklist

### DIFF
--- a/web/src/components/njwds-extended/cta/SingleCtaLink.tsx
+++ b/web/src/components/njwds-extended/cta/SingleCtaLink.tsx
@@ -20,7 +20,7 @@ export const SingleCtaLink = (props: Props): ReactElement => {
 
   return (
     <CtaContainer noBackgroundColor={props?.noBackgroundColor || false}>
-      <ActionBarLayout reverseInMobile={false} alignLeft={props?.alignLeft || false}>
+      <ActionBarLayout disableReverseOrderInMobile alignLeft={props?.alignLeft || false}>
         <PrimaryButton
           isColor={props?.buttonColor || "primary"}
           isRightMarginRemoved={true}
--- a/web/src/components/onboarding/OnboardingButtonGroup.tsx
+++ b/web/src/components/onboarding/OnboardingButtonGroup.tsx
@@ -33,7 +33,7 @@ export const OnboardingButtonGroup = (props: Props): ReactElement => {
 
   return (
     <div className="margin-bottom-2">
-      <ActionBarLayout reverseInMobile={true}>
+      <ActionBarLayout>
         {(state.page || 1) > 1 && (
           <div className="margin-top-2 mobile-lg:margin-top-0">
             <SecondaryButton isColor="primary" onClick={back} dataTestId="back">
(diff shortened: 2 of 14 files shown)


jphechter (approved):

web/src/components/njwds-extended/cta/SingleCtaLink.tsx
```diff
@@ -20,7 +20,7 @@ export const SingleCtaLink = (props: Props): ReactElement => {
 
   return (
     <CtaContainer noBackgroundColor={props?.noBackgroundColor || false}>
-      <ActionBarLayout reverseInMobile={false} alignLeft={props?.alignLeft || false}>
+      <ActionBarLayout disableReverseOrderInMobile alignLeft={props?.alignLeft || false}>
```
seidior, line 23: [dust] Any particular reason we don't want to explicitly call out the value as `true` here for readability's sake?

web/src/components/onboarding/OnboardingButtonGroup.tsx
```diff
@@ -33,7 +33,7 @@ export const OnboardingButtonGroup = (props: Props): ReactElement => {
 
   return (
     <div className="margin-bottom-2">
-      <ActionBarLayout reverseInMobile={true}>
+      <ActionBarLayout>
```
seidior, line 36: [dust] Any particular reason we don't call out the value as `false` here for readability's sake?

jphechter: I don't know that I necessarily agree. I thought the whole point was that we wanted this to be default behavior unless explicitly stated otherwise?

adidner: Pro con, Less boiler plate / mainting of things vs clarity / documentation

I'm leaning toward letting it default, less clear but I think it would hard to force people to define the prop consistently and we would end up eventually having a mix of explicit and not.

adidner: I'm good deffering to @FarazA22 as he writes the ticket personally and we don't have precedent yet.

web/src/components/njwds-extended/cta/SingleCtaLink.tsx
FarazA22: Less code can prevent the code from wrapping to a new line. If we want to enforce a value, can we add an eslint rule?

web/src/components/onboarding/OnboardingButtonGroup.tsx
FarazA22: The default design pattern is to reverse the ordering on mobile, if we make the prop required, it will cause us to do a gut check by reviewing the designs and AC. This could prevent design bugs when A/C deviating from default behavior is not called out. 

I think the disabling feature will be sparingly used. But I think it's something that can be easily missed in AC. Should we rely on design identifying it in their review? I prefer it to be optional.

jphechter: I think it's an odd precedent for us to require things that should be otherwise the default and intentional behavior.

PR #8131, opened 2024-07-02
Fixed extra webflow licenses

adidner:
## Description

Somehow we had a bunch of extra licenses in webflow, turns out we didn't have a webflow delete licenses function. 

### Ticket

This pull request resolves [#187809299](https://www.pivotaltracker.com/story/show/187809299).

### Steps to Test

Run the preview script added to package json and see that there are no files to be deleted. 

## Code author checklist

### DIFF
--- a/web/src/scripts/webflow/licenseSync.mjs
+++ b/web/src/scripts/webflow/licenseSync.mjs
@@ -168,7 +168,7 @@ const updateLicenseWithWebflowId = (webflowId, filename) => {
     webflowId: webflowId,
   };
   const stringifiedFile = writeMarkdownString(updatedMdObject);
-  fs.writeFileSync(`${filepath}/${filename}.md`, stringifiedFile, (err) => {
+  fs.writeFileSync(`${filepath}`, stringifiedFile, (err) => {
     if (err) {
       throw err;
     }
@@ -212,7 +212,6 @@ const getUnusedLicenseIds = async () => {
 
   const webflowLicenseArray = await getAllLicensesFromWebflow();
 
-
   return webflowLicenseArray.filter((item) => {
     const hasValueInLocalCmsAndWebflowRemote = localLicensesIds.has(item.id);
     return !hasValueInLocalCmsAndWebflowRemote;
@@ -237,7 +236,6 @@ const deleteLicenses = async () => {
 };
 
 const syncLicenses = async (params) => {
-
   console.log("deleting licenses");
   await deleteLicenses();
   await wait();
@@ -274,7 +272,6 @@ if (process.env.NODE_ENV === "test") {
   })();
 } else if (argsInclude("--preview")) {
   await (async () => {
-
     console.info("---- To be deleted: -----");
     console.info((await getUnusedLicenseIds()).map((it) => `${it.fieldData.name}: ${it.id}`));
     console.info("---- To be created: -----");
--- a/web/src/scripts/webflow/licenseSync.mjs
+++ b/web/src/scripts/webflow/licenseSync.mjs
@@ -20,7 +20,7 @@ import {
 } from "./helpers.mjs";
 import { wait } from "./helpers2.mjs";
 import { LicenseClassificationLookup } from "./licenseClassifications.mjs";
-import { createItem, getAllItems, modifyItem } from "./methods.mjs";
+import { createItem, deleteItem, getAllItems, modifyItem } from "./methods.mjs";
 import { licenseCollectionId } from "./webflowIds.mjs";
 
 export const LookupTaskAgencyById = (id) => {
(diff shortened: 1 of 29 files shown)


jphechter (approved):
I think this looks good. If everything runs, this is good to merge.

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -240,10 +274,12 @@ if (process.env.NODE_ENV === "test") {
...
 } else if (argsInclude("--preview")) {
   await (async () => {
+    console.info("---- To be deleted: -----");
+    console.info((await getUnUsedLicenses()).map((it) => it.fieldData.name));
     console.info("---- To be created: -----");
-    console.info((await getNewLicenses()).map((it) => it.filename));
+    console.info((await getNewLicenses()).map((it) => it.name));
     console.info("---- To be updated: -----");
-    console.info((await getLicensesAlreadyInWebflow()).map((it) => it.filename));
+    console.info((await getLicensesAlreadyInWebflow()).map((it) => it.name));
```
adidner, line 282: I think these fields are more useful as they facilitate actual verification of existence on the web-flow site. IE you can search for these.

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -205,8 +205,42 @@ const createNewLicenses = async () => {
   console.info(`Created a total of ${newLicenses.length} licenses`);
 };
 
+const getUnUsedLicenses = async () => {
+
+  let localLicensesTemp = loadAllLicenses();
+  localLicensesTemp = localLicensesTemp.filter((it) => it.webflowId === undefined);
```
adidner, line 211: So in the function `getNewLicenses` which is used in createLicenses we see similar logic here removing webflow items that don't have webfowId's from consideration here. 

I assume thats because they aren't tracked. I'm wondering why we have licenses that aren't tracked? 

Also the update section follows slightly different logic only pulling from a subset of our MD content rather than doing this filter above, I'm curious as to whether there is a concrete reason for this or not. 

As I expected when running preview now there are no files to be deleted which is good verification for me.

adidner: Okay looked into this slightly further and realized that items without webflow ID's are items that haven't been created yet and thus shouldn't be deleted because they don't exist in webflow.

AnJuHyppolite (approved):
LGTM...This is a nice add, Aaron!

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -205,8 +205,40 @@ const createNewLicenses = async () => {
   console.info(`Created a total of ${newLicenses.length} licenses`);
 };
 
+const getUnUsedLicenses = async () => {
```
seidior, line 208: [pebble] `getUnusedLicenses`

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -205,8 +205,40 @@ const createNewLicenses = async () => {
...
+  const localLicensesRaw = loadAllLicenses();
+  const localLicensesFiltered = localLicensesRaw.filter((it) => it.webflowId === undefined);
```
seidior, line 210: [sand] We can see it's filtered with .filter(). It's more important to denote what you're filtering _for_.

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -205,8 +205,40 @@ const createNewLicenses = async () => {
...
+  const localLicensesIds = new Set(localLicensesFiltered.map((i) => i.webflowId));
```
seidior, line 211: Two things:

- [sand] You're using `i` here and `it` above. I'd pick a better word for both, like `contentItem`.
- [dust] `localLicensesFiltered` appears to be a list of all licenses that have no `webflowId`. But here, we're trying to create a `Set` of all the `webflowId`s of `localLicensesFiltered`. Are you sure you want to do this? Wouldn't the set just be undefined?

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -205,8 +205,40 @@ const createNewLicenses = async () => {
...
+
+  const webflowLicense = await getAllLicensesFromWebflow();
+
+  return webflowLicense.filter((item) => {
+    const haslicense = localLicensesIds.has(item.fieldData.id);
```
seidior, line 216: [dust] I'd love a better name here for `hasLicense` that more accurately describes what value this boolean result provides.

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -205,8 +205,40 @@ const createNewLicenses = async () => {
   console.info(`Created a total of ${newLicenses.length} licenses`);
 };
 
+const getUnUsedLicenses = async () => {
+  const localLicensesRaw = loadAllLicenses();
+  const localLicensesFiltered = localLicensesRaw.filter((it) => it.webflowId === undefined);
+  const localLicensesIds = new Set(localLicensesFiltered.map((i) => i.webflowId));
+
+  const webflowLicense = await getAllLicensesFromWebflow();
```
seidior, line 213: [dust] Perhaps `webflowLicenseArray` that better describes the objects that are returned from that function.

web/src/scripts/webflow/licenseSync.mjs
```diff
@@ -205,8 +205,40 @@ const createNewLicenses = async () => {
...
+  return webflowLicense.filter((item) => {
+    const haslicense = localLicensesIds.has(item.fieldData.id);
+    return !haslicense;
+  });
+};
+
+const deleteLicenses = async () => {
+  const licenses = await getUnUsedLicenses();
+  const deleteLicense = async (license) => {
+    console.info(`Attempting to delete ${license.fieldData.slug}`);
```
seidior, line 224: [sand] As a person doing release or reading through release logs, I'd like to know _why_ a license is being deleted. Otherwise a message like "Attempting to delete cosmetology" would be unnecessarily frightening.

seidior (commented):
Thanks for taking this item on. I appreciate the work so far! A few notes for you:

somabadri (approved):
Good to merge after you address Spencer's comments regarding naming!

web/src/scripts/webflow/licenseSync.mjs
adidner: Lol, don't you think the second comment in more worth of a boulder?

I changed the logic to be the opposite because apparently I can't remember consistently how filter works. Sob

Anyways I believe this should work now and I logged out all the ID's we're going to delete and did some spot checking via global search to make sure those webflow ID's don't appear locally as validation.

PR #8241, opened 2024-07-19
chore: [#187831258] logging with env var in tests

somabadri:
### Ticket

## Code author checklist

- [ ] If I added any new collections to the CMS config, then I updated the search tool and `cmsCollections.ts` (see [CMS Additions in Engineering Reference/FAQ](https://docs.google.com/document/d/1X4iWSGmBZdHYZ0jGqkwTR3_yyyqI_TiiJptfc8pi9Po/edit#heading=h.fu5jdsrcqxbh))


web/src/scripts/webflow/helpers.mjs
```diff
@@ -36,19 +38,19 @@ export const resolveApiPromises = async (promiseArray) => {
 };
 
 export const catchRateLimitErrorAndRetry = async (error, retryFunc, ...params) => {
-  console.debug("in Catch Rate and retry");
+  isNotTestEnv ?? console.debug("in Catch Rate and retry");
```
FarazA22, line 41: curios, what value does adding this variable add? Do devs run code in a test environment? It will render in dev.

FarazA22:
The debug env should be added to the template file and bitwarden

cbrunefearless (commented):
besides Faraz's question, looks good.. great work!

api/src/libs/logWriter.ts
```diff
@@ -39,25 +41,21 @@ export const LogWriter = (groupName: string, logStream: string, region?: string)
 
   const LogError = async (message: string, details?: AxiosError): Promise<void> => {
     try {
-      console.error(`${message} - ${JSON.stringify(details?.toJSON())}`);
+      isNotTestEnv ?? console.error(`${message} - ${JSON.stringify(details?.toJSON())}`);
```
seidior, line 44: [pebble] Did you mean to use `||` here? `isNotTestEnv` will always be a boolean, so it'll never be undefined and therefore never trigger the other half of the nullish coalescing operator.

seidior:
@somabadri What do you think about adding some code like this to a Jest setup file so that we can catch any future console.log (or other console function) calls and error out and prevent this in the future?

```javascript
const originalConsoleMethods = {};

// List of console methods to intercept
const methodsToIntercept = [
  'log', 'error', 'warn', 'info', 'debug', 'trace', 'dir', 'dirxml', 
  'group', 'groupCollapsed', 'groupEnd', 'time', 'timeEnd', 'timeLog', 
  'assert', 'clear', 'count', 'countReset', 'table', 'profile', 'profileEnd'
];

methodsToIntercept.forEach(method => {
  originalConsoleMethods[method] = console[method];
  console[method] = (...args) => {
    originalConsoleMethods[method].apply(console, args);
    throw new Error(`Unexpected call to console.${method}`);
  };
});

function restoreConsoleMethods() {
  methodsToIntercept.forEach(method => {
    console[method] = originalConsoleMethods[method];
  });
}
```

consoleMethodsSetupAndBreakdown.ts
```diff
@@ -0,0 +1,42 @@
...
+  "profile",
+  "profileEnd",
+];
+
+export const throwErrorOnConsoleMethods = () => {
+  for (const method of methodsToIntercept) {
+    originalConsoleMethods[method] = console[method];
+    console[method] = (...args) => {
+      originalConsoleMethods[method].apply(console, args);
+      throw new Error(`Unexpected call to console.${method}`);
```
seidior, line 33: A couple of things here:

- [sand] You may need to (console as any)[method] on L31.
- [sand] Let's actually name this function from lines L31 to L34, pulling it up to the top level of the script for better readability, and let's add a stack trace so people can find their console method call. Something like:

```typescript
const throwErrorForConsoleMethod = (methodName: string): ((...args: any[]) => void) => {
  return (...args: any[]) => {
    const error = new Error(`Usage of console.${methodName} is not allowed. Please use the logging framework.`);
    const stackTrace = error.stack || 'Stack trace is unavailable.';
    console.error(`Forbidden console.${methodName} call detected. Stack trace:\n${stackTrace}`);
    
    // Optionally, you can throw the error to halt execution
    // throw error;
  };
};
```

- [dust] Up to you if you want to throw the error or not. I'll let you make that call.

seidior (commented):
All minor stuff! I'm super excited for this to land. Thanks Soma!

shared/setupTests.ts
```diff
@@ -0,0 +1,9 @@
+import { restoreConsoleMethods, throwErrorOnConsoleMethods } from "../consoleMethodsSetupAndBreakdown";
+
```
FarazA22, line 2: Question, who do we have this file in shared and web?

api/src/client/AirtableUserTestingClient.test.ts
```diff
@@ -32,7 +33,7 @@ describe("AirtableUserTestingClient", () => {
         baseUrl: "some-base-url",
         usersTableName: "some-users-table",
       },
-      logger
+      DEBUG ? logger : DummyLogWriter
```
FarazA22, line 36: What if we centralized this line of code in setupTest.js (define and invoke the set globally). we could mock the implementation and return Logger or dummyLogWriter based on the env variable. The though it not fully thought through. If it can be done, we wouldn't need to do this everytime, we'd need to simply import the module, and the mocking is done in setupTests.

shared/setupTests.ts
somabadri: It's because we have different setups for web and shared

PR #8333, opened 2024-07-31
feat: implement general starter kit seo landing page

alexander-littleton:
## Description

- Adds the general SEO landing page for business starter kits (is static page)

### Ticket

This pull request resolves [#187738760](https://www.pivotaltracker.com/story/show/187738760).

## Code author checklist


web/src/pages/starter-kits/general.tsx
alexander-littleton, line 1: The meat of the changes is in this file

web/src/pages/starter-kits/general.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const emptyProfileData = createEmptyProfileData();
+  const newProfileData: ProfileData = {
+    ...emptyProfileData,
+    businessPersona: "STARTING",
+    industryId: "generic",
+  };
+
+  // Next does not accept explicitly undefined values and roadmap has some by default
+  const roadmap = JSON.parse(JSON.stringify(await buildUserRoadmap(newProfileData))) as Roadmap;
+
```
alexander-littleton, line 154: It seems like the explicitly undefined culprit here is the `required` field on the `Task` object. I'll have to look into whether we need `required` to be explicitly undefined and if not, convert it to just a boolean. I'll probably be fighting [Hyrum's Law](https://www.hyrumslaw.com/) a bit so we'll see how it goes.

alexander-littleton: Turns out we had some inaccurate typing during the conversion from the roadmap json files to the `AddOn` type which resulted in that undefined field getting explicitly assigned. Commit with explanation to follow.

web/src/lib/roadmap/roadmapBuilder.ts
```diff
@@ -109,7 +109,7 @@ const addTasksFromAddOn = (builder: RoadmapBuilder, addOns: AddOn[]): RoadmapBui
     if (!task) continue;
     builder.tasks = [
       ...builder.tasks,
-      { filename: task, weight: addOn.weight, stepNumber: addOn.step, required: addOn.required },
+      { filename: task, weight: addOn.weight, stepNumber: addOn.step, required: !!addOn.required },
```
alexander-littleton, line 112: While the `RoadmapBuilder` type enforces the `required` field to be a `boolean`, when the roadmap is generated from the builder downstream, the `Roadmap` type allows it to be `undefined` again. Maybe worth looking into converting that over to just a `boolean` at some point.

alexander-littleton:
From my cursory work around those `/industry/` json files, it feels like there might be an advantage to introducing some runtime safeguards for when they are parsed a la [Zod](https://zod.dev/?id=introduction) or some other runtime type checking library. Maybe that's a bit heavy handed though.

web/src/components/starter-kits/StepInfo.tsx
```diff
@@ -0,0 +1,25 @@
+import { Heading } from "@/components/njwds-extended/Heading";
+import { Step } from "@/lib/types/types";
+import { ReactElement } from "react";
+
+export const StepInfo = (props: { step: Step; taskNames: string[] }): ReactElement => {
+  return (
+    <div className="flex-basis-100">
+      <div className="margin-top-05 margin-bottom-2">
+        <span className="border-2px border-secondary-vivid-dark text-secondary-vivid-dark radius-lg padding-x-1 padding-y-05">
+          <b>STEP {props.step.stepNumber}</b>
```
seidior, line 10: [sand] @alexander-littleton Let's use `<strong>` rather than `<b>` to embolden these.

Also @adidner, I'd imagine these would semantically be headings that should appear in the page accessibility heading structure, no?

web/src/components/starter-kits/StepInfo.tsx
```diff
@@ -0,0 +1,25 @@
...
+        </span>
+      </div>
+      <Heading level={3} styleVariant="h4">
```
seidior, line 13: [sand] On the other hand, I don't believe the step names should be headings, since they don't really define a section of content; the steps do. Or rather, the heading should read as "Step 1: Plan Your Business" somehow.

web/src/components/starter-kits/StepInfo.tsx
```diff
@@ -0,0 +1,25 @@
...
+        {props.step.name.replace("${OoS}", "")}
+      </Heading>
+      <p>{props.step.description}</p>
+      <b className="text-base-dark">{`(${props.step.timeEstimate})`}</b>
```
seidior, line 17: [sand] Another opportunity to use `<strong>` rather than `<b>`. And it should semantically probably be inside the same paragraph as the description, so that screen readers don't read it as a separate element.

web/src/lib/roadmap/roadmapBuilder.ts
seidior: You could also just switch this to an explicit test, like so:

```typescript
{
  required: addOn.required === true
}
```

web/src/pages/starter-kits/general.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { buildUserRoadmap } from "@/lib/roadmap/buildUserRoadmap";
+import { Roadmap, Task } from "@/lib/types/types";
+import analytics from "@/lib/utils/analytics";
+import { ProfileData, createEmptyProfileData } from "@businessnjgovnavigator/shared/index";
+import type { InferGetStaticPropsType } from "next";
+import { useRouter } from "next/router";
+import { ReactElement } from "react";
+
+const General = (props: InferGetStaticPropsType<typeof getStaticProps>): ReactElement => {
+  const router = useRouter();
```
seidior, line 18: [dust] I wonder if we can drop the router and route manually, either by putting the buttons inside links or by setting the window location. It's clearly not affecting the starter kit, as the page's HTML is being rendered as expected, but it's one fewer frontend JS-based component here. Do you see a benefit, or do the benefits of using the router outweigh the theoretical benefits of dropping it?

web/src/pages/starter-kits/general.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { StepInfo } from "@/components/starter-kits/StepInfo";
+import { getMergedConfig } from "@/contexts/configContext";
+import { ROUTES } from "@/lib/domain-logic/routes";
+import { buildUserRoadmap } from "@/lib/roadmap/buildUserRoadmap";
+import { Roadmap, Task } from "@/lib/types/types";
+import analytics from "@/lib/utils/analytics";
+import { ProfileData, createEmptyProfileData } from "@businessnjgovnavigator/shared/index";
+import type { InferGetStaticPropsType } from "next";
+import { useRouter } from "next/router";
+import { ReactElement } from "react";
```
seidior, line 15: [sand] `ReactElement` is a type so it should be imported using `import type`.

web/src/pages/starter-kits/general.tsx
```diff
@@ -0,0 +1,158 @@
...
+              {Config.starterKits.hero.ctaButton}
+            </PrimaryButton>
+          </div>
+
+          <div className="desktop:grid-col-6 desktop:display-block display-none">
+            <img className="width-100" src="/img/seo-img-main.webp" alt="" role="presentation" />
+          </div>
+        </section>
+        <section
+          aria-label="Solutions"
```
seidior, line 63: [dust] Does a Section need an `aria-label` when it clearly has a heading already? Also, we try not to encode bare strings in our code like this.

web/src/pages/starter-kits/general.tsx
```diff
@@ -0,0 +1,158 @@
...
+              }}
+              isRightMarginRemoved={true}
+              isLargeButton={true}
+            >
+              {Config.starterKits.solutions.ctaButton}
+            </PrimaryButton>
+          </div>
+        </section>
+        <section
+          aria-label="Example Roadmap"
```
seidior, line 104: [dust] Same comment as above.

seidior: [sand] I wonder if it would be better to separate this section out to a separate component, given that it behaves differently from the rest of the starter kit page.

web/src/styles/base/gap.scss
```diff
@@ -0,0 +1,23 @@
...
+  gap: 1.5rem;
+}
+
+.gap-4 {
+  gap: 2rem;
+}
+
+.gap-5 {
+  gap: 2.5rem;
+}
```
seidior, line 23: [dust] Are these consistent with USWDS naming conventions? Typically their increments are -1, -105, -2, -205, etc.

content/src/fieldConfig/starter-kits.json
```diff
@@ -0,0 +1,30 @@
...
+        "title": "Learn about business licenses",
+        "body": "Your personalized guide will help you understand what licenses your business needs to operate in New Jersey."
+      }
+    },
+    "steps": {
+      "title": "Get Your Business Started in 4 Easy Steps ",
+      "ctaButton": "Get My Starter Kit"
+    }
+  }
+}
```
seidior, line 30: [sand] This needs a trailing space.

seidior (approved):
Great work here both of you! Three additional things I think this needs, aside from the above, which may not be called out in the acceptance criteria (sorry):

1. I think we should change Next.js's configuration to route any traffic from /starter-kits to this page.
2. We should add this to the sitemap as part of the AC for this story (which is on the way thanks to @alekvuka).
3. I think we should have some Cypress tests that test to ensure the content is being rendered statically. (It is, but I want to make sure that doesn't break in the future.)

web/src/pages/starter-kits/general.tsx
alexander-littleton: new syntax unlocked 🧠

web/src/lib/roadmap/roadmapBuilder.ts
alexander-littleton: True. `Boolean(addOn.required)` is another option but I figured `!!` is the more common approach to Boolean conversion. Any limitations to sticking with that?

web/src/components/starter-kits/StepInfo.tsx
alexander-littleton: I do have a couple ideas of how to potentially achieve this with varying levels of complexity if @adidner prefers this approach

web/src/components/starter-kits/StepInfo.tsx
alexander-littleton: I went with `<b>` due to this being a standalone typography tag which I figured conveys some form of emphasis already for screen readers. No problem switching over to `<strong>` since I could see that being a bit more explicit about conveying "this is basically a heading".

web/src/lib/roadmap/roadmapBuilder.ts
seidior: Nope. If eslint doesn't error, you're fine!

seidior:
Also, @alexander-littleton, reading through the checklist, you did add new kinds of content to the CMS. Is it your plan to add it to the search tool, or are you holding off on that until the next story?

alexander-littleton:
> Also, @alexander-littleton, reading through the checklist, you did add new kinds of content to the CMS. Is it your plan to add it to the search tool, or are you holding off on that until the next story?
Yea this is one I should have probably checked further with somebody about. I made that change to `cmsCollections.ts` and saw starter kits search results coming through the tool so thought that might have been all I needed to touch. I can reach out when I pick this work up again for clarification.

web/src/pages/starter-kits/general.tsx
alexander-littleton: Looks like was dropped during the industry specific starter kit feature implementation 🚀

web/src/styles/base/gap.scss
alexander-littleton: I decided to just do a loose implementation of more or less what I needed instead of a more complete implementation just for a bit of time saving

web/src/pages/starter-kits/general.tsx
alexander-littleton: For acessibility navigation purposes it does but to be honest I'm not entirely sure a section tag is necessary at all

alexander-littleton: I guess for semantics during dev it's helpful. The labels aren't really meaningful as is to screen readers IMO.

PR #8391, opened 2024-08-07
chore: [#188081219] add trim to nospace and write-once-read-only widgets

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/188081219

### Approach

Add trim() to the nospace and the write once widgets. Note: the write once widget name was updated to state no spaces can be added.

## Code author checklist

### DIFF
--- a/web/src/pages/mgmt/cms.tsx
+++ b/web/src/pages/mgmt/cms.tsx
@@ -5,7 +5,7 @@ import IconWidgetEditor from "@/lib/cms/editors/icon";
 import AlertEditor from "@/lib/cms/editors/infoAlert";
 import Note from "@/lib/cms/editors/note";
 import { NoSpaceControl } from "@/lib/cms/fields/nospacefield";
-import { WriteOnceReadOnlyControl } from "@/lib/cms/fields/writeoncereadonlyfield";
+import { WriteOnceReadOnlyNoSpaceControl } from "@/lib/cms/fields/writeoncereadonlynospacefield";
 import { applyTheme } from "@/lib/cms/helpers/applyTheme";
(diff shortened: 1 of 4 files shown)


web/src/pages/mgmt/cms.tsx
```diff
@@ -63,7 +63,7 @@ const CMS = dynamic(
       // @ts-expect-error: No type definition available
       CMS.init({ CMS_CONFIG });
       // @ts-expect-error: No type definition available
-      CMS.registerWidget("write-once-read-only", WriteOnceReadOnlyControl);
+      CMS.registerWidget("write-once-read-only-no-space", WriteOnceReadOnlyNoSpaceControl);
```
jphechter, line 66: It looks like there are a couple other places where we still use `write-once-read-only`. If we're completely removing this widget, then we should update those to also use the new prop.

I just did a quick search in `config.yml` to be sure and saw a couple.

web/src/lib/cms/fields/writeoncereadonlyfield.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { Component, createRef, ReactElement } from "react";
```
jphechter, line 1: Should we rename this file?

FarazA22: changed

web/src/pages/mgmt/cms.tsx
FarazA22: There were two, i missed them cuz I did a find with quotes included, they all need to be swapped

jphechter (approved):

PR #8494, opened 2024-08-22
Create industry specific starter kits revisions

adidner:
## Description

We wanted to largely do 2 things
We wanted to have Steps be represented as Step when there was only one step with regards to the title for the Road Map Section

We also wanted all essential questions to be answered to give us the maximum number of roadmaps events for SEO

image

### Ticket

This pull request resolves [#187792208](https://www.pivotaltracker.com/story/show/187792208).

### Steps to Test

Run Local and Go here: http://localhost:3000/starter-kits/domestic-employer, see that at the bottom it says Step instead of Steps for the roadmap title (because it only has 1 step)

http://localhost:3000/starter-kits/ go here and see that it says steps

We also wanted to make sure that essential questions are being answered in the affirmative to see as many roadmap steps as possible. 

Go to any of the urls for starter kits with essential questions, 

http://localhost:3000/starter-kits/cannabis or http://localhost:3000/starter-kits/certified-public-accountant

and see that the roadmaps contains extra events due to essential questions like Apply for Your Annual Cannabis License or Apply for a Certified Public Accountant License

A thing I added is the idea that we should have a test that fails if we add an essential question that isn't accounted for in the logic of the starter-kits. 

I've added this test and have a second commit with code that you can uncomment to verify that it does indeed fail if an additional essential question is added (this will be removed prior to rebasing into main)

## Code author checklist


shared/src/profileData.ts
```diff
@@ -37,6 +37,7 @@ export interface IndustrySpecificData {
   readonly residentialConstructionType: ResidentialConstructionType;
   readonly employmentPersonnelServiceType: EmploymentAndPersonnelServicesType;
   readonly employmentPlacementType: EmploymentPlacementType;
```
adidner, line 39: For testing only: if you want to validate the test. Will be removed before a merge

shared/src/test/factories.ts
```diff
@@ -248,6 +248,7 @@ export const generateIndustrySpecificData = (
     ]),
     employmentPersonnelServiceType: randomElementFromArray(["JOB_SEEKERS", "EMPLOYERS"]),
     employmentPlacementType: randomElementFromArray(["TEMPORARY", "PERMANENT", "BOTH"]),
```
adidner, line 250: For testing only: if you want to validate the test. Will be removed before a merge

web/src/lib/domain-logic/essentialQuestions.ts
```diff
@@ -179,4 +179,11 @@ export const EssentialQuestions: EssentialQuestion[] = [
...
     fieldName: "employmentPlacementType",
   }),
+  // new EssentialQuestion({
+  //   shouldBeResetWhenIndustryChanges: true,
+  //   isQuestionApplicableToIndustry: (industry): boolean => {
+  //     return !!industry.industryOnboardingQuestions.isEmploymentAndPersonnelTypeApplicable;
+  //   },
+  //   fieldName: "todeletetype",
+  // }),
 ];
```
adidner, line 189: For testing only: if you want to validate the test. Will be removed before a merge

web/src/lib/domain-logic/starterKits.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      allStarterKitEssentialQuestions = combineObjectsKeepingTruthyValues(
+        allStarterKitEssentialQuestions,
+        data
+      );
+    }
+
+    return Object.keys(removeFalsyValues(allStarterKitEssentialQuestions));
+  };
+
+  it("fails if a new essential question is added to the code base without being explicitly handled in starter kits", () => {
```
adidner, line 48: I feel like some of the names here are kinda long winded and confusing. I didn't have any better ideas on how to name things though. Very open to suggestions

web/src/lib/domain-logic/starterKits.tsx
```diff
@@ -0,0 +1,103 @@
...
+    businessPersona: "STARTING",
+    industryId: industry.id,
+    legalStructureId: "limited-liability-company",
+  };
+
+  if (hasEssentialQuestion(industry.id)) {
+    const essentialQuestions = getEssentialQuestion(industry.id);
+    for (const essentialQuestion of essentialQuestions) {
+      // this only works if the values is defaults to a boolean, if its undefined by default as a value this will not catch it
+      if (typeof newProfileData[essentialQuestion.fieldName] === "boolean") {
```
adidner, line 39: I'm not sure if there is a better way to capture the idea of this comment, or if it should just be deleted. 

It did take me a bit to understand that this was happening and I worry that someone in the future might fall into the same minor time wasting. 

Open to suggestions and alternative.

seidior (approved):
Behavior's a lot better, and I'm glad you were able to come up with a solution that catches new essential questions. Thanks for your patience on this story!

web/src/lib/domain-logic/starterKits.test.ts
```diff
@@ -0,0 +1,64 @@
...
+import { createStarterKitProfileData } from "@/lib/domain-logic/starterKits";
+import { emptyProfileData, Industries, ProfileData } from "@businessnjgovnavigator/shared";
+
+describe("starter Kits", () => {
+  const combineObjectsKeepingTruthyValues = (obj1: ProfileData, obj2: ProfileData): ProfileData => {
+    let result: ProfileData = { ...obj1 };
+    for (const key in obj2) {
+      const typedKey = key as keyof ProfileData;
+      if (!result[typedKey]) {
+        result = {
```
FarazA22, line 11: dust - you could update logic to a single line
`result[typedKey] = obj2[typedKey]`

web/src/lib/domain-logic/starterKits.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    return result;
+  };
+
+  const removeFalsyValues = (obj: ProfileData): Partial<ProfileData> => {
+    let result: Partial<ProfileData> = {};
+    for (const key in obj) {
+      const typedKey = key as keyof ProfileData;
+      if (Object.prototype.hasOwnProperty.call(obj, typedKey) && obj[typedKey]) {
+        result = {
+          ...result,
```
FarazA22, line 26: dust - you can add logic to in the conditional to see if the value is false and use the delete keyword to remove the key

web/src/lib/domain-logic/starterKits.tsx
FarazA22: maybe update the comment to say, essential questions with default boolean value.

web/src/lib/domain-logic/starterKits.test.ts
adidner: I'm pretty sure I can't because the field is read only

web/src/lib/domain-logic/starterKits.test.ts
adidner: I'm pretty sure I can't because the field is read only, I had it initially this way but I was getting errors

PR #8497, opened 2024-08-23
feat: [#188135142] Add multiple dwelling registration status check

SirSamuelJoseph:
## Description

Adds multiple dwelling status check registration search

### Ticket

This pull request resolves [#188135142](https://www.pivotaltracker.com/story/show/188135142).

### Steps to Test

1. Create a Poppy in the "Real Estate Investing" industry
2. Answer yes to the multiple dwelling non essential question
3. Use address in the QA testing doc to see results
4. (NOTE! No addresses in the DB can get you into the incomplete or cancelled state)

## Code author checklist


api/src/client/dynamics/housing/DynamicsHousingRegistrationClient.ts
```diff
@@ -53,7 +62,7 @@ export const DynamicsHotelMotelRegistrationClient = (
           }
           return 1;
         });
-        return sortedRegistrations;
+        return sortedRegistrations.slice(0, 1);
```
SirSamuelJoseph, line 65: For this iteration, we decided to only show the most recent registration. However, I'm keeping this as an array since we have talked about wanting to show history again in the future

web/src/lib/api-client/apiClient.ts
```diff
@@ -46,11 +46,12 @@ export const checkElevatorViolations = (address: string, municipalityId: string)
   return post(`/elevator-safety/violations`, { address, municipalityId });
 };
 
-export const checkHousingHotelMotelRegistrationStatus = (
+export const checkHousingRegistrationStatus = (
   address: string,
-  municipalityId: string
+  municipalityId: string,
+  propertyInterestType: string
```
jphechter, line 52: [dust] Should this be a specific set of strings we anticipate?

api/src/client/dynamics/housing/DynamicsHousingRegistrationStatusClient.ts
```diff
@@ -1,24 +1,29 @@
...
+  HousingRegistrationClient,
   HousingRegistrationRequestLookupResponse,
 } from "@client/dynamics/housing/types";
 import { AccessTokenClient } from "@client/dynamics/types";
 import { LogWriterType } from "@libs/logWriter";
 
 type Config = {
   accessTokenClient: AccessTokenClient;
-  housingHotelMotelRegistrationClient: HotelMotelRegistrationClient;
+  housingHotelMotelRegistrationClient: HousingRegistrationClient;
```
aadedejifearless, line 11: Noticed we updated the properties in the housingRouter.ts file. We should be updating this as well `housingHotelMotelRegistrationClient`

api/src/client/dynamics/housing/DynamicsHousingRegistrationStatusClient.ts
```diff
@@ -34,10 +39,11 @@ export const DynamicsHotelMotelRegistrationStatusClient = (
       };
     }
 
-    const registrations = await config.housingHotelMotelRegistrationClient.getHotelMotelRegistration(
+    const registrations = await config.housingHotelMotelRegistrationClient.getHousingRegistration(
```
aadedejifearless, line 42: See previous comment

api/src/functions/express/app.ts
```diff
@@ -225,11 +225,11 @@ const dynamicsHousingHealthCheckClient = DynamicsHousingHealthCheckClient(logger
   accessTokenClient: dynamicsHousingAccessTokenClient,
   orgUrl: DYNAMICS_HOUSING_URL,
 });
-const dynamicsHousingHotelMotelRegistrationClient = DynamicsHotelMotelRegistrationClient(
+const dynamicsHousingHotelMotelRegistrationClient = DynamicsHousingRegistrationClient(
```
aadedejifearless, line 228: think we should update this as well.

web/src/components/tasks/MultipleDwellingRegistrationTask.tsx
```diff
@@ -0,0 +1,172 @@
...
+  };
+
+  const onEdit = (): void => {
+    setHousingAddress(undefined);
+    setHousingRegistrationLookupSummary(undefined);
+  };
+
+  if (municipalities.length === 0) {
+    setError(`SEARCH_FAILED`);
+  }
```
jphechter, line 57: We just went through something very similar with the license search. Based on what I'm seeing, the case we're looking at here displays an error message: "Sorry, something went wrong. Please try again later." This implies that the service has an issue, but it could be that the user fat-fingered some data. Trying again may not solve the issue.

It might be worth having a different error message for a successful search that returns no data.

web/src/components/tasks/MultipleDwellingRegistrationTask.tsx
```diff
@@ -0,0 +1,172 @@
...
+        })
+        .catch(() => {
+          setError("SEARCH_FAILED");
+        })
+        .finally(async () => {
+          setIsLoading(false);
+        });
+    } else {
+      setError("FIELDS_REQUIRED");
+      return;
```
jphechter, line 83: [dust] Maybe have this the first check and act like a guard clause? If you did that could you take away the first `if` and un-nest the logic in lines 61-80?

web/src/components/tasks/MultipleDwellingRegistrationTask.tsx
```diff
@@ -0,0 +1,172 @@
+import { Content } from "@/components/Content";
```
jphechter, line 1: Given the parallels between this and the other similar 2-tab status check components (License check being one), it might be a cool refactor to have a common component these are both leveraging.

web/src/components/tasks/MultipleDwellingRegistrationTask.tsx
SirSamuelJoseph: So this would be the service having an error: specifically if we can't pull the list of municipalities, they can't search because of it. It would be unlikely to actually have this happen though, since we're pulling from a static list of municipalities. We do have dedicated errors per:

1. We can't find anything for the address you put in
2. We found a property at that address but no registrations

api/src/functions/express/app.ts
SirSamuelJoseph: Totally right, I'll fix that!

api/src/client/dynamics/housing/DynamicsHousingRegistrationClient.test.ts
```diff
@@ -78,7 +112,7 @@ describe("DynamicsHousingPropertyInterestClient", () => {
 
     mockAxios.get.mockResolvedValue({ data: housingHotelMotelRegistrationMockResponse });
     expect(
-      await client.getHotelMotelRegistration(mockAccessToken, propertyInterestID, buildingCount)
+      await client.getHousingRegistration(mockAccessToken, propertyInterestID, buildingCount, "hotelMotel")
```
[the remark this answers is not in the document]
SirSamuelJoseph: Discussed this in a call

web/src/lib/api-client/apiClient.ts
SirSamuelJoseph: Yeah, makes sense, I'll update this

web/src/components/tasks/MultipleDwellingRegistrationTask.tsx
SirSamuelJoseph: Love it, I'll move that around

web/src/components/tasks/MultipleDwellingRegistrationTask.tsx
SirSamuelJoseph: I'll create a ticket for this to get tackled on a refactor Friday!

api/src/client/dynamics/housing/DynamicsHousingRegistrationClient.ts
```diff
@@ -1,31 +1,45 @@
...
   orgUrl: string
-): HotelMotelRegistrationClient => {
-  const getHotelMotelRegistration = async (
+): HousingRegistrationClient => {
+  const getHousingRegistration = async (
     accessToken: string,
     propertyInterestId: string,
-    buildingCount: number
+    buildingCount: number,
+    propertyInterestType: PropertyInterestType
```
jphechter, line 21: [dust] Is there a point we should transition from multiple positional arguments to passing in a single config object?

api/src/client/dynamics/housing/DynamicsHousingRegistrationClient.ts
```diff
@@ -1,31 +1,45 @@
...
   ): Promise<HousingRegistrationRequestResponse> => {
     const logId = logWriter.GetId();
     logWriter.LogInfo(`Dynamics Hotel Motel Registration Client - Id:${logId}`);
-    const hotelPropertyInterestType = getPropertyInterestTypeIntegerFromTitle("Hotel");
-    const motelPropertyInterestType = getPropertyInterestTypeIntegerFromTitle("Motel");
+    let propertyInterestTypeFilter = "";
+
+    switch (propertyInterestType) {
```
jphechter, line 27: Is there a particular reason you reached for a `switch` here? I'm assuming we're anticipating more `propertyInterestType`s will be added in the future?

api/src/client/dynamics/housing/DynamicsHousingRegistrationStatusClient.ts
```diff
@@ -1,24 +1,30 @@
...
 type Config = {
   accessTokenClient: AccessTokenClient;
-  housingHotelMotelRegistrationClient: HotelMotelRegistrationClient;
+  housingHousingRegistrationClient: HousingRegistrationClient;
   housingPropertyInterestClient: HousingPropertyInterestClient;
 };
 
-export const DynamicsHotelMotelRegistrationStatusClient = (
+export const DynamicsHousingRegistrationStatusClient = (
   logWriter: LogWriterType,
```
jphechter, line 17: [dust] It doesn't look like the `logWriter` is being used here.

web/src/components/tasks/CheckHousingRegistrationStatus.tsx
```diff
@@ -12,17 +12,22 @@ import { ChangeEvent, FormEvent, ReactElement, useEffect, useState } from "react
...
+> = {
   NO_PROPERTY_INTEREST_FOUND: Config.housingRegistrationSearchTask.errorTextNoPropertyInterestFound,
   NO_HOTEL_MOTEL_REGISTRATIONS_FOUND: Config.housingRegistrationSearchTask.errorTextNoHotelMotelRegistrations,
   FIELDS_REQUIRED: Config.housingRegistrationSearchTask.errorTextFieldsRequired,
   SEARCH_FAILED: Config.housingRegistrationSearchTask.errorTextSearchFailed,
+  NO_MULTIPLE_DWELLINGS_REGISTRATIONS_FOUND:
+    Config.housingRegistrationSearchTask.errorTextNoMultipleDwellingRegistrations,
 };
 
 export const CheckHousingRegistrationStatus = (props: Props): ReactElement => {
```
jphechter, line 33: Not included in the changes in this PR, but a question I had in just reviewing this file:

In the `useEffect` defined on 47, do we need the guard clauses for `communityAffairsAddress`? It looks like we're checking to see if `communityAffairsAddress` is defined and then conditionally calling `setFormValues`. Otherwise it looks like they're being initialized to undefined values.

Is there any harm in just doing this without the checks?
```typescript
setFormValues((prevValues) => {
  return {
    ...prevValues,
    address1: communityAffairsAddress.streetAddress1,
    address2: communityAffairsAddress.streetAddress2
    }
});
```

web/src/components/tasks/MultipleDwellingRegistrationTask.tsx
jphechter: Ah that's interesting. So a slightly different use case. Great to know. thank you!

api/src/client/dynamics/housing/DynamicsHousingRegistrationClient.ts
SirSamuelJoseph: Yup, for expansion potential. There's a lot of types here

api/src/client/dynamics/housing/DynamicsHousingRegistrationStatusClient.ts
SirSamuelJoseph: Good catch, I'll remove that param

api/src/client/dynamics/housing/DynamicsHousingRegistrationClient.ts
SirSamuelJoseph: This would be a good thing to add to the style guide, I think I'm fine with 4 but if it gets to five I'll pull it into an object

web/src/components/tasks/CheckHousingRegistrationStatus.tsx
SirSamuelJoseph: You're totally right, this is much simpler

PR #8534, opened 2024-08-30
Updates to Domestic Employer experience based on SME feedback

jphechter:
## Description

SMEs reviewing the Domestic Employer industry experience requested some specific changes to the roadmap.
Screenshot 2024-08-30 at 3 55 42 PM

### Ticket

This pull request resolves [#188193453](https://www.pivotaltracker.com/story/show/188193453).

### Approach

These changes largely use the "Domestic Employer" operating phase to determine what content to render. To handle the roadmap section copy changes I created a new section to be conditionally shown only for Domestic Employers.

### Steps to Test

Create a business of industry "Domestic Employer". After completing onboarding all of the changes should be immediately visible on the roadmap.

Be sure to test both desktop and mobile, as well as the mini-roadmap.

## Code author checklist

### DIFF
--- a/content/src/roadmaps/steps.json
+++ b/content/src/roadmaps/steps.json
@@ -1,6 +1,13 @@
 {
   "steps": [
     {
+      "section": "DOMESTIC_EMPLOYER_SECTION",
+      "stepNumber": 1,
+      "name": "Register as an Employer",
+      "timeEstimate": "",
+      "description": ""
+    },
+    {
       "section": "PLAN",
       "stepNumber": 1,
       "name": "Plan Your ${OoS} Business",
--- a/web/src/components/dashboard/DashboardOnDesktop.tsx
+++ b/web/src/components/dashboard/DashboardOnDesktop.tsx
@@ -6,9 +6,9 @@ import { ElevatorViolationsCard } from "@/components/dashboard/ElevatorViolation
 import { HideableTasks } from "@/components/dashboard/HideableTasks";
 import { Roadmap } from "@/components/dashboard/Roadmap";
 import { SidebarCardsContainer } from "@/components/dashboard/SidebarCardsContainer";
+import { getRoadmapHeadingText } from "@/components/dashboard/helpers";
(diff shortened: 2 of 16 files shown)


web/src/components/dashboard/DashboardOnDesktop.tsx
```diff
@@ -81,7 +81,9 @@ export const DashboardOnDesktop = (props: Props): ReactElement => {
                   <>
                     <hr className="margin-bottom-3" />
                     <Heading level={2}>
-                      {Config.dashboardRoadmapHeaderDefaults.RoadmapTasksHeaderText}
+                      {operatingPhase.id === OperatingPhaseId.DOMESTIC_EMPLOYER
```
adidner, line 84: pebble - So this code is repeated verbatim in the DashboardOnMobile code. I think as a result it would be good to make it a function. 

Also its business logic at this point and has nothing to do with the actual shape of the component and should thus, be separated into a function somewhere else anyways

content/src/roadmaps/steps.json
```diff
@@ -1,5 +1,12 @@
...
+    {
+      "section": "DOMESTIC_EMPLOYER_SECTION",
+      "stepNumber": 1,
+      "name": "Register as an Employer",
+      "timeEstimate": "",
+      "description": ""
+    },
     {
       "section": "PLAN",
       "stepNumber": 1,
```
adidner, line 12: Not sure why but it also appears to be starting with the area collapsed 
image
Probably not what we want, I don't see something obvious as to why but probably worth looking into a bit deeper

adidner (approved):
Generally looks good and performs as expected from my spot check. 

can't think of a particularly better way to do this, however I did leave a few little comments that should probably be addressed before merging

content/src/roadmaps/steps.json
jphechter: That's a great callout that I did not catch! Thank you!

PR #8672, opened 2024-09-16
feat: [#188283234] Enable contextual info in non essential questions

SirSamuelJoseph:
## Description
Enables embedding contextual links within non-essential questions

### Ticket

This pull request resolves [#188283234](https://www.pivotaltracker.com/story/show/188283234).

### Approach
Converts non-essential question to Content.

This format `text|contextual-info-text` will be displayed in the CMS as a string, not as a contextual text. Right now this only affects a single NEQ but we might need a follow-up ticket so it properly displays in the CMS

### Steps to Test

1. Create new Poppy in health care industry
2. Go to profile to see non-essential question

## Code author checklist

### DIFF
--- a/web/src/components/data-fields/non-essential-questions/NonEssentialQuestion.tsx
+++ b/web/src/components/data-fields/non-essential-questions/NonEssentialQuestion.tsx
@@ -1,3 +1,4 @@
+import { Content } from "@/components/Content";
 import { ProfileDataContext } from "@/contexts/profileDataContext";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { getNonEssentialQuestionText } from "@/lib/domain-logic/getNonEssentialQuestionText";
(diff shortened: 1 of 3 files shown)


jphechter:
I don't have a problem with this pattern. Storing strings that will be formatted, like this contextual link, in the JSON files isn't a problem. The only issue I could see is this being rendered in the CMS as just plain text.

We could change this widget from `string` to `markdown`. That will at least give the content team a greater understanding of what the string represents.
Screenshot 2024-09-16 at 1 27 03 PM

It should also help with any future edits they need to make.
Screenshot 2024-09-16 at 1 27 36 PM

This might also be a good candidate for a CMS preview on the right-hand side. I think we should be able to do something programmatic, like looping over all the objects in `nonEssentialQuestions.json`, so that this could be reflective of any new questions they add.

TL;DR version
No issues with the pattern, but I think there's more we can do to make this more helpful for the CMS users.
@SirSamuelJoseph @seidior

somabadri (approved):
+1 to what jp said but otherwise it looks good to me!

web/src/components/data-fields/non-essential-questions/NonEssentialQuestion.tsx
```diff
@@ -28,7 +29,9 @@ export const NonEssentialQuestion = (props: Props): ReactElement => {
       {nonEssentialQuestionText && (
         <>
           <div className={"margin-top-2"}>
-            <span className={"text-bold"}>{nonEssentialQuestionText}</span>
+            <span className={"text-bold"}>
+              <Content>{nonEssentialQuestionText}</Content>
+            </span>
```
seidior, line 34: [sand] I know it's mostly meaningless these days, but since `<Content>` can include block-level elements, it should be wrapped in a block-level element rather than a `<span>`.

seidior (commented):
Thanks, Sam!

PR #8692, opened 2024-09-18
feat: [#188079116] add license status to license calendar events

FarazA22:
### Approach

Note: It might be helpful to review each commit individually.

https://www.pivotaltracker.com/story/show/188259811

The nonessential question was previously created but was not mapped to the correct industry. I went ahead and added the nonessential question to the entertainment industry. Separately, I added the license name to the appropriate task and license calendar event files. I also created a test to validate the list of task ids that render the license task for completeness.

https://www.pivotaltracker.com/story/show/188339189

We noticed that there are many users that have multiple records with the same business name and address combo in RGB db. Because of our existing logic, those license searches would result in an error. We removed the logic around the error.

https://www.pivotaltracker.com/story/show/188341426

I noticed a flakey test when running this commit during CircleCI, I believe is was occurring b/c a disable industry was being selected. Logic was updated to prevent this. 

https://www.pivotaltracker.com/story/show/188408898
Remove unused randomOperatingPhaseId fn

## Code author checklist

### DIFF
--- a/api/src/client/dynamics/license-status/formatRegulatedBusinessDynamicsApplications.test.ts
+++ b/api/src/client/dynamics/license-status/formatRegulatedBusinessDynamicsApplications.test.ts
@@ -1,5 +1,4 @@
 import { formatRegulatedBusinessDynamicsApplications } from "@client/dynamics/license-status/formatRegulatedBusinessDynamicsApplications";
-import { DUPLICATE_LICENSE_TYPE_ERROR } from "@client/dynamics/license-status/regulatedBusinessDynamicsLicenseStatusTypes";
 import { generateLicenseStatusChecklistResult } from "@test/factories";
 
 describe("formatRegulatedBusinessDynamicsApplications", () => {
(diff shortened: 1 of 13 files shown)


somabadri (commented):
Just one note: I see that you removed the `randomOperatingPhaseId` function because we don't use it anymore. Could you split it into its own commit so it's clear that it's not related to the home-based industry fix commit?"

api/src/client/dynamics/license-status/formatRegulatedBusinessDynamicsApplications.test.ts
```diff
@@ -20,17 +19,4 @@ describe("formatRegulatedBusinessDynamicsApplications", () => {
...
-      professionNameAndLicenseType: "duplicate profession name and license type",
-    });
-    const checklistResult2 = generateLicenseStatusChecklistResult({
-      professionNameAndLicenseType: "duplicate profession name and license type",
-    });
-
-    expect(() => formatRegulatedBusinessDynamicsApplications([checklistResult1, checklistResult2])).toThrow(
-      DUPLICATE_LICENSE_TYPE_ERROR
-    );
-  });
```
alexander-littleton, line 35: [sand] - I wonder if it's worth expecting the inverse now as a happy path case to communicate to future devs that there can be duplicates.

alexander-littleton (approved):
LGTM

api/src/client/dynamics/license-status/formatRegulatedBusinessDynamicsApplications.test.ts
FarazA22: the duplicates get over written, i added a test. thanks for the feedback

PR #8711, opened 2024-09-24
feat: [#188172364] connect businesses table with lambda

aadedejifearless:
## Description

This PR establishes a connection between the businesses tables and the backend Lambda application (referred to as "API"). This integration enables us to read from and write to these tables in future stories.

### Ticket

This pull request resolves [#188172364](https://www.pivotaltracker.com/story/show/188172364).

### Approach

- **API Business Schema:** Created `api/businesses-dynamodb-schema.json` to define the DynamoDB schema for the businesses table.
- **DynamoDB Data Client:** Created `api/src/db/DynamoBusinessesDataClient.ts` to handle data operations related to the businesses table.
-  **DynamoDB Data Client Testing** Created `api/src/db/DynamoBusinessesDataClient.test.ts` to implement unit tests for the newly created data client.
- Updated `api/src/db/DynamoUserDataClient.test.ts` and `api/src/db/DynamoUserDataClient.ts` to ensure compatibility with the new businesses data operations.
-  Updated `api/jest-dynalite-config.js` 
-  Updated `api/serverless.ts`
-  Added `api/src/db/config/dynamoDbConfig.ts` 
- Updated `api/src/db/dynamoUserDataMigration.test.ts`
- Added interface BusinessDataClient in `api/src/domain/types.ts`
-  Updated `api/src/functions/encryptTaxId/app.ts`
-  Updated `api/src/functions/express/app.ts`
- Updated `api/src/functions/updateExternalStatus/app.ts`
- Renamed `api/dynamodb-schema.json` to `api/users-dynamodb-schema.json`
- Updated `docker-compose.yml`

## Code author checklist


api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      }),
+    });
+    const businessId = businessesData.id;
+    await dynamoBusinessesDataClient.put(businessesData);
+
+    const expectedValue = await dynamoBusinessesDataClient.get(businessId);
+    const { version, ...filteredExpectedValue } = expectedValue as unknown as {
+      version?: string;
+      [key: string]: any;
+    };
```
alexander-littleton, line 45: Huh. I didn't realize you could use the spread syntax during destructuring like this or chain type assertions. Clever!

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    return db
+      .send(new QueryCommand(params))
+      .then(async (result) => {
+        if (!result.Items || result.Items.length !== 1) {
+          return;
+        }
+        const item = unmarshall(result.Items[0]);
+        return await doMigration(item.data);
+      })
```
alexander-littleton, line 41: [dust] I like to avoid throwing `await` within `.then()` just because I find it's less cognitive load for the reader. Not a super strong opinion though.

aadedejifearless: Hey @alexander-littleton,

Thanks for the feedback! I initially used that approach for consistency with the existing code, but it’s a good call out. I've gone ahead and adjusted both the `DynamoUserDataClient` and `DynamoBusinessesDataClient` to follow the suggested approach.

Appreciate the review!

api/src/db/DynamoBusinessesDataClient.test.ts
aadedejifearless: Yup, that’s exactly what’s happening. Glad you found it interesting! The spread syntax during destructuring can be pretty handy, and chaining type assertions can help streamline things. Always learning new tricks with TS

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -0,0 +1,151 @@
...
+    };
+
+    return db
+      .send(new PutCommand(params))
+      .then(() => {
+        return migratedData;
+      })
+      .catch((error) => {
+        throw error;
+      });
```
alexander-littleton, line 110: [dust] Nitpick but I think this is redundant. It's in at least one other place as well.

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  let client: DynamoDBDocumentClient;
+  let dynamoBusinessesDataClient: BusinessesDataClient;
+
+  beforeEach(() => {
+    client = DynamoDBDocumentClient.from(new DynamoDBClient(config), dynamoDbTranslateConfig);
+    dynamoBusinessesDataClient = DynamoBusinessesDataClient(client, dbConfig.tableName);
+  });
+
+  it("Test to verify db gets inserted items", async () => {
```
somabadri, line 31: [pebble] not sure if this was left from the initial draft of the test but it doesn't read well, "it Test to....". You could rewrite it as `it("successfully inserts and retrieves items from the db")`

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    const expectedValue = await dynamoBusinessesDataClient.findByEncryptedTaxId(encryptedTaxId);
+    expect(expectedValue).toBeDefined();
+    const { version, ...filteredExpectedValue } = expectedValue as unknown as {
+      version?: string;
+      [key: string]: any;
+    };
+    expect(filteredExpectedValue).toEqual(businessData);
+  });
+
+  it("Test should delete a business by the ID", async () => {
```
somabadri, line 146: [pebble] same as above, `it("successfully deletes a business by the ID")`

api/src/db/DynamoBusinessesDataClient.ts
alexander-littleton, line 1: [sand]: Just want to call out that it seems like there's a bit of difference in how we handle errors within the methods in this file. Some places we log and then throw a new error, other places we just pass through the error. Curious if this was a philosophical decision or if it's something that we might want to give more thought to at some point.

alexander-littleton (approved):
LGTM but I defer to the leads!

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -0,0 +1,151 @@
...
+        const item = unmarshall(result.Items[0]);
+        return doMigration(item.data);
+      })
+      .catch((error) => {
+        console.log(error);
+        throw new Error("Not found");
+      });
+  };
+
+  const findByNAICSCode = (naicsCode: string): Promise<Business | undefined> => {
```
somabadri, line 48: [boulder] I'm not sure what the use cases for `findByNAICSCode` and `findByIndustry` are, especially given how their type definitions are written. Various businesses could have the same NAICS code and industry type, so it would get matched with multiple businesses if I'm understanding the code correctly, and wouldn't that cause line 36 to trigger a return? Please let know if I'm misunderstanding how it's supposed to work.

aadedejifearless: Great question! You're right that multiple businesses can have the same NAICS code or industry type, and that’s actually the use case for these functions. Since we’ve set up secondary indexes on NAICS code and industry in the database, these functions are designed to query based on those indexes and return all matching businesses.
So, rather than triggering a return at line 36 after finding one match, the intention is to fetch and return all businesses that meet the criteria. I'll update the code

api/src/db/DynamoBusinessesDataClient.ts
aadedejifearless: Good catch!,  I've removed the redundant catch block here and in the other places as well. Thanks for pointing that out!

api/src/db/DynamoBusinessesDataClient.ts
aadedejifearless: Great point!  @somabadri You're absolutely right that multiple businesses can share the same NAICS code or industry. To address that, I added a findByIndexArray function, which is designed to return multiple businesses when querying by these indexes. This should prevent any unintended behavior on line 36 by ensuring that we're handling cases where more than one business matches the query. Let me know if you have any further thoughts or suggestions!

api/src/db/DynamoBusinessesDataClient.test.ts
somabadri, line 1: [pebble] I know I initially suggested it, but I think you can get rid of the prefix successfully because we want a test to pass and that implies that it's successfully finishes the action.

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -0,0 +1,170 @@
...
+        const item = unmarshall(result.Items[0]);
+        return doMigration(item.data);
+      })
+      .catch((error) => {
+        console.log(error);
+        throw new Error("Not found");
+      });
+  };
+
+  const findByIndexArray = (
```
somabadri, line 48: [dust] `findAllByIndex`or `findBusinessByIndex` for line 21 and `findAllBusinessesByIndex` for the function above on line 48. `findByIndexArray` is just a tad confusing upon initial read, thoughts?

api/src/domain/types.ts
```diff
@@ -33,6 +33,16 @@ export interface UserDataClient {
   getNeedTaxIdEncryptionUsers: () => Promise<UserData[]>;
 }
 
+export interface BusinessesDataClient {
+  get: (businessId: string) => Promise<Business>;
+  put: (businessData: Business) => Promise<Business>;
+  deleteBusiness: (businessId: string) => Promise<void>;
```
somabadri, line 39: [dust] thoughts on `deleteBusinessById`

api/src/domain/types.ts
```diff
@@ -33,6 +33,16 @@ export interface UserDataClient {
...
+  findByBusinessName: (businessName: string) => Promise<Business | undefined>;
+  findByNAICSCode: (naicsCode: string) => Promise<Business[]>;
```
somabadri, line 41: [dust/pebble] `findAllByNAICSCode` to indicate that it's returning multiple businesses

api/src/domain/types.ts
```diff
@@ -33,6 +33,16 @@ export interface UserDataClient {
...
+  findByIndustry: (industry: string) => Promise<Business[]>;
```
somabadri, line 42: [dust/pebble] `findAllByIndustry` to indicate that it's returning multiple businesses

api/src/domain/types.ts
aadedejifearless: Updated

somabadri (approved):
great work Ade!

PR #8772, opened 2024-09-30
Dev design

FarazA22:
### Approach

Note: It might be helpful to review each commit individually.

https://www.pivotaltracker.com/story/show/188326595
I updated the aria-level attribute for improved accessibility.

https://www.pivotaltracker.com/story/show/188284207
The min-height defined on a parent div was not being respected by a child component, causing the CTA button to not stay positioned at the bottom as expected. To address this, I added a min-height CSS rule specifically to the section below the header component. After some testing, 26.6rem was determined as the appropriate value to maintain visual consistency with other screens, ensuring the container height behaves uniformly across different layouts.

https://www.pivotaltracker.com/story/show/188347978
Given our shift away from USWDS, it was necessary to refactor class names that referenced the minh utility. I renamed these custom class names to be more descriptive and meaningful to the project context. Additionally, I separated the CSS class definitions into four distinct files: height, width, margin, and padding. During this process, I also removed deprecated class names that are no longer in use.

## Code author checklist


web/src/styles/base/height-width.scss
```diff
@@ -22,6 +22,10 @@
   min-height: 38rem;
 }
 
+.minh-2606 {
```
jphechter: TIL about [Sass Functions](https://sass-lang.com/documentation/at-rules/function/). It looks like we only have 5 classes like this right now, but if we see ourselves adding or needing more in the future, maybe this is something we could be programmatically generate so they're either already available or could be extended further.

web/src/styles/base/height.scss
```diff
@@ -0,0 +1,15 @@
+.min-height-40rem {
```
jphechter, line 1: If we're breaking this out into it's own stylesheet, my comment about programmatically generating these classes on the other commit would actually be easier to implement.

jphechter:
I did a global search and found a few instances of us still referencing `minh-*` classes. I'm assuming these are coming from USWDS.

If these are coming from USWDS and we're actively renaming our classes to disassociate from USWDS, do you think now would be a good time to recreate those classes following the new pattern you've created? I'm leaning towards yes. Leaving these as is means we'd have 2 different patterns that do roughly the same thing in our application.

Screenshot 2024-09-30 at 11 03 09 AM

web/src/styles/base/height-width.scss
aadedejifearless: +1 By programmatically generating these classes, we can ensure future flexibility and consistency in our codebase as we scale.

FarazA22: Spoke to JP, we are going to hold off on implementing some of the features of SASS given the limited number of css that would change. Also, after doing some research, it looks like the best approach would have been to create a list and then iterate over the list using @ each, and defining the css

web/src/styles/base/height.scss
FarazA22: spoke to JP, leaving as is due to the limited number of css class names that will be changed

PR #8815, opened 2024-10-02
License data

FarazA22:
### Approach
https://www.pivotaltracker.com/story/show/188369690
Removed the CPA license name from the mapping object b/c the task no longer has license search enabled (we forgot to remove it).

https://www.pivotaltracker.com/story/show/188194087
I created a page that allows folks on the dev site to manipulate license data. If the environment variable is not development, the user will be routed back to the dashboard. 

I was not able to add any unit tests because the environment variable when running jest is `test` (it requires the env variable to be set to dev);  manually changing the variable causes an error because the node environment variable is read-only. Since this functionality only works on dev, I think it's OK not to have any tests. JP is comfortable with it as well.

I highly recommend visiting the site locally and making changes to license data. Once you have completed onboarding, visit http://localhost:3000/mgmt/modifyBusiness

Note: Dillon was only with not adding a link in profile that directs users to the site, the team will need to manually visit the page to make the changes. In an environment that is not dev, if a user navigates to the site they will be routed back to the dasnboard.

## Code author checklist


web/src/pages/mgmt/modifyBusiness.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const isLicenseDataUndefined = !business?.licenseData;
+
+  const [licenseName, setLicenseName] = useState("");
+  const [licenseStatus, setLicenseStatus] = useState("");
+  const [taskID, setTaskId] = useState("");
+
+  const licenseSearchEnabledTaskIds = Object.keys(taskIdLicenseNameMapping).sort();
+
+  useEffect(() => {
+    if (process.env.NODE_ENV !== "development") {
```
FarazA22, line 36: is this the right variable?

web/src/pages/mgmt/modifyBusiness.tsx
```diff
@@ -0,0 +1,208 @@
...
+          <SingleColumnContainer>
+            <PageCircularIndicator />
+          </SingleColumnContainer>
+        </main>
+      </PageSkeleton>
+    );
+  }
+
+  return (
+    <PageSkeleton showNavBar>
```
FarazA22, line 54: I'm not using `Config` since this is an internal facing page.

jphechter: That makes sense. This is a tool we're owning and operating for testing purposes.

web/src/pages/mgmt/modifyBusiness.tsx
jphechter: Pretty sure "development" will only allow this to be visible in local. Trying to focus on PI Planning, but looking for another rec here. I think what you want to check the stage.

jphechter, line 36: [pebble] I'm not super concerned about testing for the functionality in this component, but we should have a test to validate that this doesn't show up on `prod`.

adidner: So we ran into this issue when dealing with favicons initially because they're not all set the way you would intuitively imagine. 

I would instead piggy back off the new favicon logic in _document.tsx

image

web/src/pages/mgmt/modifyBusiness.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const licenseSearchEnabledTaskIds = Object.keys(taskIdLicenseNameMapping).sort();
+
+  useEffect(() => {
+    if (process.env.NODE_ENV !== "development") {
+      router.push(ROUTES.dashboard);
+    }
+  }, [router, business]);
+
+  if (process.env.NODE_ENV !== "development") {
```
adidner, line 41: dust - you could save this value as a variable since you use it twice (but you don't have to)

web/src/pages/mgmt/modifyBusiness.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const [licenseStatus, setLicenseStatus] = useState("");
+  const [taskID, setTaskId] = useState("");
+
+  const licenseSearchEnabledTaskIds = Object.keys(taskIdLicenseNameMapping).sort();
+
+  useEffect(() => {
+    if (process.env.NODE_ENV !== "development") {
+      router.push(ROUTES.dashboard);
+    }
+  }, [router, business]);
```
adidner, line 39: pebble - why business and router? why not an empty array that just runs on mount?

web/src/pages/mgmt/modifyBusiness.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  if (process.env.NODE_ENV !== "development") {
+    return (
+      <PageSkeleton showNavBar>
+        <main className="usa-section padding-top-0 desktop:padding-top-8" id="main">
+          <SingleColumnContainer>
+            <PageCircularIndicator />
+          </SingleColumnContainer>
+        </main>
+      </PageSkeleton>
```
adidner, line 49: dust - maybe we want a reusable component for this? This is exactly the same as the loading page, right?

web/src/pages/mgmt/modifyBusiness.tsx
```diff
@@ -0,0 +1,208 @@
...
+            <div className="margin-top-6">
+              Note 1: to view an anytime actions reinstatement, the license status must be set to expired.
+            </div>
+            <div>
+              Note 2: deleting the license data will reset all license search enabled task statuses to
+              NOT_STARTED
+            </div>
+            <div>
+              Note 3: to view the license details in the task, you must select an industry that has the task
+              on the roadmap
```
adidner, line 90: dust - having notes here is a good idea

web/src/pages/mgmt/modifyBusiness.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  }
+                }}
+              >
+                Delete All License Data
+              </SecondaryButton>
+            </div>
+          </div>
+        </SingleColumnContainer>
+      </main>
+    </PageSkeleton>
```
adidner, line 204: I'm suprised that when I go to the license and look at the check status tab I don't see my expired state?

image
image

did I not set this up correctly? (I did set the license status before logging in)

adidner (approved):
Super glad you advocated for this and made it happen Faraz. 

A few minor comments and tweaks and one question about how it work. Great job

web/src/pages/mgmt/modifyBusiness.tsx
FarazA22: The linter is telling me to add it.

web/src/pages/mgmt/modifyBusiness.tsx
FarazA22: I'll relay this to Dillon. It's not a bug, the tickets were created to enable it in certain places.

PR #8834, opened 2024-10-10
feat: [#186084688] callout alerts are now aria groups

adidner:
## Description

Bascially we wanted the callout inline alerts to be grouped from the perspective of a screen reader

### Ticket

This pull request resolves [#186084688](https://www.pivotaltracker.com/story/show/186084688).

### Steps to Test

Go to a page with an inline calllout, theres one on step 2 of formation under one of the optional parts below, you do have to open it via the link on the right

image

use a screen reader to navigate through it and see that it has structure relative to the screen reader

### Notes

Made a mapping for content such that they can modify these overrides and text completely. 

Also removed the role from this component it's parent handles this already

## Code author checklist

### DIFF
--- a/content/src/mappings/callout-alerts.json
+++ b/content/src/mappings/callout-alerts.json
@@ -1,0 +1,10 @@
+{
+  "calloutAlerts": {
+    "calloutAlertGroupTitle": "${alertVariant}",
+    "info": "informational",
+    "success": "success",
+    "error": "error",
+    "warning": "warning",
+    "note": "note"
+  }
+}
--- a/web/src/components/njwds-extended/Alert.tsx
+++ b/web/src/components/njwds-extended/Alert.tsx
@@ -1,4 +1,6 @@
 import { Heading } from "@/components/njwds-extended/Heading";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { modifyContent } from "@/lib/domain-logic/modifyContent";
 import { ReactElement, ReactNode } from "react";
 
 export interface AlertProps {
@@ -8,6 +10,7 @@ export interface AlertProps {
   rounded?: boolean;
   dataTestid?: string;
   className?: string;
+  noAlertRole?: boolean;
 }
 
 interface Props extends AlertProps {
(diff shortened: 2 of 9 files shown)


web/src/components/njwds-extended/Alert.tsx
```diff
@@ -18,12 +20,21 @@ export const AlertVariants = ["info", "success", "warning", "error", "note"] as
...
+    },
+  });
+};
+
 export const Alert = (props: Props): ReactElement => {
   const { variant, children, noIcon, heading, rounded, dataTestid } = props;
   const variantClass = variant ? `usa-alert--${variant}` : "";
   const noIconClass = noIcon ? " usa-alert--no-icon" : "";
   const roundedClass = rounded ? " radius-md" : "";
-  const alertRole = variant === "error" ? { role: "alert" } : {};
```
adidner, line 26: dust - removed this role as when the element is inlined I think announcing it as a group is better and the popup version has separate logic to make sure it's spoken out loud as an alert.

web/src/lib/search/cmsCollections.ts
```diff
@@ -77,6 +77,7 @@ export const cmsCollections = [
       "404 Page",
       "Global Error Messages",
       "Navigation Config",
```
adidner, line 79: dust - added a new CMS collection for Component Accessibility, this feels like something we need as some components will have specific accessibility content that won't be pairs with any other content.

FarazA22 (approved):

content/src/mappings/callout-alerts.json
```diff
@@ -0,0 +1,10 @@
+{
+  "calloutAlerts": {
+    "calloutAlertGroupTitle": "${alertVariant} alert",
+    "infoOverride": "informational",
+    "successOverride": "success",
+    "errorOverride": "error",
+    "warningOverride": "warning",
+    "noteOverride": "note"
```
jphechter, line 8: [sand] Is "override" the right name for these? It feels like this is just being used to set the type, less override a default value.

jphechter (commented):
One minor thought in addition to what's already been called out. Otherwise this looks good to me 👍

PR #8850, opened 2024-10-11
chore: [#188414852] fix url slug widget in cms for license calendar e…

FarazA22:
### Approach
https://www.pivotaltracker.com/story/show/188414852

There were a few license calendar events that had the same URL slug value, I updated the value with the correct URL slug that was provided by the content team. Separately, there was no reason to make this field a read only field, are you updated the widget so that it uses the no space widget.

## Code author checklist

### DIFF
--- a/web/public/mgmt/config.yml
+++ b/web/public/mgmt/config.yml
@@ -3152,7 +3152,7 @@ collections:
           options: ["expiration", "renewal"],
         }
       - { label: "License Name", name: "licenseName", widget: hidden, default: "" }
-      - { label: "URL Slug", name: "urlSlug", widget: "write-once-read-only-no-space", required: true }
+      - { label: "URL Slug", name: "urlSlug", widget: "no-space", required: true }
       - {
           label: "Expiration Event Label",
           name: "expirationEventDisplayName",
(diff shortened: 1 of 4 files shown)


cbrunefearless (approved):
lgtm

web/public/mgmt/config.yml
```diff
@@ -3152,7 +3152,7 @@ collections:
           options: ["expiration", "renewal"],
         }
       - { label: "License Name", name: "licenseName", widget: hidden, default: "" }
-      - { label: "URL Slug", name: "urlSlug", widget: "write-once-read-only-no-space", required: true }
+      - { label: "URL Slug", name: "urlSlug", widget: "no-space", required: true }
```
adidner, line 3155: pebble - So the only reason I can think for this being write-once-read-only is because it's being bastardized into an ID of sorts for webflow sync? Or maybe being used like an ID somewhere else. 

So I would be awesome if you could double check real quick that this isn't the case and there aren't more weird changes to be made. 

Other than that looks fine.

adidner (approved):
Approved with a comment to be looked into

web/public/mgmt/config.yml
FarazA22: confirmed, we are not using this collection in webflow

PR #8868, opened 2024-10-16
fix: [#188364329] log user out if no userData found

jphechter:
As of 2024-10-16 - this is deployed in the testing site for manual testing.

## Description

If a user attempts to log in and there is no user data found, they will be logged out of the My Account app automatically.

### Ticket

This pull request resolves [#188364329](https://www.pivotaltracker.com/story/show/188364329).

### Approach

Added logging on the backend to better capture login events
- userData fetch attempted with email instead of userID
- userData not found
- server error

Added a check in the `onSignIn` function to validate that `userData` exists. If not, trigger log out flow.

### Steps to Test

- Create a new myNJ account
- Attempt to log into the application without completing onboarding
- The application should immediately log you out

Additionally, you should still be able to complete onboarding as a guest, self-reg for myNJ, link an existing myNJ account with your My Account/Navigator userData, and log into an existing account.

### Notes

There will be further work done on this in the near future. This current workflow is not ideal. It is likely that the user will end up in a behavior loop, but this is significantly better than the software loop that users currently experience.

## Code author checklist

### DIFF
--- a/api/src/api/userRouter.ts
+++ b/api/src/api/userRouter.ts
@@ -8,6 +8,7 @@ import {
   UserDataClient,
 } from "@domain/types";
 import { encryptTaxIdFactory } from "@domain/user/encryptTaxIdFactory";
+import { LogWriterType } from "@libs/logWriter";
 import { NameAvailability } from "@shared/businessNameSearch";
 import { decideABExperience } from "@shared/businessUser";
 import { getCurrentDate, getCurrentDateISOString, parseDate } from "@shared/dateHelpers";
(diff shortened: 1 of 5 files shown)


api/src/api/userRouter.ts
```diff
@@ -136,29 +139,62 @@ export const userRouterFactory = (
...
+    that is an email address.
+
+    TODO: Possible solution - check if the userID is a uuid - if not, log the user out, direct them to onboarding step 1
+    It is likely that this user will enter into a behavior loop where they will then attempt to login again. This is
+    significantly better than the loop behavior that they would otherwise experience. Hopefully, these people will reach
+    out via a support request, at which point we can coach them through the onboarding process.
+    */
+
+    // Check if userId is an email address
+    if (signedInUserId.includes("@")) {
```
jphechter, line 155: We had initially talked about responding with a 401 here. Opted not to do that, just capture the log. Thinking was to have a more broad solution and account for a user who does not have any data in our system. This is not something I know of having come up in Intercom previously.

seidior: (That does cover the case where, like with my user where I deleted my information in DynamoDB, if a user somehow doesn't have information in DDB or it's malformed. But not a top priority to fix right now.)

seidior (commented):
I like the overall strategy, but I'm not sure this solves our immediate issue. It doesn't actually log the user out of myNJ; it instead just gives them an error page which shows this as part of its message:

> This is generally due to a misconfiguration on the part of the application and should be reported to the application's support team or owner.

Ultimately, I think this approach will increase our support burden and still lead to a bad user experience.

Should we reach out to OIT to see what they think in these circumstances? It may also help us figure out why we seem to have such an increase in support requests around this issue.

PR #8906, opened 2024-10-22
Use user data

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/188414550
https://www.pivotaltracker.com/story/show/188448903

### Approach
The issue we are solving is that the user data in the database and updateQueue are not in sync. The user data in the database gets updated with the license status results. However, the updateQueue is not updated until you refresh the page. I added logic to check to see if there is a discrepancy, and if there is then update the updateQueue. I'm also only updating the license data, there is an edge case in which the user might make changes, which would be reflected in the update queue, but not in the user data and the database. 

## Code author checklist


shared/src/test/helpers.ts
```diff
@@ -19,16 +18,3 @@ export const getThirdAnnualFiling = (formationDate: string): string => {
     ? getDateInCurrentYear(formationDate).endOf("month").add(2, "year").format(defaultDateFormat)
     : getDateInCurrentYear(formationDate).endOf("month").add(3, "year").format(defaultDateFormat);
 };
-
-export const modifyCurrentBusiness = (
```
FarazA22, line 23: I removed this from this test file b/c it's used in non-test files.

api/src/client/WebserviceLicenseStatusProcessorClient.test.ts
```diff
@@ -1,10 +1,10 @@
-import { LicenseStatusResults } from "@api/types";
```
FarazA22, line 1: This commit only contains changes to the imports, I temporary separated them so you can focus on the next commit

api/src/api/types.ts
```diff
@@ -1,16 +1,3 @@
-import { LicenseName, LicenseStatus, LicenseStatusItem } from "@shared/license";
-
 export interface ExpressRequestBody<T> extends Express.Request {
   body: T;
 }
-
-export type LicenseChecklistResponse = {
```
FarazA22, line 7: Removed from api to shared because it's now used by the front end as well

shared/src/calendarEvent.ts
```diff
@@ -0,0 +1,19 @@
+import { LicenseName } from "./license";
```
FarazA22, line 1: Due to some circular dependency issue, I move to a separate file

web/src/lib/data-hooks/useUserData.test.tsx
```diff
@@ -246,4 +246,55 @@ describe("useUserData", () => {
       expect(mockApi.postUserData).not.toHaveBeenCalled();
     });
   });
+
```
FarazA22, line 249: I need help figuring oud how to test the new functionality. The commented code is no working and will need to be removed before merging code into main.

web/src/lib/data-hooks/useUserData.ts
```diff
@@ -53,10 +55,22 @@ export const useUserData = (): UseUserDataResponse => {
...
+        businessFromDb: data?.businesses[data.currentBusinessId],
+        businessFromUpdateQueue: updateQueue?.currentBusiness(),
+      });
+    }
+
     if (updateQueue === undefined && data) {
       setUpdateQueue(new UpdateQueueFactory(data, update));
     } else if (updateQueue?.current() === undefined && data) {
       updateQueue?.queue(data);
+    } else if (licenseDataFromDatabaseDataMoreRecent && updateQueue && data) {
```
FarazA22, line 71: I'm not sure how to test this, also I noticed there are no tests for lines 67-70. I'm guessing it was intentional. Perhaps we can't create them using the existing approach.

web/src/lib/utils/licenseStatus.ts
```diff
@@ -0,0 +1,39 @@
...
+  Business,
+  getLicenseTasksProgress,
+  getNonLicenseTasks,
+  LicenseName,
+  Licenses,
+  LicenseStatusResults,
+  UserData,
+} from "@businessnjgovnavigator/shared/";
+
+export const getLicenseStatusResultsFromLicenses = (licenses: Licenses): LicenseStatusResults => {
```
FarazA22, line 11: I created this function b/c the fn that created the task checkbox status expected the argument to pass in LicenseStatusResults. To avoid creating two functions that can determine the correct status, I created a function that modifies the shape so the same one can be used.

web/src/lib/domain-logic/isLicenseDataFromDatabaseDataMoreRecent.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  it("returns true when license data is undefined in businessFromDb but defined in businessFromUpdateQueue", () => {
+    const result = isLicenseDataFromDatabaseDataMoreRecent({
+      businessFromDb: generateBusiness({ licenseData: undefined }),
+      businessFromUpdateQueue: generateBusiness({}),
+    });
+
+    expect(result).toBeTruthy();
+  });
+
+  it("returns true when licence data from businessFromUpdateQueue is not defined but licence data from businessFromDb is", () => {
```
alexander-littleton, line 18:
```suggestion
  it("returns true when license data from businessFromUpdateQueue is not defined but license data from businessFromDb is", () => {
```
[dust] Typos. Also in other tests.

shared/src/domain-logic/licenseStatusHelpers.ts
```diff
@@ -0,0 +1,37 @@
+import { LicenseName, LicenseStatus, LicenseStatusItem, taskIdLicenseNameMapping } from "../license";
+import { Business, TaskProgress } from "../userData";
+
+export const getNonLicenseTasks = (business: Business): Record<string, TaskProgress> =>
+  Object.fromEntries(
+    Object.entries(business.taskProgress).filter(
+      ([key]) => !Object.keys(taskIdLicenseNameMapping).includes(key)
+    )
+  );
```
alexander-littleton, line 9: [dust] - We can just take whatever type `taskProgress` is instead of `Business` as an argument to this function since that's the only field we use. It has the slight benefit of decoupling the function from the structure of a `Business`.

shared/src/domain-logic/licenseStatusHelpers.ts
```diff
@@ -0,0 +1,37 @@
...
+      (taskId) => taskIdLicenseNameMapping[taskId as keyof typeof taskIdLicenseNameMapping] === licenseName
+    );
+    let taskStatus: TaskProgress = "NOT_STARTED";
+    if (licenseStatusResult[licenseName]!.licenseStatus === "ACTIVE") {
+      taskStatus = "COMPLETED";
+    } else if (licenseStatusResult[licenseName]!.licenseStatus === "PENDING") {
+      taskStatus = "IN_PROGRESS";
+    }
+    if (taskId !== undefined) licenseTasksProgress[taskId] = taskStatus;
+  }
```
alexander-littleton, line 34:
```suggestion
  const licenseTasksProgress: Record<string, TaskProgress> = {};
  for (const licenseName of Object.keys(licenseStatusResult) as LicenseName[]) {
    const taskId = Object.keys(taskIdLicenseNameMapping).find(
      (taskId) => taskIdLicenseNameMapping[taskId as keyof typeof taskIdLicenseNameMapping] === licenseName
    );
    if (taskId === undefined) {
      continue
    }
    let taskStatus: TaskProgress = "NOT_STARTED";
    if (licenseStatusResult[licenseName]!.licenseStatus === "ACTIVE") {
      taskStatus = "COMPLETED";
    } else if (licenseStatusResult[licenseName]!.licenseStatus === "PENDING") {
      taskStatus = "IN_PROGRESS";
    }
   licenseTasksProgress[taskId] = taskStatus;
  }
```

I think switching to a guard clause here makes it more explicit that the result of our `find` determines whether or not the code following it is run.

alexander-littleton (approved):

web/src/lib/domain-logic/isLicenseDataFromDatabaseDataMoreRecent.test.ts
```diff
@@ -0,0 +1,80 @@
+import { isLicenseDataFromDatabaseDataMoreRecent } from "@/lib/domain-logic/isLicenseDataFromDatabaseDataMoreRecent";
+import { generateBusiness, generateLicenseData, getCurrentDate } from "@businessnjgovnavigator/shared/";
+
+const currentDate = getCurrentDate();
+const currentDateISOString = currentDate.toISOString();
+const currentDatePlusOneHourISOString = currentDate.add(1, "hour").toISOString();
+
+describe("isLicenseDataFromDatabaseDataMoreRecent", () => {
```
somabadri, line 8: [pebble] all of the expect statements should be explicitly checking `expect(result).toBe(true) or ..toBe(false)` since the function is explicitly returning a boolean

web/src/lib/domain-logic/isLicenseDataFromDatabaseDataMoreRecent.ts
```diff
@@ -0,0 +1,22 @@
...
+import dayjs from "dayjs";
+
+export const isLicenseDataFromDatabaseDataMoreRecent = ({
+  businessFromDb,
+  businessFromUpdateQueue,
+}: {
+  businessFromDb: Business;
+  businessFromUpdateQueue: Business;
+}): boolean => {
+  if (typeof businessFromDb.licenseData !== typeof businessFromUpdateQueue.licenseData) return true;
```
somabadri, line 11: [dust/pebble] I'm a little confused on how different types indicate that one is more recent than the other?

shared/src/domain-logic/licenseStatusHelpers.ts
FarazA22: Hey Lex, we have an established pattern where functions take in the Business as the argument, we typically don't provide specific attributes for the functions. This approach reduces the cognitive load on the devs. I'll leave it as is, but open to discussing during techtro if you think it's worth discussing your suggestion.

web/src/lib/domain-logic/isLicenseDataFromDatabaseDataMoreRecent.ts
FarazA22: license data is initially undefined, if license data is an object for one of the parameters and undefined for the other, then one is more recent.

web/src/lib/domain-logic/isLicenseDataFromDatabaseDataMoreRecent.test.ts
FarazA22: nice catch

PR #8999, opened 2024-10-30
chore: [#188297852] local env defaults to wiremock for formation and business search

FarazA22:
### Approach
https://www.pivotaltracker.com/story/show/188297852

Created env variable that enables wiremock to be used locally, I didn't add tests in the backend b/c we don't have any tests at the app level.

The task status does update when the business is formed.

## Code author checklist


api/src/functions/express/app.ts
```diff
@@ -209,8 +209,9 @@ const dynamicsHousingRegistrationStatusClient = DynamicsHousingRegistrationStatu
   housingPropertyInterestClient: dynamicsHousingPropertyInterestClient,
 });
 
-const BUSINESS_NAME_BASE_URL =
-  process.env.BUSINESS_NAME_BASE_URL || `http://${IS_DOCKER ? "wiremock" : "localhost"}:9000`;
+const BUSINESS_NAME_BASE_URL = process.env.USE_WIREMOCK_FOR_FORMATION_AND_BUSINESS_SEARCH
+  ? "http://localhost:9000"
+  : process.env.BUSINESS_NAME_BASE_URL || `http://${IS_DOCKER ? "wiremock" : "localhost"}:9000`;
```
seidior, line 214: [dust] Just musing here, I wonder if there's ever a situation where `USE_WIREMOCK_FOR_FORMATION_AND_BUSINESS_SEARCH` is false but we point to wiremock anyway because `IS_DOCKER` is true.

Totally fine with the code as written, just curious.

seidior (approved):
Thanks for jumping on this so quickly and helping to unblock the team, Faraz!

api/src/functions/express/app.ts
FarazA22: if that scenario is true, wouldn't everything work as is, meaning the right hand side of the ternary is checked.

aadedejifearless (approved):

PR #9035, opened 2024-11-04
feat: [#186324885] agent information checkbox missing proper notifica…

adidner:
…tion

## Description

We have a checkbox on step 3 of formation that locks some fields and pre populates them. This isn't announced and should be instead moving forward. 

### Ticket

This pull request resolves [#186324885](https://www.pivotaltracker.com/story/show/186324885).

### Steps to Test

Go to step 3 of formation, turn on screen reader
cmd + F5

Navigate to the checkbox, 

Check it, 

See that it announces additional context regarding what happened

### Notes

This was a whole mess and a half. 

We wanted to announce the change and I was super on the fence about just trying to have context visible on the page be enough vs announcing and leaving that aria live region on the page to then later be discovered and having that aria-live region disappear after a few seconds so as to not confuse a screen reader user when they returned. 

I went with the second option and I think in our case it is quite clear but I still don't love it and am not 100% confident I've made best decision. 

At a minimum I did do research, and consulted a number of people so I think I've done the best I could with the context available to me at the moment. 

(The change text is discoverable and divides the UX of the page for screen reader users in that their note block changes to be the alert message whereas a sighted user doesn't have this experience, I think this is reasonable and clear if not a little less than idea in that it's a bit complex and dividing UX is not a great practice. )

## Code author checklist

### DIFF
--- a/web/public/mgmt/config.yml
+++ b/web/public/mgmt/config.yml
@@ -894,6 +894,18 @@ collections:
                       name: "sameContactCheckbox",
                       widget: "string",
                     }
+                  - { label: "Note Title", name: "noteTitle", widget: "string" }
+                  - { label: "Note Body", name: "noteBody", widget: "markdown" }
+                  - {
+                      label: "Checkbox Unchecked Screen Reader Announcement",
+                      name: "checkboxUnCheckedScreenReaderAnnouncement",
+                      widget: "string",
+                    }
+                  - {
+                      label: "Checkbox Checked Screen Reader Announcement",
+                      name: "checkboxCheckedScreenReaderAnnouncement",
+                      widget: "string",
+                    }
               - label: Sections
                 name: sections
                 collapsed: true
--- a/content/src/fieldConfig/business-formation.json
+++ b/content/src/fieldConfig/business-formation.json
@@ -6,7 +6,11 @@
       "radioButtonNumberText": "Enter Your Registered Agent ID Number",
       "radioButtonManualText": "Manually Identify Your Registered Agent",
       "sameAddressCheckbox": "The address is the same as the business location.",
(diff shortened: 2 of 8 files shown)


web/public/mgmt/config.yml
```diff
@@ -894,6 +894,19 @@ collections:
...
+                      label: "Checkbox Unchecked Screen Reader Announcement",
+                      name: "checkboxUnCheckedScreenReaderAnnouncement
+                        ",
+                      widget: "string",
+                    }
+                  - {
+                      label: "Checkbox Checked Screen Reader Announcement",
+                      name: "checkboxCheckedScreenReaderAnnouncement",
+                      widget: "string",
+                    }
```
cbrunefearless, line 909: Did u check if these are searchable in the cms?

adidner: Great callout, I double checked it and you were totally right, I had a syntax error in the search tool. Resolved now.

somabadri (approved):
LGTM! I think you should do a show and tell on this, it can help us include these notifications in the initial iteration instead of going back to it.

adidner:
> LGTM! I think you should do a show and tell on this, it can help us include these notifications in the initial iteration instead of going back to it. Good job, Aaron! good to merge after fixing the failing integration tests

A good suggestion, I think that would be great to talk about aria-live and making sure dynamic changes are announced.  

Although to be completely honest I'm not as confident as I'd like to be that this is 100% the best way to do this. This is kind of an area where the guidelines haven't really fully caught up to how normal apps work and I'm not sure I understand edge cases well enough to advise on general best practices. 

I do want to learn and perhaps I'll consult an expert as digging hasn't gotten me much further recently

PR #9038, opened 2024-11-04
feat: [#188172536] copy businesses data from users table and sync

aadedejifearless:
### Ticket

## Code author checklist


api/src/api/selfRegRouter.test.ts
```diff
@@ -58,7 +60,8 @@ describe("selfRegRouter", () => {
       const hashedKey = generateHashedKey(myNJKey);
       expect(stubSelfRegClient.resume).toHaveBeenCalledWith(myNJKey);
 
-      const putCalledWith = getLastCalledWith(stubUserDataClient.put)[0];
+      const lastCallIndex = stubUserDataClient.put.mock.calls.length - 1;
+      const putCalledWith = stubUserDataClient.put.mock.calls[lastCallIndex][0];
```
aadedejifearless, line 64: Modified the code to achieve the same expectation as the original `getLastCalledWith` function, which retrieves the first argument from the last call to the mocked put function. This refactor simplifies the code by directly accessing the last call's first argument, removing the need for an extra utility function, and avoiding an additional function call.

api/src/api/selfRegRouter.test.ts
```diff
@@ -99,7 +102,8 @@ describe("selfRegRouter", () => {
       expect(stubSelfRegClient.grant).toHaveBeenCalledWith(stubRecordNoKey.user);
       const hashedKey = generateHashedKey(selfRegResponse.myNJUserKey);
 
-      const putCalledWith = getLastCalledWith(stubUserDataClient.put)[0];
+      const lastCallIndex = stubUserDataClient.put.mock.calls.length - 1;
+      const putCalledWith = stubUserDataClient.put.mock.calls[lastCallIndex][0];
```
aadedejifearless, line 106: See previous comment

api/src/api/userRouter.test.ts
```diff
@@ -738,12 +746,10 @@ describe("userRouter", () => {
...
-      expect(
-        parseDate(getLastCalledWith(stubUserDataClient.put)[0].lastUpdatedISO).isSame(
-          getCurrentDate(),
-          "minute"
-        )
-      ).toEqual(true);
+      const lastCallIndex = stubUserDataClient.put.mock.calls.length - 1;
+      const putCalledWith = stubUserDataClient.put.mock.calls[lastCallIndex][0];
+
+      expect(parseDate(putCalledWith.lastUpdatedISO).isSame(getCurrentDate(), "minute")).toEqual(true);
```
aadedejifearless, line 752: Similar to the previous change.

shared/src/userData.ts
```diff
@@ -11,8 +11,8 @@ export interface UserData {
   readonly lastUpdatedISO: string;
   readonly dateCreatedISO: string;
   readonly versionWhenCreated: number;
-  readonly businesses: Record<string, Business>;
-  readonly currentBusinessId: string;
+  businesses: Record<string, Business>;
+  currentBusinessId: string;
```
aadedejifearless, line 15: Removed `readonly` from 'businesses' and 'currentBusinessId' properties to allow mutability. These properties need to be updated when syncing data (e.g., from the users table to the businesses table). Removing 'readonly' ensures that we can modify these properties during the sync process without causing TypeScript errors.

.circleci/config.yml
```diff
@@ -167,6 +167,7 @@ commands:
       - run:
           name: Set Development Environment Variables
           command: |
+            echo 'export ENABLE_BUSINESS_MIGRATION_FROM_USERS_TABLE=ENABLE_BUSINESS_MIGRATION_FROM_USERS_TABLE_DEV' >> $BASH_ENV
```
aadedejifearless, line 170: Add the env variable for all the stages. Currently only Testing is set to true.

jphechter (commented):
I haven't had a chance to do a detailed review of this PR just yet, but I do have a couple thoughts I want to voice.

Firstly, I'm not entirely sure I agree with the strategy here. I think we can accomplish this with fewer overall code changes. The approach also doesn't seem completely in line with patterns we've established elsewhere in the code. I do have more specific thoughts about this and I'll try to leave detailed comments in a bit.

Secondly, the information conveyed in this PR description isn't very helpful. Listing out the files changed isn't something that we typically do and it doesn't give a lot of information. Especially in the case of a PR that's changing fundamental logic, I'd like to see more information about the decisions that were made, the strategy you took in implementing this work, and how a reviewer can approach the changes. I'm happy to talk through this more.

api/src/db/DynamoUserDataClient.ts
aadedejifearless, line 1: Hi @jphechter, I've updated the PR description. I understand that you may have some concerns with the current approach. Could you provide more details on the alternative direction you'd like to explore?

One approach we could consider is separating the concerns of user and business data operations. Right now, the BusinessesDataClient is being updated directly in the user data workflow. Instead, we could separate these operations by creating an independent UserBusinessService that handles both user and business updates separately. This would allow the user updates to trigger a sync for business data without tightly coupling the two, ensuring better separation of concerns and potentially offering more flexibility for future changes.

api/package.json
```diff
@@ -43,6 +43,7 @@
     "@businessnjgovnavigator/shared": "*",
     "@serverless/typescript": "3.38.0",
     "@smithy/node-http-handler": "2.5.0",
+    "@types/aws-lambda": "^8.10.145",
```
FarazA22, line 46: pebble - remove the carat

package.json
```diff
@@ -141,6 +141,7 @@
     "@testing-library/jest-dom": "6.6.3",
     "@testing-library/react": "15.0.7",
     "@testing-library/user-event": "14.5.2",
+    "@types/aws-lambda": "^8.10.145",
```
FarazA22, line 144: pebble - remove carat

api/src/libs/migrateBusinessesFromUsers.test.ts
```diff
@@ -0,0 +1,149 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
seidior, line 2: [dust] I'd prefer if we didn't have to do these. Are they still needed for this?

api/src/functions/oneTimeBusinessMigration/app.ts
seidior, line 1: [boulder] I'd really prefer this to be kicked off using our existing migration infrastructure, if at all possible. That way we can have it appear in our list of changes to user data, and it sets up our migrations to be able to read/write to/from the businesses table too.

seidior (commented):
Overall I like this updated strategy, just a few notes. Thanks Ade!

api/src/libs/migrateBusinessesFromUsers.ts
```diff
@@ -0,0 +1,53 @@
...
+          await businessesDataClient.put({
+            ...businessData,
+            id: businessId,
+            version: currentVersion,
+          });
+
+          await userDataClient.put({
+            ...userData,
+            businesses: { [businessId]: { ...businessData, version: currentVersion } },
+          });
```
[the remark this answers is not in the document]
seidior: This way we can keep track of which users have been migrated and which haven't. This would allow us to resume the migration in case we later need to resume the migration.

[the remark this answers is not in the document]
aadedejifearless: @seidior We are currently incrementing  the migration/version number on line 18.

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -39,6 +39,7 @@ describe("DynamoBusinessesDataClient", () => {
     taxFilingData: generateTaxFilingData({
       filings: [],
     }),
+    version: 1,
```
FarazA22, line 42: pebble - should this be the same version as userData?

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -70,23 +67,15 @@ describe("DynamoBusinessesDataClient", () => {
...
     expect(await dynamoBusinessesDataClient.findAllByNAICSCode("some-naicsCode")).toHaveLength(0);
     await dynamoBusinessesDataClient.put(businessData);
     const expectedValue = await dynamoBusinessesDataClient.findAllByNAICSCode(naicsCode);
     expect(expectedValue[0]).toBeDefined();
-    const { version, ...filteredExpectedValue } = expectedValue[0] as unknown as {
-      version?: string;
-      [key: string]: any;
-    };
-    expect(filteredExpectedValue).toEqual(businessData);
+    expect(expectedValue[0]).toEqual(businessData);
```
FarazA22, line 78: sand - does it make sense to have an assertion where expectedValue has a length of 1?

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -95,11 +84,7 @@ describe("DynamoBusinessesDataClient", () => {
     await dynamoBusinessesDataClient.put(businessData);
     const expectedValue = await dynamoBusinessesDataClient.findAllByIndustry(industry);
     expect(expectedValue[0]).toBeDefined();
```
FarazA22, line 86: sand - does it make sense to have an assertion where expectedValue has a length of 1?

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -37,7 +30,7 @@ export const DynamoBusinessesDataClient = (
           return;
         }
         const item = unmarshall(result.Items[0]);
```
FarazA22, line 32: sand - we could return `unmarshall(result.Items[0]);`

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -101,19 +94,16 @@ export const DynamoBusinessesDataClient = (
...
+      if (!result.Item) {
+        throw new Error("Not found");
+      }
+      return result.Item.data;
+    });
   };
 
   const put = async (businessData: Business): Promise<Business> => {
-    const migratedData = migrateData(businessData);
+    const migratedData = businessData;
```
FarazA22, line 106: pebble - since fn migrateData is removed, we should delete the variable name and reference businessData

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -155,7 +145,7 @@ export const DynamoBusinessesDataClient = (
     return await Promise.all(
       Items.map(async (object: any): Promise<Business> => {
         const data = unmarshall(object).data;
```
FarazA22, line 147: sand - we could return `unmarshall(object).data;`

api/src/db/DynamoUserDataClient.ts
```diff
@@ -141,6 +141,31 @@ export const DynamoUserDataClient = (
     return await search(statement);
   };
 
+  const getUsersWithBusinesses = async (
```
FarazA22, line 144: sand - I have a question about this query, it's likely due to my lack of understanding. It looks like we are performing a query by 1 or many businesses, in what scenario can we return multiple users?

api/src/db/DynamoBusinessesDataClient.test.ts
aadedejifearless: Yh we discussed yesterday, This is going to have the same version as the userData.

package.json
FarazA22, line 144: pebble - remove the carat

shared/src/userData.ts
```diff
@@ -27,9 +27,11 @@ export interface Business {
   readonly preferences: Preferences;
   readonly taxFilingData: TaxFilingData;
   readonly formationData: FormationData;
+  readonly version: number;
 }
 
 export const CURRENT_VERSION = 147;
+export const CURRENT_BUSINESS_VERSION = 1;
```
FarazA22, line 34: we are updating this to 147? Can we reference the line above?

shared/src/userData.ts
```diff
@@ -68,6 +70,7 @@ export const createEmptyBusiness = (id?: string): Business => {
       dbaBusinessNameAvailability: undefined,
       lastVisitedPageIndex: 0,
     },
+    version: 0,
```
FarazA22, line 73: pebble - This fn returns the shape of V147, wouldn't version 0 cause it to go through all the migrations? If we are keeping CURRENT_BUSINESS_VERSION = 1, shouldn't this be 1?

api/package.json
FarazA22, line 46: pebble - remove carat

api/src/libs/migrateBusinessesFromUsers.ts
```diff
@@ -0,0 +1,53 @@
...
+    for (const userData of users) {
+      const businessesData = userData.businesses;
+
+      if (businessesData && Object.keys(businessesData).length > 0) {
+        for (const businessId in businessesData) {
+          const businessData = businessesData[businessId];
+          const existingBusiness = await businessesDataClient.get(businessId);
+          const currentVersion = existingBusiness ? existingBusiness.version + 1 : 1;
+
+          await businessesDataClient.put({
```
FarazA22, line 20: Within `businessesDataClient`, we are no longer running the migration fn, but we are in `userDataClient`. Doesn't this mean the data is not in sync?

api/src/libs/migrateBusinessesFromUsers.ts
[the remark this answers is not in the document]
FarazA22: I think what Spencer is trying to say is the one time migration should be v148 (currently v147 in UserData). The next migration will jump to v149. This way we can track who has migrated. This need to overlap with the Migrations function. 

Also, if my data is on version 99, this increments it to 100, and messes the migration because it's now skipping a migration (v99 to v100).

api/src/libs/usersBusinessesMigration.ts
```diff
@@ -0,0 +1,67 @@
...
+        const businessesData = userData.businesses;
+        for (const businessId in businessesData) {
+          const businessData = businessesData[businessId];
+          const existingBusiness = await businessesDataClient.get(businessId);
+          const currentVersion = existingBusiness ? existingBusiness.version + 1 : 1;
+
+          await businessesDataClient.put({
+            ...businessData,
+            id: businessId,
+            version: currentVersion,
```
FarazA22, line 39: boulder - I'm not clear on the intent of this field. Everytime I update my data, this fn will get updated, which means the current version will increment. What is the value of this field? How does it let us know which state the data is in?

api/src/db/DynamoBusinessesDataClient.test.ts
aadedejifearless: The .toHaveLength(1) assertion is not applicable here because findByEncryptedTaxId returns an object, not an array or iterable. Objects do not have a length property, so this assertion will fail. The correct approach is to use .toEqual(businessData) to verify that the returned object matches the expected data

api/src/libs/usersBusinessesMigration.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        if (!userData || !userData.businesses || Object.keys(userData.businesses).length === 0) {
+          logger.LogError(`No businesses found for user ${userId}. Skipping.`);
+          continue;
+        }
+
+        const businessesData = userData.businesses;
+        for (const businessId in businessesData) {
+          const businessData = businessesData[businessId];
+          const existingBusiness = await businessesDataClient.get(businessId);
```
FarazA22, line 33: Boulder - businessesDataClient no longer performs the migration function, we are taking out dated data and over writing the userData object.

api/src/db/DynamoBusinessesDataClient.test.ts
aadedejifearless: The key difference here is that `findAllByIndustry`returns an array of objects, whereas `findByEncryptedTaxId` returns a single object or undefined. In this test, I'll assert based on the array's length using .toHaveLength(1) to verify that exactly one business is returned. After that, I'll assert the value of the first item in the array, ensuring it matches the businessData. This approach is consistent with the previous test where we assert based on length first and then validate the actual object

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -108,11 +93,7 @@ describe("DynamoBusinessesDataClient", () => {
     await dynamoBusinessesDataClient.put(businessData);
     const expectedValue = await dynamoBusinessesDataClient.findByEncryptedTaxId(encryptedTaxId);
     expect(expectedValue).toBeDefined();
```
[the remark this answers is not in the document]
aadedejifearless: Addressed. The key differences here is we are finding an Array of Objects for some methods and the Objects are returned individually for other methods.

api/src/libs/migrateBusinessesFromUsers.ts
FarazA22: We might store V99 in the business table and v147 (current) in the users table?

FarazA22: This isn't an issue if every user has migrated to the current version (I forgot if we've previously done that)

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -64,7 +57,7 @@ export const DynamoBusinessesDataClient = (
           return [];
         }
         const businesses = result.Items.map((item) => unmarshall(item));
-        return Promise.all(businesses.map(async (item) => await doMigration(item.data)));
+        return Promise.all(businesses.map(async (item) => await item.data));
```
seidior, line 60: Is this a situation where we want to use `Promise.allSettled` instead of `Promise.all`?

api/src/db/DynamoBusinessesDataClient.ts
aadedejifearless: Good call out! I've updated the code to remove the businessData variable and reference it directly as needed, now that the migrateData function has been removed. Thanks for pointing that out!

api/src/db/DynamoUserDataClient.ts
seidior: This code could probably be removed, no? If we're switching to a migration strategy, then we will be migrating all users to a new version rather than selecting only users with businesses.

api/src/db/DynamoBusinessesDataClient.ts
aadedejifearless: I see the suggestion, but we're not returning unmarshall(result.Items[0]) directly. Instead, we are returning the data inside the result of unmarshall(result.Items[0]). Since unmarshall is a method call, I prefer to assign it to a variable first for better clarity and readability. This makes the data transformation more explicit, and I think it improves the overall maintainability of the code. Additionally, this approach keeps up with the consistency of the other data clients.

api/src/db/DynamoUserDataClient.ts
aadedejifearless: The query is designed to fetch multiple users who have associated businesses for batch processing. The goal is to handle large datasets efficiently by processing several users at once. This is useful when migrating or syncing business data across multiple users. By returning multiple users, we can perform the necessary operations in bulk rather than individually, ensuring better scalability and performance. The getUsersWithBusinesses function is responsible for fetching users with businesses, which is why we can return multiple users in this case.

shared/src/userData.ts
aadedejifearless: This was the older implementation. We are not referencing CURRENT_BUSINESS_VERSION going forward

shared/src/userData.ts
aadedejifearless: This has been updated to the latest version.

api/src/libs/migrateBusinessesFromUsers.ts
aadedejifearless: Yh. This would no longer exist. Since we have the lambda function that migrates all the users to the latest version.

api/src/libs/migrateBusinessesFromUsers.ts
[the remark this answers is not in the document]
aadedejifearless: Got what he was saying. I'm talkiing about the concept of the incrementing the versions. From the session yesterday we need to update the versions and reference the CURRENT_VERSION across both datastore/data objects.

api/src/libs/usersBusinessesMigration.ts
aadedejifearless: @FarazA22 It seems like you're asking about the intent behind the `version` field and how it helps us track the state of a business. The current implementation increments the version whenever there's a data change for an existing business. For example, if we update the business name or any other property, it increments the business's version. This aligns with our current pattern, where the version is meant to track changes. Regardless of how many times the business is updated via the user's client, the current logic ensures that both the user and business records remain in sync. There is a test in `usersBusinessesMigration.test.ts` that validates this behavior.

I also want to point out that the requirements have shifted slightly. Initially, the logic was designed to set the version to 1 for newly migrated businesses. However, the updated approach ensures that the version is synchronized with the latest `CURRENT_VERSION` from the user object in our codebase. cc @seidior

api/src/libs/migrateBusinessesFromUsers.test.ts
aadedejifearless: Yes, the suppression (the `any`) is needed. We're disabling @typescript-eslint/no-explicit-any because we only need a partial mock of the Business object with the relevant properties for testing. Implementing all members would unnecessarily complicate the mock. Using any keeps the mock focused and simple without requiring all properties, which isn't needed for the test.

api/src/functions/migrateBusinessesOnUserChange/app.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const dynamoDb = createDynamoDbClient(IS_OFFLINE, IS_DOCKER, DYNAMO_OFFLINE_PORT);
+  const usersDbClient = DynamoUserDataClient(dynamoDb, USERS_TABLE, logger);
+  const businessesDbClient = DynamoBusinessesDataClient(dynamoDb, BUSINESSES_TABLE);
+
+  try {
+    const isBusinessesDataMigrationEnabled =
+      process.env.ENABLE_BUSINESS_MIGRATION_FROM_USERS_TABLE === "true";
+    if (!isBusinessesDataMigrationEnabled) {
+      logger.LogInfo("Businesses migration is disabled");
```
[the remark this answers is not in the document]
aadedejifearless: The current logs in this section are sufficient for identifying whether the migration flag is enabled or not. The logic here is straightforward: if the flag is enabled, we proceed to the migration logic, which already contains detailed logging for debugging. I don't believe additional logs are needed in this part of the code since the key conditions—whether the flag is enabled or not—are already captured. The migration logic itself is where the majority of the complexity lies, and that's where the relevant logs are in place to help debug any issues.

api/src/libs/usersBusinessesMigration.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
FarazA22, line 1: we should remove this b/c any unused vars should be deleted

api/src/libs/migrateBusinessesFromUsers.ts
```diff
@@ -0,0 +1,45 @@
...
+import { UserData } from "@shared/userData";
+
+export const migrateBusinessesFromUsers = async (
+  userDataClient: UserDataClient,
+  businessesDataClient: BusinessesDataClient,
+  logger: LogWriterType
+): Promise<void> => {
+  const processBatch = async (users: UserData[]): Promise<void> => {
+    for (const userData of users) {
+      logger.LogInfo(`Starting migration for userId: ${userData.user.id}`);
```
FarazA22, line 13: Boulder, there was a comment around, tracking when a migration for a user or business is complete. This way, if the fn crashes, we could rerun it and it would skip the user or business that is already copied over.

api/src/db/DynamoUserDataClient.ts
```diff
@@ -141,6 +141,31 @@ export const DynamoUserDataClient = (
...
+    ExclusiveStartKey?: any
```
FarazA22, line 145: This value can be undefined, we should hace tests, will results correctly come back from the db?

api/src/libs/migrateBusinessesFromUsers.ts
```diff
@@ -0,0 +1,45 @@
...
+          businessesDataClient,
+          logger
+        );
+      } else {
+        logger.LogError(`Businesses not found for user ID: ${userData.user.id}`);
+      }
+    }
+  };
+
+  const migrateUsersInPages = async (lastEvaluatedKey?: string): Promise<void> => {
```
FarazA22, line 29: Why is there a parameter if its never used?

api/src/libs/migrateBusinessesFromUsers.ts
```diff
@@ -0,0 +1,45 @@
...
+    const usersWithBusinesses = await userDataClient.getUsersWithBusinesses(lastEvaluatedKey);
+    await processBatch(usersWithBusinesses.users);
+
+    if (usersWithBusinesses.lastEvaluatedKey) {
+      await migrateUsersInPages(usersWithBusinesses.lastEvaluatedKey);
+    }
+  };
+
+  try {
+    await migrateUsersInPages();
```
FarazA22, line 39: to confirm, when `lastEvaluatedKey` is initially not provided, the search, getUsersWithBusinesses will return a value?

api/src/libs/migrationUtils.ts
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { BusinessesDataClient, UserDataClient } from "@domain/types";
+import { Business, CURRENT_VERSION, UserData } from "@shared/userData";
+
+export async function processBusinessesMigration(
+  userData: UserData,
+  businessesData: Record<string, Business>,
```
FarazA22, line 7: In both instances, `migrateBusinessesFromUsers` and `usersBusinessesMigration`, `businessesData` is pulled from `userData`, since that is the case, I think we can remove the parameter.

api/src/libs/migrationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+    logger.LogInfo(`Starting migration for business ID ${businessId} for user ${userData.user.id}`);
+    const businessData = businessesData[businessId];
+    try {
+      const existingBusiness = await businessesDataClient.get(businessId);
+      logger.LogInfo(
+        existingBusiness
+          ? `Found existing business data for business ID ${businessId}, current version: ${existingBusiness.version}`
+          : `No existing business found for business ID ${businessId}, setting to default version`
+      );
+      const currentVersion = existingBusiness ? existingBusiness.version + 1 : CURRENT_VERSION;
```
FarazA22, line 22: This should check if existingBusiness.version exists, so `existingBusiness.version ? ....`

api/src/libs/migrationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+  businessesData: Record<string, Business>,
+  userDataClient: UserDataClient,
+  businessesDataClient: BusinessesDataClient,
+  logger: any
+): Promise<void> {
+  for (const businessId in businessesData) {
+    logger.LogInfo(`Starting migration for business ID ${businessId} for user ${userData.user.id}`);
+    const businessData = businessesData[businessId];
+    try {
+      const existingBusiness = await businessesDataClient.get(businessId);
```
FarazA22, line 16: why do we need to retrieve data from the business table?

api/src/libs/migrationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+        existingBusiness
+          ? `Found existing business data for business ID ${businessId}, current version: ${existingBusiness.version}`
+          : `No existing business found for business ID ${businessId}, setting to default version`
+      );
+      const currentVersion = existingBusiness ? existingBusiness.version + 1 : CURRENT_VERSION;
+
+      await businessesDataClient.put({
+        ...businessData,
+        id: businessId,
+        version: currentVersion,
```
FarazA22, line 27: This value is the "Track Changes" version, where are we maintaining userData migration versioning? Should it also be maintained within each business?

api/src/libs/usersBusinessesMigration.ts
FarazA22: I think it makes sense to create a track changes version, perhaps we should update the variable name so it's more clear.

I'm not sure what you mean by `This aligns with our current pattern, where the version is meant to track changes.` We are tracking UserData version, but that version changes when the shape of the userData is changed, not when a user makes changes.

api/src/db/DynamoUserDataClient.ts
```diff
@@ -24,35 +24,13 @@ const unmarshallOptions = {
...
-      return migration(prevData);
-    } catch (error) {
-      logger.LogError(
-        `Dynamo User Migration Error - Id:${logId} - Error: ${error} - Data: ${JSON.stringify(prevData)}`
-      );
-    }
-  }, data);
-  return { ...migratedData, version: CURRENT_VERSION };
-};
-
```
aadedejifearless, line 48: Removed for reusability across both data clients.

api/src/libs/usersBusinessesMigration.ts
FarazA22: Within `migrateData` in `dynamoDBConfig.ts`, the line `const dataVersion = data.version ?? CURRENT_VERSION;` doesn't make sense to me. 

The methods used in `DynamoBusinessesDataClient` use this function. You previously mentioned the version field within the business object is used for tracking changed, but the methods in DynamoBusinessesDataClient` that use the fn pass in the business object.  We run into an issue b/c the version field in the business object is not the same as the version field for migrations.

api/src/libs/migrateBusinessesFromUsers.ts
aadedejifearless: The function will return a value even if lastEvaluatedKey is not provided initially. When not defined, the query starts from the beginning of the dataset. If additional results are available for pagination, the lastEvaluatedKey from the query response will be returned. This ensures that the function handles both the initial case (no key required) and subsequent calls when pagination is needed.

api/serverless.ts
```diff
@@ -317,6 +319,30 @@ serverlessConfiguration.functions = {
...
+    env.CI
+      ? {
+          securityGroupIds: ["${self:custom.config.infrastructure.SECURITY_GROUP}"],
+          subnetIds: [
+            "${self:custom.config.infrastructure.SUBNET_01}",
+            "${self:custom.config.infrastructure.SUBNET_02}",
+          ],
+        }
+      : undefined
+  ),
```
seidior, line 345: [dust] We're still setting up the one-time migration lambda here. Is this still needed now that we've moved our approach to a migrations-based one?

api/src/libs/migrationUtils.ts
```diff
@@ -0,0 +1,25 @@
+import { BusinessesDataClient } from "@domain/types";
+import { LogWriterType } from "@libs/logWriter";
+import { UserData } from "@shared/userData";
+
+export async function pushBusinessesInUsersDataToBusinessTable(
```
seidior, line 5: [dust] Dustiest of dust comments, but should this be called something like `copyBusinessDataFromUsersTableToBusinessesTable`?

seidior (commented):
Quick re-review for ya so I don't hold you up. Didn't investigate the migration too closely since I assume you worked with @SirSamuelJoseph on this one.

api/src/db/config/dynamoDbConfig.ts
```diff
@@ -22,11 +23,23 @@ export const unmarshallOptions = {
...
+export const migrateData = (data: any, logger: LogWriterType): any => {
+  const logId = logger.GetId();
   const dataVersion = data.version ?? CURRENT_VERSION;
   const migrationsToRun = Migrations.slice(dataVersion);
   const migratedData = migrationsToRun.reduce((prevData: any, migration: MigrationFunction) => {
-    return migration(prevData);
+    try {
+      logger.LogInfo(
+        `Database Migration - Id:${logId} - Upgrading ${data.user.id} from ${prevData.version} to ${
+          Number(prevData.version) + 1
```
FarazA22, line 34: pebble - we are not always migrating by an increment of 1. Once the once time migration occurs, the migrations will occur when they log onto the system. The correct value should be `Migrations.length + 1`

api/src/db/migrations/v149_extract_business_data.ts
```diff
@@ -0,0 +1,861 @@
...
+    version: 149,
+  } as v149UserData;
+};
+
+export const migrate_v148Business_to_v149Business = (business: v148Business): v149Business => {
+  return {
+    ...business,
+    profileData: {
+      ...business.profileData,
+      vacantPropertyOwner: undefined,
```
FarazA22, line 21: pebble - why is this line included?

api/src/db/config/dynamoDbConfig.ts
aadedejifearless: Good catch. Missed this update.

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -101,19 +97,16 @@ export const DynamoBusinessesDataClient = (
...
+      if (!result.Item) {
+        throw new Error("Not found");
+      }
+      return result.Item.data;
+    });
   };
 
   const put = async (businessData: Business): Promise<Business> => {
-    const migratedData = migrateData(businessData);
+    const migratedData = migrateData(businessData, logger);
```
FarazA22, line 109: Why is the migrate data function necessary on the put statement?

FarazA22: It should throw an error b/c migrateData expects userData

api/src/db/config/dynamoDbConfig.ts
```diff
@@ -22,11 +23,23 @@ export const unmarshallOptions = {
 
 export const dynamoDbTranslateConfig = { marshallOptions, unmarshallOptions };
 
-export const migrateData = (data: any): any => {
+export const migrateData = (data: any, logger: LogWriterType): any => {
```
FarazA22, line 26: pebble - type should be UserData not any

FarazA22, line 26: Pebble - I think this function shuould continue to live within `DynamoUserDataClient` - I think it's the only file it should be used in. I left a comment in `DynamoBusinessDataClient` the migration fn can only be used on userData, not the business object.

aadedejifearless: This is going back since its not reusable.

api/src/db/DynamoUserDataClient.ts
```diff
@@ -141,6 +119,31 @@ export const DynamoUserDataClient = (
     return await search(statement);
   };
 
+  const getUsersWithBusinesses = async (
+    ExclusiveStartKey?: any
```
FarazA22, line 123: question, is this type a string?

api/src/db/DynamoUserDataClient.ts
```diff
@@ -141,6 +119,31 @@ export const DynamoUserDataClient = (
...
+  ): Promise<{ users: UserData[]; lastEvaluatedKey: any }> => {
```
FarazA22, line 124: question, is lastEvaluatedKey type a string?

api/src/functions/migrateBusinessesOnUserChange/app.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const dynamoDb = createDynamoDbClient(IS_OFFLINE, IS_DOCKER, DYNAMO_OFFLINE_PORT);
+  const usersDbClient = DynamoUserDataClient(dynamoDb, USERS_TABLE, logger);
+  const businessesDbClient = DynamoBusinessesDataClient(dynamoDb, BUSINESSES_TABLE, logger);
+
+  try {
+    const isBusinessesDataMigrationEnabled =
+      process.env.ENABLE_BUSINESS_MIGRATION_FROM_USERS_TABLE === "true";
+    if (!isBusinessesDataMigrationEnabled) {
+      logger.LogInfo("Businesses migration is disabled");
```
FarazA22, line 27: what value is this log adding?

api/src/libs/migrateBusinessesFromUsers.test.ts
```diff
@@ -0,0 +1,149 @@
...
```
FarazA22, line 1: can we delete?

api/src/libs/migrationUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+  userData: UserData,
+  businessesDataClient: BusinessesDataClient,
+  logger: LogWriterType
+): Promise<void> {
+  for (const businessId in userData.businesses) {
+    logger.LogInfo(`Starting migration for business ID ${businessId} for user ${userData.user.id}`);
+    const businessData = userData.businesses[businessId];
+    try {
+      await businessesDataClient.put({
```
FarazA22, line 14: left a cemment in the file where the put is created, the migration fn would lead to an error

api/src/db/config/dynamoDbConfig.ts
```diff
@@ -22,11 +23,23 @@ export const unmarshallOptions = {
...
-    return migration(prevData);
+    try {
+      logger.LogInfo(
+        `Database Migration - Id:${logId} - Upgrading ${data.user.id} from ${prevData.version} to ${
+          Number(prevData.version) + 1
+        }`
+      );
+      return migration(prevData);
+    } catch (error) {
+      logger.LogError(
```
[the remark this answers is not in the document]
aadedejifearless: this migration only modifies the UserData object and doesn’t involve copying the businesses object to the businesses datastore, I don’t think setting up an SNS notification is necessary.
This process has already been executed around 150 times successfully, which demonstrates its reliability. Adding SNS for this specific scenario may introduce unnecessary complexity without providing significant benefits.

api/src/db/migrations/v149_extract_business_data.ts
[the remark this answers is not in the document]
aadedejifearless: Added a test to do this. Good call out.

api/serverless.ts
aadedejifearless: Yes, the one-time migration is still necessary. The current migrations focus solely on modifying the UserData shape. This particular migration, however, involves copying data to the new table, which is a distinct operation not covered by the existing approach. Therefore, retaining the one-time migration is essential to ensure this data transfer is properly handled.

api/src/db/migrations/v149_extract_business_data.ts
aadedejifearless: I’m not sure which line you’re specifically referring to, but the primary change in this version compared to the previous one is the addition of the version field, i.e `version: 149`. This modification is part of the update being applied to the businesses object to reflect the migration changes

api/src/db/DynamoUserDataClient.ts
aadedejifearless: No, `lastEvaluatedKey` is not a string. Its type is `Record<string, AttributeValue> | undefined`, which represents the key(s) used for continuing pagination in DynamoDB queries. I'm updating it throughout the codebase to ensure consistency and eliminate potential issues related to incorrect type handling.

api/src/functions/migrateBusinessesOnUserChange/app.ts
aadedejifearless: You're right. Removing this as it's unnecessary.

api/src/db/DynamoUserDataClient.ts
aadedejifearless: This is essentially the same as LastEvaluatedKey . Both refer to the mechanism used by DynamoDB for paginating query results and is not a string.

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -101,19 +97,16 @@ export const DynamoBusinessesDataClient = (
...
-
-      .then((result) => {
-        if (!result.Item) {
-          throw new Error("Not found");
-        }
-        return doMigration(result.Item.data);
-      });
+    return db.send(new GetCommand(params)).then((result) => {
+      if (!result.Item) {
+        throw new Error("Not found");
```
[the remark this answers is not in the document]
aadedejifearless: Updated the get method for both the userDataClient and the businessDataClient.

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -22,10 +22,12 @@ describe("DynamoBusinessesDataClient", () => {
 
   let client: DynamoDBDocumentClient;
   let dynamoBusinessesDataClient: BusinessesDataClient;
+  let logger: LogWriterType;
   const formationDate = dayjs().subtract(3, "year").add(1, "month").day(1).format("YYYY-MM-DD");
   const naicsCode = "12345";
   const industry = "test-industry";
   const encryptedTaxId = "test-id-12345";
+  const errorId = "some-id";
```
jphechter, line 30: It looks like a lot of the tests in here are testing both the positive and the negative test cases in a single test. Ex: `finds a business by the businessName` is first checking that a business doesn't exist in the db and then later on does find a business. I'd like to see us explicitly define the negative cases so that we're defining what the negative behavior should be in the test name.

api/src/db/DynamoBusinessesDataClient.test.ts
```diff
@@ -39,25 +41,25 @@ describe("DynamoBusinessesDataClient", () => {
...
     const expectedValue = await dynamoBusinessesDataClient.get(businessId);
-    const { version, ...filteredExpectedValue } = expectedValue as unknown as {
-      version?: string;
-      [key: string]: any;
-    };
-    expect(filteredExpectedValue).toEqual(businessData);
+    expect(expectedValue).toEqual(businessData);
   });
 
   it("finds a business by the businessName", async () => {
```
jphechter, line 65: Ideally our random strings should be truly random. Ex: on line 67 we say
```js
findByBusinessName("some-business-name-59157929")
```

this should be
```js
findByBusinessName("some-business-name-${randomInt()}")
```

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -64,10 +56,10 @@ export const DynamoBusinessesDataClient = (
...
-        return Promise.all(businesses.map(async (item) => await doMigration(item.data)));
+        return Promise.all(businesses.map(async (item) => await item.data));
       })
-      .catch((error) => {
-        console.log(error);
+      .catch((error: Error) => {
+        logger.LogError(`Error finding business by index: ${indexName} with error: ${error.message}`);
         throw new Error("Error retrieving items");
       });
   };
```
jphechter, line 65: [dust] It might also be really helpful to have a function `findAllByBusinessName` given there's a strong likelihood we have multiple businesses with the same name.

api/src/db/DynamoBusinessesDataClient.test.ts
aadedejifearless: Fixed.

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -1,27 +1,19 @@
...
+  tableName: string,
+  logger: LogWriterType
 ): BusinessesDataClient => {
-  const doMigration = async (data: any): Promise<Business> => {
-    const migratedData = migrateData(data);
-    await put(migratedData);
-    return migratedData;
-  };
-
   const findBusinessByIndex = (
```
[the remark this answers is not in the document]
aadedejifearless: Thanks for the feedback! The function only returns a value if exactly one business is found; if no businesses or more than one business is returned, it resolves to undefined. To make this behavior clearer, I’ll update the function name to `findSingleBusinessByIndex`, which more accurately reflects that it expects a single result. This will help prevent any confusion and ensure the function’s purpose is explicit.

api/src/db/DynamoBusinessesDataClient.ts
aadedejifearless: Added.

api/src/db/DynamoBusinessesDataClient.ts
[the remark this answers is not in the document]
jphechter: Oh, my bad, I think you're fine then. I just misread the logic.

jphechter (commented):
I'm still somewhat confused by this approach. I stated this in another comment, but I don't think I agree with this. To me, this feels like it's over-complicating the solution.

Instead of having this be event-driven and requiring additional lambdas to make changes and keep the data in-sync, I'm wondering why we don't just make modifications when we're already handling user data.

api/src/db/migrations/v150_extract_business_data.ts
```diff
@@ -0,0 +1,878 @@
+import { v149Business, v149UserData } from "@db/migrations/v149_add_environment_data";
+import { randomInt } from "@shared/intHelpers";
+
+export const migrate_v149_to_v150 = (v149Data: v149UserData): v150UserData => {
+  return {
+    ...v149Data,
+    businesses: Object.fromEntries(
+      Object.values(v149Data.businesses)
+        .map((business: v149Business) => migrate_v149Business_to_v150Business(business))
+        .map((currBusiness: v150Business) => [currBusiness.id, currBusiness])
```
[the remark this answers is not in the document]
aadedejifearless: You're right @jphechter ; this line has been in the migrations for a while, but the way I understand it is that it is reconstructing the K-V pairs ([id, business]) for Object.fromEntries. Since the id field isn’t being modified in the migration function, this line is redundant.
I’ve removed it in this migration to simplify the code while maintaining the same functionality. Thanks for pointing it out!

api/src/db/migrations/v150_extract_business_data.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        licenses: {
+          ["Pharmacy-Pharmacy"]: generatev149LicenseDetails({
+            licenseStatus: getRandomv149LicenseStatus(),
+          }),
+        },
+      },
+      id,
+    });
+
+    const v149UserData = generatev149UserData({
```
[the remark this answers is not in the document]
aadedejifearless: Updated it, but the current way we set up our migrations means we create these generators for every version. I think we should consider revisiting this approach and potentially focus on generating only the parts we need for each migration. This could help reduce unnecessary duplication and make the migration process more efficient. Thoughts? @jphechter. Point for us to discuss tomorrow.

api/src/db/DynamoBusinessesDataClient.ts
```diff
@@ -1,27 +1,19 @@
...
-import { ExecuteStatementCommand, QueryCommand, QueryCommandInput } from "@aws-sdk/client-dynamodb";
+import { AttributeValue, QueryCommand, QueryCommandInput } from "@aws-sdk/client-dynamodb";
 import { DeleteCommand, DynamoDBDocumentClient, GetCommand, PutCommand } from "@aws-sdk/lib-dynamodb";
 import { unmarshall } from "@aws-sdk/util-dynamodb";
-import { migrateData } from "@db/config/dynamoDbConfig";
 import { BusinessesDataClient } from "@domain/types";
+import { LogWriterType } from "@libs/logWriter";
 import { Business } from "@shared/userData";
 
 export const DynamoBusinessesDataClient = (
```
jphechter, line 8: [dust] This should be singular to be inline with the way we've named the `DynamoUserDataClient`
```suggestion
export const DynamoBusinessDataClient = (
```

api/src/db/UserAndBusinessSyncClient.ts
```diff
@@ -0,0 +1,100 @@
+import { BusinessesDataClient, UnifiedDataClient, UserDataClient } from "@domain/types";
+import { LogWriterType } from "@libs/logWriter";
+import { CURRENT_VERSION, UserData } from "@shared/userData";
+
+export const UserAndBusinessSyncClient = (
+  userDataClient: UserDataClient,
+  businessesDataClient: BusinessesDataClient,
+  logger: LogWriterType
+): UnifiedDataClient => {
+  const migrateUsersAndBusinesses = async (): Promise<{
```
jphechter, line 10: [dust]
```suggestion
  const migrateData = async (): Promise<{
```

api/src/db/UserAndBusinessSyncClient.ts
```diff
@@ -0,0 +1,100 @@
...
+      logger.LogError(errorMessage);
+      return { success: false, error: error instanceof Error ? error.message : "Unknown error" };
+    }
+  };
+
+  const getUserData = async (userId: string): Promise<UserData> => {
+    return await userDataClient.get(userId);
+  };
+
+  const updateUserAndBusinesses = async (userData: UserData): Promise<void> => {
```
jphechter, line 46: [dust] How do you feel about just calling these `get` and `put` respectively?

api/src/db/UserAndBusinessSyncClient.ts
```diff
@@ -0,0 +1,100 @@
...
+  const addUpdatedUserToUsersAndBusinessesTable = async (userData: UserData): Promise<UserData> => {
+    try {
+      await updateUserAndBusinesses(userData);
+      return userData;
+    } catch (error) {
+      logger.LogError(`Failed to process updated user ${userData.user.id}: ${error}`);
+      throw new Error(`Error processing user ${userData.user.id}`);
+    }
+  };
+
```
jphechter, line 89: [sand] Given this function is really just invoking `updateUserAndBusinesses` I think we could delete this function and call `updateUserAndBusinesses` directly.

api/src/db/UserAndBusinessSyncClient.ts
```diff
@@ -0,0 +1,100 @@
...
+  const getUserData = async (userId: string): Promise<UserData> => {
+    return await userDataClient.get(userId);
+  };
+
+  const updateUserAndBusinesses = async (userData: UserData): Promise<void> => {
+    try {
+      await userDataClient.put(userData);
+      logger.LogInfo(`Processed user ${userData.user.id} in the user data table`);
```
jphechter, line 49:
```suggestion
      logger.LogInfo(`Updated user ${userData.user.id} in the user data table`);
```

api/src/db/UserAndBusinessSyncClient.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      if (userData.businesses && Object.keys(userData.businesses).length > 0) {
+        for (const businessId in userData.businesses) {
+          const businessData = userData.businesses[businessId];
+          logger.LogInfo(`Processing business ID ${businessId} for user ${userData.user.id}`);
```
jphechter, line 54:
```suggestion
          logger.LogInfo(`Updated business ${businessId} for user ${userData.user.id}`);
```

api/src/db/UserAndBusinessSyncClient.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+  };
+
+  const addUpdatedUserToUsersAndBusinessesTable = async (userData: UserData): Promise<UserData> => {
+    try {
+      await updateUserAndBusinesses(userData);
+      return userData;
+    } catch (error) {
+      logger.LogError(`Failed to process updated user ${userData.user.id}: ${error}`);
+      throw new Error(`Error processing user ${userData.user.id}`);
```
jphechter, line 86: [sand] I think this should just be the error message, and then we can log a more detailed message.
```suggestion
      throw new Error(error.message);
```

api/src/db/UserAndBusinessSyncClient.ts
```diff
@@ -0,0 +1,100 @@
...
+      throw new Error(errorMessage);
+    }
+  };
+
+  const addUpdatedUserToUsersAndBusinessesTable = async (userData: UserData): Promise<UserData> => {
+    try {
+      await updateUserAndBusinesses(userData);
+      return userData;
+    } catch (error) {
+      logger.LogError(`Failed to process updated user ${userData.user.id}: ${error}`);
```
jphechter, line 85:
```suggestion
      logger.LogError(`Failed to update user ${userData.user.id}: ${error}`);
```

api/src/db/migrations/v151_extract_business_data.test.ts
```diff
@@ -0,0 +1,14 @@
+import { generatev150UserData } from "@db/migrations/v150_remove_needs_nexus_dba_name";
+import { migrate_v150_to_v151 } from "@db/migrations/v151_extract_business_data";
+
+describe("v151 migration adds version field to businesses object", () => {
+  it("should upgrade v150 user by adding version field to businesses object", () => {
+    const v149UserData = generatev150UserData({});
```
jphechter, line 6:
```suggestion
    const v150UserData = generatev150UserData({});
```

api/src/functions/express/app.ts
```diff
@@ -279,6 +278,8 @@ const airtableUserTestingClient = AirtableUserTestingClient(
 const USERS_TABLE = process.env.USERS_TABLE || "users-table-local";
 const dynamoDb = createDynamoDbClient(IS_OFFLINE, IS_DOCKER, DYNAMO_OFFLINE_PORT);
 const userDataClient = DynamoUserDataClient(dynamoDb, USERS_TABLE, dataLogger);
+const businessesDataClient = DynamoBusinessesDataClient(dynamoDb, BUSINESSES_TABLE, dataLogger);
+const unifiedDataClient = UserAndBusinessSyncClient(userDataClient, businessesDataClient, dataLogger);
```
jphechter, line 282: [dust] If we do update the type definition to `DynamoDataClient` I'd like to change this name throughout, too
```suggestion
const dynamoDataClient = UserAndBusinessSyncClient(userDataClient, businessesDataClient, dataLogger);
```

jphechter (commented):
Couple minor comments, but overall, I think this is great

api/src/db/UserAndBusinessSyncClient.ts
aadedejifearless: Fine with it.

PR #9045, opened 2024-11-05
fix: [#188414550] license data for correct business is updated when multiple businesses exist

FarazA22:
### Ticket

https://www.pivotaltracker.com/story/show/188414550

### Approach
Logic was added to ensure the correct business was updated with license data. There can be a discrepancy with the currentBusinessId being one value in the db and another in the queue because the user switched their business. 

## Code author checklist

### DIFF
--- a/web/src/lib/utils/licenseStatus.test.ts
+++ b/web/src/lib/utils/licenseStatus.test.ts
@@ -1,5 +1,12 @@
-import { getLicenseStatusResultsFromLicenses } from "@/lib/utils/licenseStatus";
-import { generateLicenseDetails, taskIdLicenseNameMapping } from "@businessnjgovnavigator/shared/";
+import { getLicenseStatusResultsFromLicenses, licenseDataModifyingFunction } from "@/lib/utils/licenseStatus";
+import {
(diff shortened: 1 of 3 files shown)


web/src/lib/utils/licenseStatus.test.ts
```diff
@@ -32,4 +40,23 @@ describe("licenseStatus", () => {
...
+        [randomElementFromArray(Object.values(taskIdLicenseNameMapping))]: {
+          ...generateLicenseDetails({}),
+        },
+      };
+
+      const userDataInDb = generateUserDataForBusiness(
+        generateBusiness({ id: "id 1", licenseData: generateLicenseData({ licenses }) })
+      );
+      const fn = licenseDataModifyingFunction(userDataInDb, "id 1");
+      const result = (fn(business).licenseData as LicenseData).licenses;
```
FarazA22, line 58: I'm not quite sure how to resolve this if we are not using the as keyword, when I remove it, typescript is screaming at me saying that license date I can be undefined, but the test are passing

jphechter (approved):
I think this all looks good

PR #9080, opened 2024-11-12
feat: [#188507862] Automatic Tax Calendar Check

Sun-Mountain:
### Ticket

This pull request resolves [#188507862](https://www.pivotaltracker.com/story/show/188507862).

## Code author checklist


web/cypress/e2e/auto-tax-filings.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    const businessName = "Make Me Pretty";
+
+    completeNewBusinessOnboarding({ industry });
+
+    completeBusinessStructureTask({ legalStructureId });
+
+    onDashboardPage.getEditProfileLink().should("exist");
+    cy.get('[data-task="apply-for-shop-license"]').first().click();
+    cy.get('button[data-testid="cta-secondary"]').first().click();
+    cy.intercept(`${Cypress.env("API_BASE_URL")}/api/users/*`).as("userAPI");
```
adidner, line 88: dust - so it looks like you're getting a license in to have something actually appear in the calendar when it's rendered rather than an empty calendar?

If thats right, is that necessary? Like would it not be slightly simpler to check for the tax calendar itself not existing or being pending and then check for it as an empty tax calendar? (Or is there some weird logic here that prevents that from being the case)?

web/cypress/e2e/auto-tax-filings.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+  completeBusinessStructureTask,
+  openFormationDateModal,
+  selectDate,
+  selectLocation,
+} from "@businessnjgovnavigator/cypress/support/helpers/helpers";
+import { completeNewBusinessOnboarding } from "@businessnjgovnavigator/cypress/support/helpers/helpers-onboarding";
+import { LookupIndustryById } from "@businessnjgovnavigator/shared/";
+import { onDashboardPage } from "cypress/support/page_objects/dashboardPage";
+
+describe("auto tax filing [feature] [all] [group4]", () => {
```
adidner, line 13: dust - cool that you added a feature test, I would definitely ping marques if you haven't for his awareness as he wanted to do something of an overhaul, so it's just good if he knows when things have appeared.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
```diff
@@ -46,8 +78,76 @@ describe("<SidebarCardFundingNudge />", () => {
...
+  }): UserData => {
+    let formationData: FormationData = {
+      formationFormData: createEmptyFormationFormData(),
+      formationResponse: undefined,
+      getFilingResponse: undefined,
+      completedFilingPayment: false,
+      businessNameAvailability: undefined,
+      dbaBusinessNameAvailability: undefined,
+      lastVisitedPageIndex: 0,
+    };
```
adidner, line 96: pebble - we have a generateEmptyFormationData function, that I think would be better here. 

That being said, couldn't you just make this let `formationData: FormationData | undefined = undefined` or does that lead to some weird type BS later?

web/src/components/dashboard/SidebarCardFundingNudge.tsx
```diff
@@ -13,16 +23,104 @@ type Props = {
 
 export const SidebarCardFundingNudge = (props: Props): ReactElement => {
   const [modalOpen, setModalOpen] = useState<boolean>(false);
+  const [profileData, setProfileData] = useState<ProfileData>(createEmptyProfileData());
```
adidner, line 26: pebble - if you were going to do this in a file where we also reference business?.profileData a bunch I would strongly suggest a more aggressive naming strategy to make it clear when you're using the temporary state variable as opposed to the global state variable. 

That being said I'm not sure what the purpose of this state variable is really? the Setter is only called on line 36 which I think implies that business?.profileData and this profileData variable should always be the same

web/src/components/dashboard/SidebarCardFundingNudge.tsx
```diff
@@ -13,16 +23,104 @@ type Props = {
...
+      profileData.taxId === business.profileData.taxId ? profileData.encryptedTaxId : undefined;
+
+    try {
+      let businessNameToSubmitToTaxApi = "";
+
+      if (displayBusinessName()) {
+        businessNameToSubmitToTaxApi = profileData.businessName;
+      }
+      if (displayResponsibleOwnerName()) {
+        businessNameToSubmitToTaxApi = profileData.responsibleOwnerName;
```
adidner, line 69: dust - This is quite nitpicky and you can totally ignore this. 

I would have structured this a bit differently personally. 

getBusinessNameToSubmitToTaxApi to hold both the getting logic for displayBusinessName and displayResponsibleOwnerName as well as choosing which string to return.

adidner: Oh, it also looks like this logic is repeated below in taxAccess so it would probably be good to throw this in a helper file and reuse it.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
```diff
@@ -46,8 +78,76 @@ describe("<SidebarCardFundingNudge />", () => {
...
+        profileData: {
+          ...business.profileData,
+          municipality: undefined,
+        },
+        taxFilingData: generateTaxFilingData({
+          state: "SUCCESS",
+          businessName: business.profileData.businessName,
+          errorField: undefined,
+        }),
+      });
```
adidner, line 150: dust - so I imagine this setup is needed to keep this from throwing an API call into the void?

web/src/components/dashboard/SidebarCardFundingNudge.tsx
```diff
@@ -13,16 +23,104 @@ type Props = {
...
+  const updateToUpAndRunningAndCompleteTaxTask = async (): Promise<void> => {
+    if (!business || !updateQueue) return;
     await updateQueue
       ?.queueProfileData({
         operatingPhase: OperatingPhaseId.UP_AND_RUNNING,
       })
       .update();
+
+    const encryptedTaxId =
+      profileData.taxId === business.profileData.taxId ? profileData.encryptedTaxId : undefined;
```
adidner, line 60: dust - I don't really understand what this is for? Especially since I think profileData and business.profileData should be the same.

web/src/components/dashboard/SidebarCardFundingNudge.tsx
```diff
@@ -13,16 +23,104 @@ type Props = {
...
+
+      const userDataToSet = await postTaxFilingsOnboarding({
+        taxId: profileData.taxId as string,
+        businessName: businessNameToSubmitToTaxApi,
+        encryptedTaxId: encryptedTaxId as string,
+      });
+
+      updateQueue.queue(userDataToSet).queueProfileData({
+        taxId: profileData.taxId,
+        encryptedTaxId: encryptedTaxId,
```
adidner, line 84: dust - I'm not sure I understand whats going on here exactly. Can you explain it further?

web/src/components/dashboard/SidebarCardFundingNudge.tsx
```diff
@@ -13,16 +23,104 @@ type Props = {
...
+      if (getCurrentBusiness(userDataToSet).taxFilingData.state === "SUCCESS") {
+        if (displayBusinessName()) {
+          updateQueue.queueProfileData({
+            businessName: profileData.businessName,
+          });
+        }
+
+        if (displayResponsibleOwnerName()) {
+          updateQueue.queueProfileData({
+            responsibleOwnerName: profileData.responsibleOwnerName,
```
adidner, line 96: dust - still confused as to why this would have changed in local state.

web/src/components/dashboard/SidebarCardFundingNudge.tsx
```diff
@@ -13,16 +23,104 @@ type Props = {
 
 export const SidebarCardFundingNudge = (props: Props): ReactElement => {
   const [modalOpen, setModalOpen] = useState<boolean>(false);
+  const [profileData, setProfileData] = useState<ProfileData>(createEmptyProfileData());
+
+  const { queueUpdateTaskProgress } = useUpdateTaskProgress();
+  const close = (): void => {};
+  const onSuccess = (): void => {};
```
adidner, line 30: dust - I'm not sure I understand the purpose of these 2 voice functions?

web/src/lib/tax-access/taxAccess.ts
```diff
@@ -0,0 +1,83 @@
...
+    "responsibleOwnerName"
+  );
+};
+
+export const taxAccess = async ({
+  business,
+  profileData,
+  setIsLoading,
+  updateQueue,
+}: TaxAccessProps): Promise<void> => {
```
adidner, line 35: We tend to prefer the props variable over this unraveled style. The reason is because it's easier to see if something is passed in. 

(props: TaxAccessProps)

if I see something like props.profileData I know it's passed vs if I see profileData somewhere in this code I might assume it comes from the global context hook

web/src/lib/tax-access/taxAccess.ts
```diff
@@ -0,0 +1,83 @@
...
+  );
+};
+
+export const displayResponsibleOwnerName = (business?: Business): boolean => {
+  return LookupLegalStructureById(business?.profileData.legalStructureId).elementsToDisplay.has(
+    "responsibleOwnerName"
+  );
+};
+
+export const taxAccess = async ({
```
adidner, line 30: dust - think you meant to use this code in the SidebarCardFunding nudge? I'm not sure where this is used at the moment as I did a global search and didn't find anything but it could be my IDE

web/cypress/support/helpers/helpers.ts
```diff
@@ -69,6 +69,24 @@ export const clickDeferredSaveButton = () => {
   return cy.get(`button[data-testid="deferred-question-save"]`).first().click();
 };
 
+export const clickModalSaveButton = (): void => {
```
adidner, line 72: dust - This is a great refactor

web/src/components/Task.tsx
```diff
@@ -49,6 +49,7 @@ export const Task = (props: Props): ReactElement => {
               href={`/tasks/${props.task.urlSlug}`}
               className={`usa-link margin-right-105 ${props.task.required ? "text-bold" : ""}`}
               data-task={props.task.id}
+              data-testid={props.task.id}
             >
```
adidner, line 53: dust - Curious about if it would be preferable to use the props.task.name below instead. No strong opinions on my end.

Sun-Mountain: Cypress was being really finicky in what it wanted to pick up and this was the most assured way I could guarantee it would select the correct link.

api/wiremock/__files/tax-calendar-lookup-success.json
```diff
@@ -0,0 +1,127 @@
+{
+  "ApiKey": "a6d23c77-483a-438d-b342-6a4241928b44",
+  "Data": [
+    {
+      "Name": "taxpayer-id",
+      "Value": "843010944"
```
jphechter, line 6: [boulder] Let's make this a more nonsense number. I have no reason to believe this is a real tax ID, but given it's a piece of private data, let's make sure it's definitely not something that could be mistaken as real.

Maybe something like `000000000`

api/wiremock/__files/tax-calendar-lookup-success.json
```diff
@@ -0,0 +1,127 @@
...
+    },
+    {
+      "Name": "taxpayer-location",
+      "Value": "003"
+    },
+    {
+      "Name": "tax-business-name",
+      "Value": "NEW MAGIC FINGER MASSAGE CENTER CORP."
```
jphechter, line 14: Is this a real business, or is this something we made up? If it's real, this should be changed too.

web/cypress/e2e/auto-tax-filings.spec.ts
jphechter: @mjackson21 @FarazA22

web/cypress/e2e/auto-tax-filings.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    cy.get('[data-testid="register-for-taxes"]').first().click();
+    cy.get('input[name="taxId"]').type("123456789098");
+    cy.get("button").contains("Save").click();
+    cy.get(`[data-testid="back-to-dashboard"]`).first().click({ force: true });
+    cy.get('[data-testid="cta-funding-nudge"]').first().click();
+    cy.get('[data-testid="get-tax-access"]').should("not.exist");
+    cy.get('[data-testid="alert-content-container"]').should("exist");
+  });
+
+  it("does not complete auto tax filing if missing business name", () => {
```
jphechter, line 49: [sand]
```suggestion
  it("does not automatically register for Gov2Go and retrieve tax filing if missing business name", () => {
```

web/cypress/e2e/auto-tax-filings.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+import { completeNewBusinessOnboarding } from "@businessnjgovnavigator/cypress/support/helpers/helpers-onboarding";
+import { LookupIndustryById } from "@businessnjgovnavigator/shared/";
+import { onDashboardPage } from "cypress/support/page_objects/dashboardPage";
+
+describe("auto tax filing [feature] [all] [group4]", () => {
+  beforeEach(() => {
+    cy.loginByCognitoApi();
+  });
+
+  it("successfully completes auto tax filing with business name and tax Id", () => {
```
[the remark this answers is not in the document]
jphechter: We don't actually do anything to register the user for taxes. We do register them for Gov2Go which is a third party service. The only thing we're getting back are tax filing events if the user has already registered their business to pay taxes.

web/cypress/e2e/auto-tax-filings.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    const legalStructureId = "limited-liability-company";
+    const businessName = "Make Me Pretty";
+
+    completeNewBusinessOnboarding({ industry });
+
+    completeBusinessStructureTask({ legalStructureId });
+
+    onDashboardPage.getEditProfileLink().should("exist");
+    cy.get('[data-task="apply-for-shop-license"]').first().click();
+    cy.get('button[data-testid="cta-secondary"]').first().click();
```
jphechter, line 29: Why are we interacting with a license task? It doesn't look like this is doing anything to modify the data.

web/cypress/e2e/auto-tax-filings.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    clickModalSaveButton();
+
+    cy.get('[data-testid="register-for-taxes"]').first().click();
+    cy.get('input[name="taxId"]').type("123456789098");
+    cy.get("button").contains("Save").click();
+    cy.get(`[data-testid="back-to-dashboard"]`).first().click({ force: true });
+    cy.get('[data-testid="cta-funding-nudge"]').first().click();
+    cy.get('[data-testid="get-tax-access"]').should("not.exist");
+    cy.get('[data-testid="alert-content-container"]').should("exist");
+  });
```
jphechter, line 47: Should we add something in here to check that one of the tax filing events is listed in the calendar?

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
Sun-Mountain: Yes.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
jphechter: [boulder] we should lean into the generators wherever possible. If the structure changes later it will cause this test to fail.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
```diff
@@ -46,8 +78,76 @@ describe("<SidebarCardFundingNudge />", () => {
     card = generateSidebarCardContent({ id: fundingNudge });
   });
 
+  const generateTaxFilingUserData = (params: {
```
jphechter, line 81: [sand] It looks like this is only being used in one place. If that's the case I think there's some logic in here that isn't necessary and may over-complicate the function.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
```diff
@@ -46,8 +78,76 @@ describe("<SidebarCardFundingNudge />", () => {
...
+        taxFilingData: generateTaxFilingData({ state: undefined }),
+      })
+    );
+  };
+
   describe("when clicking funding button for non-generic industry", () => {
+    const userDataWithPrefilledFields = generateTaxFilingUserData({
+      publicFiling: true,
+      taxId: "123456789123",
+      businessName: "MrFakesHotDogBonanza",
```
jphechter, line 135: [sand] If we're using the same setup every time, I don't know that it necessarily makes sense for us to pass these as props.

jphechter: Thinking specifically because this function is only used in one place.

web/src/components/dashboard/SidebarCardFundingNudge.tsx
Sun-Mountain: Renamed to `tempProfiledata` in both `SidebarCardFundingNudge.tsx` and `TaxAccessStepTwo.tsx`.

web/src/components/dashboard/SidebarCardFundingNudge.tsx
Sun-Mountain: Attempts at refactoring this (i.e. -- moving this logic to a `lib` folder to reuse in `SidebarCardFundingNudge.tsx`) or to put into a switch function (or similar) caused the jest tests to fail.

web/src/components/dashboard/SidebarCardFundingNudge.tsx
Sun-Mountain: Attempts at refactoring this (i.e. -- moving this logic to a `lib` folder to reuse in `SidebarCardFundingNudge.tsx`) caused the jest tests to fail.

api/wiremock/__files/tax-calendar-lookup-success.json
jphechter: Marked unresolved. Just want to confirm that this actually isn't a real business name.

jphechter (commented):
@Sun-Mountain Did you recently rebase? It looks like there are a number of changes in here that should be in `main` but are represented in this diff.

jphechter (commented):
Minor note - every time we push we trigger our CI pipelines. Normally this is fine. We want to see our tests run. The kicker is that we are charged for the minutes required to run the workflow.

We don't want to discourage people from pushing work in progress, there's value there. Instead, if you know the work is in progress and you don't need the full pipeline to run, you can add "[skip ci]" to your commit message to skip the CI pipeline.

Screenshot 2024-11-19 at 1 11 22 PM

api/wiremock/__files/tax-calendar-lookup-success.json
jphechter: I'm taking this being marked "resolved" again that this is in-fact not a real business name. Just wanted the verbal confirmation.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
jphechter: This is still true. I think we can simplify this function.

I'm assuming the intention was to write something in a way that it can be reused if needed. Given this is only being used in one place, I'd opt for making this as simple as possible.

web/src/components/dashboard/SidebarCardFundingNudge.tsx
```diff
@@ -13,16 +29,93 @@ type Props = {
...
+
+    if (taxFilingData.state === "PENDING") {
+      analytics.event.tax_calendar_modal.submit.business_exists_but_not_in_Gov2Go();
+      close();
+    }
+
+    if (taxFilingData.state === "FAILED") {
+      analytics.event.tax_calendar_modal.submit.tax_calendar_business_does_not_exist();
+    }
+
```
jphechter, line 118: [boulder] This whole block of code is almost identical to what's in `TaxAccessStepTwo`. I don't think I realized this before looking at this a second time. There's a good opportunity to do this DRY.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
```diff
@@ -46,8 +88,68 @@ describe("<SidebarCardFundingNudge />", () => {
...
+          businessName: params.businessName || "",
+          responsibleOwnerName: params.responsibleOwnerName || "",
+        }),
+        formationData: formationData,
+        taxFilingData: generateTaxFilingData({ state: undefined }),
+      })
+    );
+  };
+
   describe("when clicking funding button for non-generic industry", () => {
```
jphechter, line 133: [boulder] We should talk about the testing strategy here. We're mocking the API response in one of the tests, but we're not actually doing anything to assert that logic is executing aside from making the tests pass.

web/src/components/filings-calendar/tax-access-modal/TaxAccessStepTwo.test.tsx
```diff
@@ -42,6 +42,7 @@ const mockApi = api as jest.Mocked<typeof api>;
 const Config = getMergedConfig();
 
 const renderComponent = (initialUserData?: UserData): void => {
+  console.log(initialUserData?.businesses[initialUserData.currentBusinessId].profileData.businessName);
```
jphechter, line 45: Delete

web/src/components/filings-calendar/tax-access-modal/TaxAccessStepTwo.tsx
```diff
@@ -44,7 +51,7 @@ export const TaxAccessStepTwo = (props: Props): ReactElement => {
   const { queueUpdateTaskProgress } = useUpdateTaskProgress();
   const business = props.CMS_ONLY_fakeBusiness ?? updateQueue?.currentBusiness();
 
-  const [profileData, setProfileData] = useState<ProfileData>(createEmptyProfileData());
+  const [tempProfileData, setTempProfileData] = useState<ProfileData>(createEmptyProfileData());
```
jphechter, line 54: [sand] I know this was discussed in a previous comment, but I'm not sure that I agree with this name change. Fully recognize that the naming might be confusing, but this is a pattern we follow throughout the application. If we want to change how this is handled, I'd rather us address that in a single refactor and keeping naming consistent.

Same comment for `SidebarCardFundingNudge`.

web/src/components/dashboard/SidebarCardFundingNudge.test.tsx
jphechter: [boulder] OK, after pulling up the code locally to look into this further, I'm realizing that a lot of this setup is being copied directly from `TaxAccessStepTwo.test.tsx`. Additionally, there's a lot of logic we're pulling from `TaxAccessStepTwo` into `SidebarCardFundingNudge` without actually writing any tests around it.

I think we should talk about the approach here. We can write this code in a way that's DRY and consolidate our test logic here.

PR #9141, opened 2024-11-22
chore: [#188526874] fix action bars,  tab styling, and external link buttons for consistency

FarazA22:
### Approach
https://www.pivotaltracker.com/story/show/188526874

There are some changes that I've made within this PR that addressed a few different things:
1. In some instances we are using a primary button that sends a user to an external site. I updated the logic so that the icon that represents the external link is used consistently across the app. When searching for code in our code base, the code with styling is applied in the same manner.

2. In many parts of the app were use material UI tabs; some instances contain overriding the padding to 0, I applied global styling to the component to zero. Now we can apply styling per the designs as needed. I went ahead and updated certain elements to ensure that the styling looks consistent.

3. Some changes were made in a previous story that caused an issue with the CTA container component. I reversed these changes because the implementation did not account for all the CTA's that exist in the application. The button can float to the left (without the action bar), float to the right (with the action bar) and float left and right (with the action bar)

4. I noticed that one of the external link buttons was not opening a new tab, I went ahead and added the functionality and added a test.

## Code author checklist


web/src/lib/muiTheme.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
FarazA22, line 2: This has been added because typescript is letting me know that MuiTabPanel doesn't exist, but it actually does.

web/src/styles/components/buttons.scss
```diff
@@ -249,3 +249,10 @@
   padding: 0;
   text-decoration: underline !important;
 }
+
+.usa-icon-button-margin {
```
FarazA22, line 253: This is used to control the spacing and padding when a launch icon is used inside a button

PR #9150, opened 2024-11-25
feat: [#187194162] add profile address to dakota persona

AnJuHyppolite:
## Description

This feature allows a Dakota user to access address fields in their profile with the ability to toggle between U.S. and international fields. 

### Ticket

This pull request resolves [#187194162](https://www.pivotaltracker.com/story/show/187194162).

### Approach
**NOTE:** This story incorporates a refactor that merges the logic from `ProfileAddressNewJersey` and the _new_ U.S. and international logic into `ProfileAddress`. There is also specific logic for `Starting` and `Up-And-Running` businesses to ensure `businessLocationType` for these personas defaults to `NJ`. Otherwise, the U.S. and International radio buttons will render. 

I also updated `AddressContext` to point to FormationAddress because the Address interface doesn't include `businessLocationType`.

### Steps to Test

**DAKOTA NEXUS**
1. Onboard as a Dakota Nexus business.
2. Go to the business profile. 
3. There is a Business Address section that has two radio buttons: U.S. and international.
4. Select a location.
5. The corresponding address fields will render.
6. Input data. Note that error messages will render if the address is incomplete.
7. Experiment with toggling locations. Doing so will clear any data you entered.
8. Once you've selected a location, enter an address, and save.
9. Complete the business structure so you can complete formation.
10. Once in the `Business Step`, the address you entered in the profile will persist. 
- Any changes made in this step will persist in the profile (and vice versa)
11. Complete formation.
12.  Go to the business profile. 
13. The address you entered will be read-only.
**NOTE:** If you do not enter an address and the business forms. `Address Not Provided` will render in the read-only fields.

**REMOTE WORKER/SELLER**
1. Onboard as a Remote Seller or Worker.
2. Go to the business profile. 
3. There is a Business Address section that has two radio buttons: U.S. and international.
4. Select a location.
5. The corresponding address fields will render.
6. Input data. Note that error messages will render if the address is incomplete.
7. Experiment with toggling locations. Doing so will clear any data you entered.

## Code author checklist


api/src/db/migrations/migrations.ts
```diff
@@ -298,6 +299,7 @@ export const Migrations: MigrationFunction[] = [
   migrate_v144_to_v145,
   migrate_v145_to_v146,
   migrate_v146_to_v147,
+  migrate_v147_to_v148,
 ];
 
 export { generatev144UserData as CURRENT_GENERATOR } from "@db/migrations/v144_add_raffle_bingo_games";
```
jphechter, line 305: IF we add a migration, then we need to advance this line.

It looks like that's been missed a couple times. There's a test that should be checking this, but for some reason that didn't trigger.

jphechter (changes requested):
Talked about this with @AnJuHyppolite on Zoom. I think there's a better way to go about addressing this work without modifying types and needing a migration.

We talked about expanding the `AddressContext` to use `FormationAddress` to account for the additional fields needed in the Dakota Business Profile international address.

web/src/components/profile/ProfileAddress.test.tsx
```diff
@@ -0,0 +1,824 @@
...
+        })
+      );
+      renderNjBusinessComponent({});
+
+      expect(screen.getByLabelText("Address line1")).toHaveValue("");
+      fillText("Address line1", "1111 Home Alone");
+      expect(screen.getByLabelText("Address line1")).toHaveValue("1111 Home Alone");
+    });
+
+    it("renders profile address line 1 error", () => {
```
alexander-littleton, line 105:
```suggestion
    it("displays an error on address line 1 when field is blurred while empty", () => {
```

This seems to align a bit closer with what is in the test.

web/src/components/profile/ProfileAddress.test.tsx
```diff
@@ -0,0 +1,824 @@
...
+          }),
+        })
+      );
+      renderNjBusinessComponent({});
+
+      blurInput("Address line1");
+      expect(screen.getByText(Config.formation.fields.addressLine1.error)).toBeInTheDocument();
+    });
+
+    it("renders profile address line 2 field with input value", () => {
```
alexander-littleton, line 123: [sand]: the tests for address line 2 could benefit from similar updates to what was in my line 1 suggestions

web/src/components/profile/ProfileAddress.test.tsx
```diff
@@ -0,0 +1,824 @@
...
+            formationFormData: generateFormationFormData({ ...address }),
+          }),
+        })
+      );
+      renderNjBusinessComponent({});
+
+      blurInput("Address line2");
+      expect(screen.getByText(Config.formation.fields.addressLine1.error)).toBeInTheDocument();
+      expect(screen.getByText(Config.formation.fields.addressMunicipality.error)).toBeInTheDocument();
+      expect(screen.getByText(Config.formation.fields.addressZipCode.error)).toBeInTheDocument();
```
alexander-littleton, line 177: [pebble]: I was anticipating an expect statement for `addressLine2.error` in here. That probably needs to be the only expect statement here or the test name should probably be changed to reflect these 3 expects.

api/src/api/userRouter.ts
```diff
@@ -253,13 +253,19 @@ export const userRouterFactory = (
...
         addressLine1: formationFormData.addressLine1,
         addressLine2: formationFormData.addressLine2,
         addressCity: formationFormData.addressCity,
         addressMunicipality: formationFormData.addressMunicipality,
+        addressState: formationFormData.addressState,
+        addressCountry: formationFormData.addressCountry,
         addressZipCode: formationFormData.addressZipCode,
+        addressProvince: formationFormData.addressProvince,
+        businessLocationType: formationFormData.businessLocationType,
       };
```
jphechter, line 267: I'm confused what the point of this object is. It doesn't look like it's doing anything.

shared/src/formationData.ts
```diff
@@ -104,6 +104,14 @@ export const emptyAddressData: Address = {
   addressZipCode: "",
 };
 
+export const emptyFormationAddressData: FormationAddress = {
+  ...emptyAddressData,
+  addressCity: "",
+  addressProvince: "",
```
jphechter, line 110: [sand]
```suggestion
  addressCity: undefined,
  addressProvince: undefined,
```

web/src/components/data-fields/address/AddressMunicipalityDropdown.test.tsx
```diff
@@ -10,12 +10,12 @@ describe("<AddressMunicipalityDropdown  />", () => {
     addressData,
     municipalities,
   }: {
-    addressData?: Address;
+    addressData?: FormationAddress;
```
jphechter, line 13: [sand] If we're changing this type, it might make sense to update the variable name to match.
```suggestion
    formationAddressData?: FormationAddress;
```

web/src/components/data-fields/address/AddressTextField.test.tsx
```diff
@@ -11,13 +11,13 @@ describe("<AddressTextField  />", () => {
     addressData,
     municipalities,
   }: {
-    fieldName: keyof Address;
-    addressData?: Address;
+    fieldName: keyof FormationAddress;
+    addressData?: FormationAddress;
```
jphechter, line 15: [sand] Same comment as above
```suggestion
    formationAddressData?: FormationAddress;
```

web/src/components/profile/getErrorStateForAddressField.ts
```diff
@@ -1,21 +1,23 @@
...
-import { Address } from "@businessnjgovnavigator/shared/index";
+import { FormationAddress } from "@businessnjgovnavigator/shared/index";
 
 export const getErrorStateForAddressField = ({
   field,
   addressData,
 }: {
   field: FieldsForAddressErrorHandling;
-  addressData: Address;
+  addressData: FormationAddress;
```
jphechter, line 20: [sand] Same comment as above
```suggestion
  formationAddressData: FormationAddress;
```

web/src/components/profile/ProfileAddressLockedFields.test.tsx
```diff
@@ -17,17 +23,21 @@ function mockMaterialUI(): typeof materialUi {
...
+  beforeEach(() => {
+    useMockBusiness({});
+  });
+
   const renderComponent = ({
     address,
     municipalities,
   }: {
-    address?: Address;
+    address?: FormationAddress;
```
jphechter, line 34: @AnJuHyppolite I'm not going to tag each instance in the PR, but my rec here would be that any place we use the type `FormationAddress` the corresponding variable name should probably reference `formationAddress`.

web/src/components/profile/ProfileAddressLockedFields.test.tsx
```diff
@@ -36,57 +46,237 @@ describe("ProfileAddressLockedFields", () => {
     );
   };
```
jphechter, line 48: This is a test

web/src/components/profile/ProfileAddressLockedFields.test.tsx
```diff
@@ -36,57 +46,237 @@ describe("ProfileAddressLockedFields", () => {
...
+
+      expect(screen.getByTestId("locked-profileBusinessAddressTooltip")).toBeInTheDocument();
+      const addressLine1 = screen.getByTestId("locked-profileAddressLine1");
+      expect(within(addressLine1).getByText("1111 Home Alone")).toBeInTheDocument();
+      const addressMuniStateZip = screen.getByTestId("locked-profileAddressMuniStateZip");
+      expect(within(addressMuniStateZip).getByText("Allendale, NJ 00893")).toBeInTheDocument();
+      expect(screen.queryByTestId("locked-profileAddressLine2")).not.toBeInTheDocument();
+    });
+
+    it("renders profile address locked fields with Address Line 2", () => {
```
jphechter, line 75: [sand] This test is asserting that all possible information is being rendered. I think I'd like this to be the first test and the base case, and then the existing test `renders required profile address locked fields` be changed to assert that addressLine2 is not rendered when the data isn't present.

web/src/components/profile/ProfileAddressLockedFields.test.tsx
```diff
@@ -36,57 +46,237 @@ describe("ProfileAddressLockedFields", () => {
...
+      const address = generateAddress({
+        addressLine1: "",
+        addressLine2: "",
+        addressMunicipality: undefined,
+        addressState: undefined,
+        addressZipCode: "",
+      });
+      renderComponent({ address });
+
+      expect(screen.getByTestId("locked-profileBusinessAddressTooltip")).toBeInTheDocument();
```
jphechter, line 119: [sand] add a check to confirm `profileAddressNotProvided` copy is present.

web/src/components/profile/ProfileAddressLockedFields.tsx
```diff
@@ -2,12 +2,63 @@ import { ArrowTooltip } from "@/components/ArrowTooltip";
...
 export const ProfileAddressLockedFields = (): ReactElement => {
   const { state } = useContext(AddressContext);
+  const { business } = useUserData();
 
   const Config = getMergedConfig();
+
+  const displayAddress = (): ReactNode => {
+    const businessLocation = business?.formationData.formationFormData.businessLocationType;
+    switch (businessLocation) {
+      case "NJ":
```
jphechter, line 17: [pebble] I think this `switch` could be further simplified. NJ and US addresses are rendered identically. Additionally, all 3 cases have the same logic for how `addressLine2` is conditionally displayed.

Doing this would also allow you to reduce the number of test cases you have to manage and account for. Right now, there's a describe for each case and you're covering these conditions:
- complete address is present
- missing addressLine2
- no address provided

That comes out to 9 test cases for what I'm seeing as 4 conditions.
- complete address
- missing addressLine2
- International address uses province
- no address provided

web/src/components/profile/ProfileAddressLockedFields.test.tsx
```diff
@@ -17,17 +23,21 @@ function mockMaterialUI(): typeof materialUi {
 }
 
 describe("ProfileAddressLockedFields", () => {
+  beforeEach(() => {
+    useMockBusiness({});
+  });
```
alexander-littleton, line 28: [sand] - I did notice that we're calling useMockBusiness at the beginning of each test so this beforeEach call might be redundant

web/src/components/profile/getErrorStateForAddressField.ts
```diff
@@ -96,24 +105,80 @@ export const getErrorStateForAddressField = ({
...
+    const isValidNjZipCode = isZipCodeNj(formationAddressData[field]);
+    const isIntlPostalCode = isZipCodeIntl(formationAddressData[field]);
+
+    if (formationAddressData.addressState?.shortCode === "NJ") {
+      inRange = isValidNjZipCode;
+    } else if (formationAddressData.addressState?.shortCode) {
+      inRange = isValidUsZipCode;
+    } else {
+      inRange = isIntlPostalCode;
+    }
```
jphechter, line 161: [dust] You're already tracking a var `inRange` that is getting set based on the result from each of these `isZipCodeX` functions. I don't think you need those additional vars. Personal preference would be to trim it down a bit.

```suggestion
    if (formationAddressData.addressState?.shortCode === "NJ") {
      inRange = isZipCodeNj(formationAddressData[field]);
    } else if (formationAddressData.addressState?.shortCode) {
      inRange = isZipCodeUs(formationAddressData[field]);
    } else {
      inRange = isZipCodeIntl(formationAddressData[field]);
    }
```

web/src/components/profile/getErrorStateForAddressField.ts
```diff
@@ -96,24 +105,80 @@ export const getErrorStateForAddressField = ({
...
+
+    const hasError = exists && !inRange;
+
+    const zipCodeErrorLabel =
+      formationAddressData.addressState?.shortCode === "NJ"
+        ? (Config.formation.fields as any)[field].error
+        : (Config.formation.fields as any)[field].foreign.errorUS;
```
jphechter, line 168: [boulder] By this logic, an International address will get the US error message. There's an `errorIntl` error message that they should see.

web/src/components/tasks/business-formation/review/ReviewMainBusinessLocation.tsx
```diff
@@ -29,10 +29,19 @@ export const ReviewMainBusinessLocation = (): ReactElement => {
...
+          <ReviewLineItem
+            label={Config.formation.fields.addressProvince.label}
+            value={getAddressState(state.formationFormData)}
+          />
+        ) : (
+          <ReviewLineItem
+            label={Config.formation.fields.addressState.label}
+            value={getAddressState(state.formationFormData)}
+          />
+        )}
```
jphechter, line 43: Nice catch!

web/src/lib/data-hooks/useFormContextFieldHelpers.ts
```diff
@@ -65,10 +65,10 @@ export const useFormContextFieldHelpers = <T, FieldError = FieldErrorType>(
   // eslint-disable-next-line react-hooks/rules-of-hooks
   const isFormFieldInValid = useMemo(
     () => {
-      return fieldStates[fieldName].invalid === true;
+      return fieldStates[fieldName]?.invalid === true;
```
jphechter, line 68: Why did we need to make this change?

I get wary when we start modifying the underlying data-hooks just because how much we use these throughout the application.

web/test/pages/profile/profile-foreign.test.tsx
```diff
@@ -282,6 +282,59 @@ describe("profile-foreign", () => {
...
+          }),
+        }),
+      });
+      renderPage({ business });
+
+      expect(screen.getByTestId("locked-profileAddressLine1")).toBeInTheDocument();
+      expect(screen.getByTestId("locked-profileAddressCityStateZip")).toBeInTheDocument();
+    });
+
+    it("does not render locked profile address fields when business has not been formed", () => {
```
jphechter, line 311: [sand] The goal of this test should be to assert that there are editable fields on the page, not just that we don't see the locked form correct?

```suggestion
    it("renders editable profile address fields when business has not been formed", () => {
```

We'll also want to validate that those editable fields are on the page.

jphechter: @AnJuHyppolite I'd like for us to address this still.

api/src/api/userRouter.ts
jphechter: Re-reading this I understand. We're purposefully clearing formation data but trying to persist the address.

I thought this was doing something different.

web/src/components/profile/ProfileAddress.tsx
```diff
@@ -0,0 +1,383 @@
+import { CountryDropdown } from "@/components/CountryDropdown";
+import { AddressMunicipalityDropdown } from "@/components/data-fields/address/AddressMunicipalityDropdown";
+import { AddressTextField } from "@/components/data-fields/address/AddressTextField";
+import { ModifiedContent } from "@/components/ModifiedContent";
+import { ProfileAddressLockedFields } from "@/components/profile/ProfileAddressLockedFields";
+import { StateDropdown } from "@/components/StateDropdown";
+import { WithErrorBar } from "@/components/WithErrorBar";
```
cbrunefearless, line 7: Great job refactoring everything into a single component!

jphechter (commented):
1 minor comment I'd still like addressed in `profile-foreign.test.tsx`, but I'm happy with the rest. Nice work!

PR #9159, opened 2024-11-26
feat: [#188540248] add check email endpoint

hinzed1127:
### Ticket

This pull request resolves [#188540248](https://www.pivotaltracker.com/story/show/188540248).

#### POST vs GET

~The original spec had the email being searched being supplied via request body. I think a GET + query parameter makes more semantic sense for what we’re trying to do here, so updating to that approach.~ As detailed [below](https://github.com/newjersey/navigator.business.nj.gov/pull/9159#discussion_r1868231687), the endpoint is a POST endpoint which takes the email via a JSON request body. This is to ensure that requests don't get cached.

#### `findByEmail`

The `findByEmail` method is only getting used here currently, so having it return the first match if there's multiple (which corresponds to the most recently updated match) seems fine for now, although @alexander-littleton's suggestion of a newer `findAllByEmail` method might be a good thing to add.

### Steps to Test

Use the [shared endpoints config ](https://github.com/newjersey/business-endpoints)to test with Bruno.

## Code author checklist

### DIFF
--- a/api/src/db/DynamoUserDataClient.ts
+++ b/api/src/db/DynamoUserDataClient.ts
@@ -69,7 +69,8 @@ export const DynamoUserDataClient = (
     return db
       .send(new QueryCommand(params))
       .then((result) => {
-        if (!result.Items || result.Items.length !== 1) {
+        // implicitly returns only the first match if multiple matches are found
+        if (!result.Items || result.Items.length === 0) {
           return;
         }
         return doMigration(unmarshall(result.Items[0], unmarshallOptions).data);
--- a/api/.dependency-cruiser.js
+++ b/api/.dependency-cruiser.js
@@ -201,6 +201,10 @@ module.exports = {
       from: { path: "src/db" },
       to: { path: "src/libs" },
     },
+    {
+      from: { path: "src/api" },
+      to: { path: "src/libs" },
+    },
   ],
   forbidden: [
     /* rules from the 'recommended' preset: */
--- a/api/src/api/userRouter.test.ts
+++ b/api/src/api/userRouter.test.ts
@@ -3,6 +3,7 @@
 import { userRouterFactory } from "@api/userRouter";
 import { EncryptionDecryptionClient, TimeStampBusinessSearch, UserDataClient } from "@domain/types";
 import { setupExpress } from "@libs/express";
+import { DummyLogWriter } from "@libs/logWriter";
 import { getCurrentDate, parseDate } from "@shared/dateHelpers";
 import { getCurrentBusiness } from "@shared/domain-logic/getCurrentBusiness";
 import { modifyCurrentBusiness } from "@shared/domain-logic/modifyCurrentBusiness";
@@ -32,7 +33,7 @@ import dayjs from "dayjs";
 import { Express } from "express";
 import { StatusCodes } from "http-status-codes";
 import jwt from "jsonwebtoken";
-import request from "supertest";
+import request, { type Response } from "supertest";
 
 jest.mock("jsonwebtoken", () => {
   return {
@@ -107,7 +108,8 @@ describe("userRouter", () => {
         stubUpdateRoadmapSidebarCards,
         stubUpdateOperatingPhase,
         stubEncryptionDecryptionClient,
-        stubTimeStampBusinessSearch
+        stubTimeStampBusinessSearch,
+        DummyLogWriter
       )
     );
   });
@@ -118,7 +120,7 @@ describe("userRouter", () => {
     });
   });
 
-  describe("GET", () => {
+  describe("GET /users/:userId", () => {
     it("gets user with id", async () => {
       const userData = generateUserData({});
       stubUserDataClient.get.mockResolvedValue(userData);
@@ -704,6 +706,41 @@ describe("userRouter", () => {
           ).toEqual(true);
         });
       });
+    });
+  });
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const sendRequest = async (body: any): Promise<Response> => {
+    return request(app).post(`/users/emailCheck`).send(body);
+  };
+
+  describe("POST /users/emailCheck", () => {
+    it("is called without an email property and return an error", async () => {
+      const response = await sendRequest({ notEmail: "user@example.org" });
+
+      expect(response.status).toEqual(StatusCodes.BAD_REQUEST);
+      expect(response.body).toEqual({ error: "`email` property required." });
+    });
+
+    it("looks up a user by email that does not exist and returns an error", async () => {
+      // eslint-disable-next-line unicorn/no-useless-undefined
+      stubUserDataClient.findByEmail.mockResolvedValue(undefined);
+
+      const response = await sendRequest({ email: "user@example.org" });
+
+      expect(response.status).toEqual(StatusCodes.NOT_FOUND);
+      expect(response.body).toEqual({ email: "user@example.org", found: false });
+    });
+
+    it("looks up a user by email that does exist and returns successful", async () => {
+      const mockUser = generateUserData({});
+
+      stubUserDataClient.findByEmail.mockResolvedValue(mockUser);
+
+      const response = await sendRequest({ email: mockUser.user.email });
+
+      expect(response.status).toEqual(StatusCodes.OK);
+      expect(response.body).toEqual({ email: mockUser.user.email, found: true });
     });
   });
 
--- a/api/src/api/userRouter.ts
+++ b/api/src/api/userRouter.ts
@@ -8,6 +8,7 @@ import {
   UserDataClient,
 } from "@domain/types";
 import { encryptTaxIdFactory } from "@domain/user/encryptTaxIdFactory";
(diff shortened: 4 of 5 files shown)


api/src/api/emailCheckRouter.ts
```diff
@@ -0,0 +1,34 @@
+import { UserDataClient } from "@domain/types";
+import type { LogWriterType } from "@libs/logWriter";
+import { Router } from "express";
+import { StatusCodes } from "http-status-codes";
+
+export const emailCheckRouterFactory = (userDataClient: UserDataClient, logWriter: LogWriterType): Router => {
```
jphechter, line 6: [pebble] How would y'all feel about moving this under `userRouterFactory`? We have a number of other endpoints that deal with getting or updating user data in that factory and this feels like an extension of that existing logic.

api/src/api/emailCheckRouter.ts
```diff
@@ -0,0 +1,34 @@
...
+  const router = Router();
+
+  router.post("/emailCheck", async (req, res) => {
```
jphechter, line 9: [sand] If you do opt to move this under `userRouterFactory` I'd vote for also nesting the endpoint like this:
```suggestion
  router.post("/users/emailCheck", async (req, res) => {
```

jphechter, line 9: [pebble] Given we're not actually modifying any information, really just trying to assert whether a value exists or not, I think this should be a `GET`.
```suggestion
  router.get("/emailCheck", async (req, res) => {
```

api/src/api/emailCheckRouter.ts
```diff
@@ -0,0 +1,34 @@
...
+    const { email } = req.body;
+
+    const logId = logWriter.GetId();
+    let status;
+
+    if (email === undefined) {
+      status = StatusCodes.BAD_REQUEST;
+      res.status(status).send({ error: "`email` property required." });
+    } else {
+      const userData = await userDataClient.findByEmail(email);
```
jphechter, line 19: [boulder] I think it's unlikely, but this logic may trigger some false negative cases. I just looked into the `findByEmail` because I'm not super familiar and it looks like if there is a match on an email address, it will return userData. IF there are multiple matches, it will return undefined.

Based on your current logic, an email address associated with multiple accounts will return undefined and the endpoint will return a 404.

I think we're either going to need to wrap this call in a try/catch or chain then/catch calls. Not super partial on the solution, but I do think this is an issue worth addressing.

jphechter: The key piece is that we need to be looking for a resolved promise not necessarily checking for userData being undefined.

api/src/api/emailCheckRouter.ts
```diff
@@ -0,0 +1,34 @@
...
+
+      if (userData) {
+        status = StatusCodes.OK;
+        res.status(status).send({ email, found: true });
+      } else {
+        status = StatusCodes.NOT_FOUND;
+        res.status(status).send({ email, found: false });
+      }
```
jphechter, line 27: Just kindof a general thought, do we need a response body or is the status code enough for us to work with on the frontend?

api/src/api/emailCheckRouter.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    };
+    app = setupExpress(false);
+    app.use(emailCheckRouterFactory(stubUserDataClient, DummyLogWriter));
+  });
+
+  afterAll(async () => {
+    await new Promise((resolve) => {
+      return setTimeout(resolve, 500);
+    });
+  });
```
jphechter, line 34: Why do we need this?

api/src/api/emailCheckRouter.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  describe("POST emailCheck", () => {
+    it("checks without email property and gets error", async () => {
+      const response = await sendRequest({ notEmail: "user@example.org" });
+      expect(response.status).toEqual(StatusCodes.BAD_REQUEST);
+      expect(response.body).toEqual({ error: "`email` property required." });
+    });
+
+    it("checks for user's existence, looking them up by their email, and they don't exist", async () => {
```
jphechter, line 48:
```suggestion
    it("looks up a user by email that does not exist and receives an error", async () => {
```

api/src/api/emailCheckRouter.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      // eslint-disable-next-line unicorn/no-useless-undefined
+      stubUserDataClient.findByEmail.mockResolvedValue(undefined);
+
+      const response = await sendRequest({ email: "user@example.org" });
+
+      expect(response.status).toEqual(StatusCodes.NOT_FOUND);
+      expect(response.body).toEqual({ email: "user@example.org", found: false });
+    });
+
+    it("checks for user's existence, looking them up by their email, and they do exist", async () => {
```
jphechter, line 58:
```suggestion
    it("looks up a user by email that does exist and returns successful", async () => {
```

api/src/api/emailCheckRouter.ts
jphechter: This might inadvertently fix your lint issue.

api/src/api/emailCheckRouter.test.ts
hinzed1127: This was a copy-paste casualty, but now that you mention it, I'm not sure why this is needed in other files, either? It's in all the `api/src/api/*.test.ts` files, but I commented it out in every test file in the directory and the whole test suite still passed.

My guess is that it was added to avoid some asynchronous timing issues causing tests to fail, but maybe that's a dated assumption and they're safe to remove?

The point is moot for this file, since I'm fine consolidating with `userRouterFactory`, but this could be a good follow-up refactor for the other test files

api/src/api/emailCheckRouter.ts
alexander-littleton: @jphechter Is this opportunity for a new `findAllByEmail` method?

api/src/api/emailCheckRouter.ts
hinzed1127: I’m fine with that. The original spec had the email being searched being supplied via request body. I think a GET + query parameter makes more semantic sense for what we’re trying to do here, so updating to that approach

api/src/api/emailCheckRouter.ts
hinzed1127: I think @alexander-littleton's suggestion is worth considering. Right now, though, it looks like `findByEmail` isn't currently being used anywhere else. I'm guessing it either got made as a utility or was used somewhere previously and then the call got removed without removing the function.

Given that, I think it's ok to update this to return the first email match if multiple users are found. To @jphechter's point about it being unlikely, I'm guessing this is more of a testing environment occurrence rather than a prod problem.

api/src/api/emailCheckRouter.ts
seidior: @jphechter Spec on the ticket says `POST`, and that was my choice. `GET` requests are often cached, and rather than spend time on doing cache busting and diagnostics around cache busting, it was easier to choose `POST` even though it's not REST-y, since `POST` requests don't get cached by default.

If it was an authenticated endpoint, then I'd say `GET` all the way. Open to the conversation tho.

@hinzed1127 FYI, it's unusual but not non-standard to send a body in a `GET` request. `GET` doesn't _necessarily_ require query parameters. Still, my suggestion would be to stick with a `POST` request.

api/src/db/DynamoUserDataClient.ts
```diff
@@ -69,9 +69,10 @@ export const DynamoUserDataClient = (
     return db
       .send(new QueryCommand(params))
       .then((result) => {
-        if (!result.Items || result.Items.length !== 1) {
+        if (!result.Items || result.Items.length === 0) {
           return;
         }
+        // implicitly returns only the first match if multiple matches are found
```
jphechter, line 75: Good use of comment IMO. This is something that might not be immediately obvious to someone, but is definitely worth noting.

jphechter: [dust] Totally a nit, I might move this up to 72 to be the first line of this callback function.

api/src/api/emailCheckRouter.ts
jphechter: Cool with me. Semantically, I'd lean on `GET` here, but for ease of implementation I don't have a strong feeling. IF we ever do come back and start to build out our API in some way, it is a change I'd like to see happen.

But that's probably a long, long time from now.

jphechter:
[sand] I noticed there are 5 commits here. Do we need all 5? Are each of these independently deployable? Is it worth squashing these into a single commit?

jphechter:
Happy with the end result. Minor comment regarding the number of commits. Otherwise, LGTM

hinzed1127:
> Is it worth squashing these into a single commit?

Yeah, I used semantic commits just to make the changes a little clearer to review, but I'm gonna squash it down to one before merging 👍

PR #9251, opened 2024-12-09
fix: [#188464297] update design and content & feat: [#188651554] remove registration modal & amend button text

somabadri:
### Ticket

This pull request resolves [#188464297](https://www.pivotaltracker.com/story/show/188464297) and resolves [#188651554](https://www.pivotaltracker.com/story/show/188651554).

### Approach

Made updates based on review comments in the fix commit and the feat commit includes improvements to the feature based on further discussion with the team.

### Steps to Test

1) Select `all other industries` during onboarding.
2) The task is found under Step 3 of the roadmap.
## Code author checklist

### DIFF
--- a/web/src/components/tasks/environment-questionnaire/CheckWastePermitsResults.tsx
+++ b/web/src/components/tasks/environment-questionnaire/CheckWastePermitsResults.tsx
@@ -111,7 +111,13 @@ export const CheckWastePermitsResults = (props: Props): ReactElement => {
   const highApplicability = (): ReactElement => {
     return (
       <>
-        <Content className={"padding-bottom-3"}>{Config.wasteQuestionnaireResultsPage.summary}</Content>
+        <div className={"padding-bottom-3"}>
+          {Config.wasteQuestionnaireResultsPage.summaryPtOne}
+          <span className={"text-bold margin-x-05"}>
+            {Config.wasteQuestionnaireResultsPage.summaryMediaArea}
+          </span>
+          {Config.wasteQuestionnaireResultsPage.summaryPtTwo}
+        </div>
         {personalizedSupport()}
         {contactDep()}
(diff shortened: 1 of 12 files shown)


web/src/components/tasks/environment-questionnaire/CheckWastePermitsResults.tsx
```diff
@@ -111,7 +111,13 @@ export const CheckWastePermitsResults = (props: Props): ReactElement => {
   const highApplicability = (): ReactElement => {
     return (
       <>
-        <Content className={"padding-bottom-3"}>{Config.wasteQuestionnaireResultsPage.summary}</Content>
+        <div className={"padding-bottom-3"}>
```
alexander-littleton, line 114: [dust] - Outside of the scope of this ticket but just hopping on my soapbox here - I find it easier to reason through a component when the body (before the return) is reserved for defining behavior instead of functions that return JSX. It's more of a consistency thing for me than anything personally. When I implemented this pattern in the past, it was typically because the component felt too small to break out into its own file. I actually find it easier to reason about the component lifecycle and data flow when these components are inlined or broken out into their own files. There are performance considerations as well but that seems to be less of a consideration for our codebase with how sparingly we use memoization in general. Just food for thought as others write/read through our codebase!

alexander-littleton (approved):
Quick + easy read - approved!

web/src/components/tasks/environment-questionnaire/CheckWastePermitsResults.tsx
somabadri: I was debating whether to break them out into their own files or keep them in the same file. I ultimately decided to reconsider this when they become more complex with the tickets that are to follow. However, I'm glad you brought this up because we were thinking along similar lines.

PR #9277, opened 2024-12-16
feat: [#187466951] business structure task, error is now auto scroll and screen reader accessible

adidner:
## Description

basically we wanted screen readers getting errors on the business structure selection page to get refocused to the error at the top instead of just staying at the bottom of the page. 

### Ticket

This pull request resolves [#187466951](https://www.pivotaltracker.com/story/show/187466951).

### Approach

Used the scroll to top and focus function I made early which is awesome

### Steps to Test

Go to the business structure task with a new business. Turn on a screen reader, save without selecting and see that scrolls  to top and reads out information

### Notes

This was a real weird story in some ways. 

I kinda had a set of weird decision to make in terms of functionality here. I ended up deciding to focus the error group because if I focus any content in the group itself it starts getting super verbose and redundant reading out the group and the error itself and it gets messy. 

I thought about making the error group not a group so that it would read cleaner which is still an option but kinda also feels a bit wrong? Idk

probably also need a bit of a content change to the naming of the error groups or perhaps a follow up to add a bit more content to the error group name here. 

## Code author checklist

### DIFF
--- a/web/src/components/njwds-extended/Alert.tsx
+++ b/web/src/components/njwds-extended/Alert.tsx
@@ -1,7 +1,7 @@
 import { Heading } from "@/components/njwds-extended/Heading";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { modifyContent } from "@/lib/domain-logic/modifyContent";
-import { ReactElement, ReactNode } from "react";
+import { ForwardedRef, forwardRef, ReactElement, ReactNode } from "react";
 
 export interface AlertProps {
   variant: AlertVariant;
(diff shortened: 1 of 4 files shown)


web/src/components/njwds-extended/Alert.tsx
```diff
@@ -77,4 +79,8 @@ export const Alert = (props: Props): ReactElement => {
       </div>
     </div>
   );
-};
+});
+
+Alert.displayName = "Alert";
+
+export { Alert };
```
adidner, line 86: Did this, because of some ESlint complaints due to using forwardRef (apparently its to make debugging easier)

FarazA22 (approved):
nice work

PR #9283, opened 2024-12-17
chore:[#186705729] add query find user by business name

aadedejifearless:
## Description

This PR introduces a new feature to search for users by their business name using a query in AWS DynamoDB. The approach involves creating an SQL query statement in the DynamoUserDataClient that searches for users with businesses. The query checks for the existence of businesses and ensures that the businessName field is not missing.

Once the query is executed, the results are ingested into the DynamoDataClient. There, the business names are normalized (trimmed and converted to lowercase) to ensure case-insensitive matching when filtering for users with a business name that matches the search term.

### Ticket

This pull request resolves [#186705729](https://www.pivotaltracker.com/story/show/186705729).

### Approach

- Added `queryUsersWithBusinesses` in `DynamoUserDataClient` to execute a DynamoDB query that retrieves users with businesses.
- Implemented `findUserByBusinessName` in `DynamoDataClient`, which filters users based on a normalized, case-insensitive business name. 
- Also updated the Router tests where the mock/stub `DynamoDataClient` lives as well as the Batch files where the mock/stub DynamoUserDataClient lives. 

### Steps to Test
1.  Run Unit Tests
 The following tests were created to verify the functionality of the changes here and previous existing logic in the DynamoDataClient.
 
 - `should find user by businessName ...`: This test ensures that the findUserByBusinessName method filters users correctly by matching business names, including case-insensitive matches.
 - `should migrate data for multiple users with outdated versions...`: This test verifies the migration of outdated user data, ensuring the business data is updated and logged appropriately
 
 2. Run the test suite with `yarn test` and ensure the tests pass successfully. 
 

## Code author checklist


api/src/db/DynamoUserDataClient.ts
```diff
@@ -137,13 +137,19 @@ export const DynamoUserDataClient = (
...
+  };
+
+  const queryUsersWithBusinesses = (): Promise<UserData[]> => {
+    const statement = `SELECT data FROM "${tableName}" WHERE data["businesses"] IS NOT MISSING AND size(data["businesses"]) > 0`;
+    return search(statement);
   };
 
   const search = async (statement: string): Promise<UserData[]> => {
     const { Items = [] } = await db.send(new ExecuteStatementCommand({ Statement: statement }));
+    console.log("ITEEEEEMS", Items);
```
seidior, line 152: [pebble] Did we mean to include this?

api/src/db/DynamoDataClient.ts
```diff
@@ -72,6 +72,16 @@ export const DynamoDataClient = (
     }
   };
 
+  const findUserByBusinessName = async (businessName: string): Promise<UserData[]> => {
+    const usersWithBusinesses = await userDataClient.queryUsersWithBusinesses();
+    const normalizedBusinessName = businessName.trim().toLowerCase();
+    return usersWithBusinesses.filter((user: UserData) => {
+      return Object.values(user.businesses).some(
+        (business: Business) => business.profileData?.businessName?.toLowerCase() === normalizedBusinessName
```
seidior, line 80: [boulder] How will this work when business data is no longer inside user data?

api/src/db/DynamoUserDataClient.ts
```diff
@@ -137,13 +137,19 @@ export const DynamoUserDataClient = (
     return search(statement);
   };
 
-  const getUsersWithOutdatedVersion = async (latestVersion: number): Promise<UserData[]> => {
+  const getUsersWithOutdatedVersion = (latestVersion: number): Promise<UserData[]> => {
```
aadedejifearless, line 140: Removed redundant async keyword in getUsersWithOutdatedVersion function.

seidior (commented):
Something I forgot to mention in the review of the last DB split ticket: is there anything in the new business structure that points back to the UUID of the user? This may make the work in this ticket a bit easier, since you in theory would just be able to query the secondary index on the businesses table, get the business data, then have the UUID of its owner, and return that object.

api/src/db/DynamoDataClient.ts
```diff
@@ -72,6 +72,16 @@ export const DynamoDataClient = (
     }
   };
 
+  const findUserByBusinessName = async (businessName: string): Promise<UserData[]> => {
+    const usersWithBusinesses = await userDataClient.queryUsersWithBusinesses();
```
aadedejifearless, line 76: Added normalization (trimming and lowercasing) of businessName for case-insensitive and whitespace-agnostic comparison. to prevent false mismatches caused by whitespace/case sensitivity.

seidior: We may want some sort of 'fuzzy' matching, since we may not have the exact business name when looking up businesses.

api/src/db/DynamoUserDataClient.ts
aadedejifearless: Removed this.

api/src/db/DynamoDataClient.ts
aadedejifearless: Currently, the workflow relies on business data being inside the user data structure. In the planned migration, we will remove the businesses object from userData and make slight updates to the code.

We will introduce a GSI (`currentBusinessId`), the common field between the users and businesses tables. The updated flow will:
- Query the businessesTable to find businesses by name with fuzzy (matching: this is now implemented in the current state).
- Extract the currentBusinessId from the matched business.
- Query the usersTable using currentBusinessId to find the associated users.
let me know your thoughts @seidior . this also answers your previous questions on what field do we have that ties both user and businesses currently.

seidior: To sum up what we discussed in our meeting just now:

- The current migration for splitting out users and businesses and keeping them in sync should be modified (or a follow-up migration should be added) so that, in addition to a version field being added to the business record, we also add a field that is called something like "owningUser" that is the UUID of the user data record.
- We then can modify this work to use the secondary index for business names on the businesses table to find the corresponding business record, then use this "owningUser" field in the business record to get the user data and return that.
  - This means we don't have to search through all businesses of all users for the correct user data, we just have to find the fuzzy-matching businesses and return those user data objects. This should _vastly_ speed up the performance of this endpoint.

PR #9292, opened 2024-12-19
fix: [#188172536] add userId to business object

aadedejifearless:
## Description
This PR introduces a major change by adding a userId field to the business object. This addition ensures that each business is associated with a specific user, which will support future features involving the business-user relationship. The generateBusiness method has been updated to take an argument of UserData to include the userId in the business object

### Ticket

This pull request resolves [#188172536](https://www.pivotaltracker.com/story/show/188172536).

### Approach

Updated the generateBusiness method to accept UserData as an argument, allowing the userId to be added to the business object.
All instances where generateBusiness is used have been updated to reflect this change, ensuring that the userId is properly included in the business object.
This ensures consistency across the codebase and sets up the necessary foundation for future work that depends on the userId field in the business object

### Steps to Test

n/A

## Code author checklist

### DIFF
--- a/shared/src/userData.ts
+++ b/shared/src/userData.ts
@@ -30,6 +30,7 @@ export interface Business {
   readonly formationData: FormationData;
   readonly environmentData: EnvironmentData | undefined;
   readonly version: number;
+  readonly userId: string;
 }
 
 export const CURRENT_VERSION = 151;
@@ -73,6 +74,7 @@ export const createEmptyBusiness = (id?: string): Business => {
     },
     environmentData: undefined,
     version: CURRENT_VERSION,
+    userId: "",
   };
 };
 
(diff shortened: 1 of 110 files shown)


shared/src/userData.ts
```diff
@@ -30,6 +30,7 @@ export interface Business {
   readonly formationData: FormationData;
   readonly environmentData: EnvironmentData | undefined;
   readonly version: number;
+  readonly userId: string;
```
jphechter, line 33: Just kindof a musing... How would y'all think about making this into an array value? This could potentially allow us to have a 1:many relationship between business:user. Maybe a little bit of YAGNI, but I could see this being a nice support feature in the future even if we don't want to work on building user permissions into the system at this present time.

@aadedejifearless @seidior

PR #9337, opened 2024-12-27
chore: [#188726526] swap useRouter with compat/router to tix cms preview

FarazA22:
### Ticket

## Code author checklist


web/src/components/tasks/business-formation/BusinessFormationPaginator.tsx
```diff
@@ -297,7 +297,7 @@ export const BusinessFormationPaginator = (): ReactElement => {
       newBusiness.formationData.formationResponse?.redirect
     ) {
       analytics.event.business_formation.submit.go_to_NIC_formation_processing();
-      await router.replace(newBusiness.formationData.formationResponse.redirect);
+      router && (await router.replace(newBusiness.formationData.formationResponse.redirect));
```
hinzed1127, line 300: I would move this one up to the existing `if`:
```js
    if (
      newBusiness.formationData.formationResponse?.success &&
      newBusiness.formationData.formationResponse?.redirect &&
      router
    )
```

web/src/pages/account-setup.tsx
```diff
@@ -69,7 +69,7 @@ const AccountSetupPage = (): ReactElement => {
 
       await updateQueue.queue(userDataWithUser).update();
       analytics.event.finish_setup_on_myNewJersey_button.submit.go_to_myNJ_registration();
-      onSelfRegister({ router, updateQueue, userData: userDataWithUser, setRegistrationStatus });
+      router && onSelfRegister({ router, updateQueue, userData: userDataWithUser, setRegistrationStatus });
```
hinzed1127, line 72: Same `analytics.event` concern, but maybe group this check in the `if` at the top of the block? e.g.
```js
if (!updateQueue || !router) return;
```

web/src/pages/loading.tsx
```diff
@@ -17,16 +17,16 @@ const LoadingPage = (): ReactElement => {
   const { updateQueue, userData } = useUserData();
   const router = useRouter();
   const { dispatch } = useContext(AuthContext);
-
   useEffect(() => {
-    if (!router.isReady) {
+    if (!router) {
```
hinzed1127, line 21: should this still include the previous condition? i.e.,
```js
if (!router || !router.isReady) {
     // ...
```

web/src/pages/loading.tsx
```diff
@@ -43,18 +43,18 @@ const LoadingPage = (): ReactElement => {
...
-          router.push(pageLink);
+          router && router.push(pageLink);
         });
     } else {
       updateQueue.update().then(() => {
-        router.push(ROUTES.dashboard);
+        router && router.push(ROUTES.dashboard);
       });
     }
   }, userData);
```
hinzed1127, line 60: could this one just be `userData && router` as the second argument to `useMountEffectWhenDefined` instead of the other checks

hinzed1127 (commented):
I don't have an issue with the `router &&` approach. There's a handful of times where I'm suggesting an `if` statement to either:
- streamline logic and reduce multiple checks of `router`
-  avoid `analytics.event` calls that I don't think we'd want to make if `router` is undefined.
- leverage existing `if` blocks

Other than that, this seems good to me!

web/src/pages/loading.tsx
FarazA22: isReady is not used within this component/ function. I don't think it's necessary.

web/src/pages/loading.tsx
FarazA22: useMountEffectWhenDefined can only take one dependency variable, adding an array with two variables would cause issues. There is a ticket to remove useMountEffectWhenDefined because it leads to pages not reloading.  I think we can address it there.

PR #9352, opened 2025-01-03
CMS unifying filename and id

adidner:
_Note:_ There are almost 1200 file changes here. Most of them are 1 line changes. You are not expected to read through every single one. Please make a point to spot check and identify any major areas of concern, but this work has been tested fairly thoroughly by both @adidner and @jphechter already.

## Description

We wanted to unify filename and id into just id across the CMS. This is because we've had some bugs over time that are a result of this and we want a single source of truth moving forward. 

### Ticket

This pull request resolves [#188347468](https://www.pivotaltracker.com/story/show/188347468) and [#188671780](https://www.pivotaltracker.com/story/show/188671780).

### Approach

We used vscode global search and regex's to find and replace `filename` with `id` where relevant or delete `filename` where both existed. Then we used a script to double check that all id's matched actual in-code filenames. [validate-id-and-filename-match.py](https://github.com/newjersey/navigator.business.nj.gov/pull/9352/files#diff-e6d8469d22d748264697421cafa3c6c0931554d701155cdaf6703276a013d4f4)

Also things like `label` and `displayname` that previously referenced `filename` now reference `id`.

We also changed all our dropdowns in the `config.yml` file to reference `id` as the field value.

Industries have a list of tasks and license tasks that are used to build the respective roadmap. There was only one place where we needed to modify an industry and that was `home-contractor.json`. All others remained the same.

_Note:_ There are many places within the application where we reference the key `filename` on an object. EX: `action.filename` when dealing with an Anytime Action. We did not modify these. When the application is displaying content to the user, we first read-in files. The Markdown or JSON content is then turned into a object. The file's filename is added as a prop on that object. Because filename is determined by the `id` within markdown, this is still perfectly fine. As a result there should not be any differences with existing application behavior.

### Steps to Test

**General Spot Checking**
- Look through CMS and make sure things look alright.
  - We really just need to confirm that there are no major discrepancies between the values of `id` and `filename`. The big collections to consider are tasks and license tasks.

**In-App Testing**
- poke around the app to make sure the various types of things aren't broken (look for a license, look for a generic task, look for some roadmap variations etc)
  - If something doesn't load, the app should break. None of these changes should result in a silent failure. It should be fairly obvious if something is broken.

**CMS Testing**
- If you want to go above and beyond, you could validate that you can modify and publish changes within the CMS. This is something that we have already done for all modified collections.

### Notes

During the course of this work we found multiple collections that are labeled "archived" collections. It's worth noting to the content team that these are unused resources. We should identify if these are things we want to keep around. Should be a follow up item with the content team.

Within `loadAllFilings` and `loadAllFilingsbyUrlSlug` we have some wonky pattern stuff going on where we need both `loadAllFilingsByUrlSlug` and `loadAllFilingsByFilename` but really we probably shouldn't and it can be made smoother. This could be addressed further in a later refactor ticket.

## Code author checklist


content/src/roadmaps/tasks/task-test.md
```diff
@@ -4,25 +4,25 @@ summaryDescriptionMd: |-
 
   A few sentences with the most important information is enough.
 urlSlug: task-test
-filename: task-test
```
jphechter, line 7: @adidner we need to delete this

web/public/mgmt/config.yml
```diff
@@ -1332,12 +1320,13 @@ collections:
           - { label: "Disabled Veteran Owned", value: "disabled-veteran" }
           - { label: "LGBTQ Owned", value: "lgbtq-owned" }
       - { label: "is SBE (show if fewer than 120 employees)", name: "isSbe", widget: boolean, default: false }
+  # TODO: Do we want to keep "archived" collections? - idk but its a sand, because I don't want to talk to content more during this ticket, will slow things down unnecessarily. So I'mma fix it for now real quick
```
jphechter, line 1323: @adidner can you add information about this to the PR description, please?

web/public/mgmt/config.yml
```diff
@@ -2482,12 +2453,12 @@ collections:
                   required: false,
                 }
 
-  - name: "tasks"
+  - name: "tasks" # TODO: also some webflow stuff and sync script stuff to do
```
jphechter, line 2456:
```suggestion
  - name: "tasks"
```

web/public/mgmt/config.yml
```diff
@@ -2693,20 +2632,20 @@ collections:
       - { label: "Call To Action Text", name: "callToActionText", widget: "string", required: false }
       - { label: "Call To Action Link", name: "callToActionLink", widget: "string", required: false }
 
-  - name: "webflow-licenses"
+  - name: "webflow-licenses" # TODO: fuck with sync scripts, also there is no ID here soooo, are we replacing filename with ID?
```
jphechter, line 2635:
```suggestion
  - name: "webflow-licenses"
```

PR #9358, opened 2025-01-09
chore: [#188347470] new script to check for unused CMS roadmap task f…

alexander-littleton:
…iles

## Description

This Python script checks for any unused CMS roadmap task markdown files in the content folder for the purpose of bringing the list to content so they may confirm and remove unused files. Specifically, this script checks for any markdown files for tasks, then checks them against all tasks mentioned in roadmap files.

### Ticket

This pull request resolves [#188347470](https://www.pivotaltracker.com/story/show/188347470).

### Steps to Test

run `python ./content/scripts/find_unused_tasks.py` and `yarn test:python`

### Notes

With the end of dev sprint coming up, I didn't get around to confirming whether or not a file with common utilities for this scripts would be helpful. Something worth noting for the future. There's probably a decent amount of duplication around traversing the filesystem between scripts.

## Code author checklist

### DIFF
--- a/content/scripts/find_unused_tasks.py
+++ b/content/scripts/find_unused_tasks.py
@@ -1,0 +1,85 @@
+import os
+import json
+
+
+def find_unused_task_files(root: str, directories: list[str]) -> None:
+    task_names = get_all_tasks_in_roadmaps(root, directories)
+    filenames = get_all_task_markdown_filenames(root)
+
+    task_names_not_in_filenames = find_task_names_not_in_filenames(
+        task_names, filenames
+    )
+    print(
+        f"Markdown files not in any roadmap: \n{list_to_output_str(task_names_not_in_filenames)}"
+    )
+    return
+
+
+def find_task_names_not_in_filenames(
+    tasks_in_roadmap: set[str], filenames: list[str]
+) -> list[str]:
+    for task in tasks_in_roadmap:
+        try:
+            filenames.remove(task)
+        except ValueError:
+            print(f"No matching markdown file for task: {task}")
+    return filenames
+
+
+def list_to_output_str(filenames: list[str]) -> str:
+    output = ""
+    for filename in filenames:
+        output += f"{filename}\n"
+    return output
+
+
+def get_all_tasks_in_roadmaps(root: str, directories: list[str]) -> set[str]:
+    tasks: set[str] = set()
+    for directory in directories:
+        dir_path = os.path.join(root, directory)
+        filenames = os.listdir(dir_path)
+        for filename in filenames:
+            full_path = os.path.join(dir_path, filename)
+            tasks = tasks.union(get_task_names_from_roadmap_file(full_path))
+    if "" in tasks:
+        tasks.remove("")
+    return tasks
+
+
+def get_task_names_from_roadmap_file(file_path: str) -> set[str]:
+    with open(file_path) as file:
+        roadmap_config = json.load(file)
+
+        task_names: set[str] = set()
+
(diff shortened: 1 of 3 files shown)


content/scripts/find_unused_tasks.py
```diff
@@ -0,0 +1,84 @@
...
+
+def get_all_task_markdown_filenames(root: str) -> list[str]:
+    filenames: list[str] = []
+    for _, _, files in os.walk(root):
+        filenames.extend(file[:-3] for file in files if file.endswith(".md"))
+    return filenames
+
+
+if __name__ == "__main__":
+    find_unused_task_files("../src/roadmaps", ["add-ons", "industries"])
```
alexander-littleton, line 84: I made subdirectories an argument because it seemed to help with testability. Not entirely sure if it makes total sense in the context but I think it's fine considering it's just a script.

content/scripts/find_unused_tasks.py
```diff
@@ -0,0 +1,84 @@
...
+            roadmap_steps = roadmap_config["roadmapSteps"]
+
+            task_names.update(
+                {step["task"] for step in roadmap_steps if "task" in step},
+                {
+                    step["licenseTask"]
+                    for step in roadmap_steps
+                    if "licenseTask" in step
+                },
+            )
```
alexander-littleton, line 65: Just a reminder that the curly braces are the syntax for a set literal here and not objects like in JS.

content/scripts/find_unused_tasks.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def find_task_names_not_in_filenames(
+    tasks_in_roadmap: set[str], filenames: list[str]
+) -> list[str]:
+    for task in tasks_in_roadmap:
+        try:
+            filenames.remove(task)
+        except:
+            print(f"No matching markdown file for task: {task}")
```
adidner, line 25: dust - interesting choice to write it with a try catch rather than if's and a contains.

content/scripts/find_unused_tasks.py
adidner: dust - I would think about making these constants instead of inline argument. I think it would just be a tad more readable / obvious what these represent

content/scripts/test_find_unused_tasks.py
```diff
@@ -0,0 +1,125 @@
+import unittest
+import os
+import tempfile
+import json
+from find_unused_tasks import (
+    find_unused_task_files,
+)
+
```
adidner, line 8: Well this is pretty awesome. Love that you wrote tests for this.

alexander-littleton: Mostly Claude here. There was some minor iteration on my part but I'm less familiar with unit testing in Python.

content/scripts/find_unused_tasks.py
alexander-littleton: I hear ya. I think the print statement keeps this as readable as an `if _ in _/else` block.

That being said, I'll be changing this to `except ValueError:` to narrow down the error. It will probably only ever propagate one error type from my understanding but I prefer to be explicit.

content/scripts/find_unused_tasks.py
alexander-littleton: Yea that's fair.

content/scripts/find_unused_tasks.py
alexander-littleton: I'll make the parameter assignment explicit here as a middle ground. I would also prefer constants as well but that was my first approach and I could not get it to work with tests.

PR #9367, opened 2025-01-10
Migration generator

SirSamuelJoseph:
## Description

New migrations involve a lot of boilerplate and are susceptible to mistakes/flakiness. This creates a migration generator script that automates most of that tedium away.

Updates to the docs site with this info found in [this PR](https://github.com/newjersey/business-dev-docs/pull/68).

### Ticket

This pull request resolves [#188347493](https://www.pivotaltracker.com/story/show/188347493).

### Approach

As an MVP, we decided to write this as a bash script. The main considerations were:

1. bash has built-ins to easily do what we want. Since the goals of this functionality are all file and directory manipulation, it's a logical fit
2. Most of the existing code in `scripts/` are bash

We're not opposed to refactoring it to TypeScript or Python in the future if that makes it easier for other engineers to update

### Steps to Test

1. In a shell from the root directory, run `./scripts/generate-new-migration.sh`
4. Follow the prompt to input a migration filename.
5. There are some instructions at the end of the script, but mainly validate that the following files were created/updated and look correct:
   - The new migration file (currently this would be `api/src/db/migrations/v152_your_new_migration.ts`)
        - The migration functions themselves are just copied from the previous migration, but all of the variable names should be incremented (for this example, they would all be `v152` with any references to the previous migration being `v151`
   - The migration and `userData` are exported properly from `api/src/db/migrations/migrations.ts`
   - The version number is properly incremented in `shared/src/userData.ts`

### Notes

- We also ran and validated this in workspaces to insure interoperability. 

## Code author checklist


scripts/generate-new-migration.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+
```
seidior, line 2: [dust] Sounds like you could use a `mktemp` in here somewhere. Also, any particular reason we're using Bash instead of TypeScript or Python to generate these?

hinzed1127: > Sounds like you could use a mktemp in here somewhere

We modified our `sed` calls to use `-i.bak`, since `-i ''` is Mac-specific and doesn't play nice with Linux. This creates temporary backups of any files the script touches. In the happy path scenario that the script executes without error, [those just get cleaned up in the end](https://github.com/newjersey/navigator.business.nj.gov/pull/9367/files#diff-0d8998bd3f139f6da42c0956da32f0652647a520039ff0e5db00eff4719fa4c0R61-R63).

> any particular reason we're using Bash

I updated the description to include that context, but the short answer is:

- Bash does a lot of the file manipulation stuff well
- most of the existing scripts are also bash

We could totally refactor this down the line if it helps others to grok and/or contribute potential updates

api/src/db/migrations/migrations.ts
hinzed1127, line 1: `prettier-plugin-organize-imports` organizes alphabetically, but we want to have this ascend numerically with the versions. I was having issues with `prettier-vscode` trying to keep sorting ([this seems to be a known issue](https://github.com/prettier/prettier-vscode/issues/716)), but with this script, subsequent updates should only really be expected via the generator now, anyway

scripts/generate-new-migration.sh
```diff
@@ -0,0 +1,75 @@
...
+NEW_VERSION_NUMBER=$(($MOST_RECENT_VERSION_NUMBER + 1))
+
+NEW_FILENAME_COMPLETE="v${NEW_VERSION_NUMBER}_${NEW_FILENAME}.ts"
+
+## Create new file
+cp "$MOST_RECENT_FILENAME" "$NEW_FILENAME_COMPLETE"
+echo "New migration created at: api/src/db/migrations/$NEW_FILENAME_COMPLETE"
+
+# Replace version numbers
+sed -i.bak "s/$MOST_RECENT_VERSION_NUMBER/$NEW_VERSION_NUMBER/g" "$NEW_FILENAME_COMPLETE"
```
hinzed1127, line 33: We initially using `sed -i ''` but it turns out that this flag works a little differently between Mac and Linux and isn't a cross-platform approach. This solution was found [here](https://stackoverflow.com/a/22084103/5469511).

hinzed1127 (commented):
Self-review

PR #9276, opened 2024-12-16
feat: [#188540268] Implement new login navigation pattern

hinzed1127:
## Description

Adds a new login page that checks for a user's email in our database before triggering a login via myNJ.

### Ticket

This pull request resolves [#188540268](https://www.pivotaltracker.com/story/show/188540268).

### Approach

Relatively straightforward. The main thing was making sure that all access points get updated from the old way, getting directed straight to myNJ login, to being directed to this intermediary page first.

### Steps to Test

Manually checking the flow:
1. Go through onboarding
2. Click "Log In" from nav bar
 3. Input a bad email for various error states:
	1. a string that isn't a properly formatted email
	2. an email that hasn't been linked to myNJ
4. Go to "Link with myNewJersey" from nav bar dropdown
5. Fill out and submit form, get redirected to profile page
6. Go back to "Log in"
7. This time, submitting that self-registered email should route you to myNJ to log in and successfully get redirected back to navigator upon completion

## Code author checklist

- [x] I have rebased this branch from the latest main branch
- [x] I have performed a self-review of my code
- [x] I have created and/or updated relevant documentation on the engineering documentation website
- [x] I have not used any relative imports
- [x] I have pruned any instances of unused code
- [x] I have not added any markdown to labels, titles and button text in config
- [x] If I added/updated any values in `userData` (including `profileData`, `formationData` etc), then I added a new migration file
- [x] I have checked for and removed instances of unused config from CMS
- [x] If I added any new collections to the CMS config, then I updated the search tool and `cmsCollections.ts` (see CMS Additions in Engineering Reference/FAQ on the engineering documentation site)
- [x] I have updated relevant `.env` values in both `.env-template` and in Bitwarden

PR #9389, opened 2025-01-14
chore: put login page behind feature flag

seidior:
## Description

Put the new login page behind a feature flag for quick enable/disable later. Default to true, but open to defaulting to false as default.


jphechter:
Missing coverage for line 62 of [NeedsAccountModal.tsx](https://github.com/newjersey/navigator.business.nj.gov/pull/9276/files#diff-e669a443d5ecdbd3d00ec52fe32c29bde1ad459e191fc94c9bba2ded66ec37af).

Aside from that, I think everything else looks good.

PR #9418, opened 2025-01-22
Fix the login flow flicker/infinite redirect loops

hinzed1127:
## Description

This fixes a few logical errors still causing infinite render loops for some users after logging in. More details provided in context via the self review, but at a high-level, logic for the `/loading` and `/login` pages has been corrected to make sure redirects are appropriately applied for users in varying combinations of the following states:
- is authenticated or unauthenticated
- has or hasn't completed onboarded
### Ticket

This pull request resolves [#188540268](https://www.pivotaltracker.com/story/show/188540268).

### Approach

- `console.log` ourselves into oblivion
- manually test on a combination of local development and on the test site
- update logic and re-test
- maintain baseline levels of sanity
### Steps to Test

This should be what's currently deployed to the test site if you want to check there and not just locally.

Routing paths that were tested:
- Go through normal login flow with preexisting account → end up on dashboard
- While logged in: 
	- Load home, i.e. `/` → redirect to dashboard
	- Load `/onboarding` for onboarded account → redirects to dashboard
	- Load `/dashboard` for onboarded account→ stays on dashboard
	- Load `/loading` 
	    - if login page is enabled → navigates to `/login` → authenticated user is fetched → ends up on dashboard
	    - if login page is disabled → calls `triggerSignIn` → which refreshes with a new `code` from myNJ (as expected by the current `loading.tsx` logic → ends up on dashboard ([I addressed this in my self-review](https://github.com/newjersey/navigator.business.nj.gov/pull/9418#discussion_r1925687689) before thinking about it more and resolving)
### Notes

See self-review

## Code author checklist


web/src/lib/utils/helpers.ts
```diff
@@ -34,14 +34,22 @@ export const useOnWindowResize = (fun: () => void): void => {
...
+      : thingToBeDefined !== undefined;
+
+    if (isDefined && !effectOccurred.current) {
       effectOccurred.current = true;
-      fun();
+      func();
     }
-  }, [thingToBeDefined, fun]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [func, ...(Array.isArray(thingToBeDefined) ? thingToBeDefined : [thingToBeDefined])]);
```
hinzed1127, line 52: I don't think this was directly causing any of the issues we were encountering, but we did observe that `dashboard.tsx` [calls this with an array of dependencies](https://github.com/newjersey/navigator.business.nj.gov/blob/e3d87f4d629828233ed2cb763be4824f451b85f1/web/src/pages/dashboard.tsx#L71-L123). While it doesn't cause a type error, this does introduce a runtime error where `func` doesn't rerun if `thingToBeDefined`'s contents changes, since the array itself is already defined.

The only other place I found where an array is passed here is `web/src/components/tasks/cannabis/CannabisPriorityRequirements.tsx`, so it's not a pervasive issue, but something to be mindful of. I know getting rid of this method is in the backlog, but if we want to move these modifications into a new method that handles the array case, something along the lines of `useMountEffectWhenAllAreDefined`, I'm down for that update as an alternative

web/src/pages/_app.tsx
```diff
@@ -28,8 +28,8 @@ import { StyledEngineProvider, ThemeProvider } from "@mui/material";
 import "@newjersey/njwds/dist/css/styles.css";
 import { DefaultSeo } from "next-seo";
 import { AppProps } from "next/app";
-import Head from "next/head";
 import { useRouter } from "next/compat/router";
+import Head from "next/head";
```
hinzed1127, line 32: automated import sorting from a [prettier-plugin](https://github.com/simonhaenisch/prettier-plugin-organize-imports)

web/src/pages/loading.tsx
```diff
@@ -20,7 +20,12 @@ const LoadingPage = (): ReactElement => {
   const loginPageEnabled = process.env.FEATURE_LOGIN_PAGE === "true";
 
   useEffect(() => {
-    if (!router) {
+    /**
+     * For client-side rendering of the Next.js router, the `query` property
+     * is first initialized as an empty object. This `isReady` check ensures
+     * we don't have false negatives when checking `router.query`.
+     */
+    if (!router?.isReady) {
```
hinzed1127, line 28: the absence of this check, combined with the catch-all `else` condition, was causing the following bug:
1. Authenticated user is redirected back from myNJ
2. [The prerendering of the page will initialize `router.query` to `{}`](https://nextjs.org/docs/pages/api-reference/functions/use-router#router-object)
3. This means the `if` condition will be false, even though they are routed to the page with a `code` query parameter
4. The `else` condition triggers sign in, which sends us back to (1) with a new code.

This situation is avoided with the `isReady` check

web/src/pages/loading.tsx
```diff
@@ -37,18 +42,14 @@ const LoadingPage = (): ReactElement => {
         encounteredMyNjLinkingError: true,
       });
     } else {
-      if (loginPageEnabled) {
-        router && router.push(ROUTES.login);
-      } else {
-        triggerSignIn();
-      }
+      triggerSignIn();
```
hinzed1127, line 45: tolerable but non-ideal edge case:
- an authenticated user navigates to `/loading`
- the other `if` conditions are false, so `triggerSignIn` is called
- user gets redirected from myNJ with a new `code` query parameter
- user ends up on dashboard

it's unlikely someone will just navigate to `/loading` by themselves, and it at least lands them in the right spot, but I don't love it. Adding this should avoid it:
```ts
getActiveUser().then(() => {
// The dispatch call should have already happened on
// on line 32 in a previous render, so do nothing and defer
// any routing logic to `useMountEffectWhenDefined` below
}.catch(() => {
    triggerSignIn();
})
```
but maybe that's overkill for an unlikely scenario?

web/src/pages/loading.tsx
```diff
@@ -37,18 +42,14 @@ const LoadingPage = (): ReactElement => {
...
     }
   }, [router, dispatch, loginPageEnabled]);
 
   useMountEffectWhenDefined(() => {
     if (!updateQueue) return;
     const business = updateQueue.currentBusiness();
-    if (!onboardingCompleted(business)) {
+    if (business && !onboardingCompleted(business)) {
```
hinzed1127, line 52: this was prematurely pushing users to `/onboarding` with some false positives

web/src/pages/onboarding.tsx
```diff
@@ -217,7 +217,7 @@ const OnboardingPage = (props: Props): ReactElement => {
 
       const currentBusiness = currentUserData.businesses[currentUserData.currentBusinessId];
       if (currentBusiness.onboardingFormProgress === "COMPLETED") {
-        await router.replace(ROUTES.profile);
+        await router.replace(ROUTES.dashboard);
```
hinzed1127, line 220: updated based on [this discussion](https://njcio.slack.com/archives/C05BCE7A2KB/p1737481635280729)

hinzed1127 (commented):
self-review

web/src/pages/loading.tsx
hinzed1127: Never mind. I reverted this logic to what it was before. If the login page is enabled, then the user will get directed to login and avoid the additional `triggerSignIn` call. The small amount of cases left in the `else` case is negligible for now.

hinzed1127:
Note: Unit test failure appears to be a flakey test not related to these changes

web/src/pages/loading.tsx
seidior: [dust] You may want to take a guard clause approach here, and make the check more explicit rather than through optional chaining for readability's sake.

web/src/pages/_app.tsx
seidior: [dust] Our VSCode settings also provide this functionality.

web/src/pages/login.tsx
```diff
@@ -1,23 +1,38 @@
...
-      .then(() => {
-        router && router.push(ROUTES.dashboard);
+      .then((activeUser) => {
+        dispatch({
+          type: "LOGIN",
+          activeUser: activeUser,
+        });
       })
-      .catch(() => {});
+      .catch(() => (): void => {});
```
seidior, line 27: [dust] Do we want to find a way to make this more readable and to have better diagnostics around this?

web/src/pages/onboarding.tsx
seidior: Thank you!

seidior (commented):
Minor notes; please don't consider this a full PR review. That said, I'd love to know how it feels in the application for this behavior to change versus before, especially with the reports of other flashing throughout the application.

web/src/lib/utils/helpers.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    );
+
+    expect(mockFn).not.toHaveBeenCalled();
+
+    rerender({ value: "defined" });
+    expect(mockFn).toHaveBeenCalledTimes(1);
+
+    rerender({ value: "changed" });
+    expect(mockFn).toHaveBeenCalledTimes(1);
+  });
```
jphechter, line 27: [dust] Should there also be a version of this where `initialProps` is defined from the beginning?

web/src/lib/utils/helpers.ts
jphechter: That's a fun nuance. Nice find.

web/test/pages/loading.test.tsx
```diff
@@ -53,6 +55,10 @@ describe("loading page", () => {
     useMockRouter({});
   });
 
+  afterEach(() => {
+    process.env.FEATURE_LOGIN_PAGE = "true";
```
FarazA22, line 59: pebble - When dealing with environment variables in test files, it’s essential to ensure their values are reverted to their original state after each test. Otherwise, the environment variable changes could persist across tests, leading to unexpected behaviors and flaky tests.

web/src/pages/loading.tsx
jphechter: I'm somewhat surprised this could even work without throwing an error. It looks like `onboardingCompleted` is expecting `business` to be defined.

```typescript
export const onboardingCompleted = (business: Business): boolean => {
  return business.onboardingFormProgress === "COMPLETED";
};
```

jphechter (approved):
Minor notes, but overall this looks like an improvement. Thanks for sinking time into this!

Assuming this all works, I have no major issues.

adidner (approved):
looks good to me and works well on my machine locally. Nice job. 

Should be good to go once other comments are addressed or ignored on your end (depending on level of comment)

web/test/pages/loading.test.tsx
hinzed1127: Good callout. This line _is_ reverting it to the original state. [I added a test](https://github.com/newjersey/navigator.business.nj.gov/pull/9418/files/246ca6975ea92409b2845a77f0c4998a25802cc4#diff-c9b586be7002a1d5f2f315cb3555672e97bf490411d5ea1660443710151ebe9aR137-R144) for different expected behavior in the `false` case of the feature flag and needed to clean up afterwards, threw it in an `afterEach` because that seemed more readable for the full test file.

web/src/pages/_app.tsx
hinzed1127: I noticed this issue when working in another file, and it appears to be some conflict between our [vscode setting](https://github.com/newjersey/navigator.business.nj.gov/blob/main/.vscode/settings.json#L22) and the prettier plugin we're using, even though [we're actually disabling that plugin for this file](https://github.com/newjersey/navigator.business.nj.gov/blob/246ca6975ea92409b2845a77f0c4998a25802cc4/web/src/pages/_app.tsx#L2) 🤷

web/src/lib/utils/helpers.test.ts
hinzed1127: good call, updated these tests to include that scenario

web/src/pages/login.tsx
hinzed1127: In the interest of getting this in, gonna defer this one to a later update, but do you have any thoughts on what could go here? I thought about adding a comment to clarify the business logic, something along the lines of:
```ts
.catch(() => (): void => {
     // getActiveUser` throws an error for unauthenticated users.
     // In that case, we don't want to take any action, since 
     // being at the `/login` page makes sense for them.
})
```
But perhaps some conditional logging might be helpful here, too?

seidior: - Conditional logging, like behind a debugging feature flag
- Naming both functions
- Making the error message checks more explicit by creating explicit subclasses of `Error` and checking for those with differential logging based on error message type

These are all valid approaches and the true answer may be some combination thereof. I know we have some outstanding opportunities to be more explicit with catching errors in the frontend, but I'm open to any possibility to make changes to this code more approachable.

PR #9442, opened 2025-01-29
fix: [#188844549] add userId, versionWhenCreated and version to business object

somabadri:
## Description
Added a migration to add missing fields (`userId`, `version` & `versionWhenCreated`) in the `business` object from previous migrations.

### Steps to Test
Review the migrations from `v151` onwards to ensure that this migration fixes any missing fields.

## Code author checklist

### DIFF
--- a/api/src/db/migrations/v155_add_user_id_and_version_to_business.ts
+++ b/api/src/db/migrations/v155_add_user_id_and_version_to_business.ts
@@ -1,0 +1,929 @@
+import {
+  v154Business,
+  v154UserData,
+} from "@db/migrations/v154_add_business_operating_length_and_nonprofit_status";
+import { randomInt } from "@shared/intHelpers";
+
+export const migrate_v154_to_v155 = (v154Data: v154UserData): v155UserData => {
+  return {
+    ...v154Data,
+    businesses: Object.fromEntries(
+      Object.values(v154Data.businesses)
+        .map((business: v154Business) => migrate_v154Business_to_v155Business(business, v154Data))
+        .map((currBusiness: v155Business) => [currBusiness.id, currBusiness])
+    ),
+    version: 155,
+  } as v155UserData;
+};
+
+export const migrate_v154Business_to_v155Business = (
+  business: v154Business,
+  userData: v154UserData
+): v155Business => {
+  return {
+    ...business,
+    version: 155,
+    versionWhenCreated: 155,
+    userId: userData.user.id,
+  } as v155Business;
+};
+
+export interface v155IndustrySpecificData {
+  liquorLicense: boolean;
+  requiresCpa: boolean;
+  homeBasedBusiness: boolean | undefined;
+  providesStaffingService: boolean;
+  certifiedInteriorDesigner: boolean;
+  realEstateAppraisalManagement: boolean;
+  cannabisLicenseType: v155CannabisLicenseType;
+  cannabisMicrobusiness: boolean | undefined;
+  constructionRenovationPlan: boolean | undefined;
+  carService: v155CarServiceType | undefined;
+  interstateTransport: boolean;
+  interstateLogistics: boolean | undefined;
+  interstateMoving: boolean | undefined;
+  isChildcareForSixOrMore: boolean | undefined;
+  petCareHousing: boolean | undefined;
+  willSellPetCareItems: boolean | undefined;
+  constructionType: v155ConstructionType;
+  residentialConstructionType: v155ResidentialConstructionType;
+  employmentPersonnelServiceType: v155EmploymentAndPersonnelServicesType;
+  employmentPlacementType: v155EmploymentPlacementType;
+  carnivalRideOwningBusiness: boolean | undefined;
+  propertyLeaseType: v155PropertyLeaseType;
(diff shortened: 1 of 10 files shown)


aadedejifearless (approved):

SirSamuelJoseph (approved):

shared/src/test/factories.ts
```diff
@@ -457,6 +457,7 @@ export const generateBusiness = (overrides: Partial<Business>): Business => {
     taxFilingData: generateTaxFilingData({}),
     environmentData: generateEnvironmentData({}),
     version: CURRENT_VERSION,
+    userId: generateUser({}).id,
```
jphechter, line 460: This isn't going to break functionality today, but in the future we're going to start relying on the `userId` in the business matching the user's ID. We may want `generateBusiness` to optionally take a user so that it can extract the ID here.

shared/src/userData.ts
```diff
@@ -73,6 +74,7 @@ export const createEmptyBusiness = (id?: string): Business => {
     },
     environmentData: undefined,
     version: CURRENT_VERSION,
+    userId: user?.id || createEmptyUser().id,
```
jphechter, line 77: [boulder] We invoke `createEmptyBusiness` in `addAdditionalBusiness.ts` and are not passing a user there. Every time an Ethan adds a business they will be creating the business with a random user id.

To safeguard against us doing this in the future, we should require `user` as a param here.

jphechter: It might be good to add a test in `addAdditionalBusiness.test.ts` that validates that all additional businesses have the same user id as the associated user.

I'm ok with that being added as a fast-follow given we're trying to get this out as part of the release.

shared/src/userData.ts
```diff
@@ -73,6 +80,7 @@ export const createEmptyBusiness = (id?: string): Business => {
     },
     environmentData: undefined,
     version: CURRENT_VERSION,
```
jphechter, line 82: @somabadri I just thought about this, but we should probably also have `versionWhenCreated` as well.

This would be a natural time to add this to the object. Sorry I didn't think about this earlier.

shared/src/test/factories.ts
somabadri: agreed that this will be addressed in a later ticket!

web/src/lib/domain-logic/addAdditionalBusiness.ts
```diff
@@ -1,7 +1,10 @@
 import { createEmptyBusiness, UserData } from "@businessnjgovnavigator/shared/userData";
 
 export const addAdditionalBusiness = (userData: UserData): UserData => {
-  const additionalBusiness = createEmptyBusiness();
+  const additionalBusiness = createEmptyBusiness({
+    userId: userData.user.id,
+    versionWhenCreated: userData.versionWhenCreated,
```
jphechter, line 6: @somabadri wouldn't this be `CURRENT_VERSION`? If we're generating the business today, the version of the business when it's created is 155.

web/src/lib/domain-logic/addAdditionalBusiness.test.ts
```diff
@@ -19,4 +19,33 @@ describe("addAdditionalBusiness", () => {
...
+    const userData = generateUserDataForBusiness(firstBusiness, {
+      versionWhenCreated: 110,
+    });
+    const newUserData = addAdditionalBusiness(userData);
+
+    expect(Object.keys(newUserData.businesses)).toHaveLength(2);
+    const newBusinessId = Object.keys(newUserData.businesses).find((id) => id !== firstBusiness.id)!;
+
+    expect(newUserData.businesses[newBusinessId].versionWhenCreated).toEqual(110);
+  });
```
jphechter, line 50: If we change how `business.versionWhenCreated` works, this test will likely need to change.

web/src/lib/domain-logic/addAdditionalBusiness.ts
SirSamuelJoseph: Instead of `CURRENT_VERSION` can we use `userData.user.version`? While they're going to be the same value in this particular case, this highlights what we're actually doing (assigning the USER's current version to the newly created business)

somabadri: @SirSamuelJoseph This function doesn't need `versionWhenCreated` to be passed in as an argument anymore since it's defaulted to `CURRENT_VERSION` and that's handled in `userData.ts`.

api/src/db/migrations/v155_add_user_id_and_version_to_business.ts
```diff
@@ -0,0 +1,929 @@
...
+};
+
+export const migrate_v154Business_to_v155Business = (
+  business: v154Business,
+  userData: v154UserData
+): v155Business => {
+  return {
+    ...business,
+    version: 155,
+    versionWhenCreated: userData.versionWhenCreated,
```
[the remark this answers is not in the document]
somabadri: Leaving a comment incase we come back to this. We discussed that it makes more sense that field called  `versionWhenCreated` within the business object should reflect just that and doesn't need to match that of `userData`, especially given that eventually these will be broken into different tables.

PR #9494, opened 2025-02-04
test: [#188858472] update random home based industry function to allow two industry id to exclude

FarazA22:
### Ticket

This pull request resolves https://www.pivotaltracker.com/story/show/188858472

### Approach

I identified the root cause of the flaky test after running it 2000 times and observing a consistent failure pattern under specific conditions:
	1.	The initial industry ID returned from the randomHomeBased function is Acupuncture with homeBased value in userdata set to true.
	2.	During the next step, when selecting an industry from the dropdown, Acupuncture is excluded.
	3.	On the third attempt, the test tries to select Acupuncture again.

This sequence causes the test to fail because Acupuncture with homeBased: true is not a valid scenario (as confirmed in acupuncture.json).

Fix:
To address this, I modified the test to:
	•	Initially exclude Acupuncture from being returned by the randomHomeBased function.
	•	Ensure that when selecting from the dropdown, both Acupuncture and the initially selected industry are excluded.

These changes should eliminate the flakiness by preventing impossible scenarios from being tested.

## Code author checklist

### DIFF
--- a/web/test/factories.ts
+++ b/web/test/factories.ts
@@ -500,21 +500,19 @@ export const randomNegativeFilteredIndustry = (func: (industry: Industry) => boo
   });
 };
 
-export const randomHomeBasedIndustry = (excludeIndustryId?: string): string => {
+export const randomHomeBasedIndustry = (excludedIndustryIds?: string[]): string => {
   const filter = (it: Industry): boolean => {
     const filterCriteria = !!it.industryOnboardingQuestions.canBeHomeBased && it.isEnabled;
-
-    return excludeIndustryId ? filterCriteria && it.id !== excludeIndustryId : filterCriteria;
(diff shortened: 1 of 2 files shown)


web/test/factories.ts
```diff
@@ -500,11 +500,19 @@ export const randomNegativeFilteredIndustry = (func: (industry: Industry) => boo
   });
 };
 
-export const randomHomeBasedIndustry = (excludeIndustryId?: string): string => {
+export const randomHomeBasedIndustry = (
+  excludeIndustryId?: string,
+  excludeSecondIndustryId?: string
+): string => {
```
jphechter, line 506: [dust] I'd like the props for this to be more explicit than positional args. 

What if we turn the prop for `randomHomeBasedIndustry()` into an object with a key `industriesToExclude` whos value is an array of industry ids?

```typeScript
randomHomeBasedIndustry({
  industriesToExclude: ["acupuncture"]
})
```

PR #9644, opened 2025-02-24
chore: [#186705729] add query to find user by business name

aadedejifearless:
### Ticket

## Code author checklist


api/src/db/DynamoDataClient.ts
```diff
@@ -87,10 +88,41 @@ export const DynamoDataClient = (
     return await userDataClient.findByEmail(email);
   };
 
+  const findUsersByBusinessName = async (businessName: string): Promise<UserData[]> => {
+    const normalizedBusinessName = businessName.trim().toLowerCase();
+
+    const businesses = await businessesDataClient.findAllByBusinessName(businessName);
```
josephine-njooi, line 94: Blocking question: I'm confused here, is `businessesDataClient.findAllByBusinessName(businessName)` an exact match query? And if so, wouldn't this line already narrow down to exact matches?
https://github.com/newjersey/navigator.business.nj.gov/blob/186705729_CreateDynamoDBQuery/api/src/db/DynamoBusinessDataClient.ts#L67

aadedejifearless: @josephine-njooi  you're absolutely right to raise this concern! The current implementation of `businessesDataClient.findAllByBusinessName(businessName)` is performing an exact match query, which means that only businesses with the exact businessName value will be returned. Need to adjust my approach here to apply fuzzy matching before querying the database rather than after. Since findAllByBusinessName(businessName) only returns exact matches, we won’t be able to retrieve businesses that are close but not identical.

josephine-njooi: 👍 Let us know when it's ready for review :). Unless it already is?

PR #9701, opened 2025-03-03
Updated pre-commit scripts to track both Pivotal and ADO tickets

adidner:
## Description

Updated pre commit script to track both pivotal and ADO id's 

### Ticket

This pull request resolves [#8681](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/8681).

### Steps to Test

Go onto this branch and install
Then try and commit with a variety of manual testing cases

This is correct feat: [#123455 AB#1234] updated some cool stuff
So trie variations that are correct and not correct, try unknown prefixes, try weird ID's, try removing the AB or the # or the number or the message after or whatever.

## Code author checklist

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,16 +1,17 @@
 #!/bin/sh
 
 # Define the regex pattern for commit messages
-COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): \[AB#[0-9]+\] .+"
+COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): (\[#[0-9]+ \| AB#[0-9]+\] .+|spell check fixes)$"
 COMMIT_MSG_FILE="$1"
 COMMIT_MSG=$(cat "$COMMIT_MSG_FILE")
 
 # Check if the commit message matches the pattern
 if ! echo "$COMMIT_MSG" | grep -Eq "$COMMIT_MSG_REGEX"; then
   echo "❌ Commit message does not match the required format!"
-  echo "❌ Expected format: 'fix|feat|chore|docs|style|refactor|test: [AB#<ticket ID>] <message here>'"
-  echo "❌ Example: 'feat: [AB#1234] added status page'"
-  echo "❌ Alternatively you can ignore this by having '--no-verify' as a commandline argument to the commit command'"
+  echo "❌ Expected format: 'fix|feat|chore|docs|style|refactor|test: [#<Pivotal ID> | AB#<ticket ID>] <message here>'"
+  echo "❌ Example: 'feat: [#123456 | AB#1234] added status page'"
+  echo "---- Alternatively you can ignore this by having '--no-verify' as a commandline argument to the commit command'"
+  echo "---- We also support 'fix: spell check fixes' as a message"
   exit 1
 fi
 


.husky/commit-msg
```diff
@@ -1,15 +1,15 @@
 #!/bin/sh
 
 # Define the regex pattern for commit messages
-COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): \[AB#[0-9]+\] .+"
+COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): \[#[0-9]+ AB#[0-9]+\] .+"
```
aadedejifearless, line 4: [dust]: Can we update the commit message regex to enforce a delimiter between the Pivotal issue ID and the ADO work item? Since they come from different portals, it would help make the distinction clearer.
```
COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): \[#[0-9]+ \| AB#[0-9]+\] .+" 
```

The line changes on line 11-12 will also be updated if we make this change.

aadedejifearless: Also we are moving away from using commitlint/config-conventional, but currently, this will cause an issue with resolving spellcheck errors in the content repo branch. Can the script be updated to allow the commit message format chore: spell check fixes? This would help accommodate the need for spellcheck-related commits while still enforcing other standard formats for tracking IDs (e.g., [#[Pivotal ID] | AB#[ADO ID]]). 

Refer to https://engineering.docs.business.nj.gov/developer-operations/release-process/#standard-release-process


`COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): (\[#[0-9]+ \| AB#[0-9]+\] |spell check fixes )?.+"
`

.husky/commit-msg
```diff
@@ -1,16 +1,17 @@
 #!/bin/sh
 
 # Define the regex pattern for commit messages
-COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): \[AB#[0-9]+\] .+"
+COMMIT_MSG_REGEX="^(fix|feat|chore|docs|style|refactor|test): (\[#[0-9]+ \| AB#[0-9]+\] .+|spell check fixes)$"
```
jphechter, line 4: I know we want to enforce a ticket number being in here. There are instances where I do not have a ticket for something I want to commit. EX: quick fix to CMS content metadata.

`spell check fixes` is a single use case – I'd like this to allow for any message that is properly formatted with a leading semantic tag.

The alternative is to say that "everything should have a ticket associated with it" which just isn't the case, but maybe it should be?

I don't think this is a showstopper for merging this change, but it does enforce something new.

aadedejifearless (approved):

.husky/commit-msg
adidner: Personally, I think that the more work tied to tickets the better. Generally visibility into what devs actually do with their time is ideal in my eyes. 

The other things I was thinking is that if there is some weird exception for a case we could use --no-verify. Because I really think almost all work should be tracked in some capacity. 

If you want things more the way you described I'm not gonna die on this hill but I have a mild preference for this. (Also happy to do your way and then bring it up in techtro next time). 

Just let me know what you want me to do here. @jphechter

jphechter: Agreed. Overall, there is a need for us to document.

Let's leave it like this for now, note any instances (if any) where this creates issues, and we can make a point to discuss.

PR #9702, opened 2025-03-03
fix: [AB#12799] update query to use current version dynamically

aadedejifearless:
## Description

This PR updates the `getUsersWithOutdatedVersion`  to use current_version dynamically instead of a fixed value. This ensures that the query always reflects the correct version during deployments, including weekend migrations.
Previously, the query used a hardcoded version, which could lead to outdated checks and inconsistent migrations. This update ensures that only users with an outdated version are correctly identified and migrated

Changes:

- Updated the query in getUsersWithOutdatedVersion to use the latestVersion parameter dynamically.
- Improved logging by including current_version for better monitoring.
- Added a unit test to verify that the correct log message is generated when no users need migration.

### Ticket

### Steps to Test
 - Run unit test to ensure the existing functionality works as expected. 

## Code author checklist

### DIFF
--- a/api/src/db/DynamoUserDataClient.ts
+++ b/api/src/db/DynamoUserDataClient.ts
@@ -138,7 +138,7 @@ export const DynamoUserDataClient = (
   };
 
   const getUsersWithOutdatedVersion = async (latestVersion: number): Promise<UserData[]> => {
-    const statement = `SELECT data FROM "${tableName}" WHERE data["version"] < ${CURRENT_VERSION}`;
+    const statement = `SELECT data FROM "${tableName}" WHERE data["version"] < ${latestVersion}`;
     return await search(statement);
   };
 
--- a/api/src/db/DynamoDataClient.test.ts
+++ b/api/src/db/DynamoDataClient.test.ts
@@ -15,7 +15,7 @@ import {
   generateTaxFilingData,
   generateUserDataForBusiness,
 } from "@shared/test";
-import { UserData } from "@shared/userData";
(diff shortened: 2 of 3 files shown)


jphechter (commented):
I'm not sure I understand what this is actually changing beyond the newly logged message

api/src/db/DynamoUserDataClient.ts
```diff
@@ -138,7 +138,7 @@ export const DynamoUserDataClient = (
   };
 
   const getUsersWithOutdatedVersion = async (latestVersion: number): Promise<UserData[]> => {
-    const statement = `SELECT data FROM "${tableName}" WHERE data["version"] < ${CURRENT_VERSION}`;
+    const statement = `SELECT data FROM "${tableName}" WHERE data["version"] < ${latestVersion}`;
```
aadedejifearless, line 141: The original code wasnt neccesarily incorrect, but its using a hardcoded CURRENT_VERSION in the query. Updating the query to reference latestVersion dynamically, we're ensuring that it uses the correct version at runtime, which aligns with the intended behavior.

We can't say for sure that this change will fix the issue, but not dynamically referencing the version could have been a contributing factor.

somabadri:
I understand this is mostly housekeeping to update the query to use a dynamic version of CURRENT VERSION. I'm okay with this merging this but this is not a fix to issue at the hand correct?

aadedejifearless:
> I understand this is mostly housekeeping to update the query to use a dynamic version of CURRENT VERSION. I'm okay with this merging this but this is not a fix to issue at the hand correct?

Correct.

PR #9707, opened 2025-03-05
Disable Auto-Tax Filing Cypress Test

jphechter:
## Description

This test is consistently flaking an causing ❌ in our CI Pipeline. I think we're all on the same page that this test works inconsistently, might have been someone hastily made, and isn't bringing a ton of value. I'd like to disable this until a time when we're able to better address so that we can more confidently rely on CI as an indicator of codebase health and hopefully not introduce additional breaking changes.

Specifically, this test was causing `Integration Tests - Group 4-1` to fail fairly consistently.


josephine-njooi (approved):
[pebble] Create a followup ticket to remove/rework/etc this?

Otherwise I'm a fan!

PR #9765, opened 2025-03-07
Normalize emails to lowercase for login/signup

ZacharyRizer:
## Description

* normalized the user email to lowercase in two different locations in the api to fix the login bug: `/users/emailCheck` and `/self-reg`
* added tests in these two locations for the lowercase changes
* made a database migration to change all previous `userData.user.email` entries to lowercase

### Ticket

This pull request resolves [#12673](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/12673).

### Approach

After discussion with a few of the team members it was decided to only change to lowercase in the api layer, rather than have form field validators that changed the input text to lowercase. 

### Steps to Test

After the migration has run, attempt to login with a version of your email with mixed or upper casing.

### Notes

This migration is changing the value of the user's email in DynamoDB. There is an assumption that this will automatically update the index based on that particular field. 

This PR is split into two different commits: 
* one for the code/typescript changes
* one for the data migration

## Code author checklist

### DIFF
--- a/api/src/api/selfRegRouter.ts
+++ b/api/src/api/selfRegRouter.ts
@@ -13,12 +13,19 @@ export const selfRegRouterFactory = (
 
   router.post("/self-reg", async (req, res) => {
     const userData = req.body as UserData;
+    const cleanedUserData: UserData = {
+      ...userData,
+      user: {
+        ...userData.user,
+        email: userData.user.email.toLowerCase(),
+      },
+    };
 
     try {
-      const selfRegResponse = await (userData.user.myNJUserKey
-        ? selfRegClient.resume(userData.user.myNJUserKey)
-        : selfRegClient.grant(userData.user));
-      const updatedUserData = await updateMyNJKey(userData, selfRegResponse.myNJUserKey);
+      const selfRegResponse = await (cleanedUserData.user.myNJUserKey
+        ? selfRegClient.resume(cleanedUserData.user.myNJUserKey)
+        : selfRegClient.grant(cleanedUserData.user));
+      const updatedUserData = await updateMyNJKey(cleanedUserData, selfRegResponse.myNJUserKey);
       res.json({ authRedirectURL: selfRegResponse.authRedirectURL, userData: updatedUserData });
     } catch (error) {
       const message = (error as Error).message;
--- a/api/src/api/selfRegRouter.ts
+++ b/api/src/api/selfRegRouter.ts
@@ -5,6 +5,10 @@ import { Router } from "express";
 import { StatusCodes } from "http-status-codes";
 import { createHmac } from "node:crypto";
 
+type Mutable<T> = {
+  -readonly [P in keyof T]: T[P] extends object ? Mutable<T[P]> : T[P];
+};
+
 export const selfRegRouterFactory = (
   databaseClient: DatabaseClient,
   selfRegClient: SelfRegClient
@@ -13,13 +17,9 @@ export const selfRegRouterFactory = (
 
   router.post("/self-reg", async (req, res) => {
     const userData = req.body as UserData;
-    const cleanedUserData: UserData = {
-      ...userData,
-      user: {
-        ...userData.user,
-        email: userData.user.email.toLowerCase(),
-      },
-    };
+    const mutableClone = structuredClone(userData) as Mutable<UserData>;
+    mutableClone.user.email = mutableClone.user.email.toLowerCase().normalize();
+    const cleanedUserData: UserData = mutableClone as UserData;
 
     try {
       const selfRegResponse = await (cleanedUserData.user.myNJUserKey
--- a/api/src/api/userRouter.test.ts
+++ b/api/src/api/userRouter.test.ts
@@ -744,6 +744,15 @@ describe("userRouter", () => {
       expect(response.body).toEqual({ error: "`email` property required." });
     });
 
+    it("converts email to lowercase before lookup", async () => {
+      const mixedCaseEmails = ["User@EXAMPLE.com", "USERNAME@gmail.COM", "MixedCase@COMPANY.com"];
+
+      for (const mixedCaseEmail of mixedCaseEmails) {
+        await sendRequest({ email: mixedCaseEmail });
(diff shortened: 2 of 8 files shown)


api/src/api/selfRegRouter.ts
```diff
@@ -13,12 +13,19 @@ export const selfRegRouterFactory = (
 
   router.post("/self-reg", async (req, res) => {
     const userData = req.body as UserData;
+    const cleanedUserData: UserData = {
+      ...userData,
+      user: {
+        ...userData.user,
+        email: userData.user.email.toLowerCase(),
+      },
+    };
```
seidior, line 22: [dust] Not sure this works with our data structure. You may want to use `structuredClone()` to deep copy here.

api/src/api/userRouter.test.ts
```diff
@@ -744,6 +744,15 @@ describe("userRouter", () => {
       expect(response.body).toEqual({ error: "`email` property required." });
     });
 
+    it("converts email to lowercase before lookup", async () => {
```
seidior, line 747: [dust] I wonder if this is an opportunity to use our `it.each()` pattern?

api/src/api/selfRegRouter.ts
seidior: [dust] Also this is usually a situation I'd recommend [`string.prototype.toLocaleLowerCase()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase), but given these are emails, I'm not too worried and think this is the more appropriate solution here.

I do wonder if we should take the opportunity to normalize the input using [`string.prototype.normalize()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize). I don't *think* it's a problem in the database right now, but if we're doing a different kind of normalization *anyway*, I'm open to the idea of normalizing the string here.

seidior (commented):
Super-quick turnaround on this! Thanks @ZacharyRizer. I think you've covered a lot more edge cases here than I thought of when writing the ticket. Just a few notes for you to consider but thank you for thinking through this so thoroughly, for advocating for a simpler and more agile approach, and for covering not only your code but your migration with tests. Great work!

api/src/api/selfRegRouter.ts
ZacharyRizer: Can't hurt to put a normalize on there as well, so that would be `userData.user.email.toLowerCase().normalize()` assuming we just use the default settings for normalize? In regards to structuredClone(), are you referring to the spread operation not being supported by our data structure? Because the spread syntax is used in a very similar fashion further down in that same file here: 

```
  const updateMyNJKey = (userData: UserData, myNJUserKey: string): Promise<UserData> => {
    const hmac = createHmac("sha256", process.env.INTERCOM_HASH_SECRET || "");
    const hash = hmac.update(myNJUserKey).digest("hex");
    return databaseClient.put({
      ...userData,
      user: {
        ...userData.user,
        myNJUserKey: myNJUserKey,
        intercomHash: hash,
      },
      dateCreatedISO: dayjs().toISOString(),
      lastUpdatedISO: dayjs().toISOString(),
    });
  };
```

seidior: That part is fine, because at that point, I don't believe we have any businesses yet. But once we introduce `businesses`, which is a complex rather than shallow structure, we would have to account for that too. But for the most part, [I'd like to see us use `structuredClone()` a lot more in these files](https://www.pivotaltracker.com/story/show/187517776) (though we may have to do some TypeScript wizardry in some places) so that people who work on these migration files don't have to worry too much about the complexity of the data structures and how spread syntax only does a shallow copy.

(When the migration system was first developed, `structuredClone()` support was either limited or non-existent.)

That said, I'm going to defer to you on what you think is the best approach here. Do you want to add the businesses accommodation with spread syntax, or do you think `structuredClone()` is more appropriate in this circumstance.

ZacharyRizer: I agree that using structuredClone is the better approach. The ambiguous nature of the spread operator never sits well with me. However, how do we handle the read only properties on these objects? For instance in this case `user.email` is a read only property so once we copy the object it won't let you directly change the email: 

```
const userData = req.body as UserData;
const cleanedUserData: UserData = structuredClone(userData);
cleanedUserData.user.email = userData.user.email.toLowerCase().normalize(); // this would error
```

api/src/api/userRouter.ts
```diff
@@ -144,7 +144,7 @@ export const userRouterFactory = (
       status = StatusCodes.BAD_REQUEST;
       res.status(status).send({ error: "`email` property required." });
     } else {
-      const userData = await dynamoDataClient.findByEmail(email);
+      const userData = await dynamoDataClient.findByEmail(email.toLowerCase());
```
adidner, line 147: pebble - is this right?

So I could be totally wrong but I want to double check here. 

So my impression is that the migration script only runs once a user is logged in. 

So when a user is trying to log in, if the have a capital letter already in their dynamo client, wouldn't we not want to lowercase that here?

If we did lowercase it wouldn't they never match with their account? 

(I could be wrong in that the migration script runs at a different time or there could be others reaons but I'm a bit confused, please clarify)

Also lmk if that doesn't make sense. 

@ZacharyRizer @seidior

seidior: Great question. Zach brought it up too. However, Ade's working on improving our code that auto-migrates users on weekends. So hopefully once we deploy Zach's code, every user will be migrated past this migration and have their emails lowercased.

(But if it doesn't, it will break login for the users with capital letters, although you could argue their logins are broken now anyway.)

adidner (approved):

api/src/api/userRouter.ts
ZacharyRizer: I was curious about this as well, and honestly I'm not sure on the process of when the migration runs. If what Spencer is saying is indeed merged in, that would be ideal. I was under the assumption (although it sounds like it was wrong), that when we do a migration in a PR like this, it would migrate all data when the PR is merged. But it sounds like that is not the case? So maybe I should wait on merging this under Ade's PR is pulled in?

PR #9823, opened 2025-03-10
fix: [AB#12799] fix user data migration

aadedejifearless:
## Description

This PR updates the getUsersWithOutdatedVersion function to correctly migrate users with outdated versions by adding pagination support. It also includes changes to improve local development and testing.

**Changes:**
- Updated getUsersWithOutdatedVersion function to support pagination with an optional nextToken argument.
- Modified tsconfig.json to include additional paths for scripts and tests.
- Added scripts that simulate the expected changes locally (see imports for details).
- Updated DynamoDB configuration in dynamoDbConfig.ts to specify the correct region for local testing.
-Added unit tests to verify pagination and migration logic.
- Removed the migrateAllUserData.ts script, as it is no longer in use.
- These changes ensure better scalability with pagination and improve the local development setup.

### Ticket

### Steps to Test
1. Open three terminal windows on your local machine.
2. In the first terminal, run the following command to start the server:
  ```
  yarn start:dev
 ```
3. In the second terminal, run the following command to start the local DynamoDB instance:
  ```
  yarn dynamo
 ```
4. In the third terminal, run the following command to populate local DynamoDB with the required data:

  ```
cd api

npx ts-node -r tsconfig-paths/register scripts/populateLocalDynamo.ts
 ```
5. Can also validate changes in the  the testing environment with the migrateUsersVersion Lambda, where it works as expected.

## Code author checklist


api/src/db/DynamoUserDataClient.ts
```diff
@@ -136,10 +136,32 @@ export const DynamoUserDataClient = (
...
+        NextToken: nextToken,
+      })
+    );
+
+    const usersToMigrate = await Promise.all(
+      Items.map(async (object: any): Promise<UserData> => {
+        const data = unmarshall(object).data;
+        return await doMigration(data);
+      })
+    );
```
hinzed1127, line 163: [Important note on Promise.all:
](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#description)
> `Promise.all()` will reject immediately upon any of the input promises rejecting. In comparison, the promise returned by [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) will wait for all input promises to complete, regardless of whether or not one rejects. Use `allSettled()` if you need the final result of every promise in the input iterable.

I'm not completely sure of the performance considerations here, but do we need to consider one of the following scenarios?

1. As it's currently written, suppose a migration fails for some reason. Since `Promise.all` rejects immediately, I think we would need some error handling added in, like [this example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#promise.all_fail-fast_behavior) on the same page previously linked, as well as tracking those failures.
2. If we switch to `Promise.allSettled`, then we still need to aggregate the failures at the end, but the fast rejection no longer applies.

This might be an overly cautious concern, but I'm just thinking of the case in which a nontrivial number of migrations fail for some reason.

api/scripts/populateLocalDynamo.ts
hinzed1127, line 1: [dust] Judging on the actions/names here, this seems like a script purely intended for testing purposes. Perhaps the file name or location could better reflect that, e.g. `populateLocalDynamoWithMassMigrations.ts`?

hinzed1127 (commented):
A few questions for clarification

api/scripts/populateLocalDynamo.ts
```diff
@@ -0,0 +1,98 @@
...
+// Run yarn ts-node -r tsconfig-paths/register scripts/populateLocalDynamo.ts
+import { BatchWriteItemCommand } from "@aws-sdk/client-dynamodb";
+import { marshall } from "@aws-sdk/util-dynamodb";
+import { generateUserData } from "../../shared/src/test";
+import { CURRENT_VERSION } from "../../shared/src/userData";
+import { createDynamoDbClient } from "../src/db/config/dynamoDbConfig";
+import { DynamoBusinessDataClient } from "../src/db/DynamoBusinessDataClient";
+import { DynamoDataClient } from "../src/db/DynamoDataClient";
+import { DynamoUserDataClient } from "../src/db/DynamoUserDataClient";
+import { BUSINESSES_TABLE, USERS_TABLE } from "../src/functions/config";
```
jphechter, line 12: [dust] Can we make these absolute imports?

for example
```typescript
import { generateUserData } from "@shared/test";
```

api/scripts/populateLocalDynamo.ts
```diff
@@ -0,0 +1,98 @@
...
+  const outdatedVersionCount = await batchWriteUsers(users);
+  console.log("Done inserting users.");
+  console.log(`starting migration ... of ${outdatedVersionCount} users`);
+  const userDataClient = DynamoUserDataClient(dynamoDb, USERS_TABLE, logger);
+  const businessesDataClient = DynamoBusinessDataClient(dynamoDb, BUSINESSES_TABLE, logger);
+  const dynamoDataClient = DynamoDataClient(userDataClient, businessesDataClient, logger);
+  await dynamoDataClient.migrateData();
+};
+// @ts-expect-error: Ignoring top-level await error for this script
+await run().catch(console.error);
```
jphechter, line 98:
```suggestion
run().catch(console.error);
```

api/scripts/populateLocalDynamo.ts
jphechter: Just noting as a follow-up from our sync...

I was able to replace all of the relative imports with absolute imports and the code worked as expected.

Not a massive deal, but if we are intending to keep this around long-term it might be nice to include.
```typescript
import { createDynamoDbClient } from "@db/config/dynamoDbConfig";
import { DynamoBusinessDataClient } from "@db/DynamoBusinessDataClient";
import { DynamoDataClient } from "@db/DynamoDataClient";
import { DynamoUserDataClient } from "@db/DynamoUserDataClient";
import { BUSINESSES_TABLE, USERS_TABLE } from "@functions/config";
import { generateUserData } from "@shared/test";
import { CURRENT_VERSION } from "@shared/userData";
```

api/src/db/DynamoDataClient.ts
```diff
@@ -13,27 +13,36 @@ export const DynamoDataClient = (
...
+          CURRENT_VERSION,
+          nextToken
+        );
 
-      const usersNeedingUpdates = usersToMigrate.map(async (user) => {
-        try {
+        if (usersToMigrate.length === 0) {
+          logger.LogInfo(`No users need migration. Current version: ${CURRENT_VERSION}`);
+          break;
+        }
```
jphechter, line 27: [boulder] I don't think this check works.

If the first set of user data queried has 0 users that need migration, then we break out of the do/while. I think we need to expand the condition to also check if there are any remaining pages of results in the database.

api/src/db/DynamoDataClient.ts
```diff
@@ -13,27 +13,36 @@ export const DynamoDataClient = (
...
+      do {
+        const { usersToMigrate, nextToken: newNextToken } = await userDataClient.getUsersWithOutdatedVersion(
+          CURRENT_VERSION,
+          nextToken
+        );
 
-      const usersNeedingUpdates = usersToMigrate.map(async (user) => {
-        try {
+        if (usersToMigrate.length === 0) {
+          logger.LogInfo(`No users need migration. Current version: ${CURRENT_VERSION}`);
```
jphechter, line 25: I also don't think we need this message to be logged for every page worth of data. We have an identical message on line 37 which would apply if there were no users updated.

api/src/db/DynamoDataClient.ts
```diff
@@ -13,27 +13,36 @@ export const DynamoDataClient = (
...
+          break;
+        }
+        for (const user of usersToMigrate) {
           await updateUserAndBusinesses(user);
           logger.LogInfo(`Migrated user ${user.user.id} to version ${CURRENT_VERSION}`);
-        } catch (error) {
-          logger.LogError(`Failed to migrate user ${user.user.id}: ${error}`);
+          migratedCount++;
```
jphechter, line 31: [sand] Instead of incrementing this value every iteration of this for loop, you could just increment once by `usersToMigrate.length`.

api/src/db/DynamoDataClient.ts
jphechter: The more I think about this, I don't think we need this conditional altogether.

If `usersToMigrate` is empty, the for loop never runs and the `do/while` just continues paginating results. I would just remove these lines.

api/src/db/DynamoUserDataClient.ts
```diff
@@ -136,10 +136,32 @@ export const DynamoUserDataClient = (
...
+      new ExecuteStatementCommand({
+        Statement: statement,
+        NextToken: nextToken,
+      })
+    );
+
+    const usersToMigrate = await Promise.all(
+      Items.map(async (object: any): Promise<UserData> => {
+        const data = unmarshall(object).data;
+        return await doMigration(data);
```
jphechter, line 161: Why do we need to invoke `doMigration` here? Aren't we doing that when we invoke `put` already?

jphechter: [boulder] I think we should just be returning the users.

api/scripts/populateLocalDynamo.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const batchWriteUsers = async (users: any[]): Promise<number> => {
+  const chunks = [];
+  let outdatedVersionCount = 0;
+
+  for (let i = 0; i < users.length; i += BATCH_SIZE) {
+    chunks.push(users.slice(i, i + BATCH_SIZE));
+  }
+  for (const chunk of chunks) {
+    const putRequests = chunk.map(({ userData, isOutdated }) => {
```
jphechter, line 55: This really doesn't impact the functionality for testing purposes, but what is the point of the chunking if we're just going to pass `putRequests` as a single param? This just feels confusing to me.

api/src/db/DynamoDataClient.ts
```diff
@@ -13,27 +13,36 @@ export const DynamoDataClient = (
     error?: string;
   }> => {
```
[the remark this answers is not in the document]
aadedejifearless: I don’t think `migrateAllData` is the right name since this function only processes a subset of users with outdated versions, not all data. To make it clearer, I'll update the name to `migrateOutdatedVersionUsers`.

api/src/db/DynamoUserDataClient.ts
aadedejifearless: You're correct that `doMigration` is redundant in the current flow within `searchWithPagination`. However a point to note here is that our current pattern in most of the code follows a similar approach. when we fetch users with these ` PartiQL queries` we are calling the doMigration function across them. I've updated this but gonna create a refactor ticket for the others.

api/scripts/populateLocalDynamo.ts
aadedejifearless: Thanks for the feedback. The reason we are using chunking in this case is because of DynamoDb's `BatchWriteItem` limit of 25 items per request. In this script we are simulating inserting a large number of users (2000) and chunking allows us to split the users into smaller batches that fit with DDB's constraints. This allows us to simulate bulk writes without running into the limitation of trying to insert all 2000 users in a single request. While it may seem unnecessary for testing its the way to handle batch limits efficiently. 
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html

api/src/db/DynamoDataClient.ts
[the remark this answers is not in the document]
aadedejifearless: This would be updated in a refactor ticket.

api/src/db/DynamoDataClient.ts
aadedejifearless: Removed this so resolving it.

api/src/db/DynamoDataClient.ts
aadedejifearless: Resolved

api/src/db/DynamoDataClient.ts
aadedejifearless: Processing in batches so its getting incremented by the size. Similar approach.

hinzed1127 (approved):
Looking good after the updates + walking through them with you

PR #9937, opened 2025-03-14
feat: [AB#12716] Enable reproducing test randomness in `api/`

josephine-njooi:
## Description

This commit
- For each test in `api/`, choose a random seed, and seed all subsequent calls to `Math.random()` in the test with the seed
- Log this seed to the console if the test fails

It also
- Modifies all test names within `api` to be unique
- Moves/renames some jest config files to better reflect the configuration that they specify
- Updates the `jest-dynalite` configuration to [work with a custom test environment](https://www.npmjs.com/package/jest-dynalite#advanced-setup)

The intention is to do this for `web/` and cypress as well, but putting up this PR first to get feedback on code and approach.

Thanks to @hinzed1127 and @SirSamuelJoseph for great ideas included in this approach!

### Ticket

This commit contributes to #[12716](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/12716).

### Approach
As background, seeding random allows such that if the seed is e.g. `"fxzb6150sqm"`, calling `Math.random()` will always produce e.g. `0.7066288397934233` for the first call, `0.37136806641527026` for the second call, `0.007527821813032584` for the third, etc. Different seeds will produce different deterministic sequences of "random" numbers.

1. Seed each test individually with a different random seed, by doing so in a `beforeEach` specified in the `setupFilesAfterEnv` jest configuration. The chosen seed is stored in a global map.
1. Create and uses a custom `testEnvironment` that extends the default `NodeEnvironment`. This environment has access to both 1) whether the test has failed, and 2) the global map with the random seeds, and is thereby able to retrieve and console.log the seed when a test fails.

#### Graveyard of alternatives considered
1. Seed all calls to `Math.random()` once at the beginning of the process, then use the single deterministic sequence from that seed to apply to all subsequent tests. The "seed it once" it can be achieved by the `globalSetup` jest configuration, and it it would be easy to just always log the seed. However, running tests locally I could not get random numbers in tests to reproduce despite using the same seed.
2. Pick one random seed, log it once, and use that same seed to seed at the beginning of all tests. This would have the benefit that we can just unconditionally log the seed once. However, if the e.g. the first call to multiple tests was to generate a random business, all these tests would generate the same business. The team finds value in the current state of randomness and would not like it reduced.
3. Use a [custom reporter](https://jestjs.io/docs/configuration#reporters-arraymodulename--modulename-options) instead of a custom node environment to conditionally log the seed. This would have the benefit that we can define one custom reporter for both `api/` and `web/`, instead of extending the two different test environments that they use. However, reporters [do not have access](https://github.com/jestjs/jest/issues/7421#issuecomment-442438684) to mutations on the global object (confirmed by having attempted this).
4. Conditionally log the seed in an `afterEach` if the test fails. However, `afterEach` unfortunately does not know if a test has failed. It [knows the number of assertions](https://stackoverflow.com/questions/42000137/check-if-test-failed-in-aftereach-of-jest) that have failed, but tests can also fail through timeouts and exceptions.
5. Use a string tag, e.g. `[logRandomSeed]`, like we do for [[group2] here](https://github.com/newjersey/navigator.business.nj.gov/blob/main/web/cypress/e2e/guest.spec.ts), and conditionally log the seed if the string tag is present. This would save us from having to use a custom test environment. However, this would require us to have to first observe a test fail/flake, then make and push a code change, then wait for the next failure before we were able to get a seed.

### Steps to Test

1. Pick a test in `api/`, and add `expect(Math.random()).toEqual(4);` so that the test fails and we get a random number. 
2. Run the test and note the random number that failed the `expect`. There should also be a console.log like `Test failed, reproduce randomness by running with RANDOM_SEED=<the seed> (<test name>)`.
3. If you run the test a second time, the random number and the seed should be different.
4. If you run the test with the env var RANDOM_SEED=<the seed>, it should reproduce the random number that failed the `expect`

This also works if running multiple tests.

First failure:
image

We can reproduce this random number:
image

## Code author checklist


api/test/customNodeEnvironment.ts
```diff
@@ -0,0 +1,24 @@
...
+
+class CustomTestEnvironment extends NodeEnvironment {
+  constructor(config: JestEnvironmentConfig, context: EnvironmentContext) {
+    super(config, context);
+    this.global.testRandomSeeds = new Map<string, string>();
+  }
+
+  async handleTestEvent(event: Circus.Event): Promise<void> {
+    if (event.name === "hook_failure" || event.name === "test_fn_failure") {
+      const testRandomSeeds = this.global.testRandomSeeds as Map<string, string>;
```
josephine-njooi, line 13: Without this cast, typescript was complaining about `testRandomSeeds` being untyped

api/test/setupFile.ts
```diff
@@ -0,0 +1,3 @@
+import { setup } from "jest-dynalite";
+
+setup("api/");
```
josephine-njooi, line 3: https://www.npmjs.com/package/jest-dynalite#advanced-setup

api/test/setupFileAfterEnv.ts
```diff
@@ -0,0 +1,23 @@
...
+
+type TestGlobalThis = {
+  testRandomSeeds: Map<string, string>;
+};
+
+beforeEach(function () {
+  const currentTestName = expect.getState().currentTestName;
+  if (currentTestName === undefined) {
+    throw new Error("expect.getState().currentTestName is undefined");
+  }
```
josephine-njooi, line 12: To satisfy typescript

api/test/setupFileAfterEnv.ts
```diff
@@ -0,0 +1,23 @@
+import "jest-dynalite/withDb";
+import seedrandom from "seedrandom";
+
+type TestGlobalThis = {
+  testRandomSeeds: Map<string, string>;
+};
```
josephine-njooi, line 6: This complaint is otherwise that `global` is untyped. The [internet says](https://stackoverflow.com/questions/68481686/type-typeof-globalthis-has-no-index-signature#70935949) to define a `global.d.ts`, but I didn't want to define `testRandomSeeds` on global for our entire app, and I couldn't figure out how to only include the type definition in tests.

I'm not opinionated about whether to do `(global as unknown as TestGlobalThis)`, or to e.g. eslint suppress the original `Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature` complaint on `global.testRandomSeeds`

josephine-njooi (commented):
Any feedback on approach/typescript suggestions?

jennyverdeyen (approved):
LGTM!

somabadri (commented):
LGTM! Great work. This should help us immensely when debugging failed flaky tests. Thank you pushing for this and setting it up.

josephine-njooi:
Thanks Jenny and Soma!

> Good to merge after addressing issues in CI.

The CI failures were existing flaky tests (is it okay to link to slack here?), being addressed (skipped) in https://github.com/newjersey/navigator.business.nj.gov/pull/9938

PR #9955, opened 2025-03-17
feat: [AB#12716] Enable reproducing test randomness in `web/`

josephine-njooi:
## Description

This commit
- For each test in `web/`, choose a random seed, and seed all subsequent calls to `Math.random()` in the test with the seed
- Logs this randomly chosen seed to the console if the test fails

It also
- Modifies all test names within `web` to be unique
- Refactors the Math.random() seeding code from `api/` to `shared/`, breaking up the jest-dynalite setup in `api/`

This was done for `api` in https://github.com/newjersey/navigator.business.nj.gov/pull/9937. The intention is to do this for cypress as well, but that will take more finagling.

### Ticket

This pull request contributes to [#12716](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/12716).

### Approach

- Seeds each test individually with a different random seed, by doing so in a `beforeEach` specified in the `setupFilesAfterEnv` jest configuration. The chosen seed is stored in a global map.
- Creates and uses a custom `testEnvironment` that extends the default `JSDOMEnvironment`. This environment has access to both 1) whether the test has failed, and 2) the global map with the random seeds, and is thereby able to retrieve and console.log the seed when a test fails.

For a graveyard of other alternative approaches considered, see the description section in https://github.com/newjersey/navigator.business.nj.gov/pull/9937

### Steps to Test

1. Pick a non-cypress test in `web/`, and add `expect(Math.random()).toEqual(4);` so that the test fails and we get a random number. 
2. Run the test and note the random number that failed the `expect`. There should also be a console.log like `Test failed, reproduce randomness by running with RANDOM_SEED=<the seed> (<test name>)`.
3. If you run the test a second time, the random number and the seed should be different.
4. If you run the test with the env var `RANDOM_SEED=<the seed>`, it should reproduce the random number that failed the `expect`

This also works if running multiple tests.

## Code author checklist

### DIFF
--- a/web/src/components/tasks/business-formation/getErrorStateForFormationField.test.ts
+++ b/web/src/components/tasks/business-formation/getErrorStateForFormationField.test.ts
@@ -1235,7 +1235,6 @@ describe("getErrorStateForField", () => {
     runTests(onlyHasErrorIfUndefinedTest);
 
     runTests(["foreignStateOfFormation"], Config.formation.fields.foreignStateOfFormation.error);
-    runTests(["foreignStateOfFormation"], Config.formation.fields.foreignStateOfFormation.error);
   });
 
   describe("foreignGoodStandingFile", () => {
--- a/api/jest.config.ts
+++ b/api/jest.config.ts
@@ -9,7 +9,10 @@ export default {
   displayName: "api",
   testEnvironment: "<rootDir>/test/customNodeEnvironment.ts",
   setupFiles: ["<rootDir>/test/setupFile.ts"],
-  setupFilesAfterEnv: ["<rootDir>/test/setupFileAfterEnv.ts"],
+  setupFilesAfterEnv: [
+    "<rootDir>/test/setupJestDynalite.ts",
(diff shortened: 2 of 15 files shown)


web/src/components/tasks/business-formation/getErrorStateForFormationField.test.ts
```diff
@@ -1235,7 +1235,6 @@ describe("getErrorStateForField", () => {
     runTests(onlyHasErrorIfUndefinedTest);
 
     runTests(["foreignStateOfFormation"], Config.formation.fields.foreignStateOfFormation.error);
-    runTests(["foreignStateOfFormation"], Config.formation.fields.foreignStateOfFormation.error);
```
josephine-njooi, line 1238: This was introduced in https://github.com/newjersey/navigator.business.nj.gov/commit/dd038a7aada3b1b82f009cd4b8cfc89966ed6544#diff-8eabcc1c76cc4f1d2d9ee1b6c445629a9caffccc4a168b7135f5e654ae7357dbR642, in looks like it is genuinely a duplicate.

PR #10011, opened 2025-03-24
fix: [AB#12799] update invocation freq in prod for migrateusersVersion

aadedejifearless:
### Increase Lambda Invocation Frequency for User Migration
This PR addresses issues related to provisioned throughput limits on our DynamoDB table, which caused migration failures due to exceeding the configured limits.

1. Reduced batch size from the previous value to 25 to minimize throttling.

2. Updated processBatch to use Promise.allSettled, ensuring partial failures don’t halt the entire migration process.

3. Increased Lambda invocation frequency from a weekly schedule to every 5 minutes daily until all users are migrated in production only. 

4. The previous weekly updates, especially in production, is insufficient to complete the migration in a reasonable timeframe.

5. This change allows us to process users more gradually while working within throughput limits rather than relying on a single large execution.

6. Updated the dataLogger path from aws/${STAGE} to NavigatorDBClient/${STAGE} to ensure logs are written to the correct CloudWatch log group.

 - The previous path incorrectly suggested an AWS resource, which it is not.

- This aligns with the existing aws_cloudwatch_log_group.data_migration_logs configuration, ensuring migration errors are logged correctly and not missed.

### Ticket

### Steps to Test
1. Open three terminal windows on your local machine.
2. In the first terminal, run the following command to start the server:
  ```
  yarn start:dev
 ```
3. In the second terminal, run the following command to start the local DynamoDB instance:
  ```
  yarn dynamo
 ```
4. In the third terminal, run the following command to populate local DynamoDB with the required data:

  ```
cd api

npx ts-node -r tsconfig-paths/register scripts/populateLocalDynamo.ts
 ```
5. Can also validate changes in the  the testing environment with the migrateUsersVersion Lambda, where it works as expected.
 Note: The cadence of invocation (every 5 minutes) cannot be directly tested locally. 

## Code author checklist

### DIFF
--- a/api/src/db/DynamoDataClient.ts
+++ b/api/src/db/DynamoDataClient.ts
@@ -16,7 +16,7 @@ export const DynamoDataClient = (
     try {
       let nextToken: string | undefined = undefined;
       let migratedCount = 0;
-      const batchSize = 100;
+      const batchSize = 25;
 
       do {
         const { usersToMigrate, nextToken: newNextToken } = await userDataClient.getUsersWithOutdatedVersion(
@@ -43,9 +43,18 @@ export const DynamoDataClient = (
   };
 
   const processBatch = async (usersToMigrate: UserData[]): Promise<void> => {
-    for (const user of usersToMigrate) {
-      await updateUserAndBusinesses(user);
-      logger.LogInfo(`Migrated user ${user.user.id} to version ${CURRENT_VERSION}`);
+    const results = await Promise.allSettled(
+      usersToMigrate.map(async (user) => {
+        await updateUserAndBusinesses(user);
+        logger.LogInfo(`Migrated user ${user.user.id} to version ${CURRENT_VERSION}`);
+      })
+    );
+
+    for (const [index, result] of results.entries()) {
+      const user = usersToMigrate[index];
+      if (result.status === "rejected") {
+        logger.LogError(`Failed to migrate user ${user.user.id}: ${result.reason}`);
+      }
     }
   };
 
--- a/api/src/functions/migrateUsersVersion/index.ts
+++ b/api/src/functions/migrateUsersVersion/index.ts
@@ -2,13 +2,14 @@ import { FnType } from "@functions/fnType";
 import { handlerPath } from "@libs/handlerResolver";
 
 export default (vpcConfig: FnType["vpc"]): FnType => {
+  const isProd = process.env.STAGE === "prod";
   return {
     handler: `${handlerPath(__dirname)}/app.default`,
     timeout: 30,
     events: [
       {
         schedule: {
-          rate: ["cron(0 6 ? * SUN *)"],
+          rate: [isProd ? "cron(*/5 18-23,0-5 * * ? *)" : "cron(*/5 0-5 ? * SUN *)"],
           enabled: true,
         },
       },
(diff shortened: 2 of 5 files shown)


api/src/db/DynamoDataClient.ts
```diff
@@ -16,7 +16,7 @@ export const DynamoDataClient = (
     try {
       let nextToken: string | undefined = undefined;
       let migratedCount = 0;
-      const batchSize = 100;
+      const batchSize = 25;
```
seidior, line 19: [dust] How many in a page of results? I figured we could just do a page of results at a time, since it consumes the same amount of read units. If we do batch sizes of 25 and we have 180,000+ users to migrate (which we'll have every time we add a new migration), that will take a long time.

api/src/db/DynamoDataClient.ts
```diff
@@ -43,9 +43,18 @@ export const DynamoDataClient = (
   };
 
   const processBatch = async (usersToMigrate: UserData[]): Promise<void> => {
-    for (const user of usersToMigrate) {
-      await updateUserAndBusinesses(user);
-      logger.LogInfo(`Migrated user ${user.user.id} to version ${CURRENT_VERSION}`);
+    const results = await Promise.allSettled(
```
seidior, line 46: [dust] Won't `promise.allSettled()` try to concurrently run all the migrations? If so, how does that square with the "if we don't run migrations serially, we consume all the write units" strategy that was being pursued?

api/src/functions/migrateUsersVersion/index.ts
```diff
@@ -2,13 +2,14 @@ import { FnType } from "@functions/fnType";
...
 export default (vpcConfig: FnType["vpc"]): FnType => {
+  const isProd = process.env.STAGE === "prod";
   return {
     handler: `${handlerPath(__dirname)}/app.default`,
     timeout: 30,
     events: [
       {
         schedule: {
-          rate: ["cron(0 6 ? * SUN *)"],
+          rate: [isProd ? "cron(*/5 * * * ? *)" : "cron(0 6 ? * SUN *)"],
```
seidior, line 12: [dust] You may want to schedule it in prod for every 5 minutes from 6PM to 6AM, something like:

```typescript
"cron(*/5 18-23,0-5 * * ? *)"
```

Additionally, we have more than 25 users in our other environments, so scheduling the lambda to run only once on Sundays may not be good. How about every 5 minutes from midnight to 6AM on Sunday?

```typescript
"cron(*/5 0-5 ? * SUN *)"
```

seidior (commented):
Even for such minor code changes, you still took the time to create a great PR here, a standard that I aspire to when I file PRs. Thanks for being so thorough in your documentation.

I did have one note about your PR description though. You mention

```shell
npx ts-node -r tsconfig-paths/register scripts/populateLocalDynamo.ts
```

Should we consider switching that to something like this to mirror our other typescript script execution:

```shell
yarn dlx tsx path/to/your-script.ts
```

Anyway, a few comments for you to consider! Otherwise, great work thinking through all these weird edge cases, doing the research, and turning around code so quickly!

api/src/db/DynamoDataClient.ts
aadedejifearless: We're using Promise.allSettled() to execute migrations in parallel while ensuring that failures in individual items don’t stop the entire batch. This allows us to track both successful and failed migrations without disrupting the process. However, since DynamoDB has a limited write throughput, running all migrations in parallel within a batch could still exceed the provisioned capacity, leading to throttling. To mitigate this, we’ve reduced the batch size to 25, limiting the number of parallel writes at any given time. Additionally, we've increased the invocation frequency to every 5 minutes, allowing us to process smaller chunks more frequently rather than overwhelming the system with large batches all at once.

PR #10038, opened 2025-04-02
feat: [AB#7752] successful tax clearance document response

FarazA22:
### Ticket

https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7752

### Approach

The goal of this ticket was to implement the submission of an API request to the tax clearance API. When a valid tax PIN and tax ID are provided—along with all required inputs—the API returns a successful response containing a certificate key as an array of numbers.

Upon receiving this array, the front end updates the local state with this value. However, there are no plans to store this value in user data. Instead, in a future ticket, this value will be used to generate a PDF document.

Additionally, I identified that the RepID and RepName fields are sourced from user data. Specifically, RepName comes from the name the user enters during sign-up. If tested in guest mode, this will result in an error, as authentication is required to provide the necessary user details.

This ticket specifically focuses on handling successful responses and error handling. 
If the Tax Id field is missing, the api returns the TAX_ID_MISSING_FIELD error message. 
If the Tax Id and Tax Pin validation fail the api returns the FAILED_TAX_ID_AND_PIN_VALIDATION error message. 
If any other field is missing, the api returns the MISSING_FIELD error message. 
If there is a downstream system error,  the api returns the SYSTEM_ERROR error message. 
There is a INELIGIBLE_TAX_CLEARANCE_FORM error, but I'm not 100% sure how to test this.

## Code author checklist


api/src/client/ApiTaxClearanceCertificateClient.ts
```diff
@@ -0,0 +1,71 @@
...
+    };
+
+    logWriter.LogInfo(
+      `Tax Clearance Certificate Client - Id:${logId} - Request Sent to ${
+        config.orgUrl
+      }//TYTR_ACE_App/ProcessCertificate/businessClearance data: ${JSON.stringify(postBody)}`
+    );
+
+    try {
+      const response = await axios.post(
```
FarazA22, line 47: The docs reference an endpoint with 2 `//` before TYTR_ACE_App, is this correct? Should we use one? It works in bruno with 1 and 2.
`${config.orgUrl}//TYTR_ACE_App`

api/src/client/ApiTaxClearanceCertificateClient.ts
```diff
@@ -0,0 +1,71 @@
...
+        {
+          auth: {
+            username: config.userName,
+            password: config.password,
+          },
+        }
+      );
+
+      logWriter.LogInfo(
+        `Tax Clearance Certificate Client - Id:${logId} - Response received: ${JSON.stringify(response.data)}`
```
FarazA22, line 59: don't log the data, instead log successful response received

api/src/client/ApiTaxClearanceCertificateClient.ts
FarazA22: use 1 /

api/.env-template
```diff
@@ -44,3 +44,6 @@ AWS_PROFILE_PROD=
 AWS_PROFILE_DEV=
 DYNAMODB_TABLE_DEV=
 DYNAMODB_TABLE_PROD=
+TAX_CLEARANCE_CERTIFICATE_URL=
```
FarazA22, line 47: remove since we are not using this locally

josephine-njooi (approved):
Just dust, thanks and LGTM!

api/.env-template
jphechter: This feels similar to the ABC stuff we were talking about in Slack. I really could see this either way. I just think it should be consistent.

api/src/api/taxClearanceCertificateRouter.ts
```diff
@@ -0,0 +1,25 @@
...
+  router.post("/postTaxClearanceCertificate", async (req: ExpressRequestBody<UserData>, res) => {
+    const userData = req.body;
+
+    try {
+      const response = await taxClearanceCertificateClient.postTaxClearanceCertificate(userData);
+      res.json(response);
+    } catch (error) {
+      // TODO: Missing Field Error
+      // TODO: Validation Error
+      // TODO: Server Error
```
jphechter, line 20: Is there a reason we're not handling for these in this work?

api/src/client/ApiTaxClearanceCertificateClient.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        zip: "12345",
+        agencyName: LookupTaxClearanceCertificateAgenciesById("newJerseyBoardOfPublicUtilities").name,
+      },
+      {
+        auth: { password: "fakePassword", username: "fakeUserName" },
+      }
+    );
+  });
+
+  it("should spy on GetId, return a value and invoke LogInfo", () => {
```
jphechter, line 79: It looks like this is the first place where we have a test to assert a log is firing. I'm not opposed, but it is a new pattern.

I might change the test name to something explicitly stating the logs should fire.

api/src/client/ApiTaxClearanceCertificateClient.ts
```diff
@@ -0,0 +1,71 @@
...
+      );
+      return { certificatePdfArray: response.data.certificate };
+    } catch {
+      // TODO: Missing Field Error
+      // TODO: Validation Error
+      // TODO: Server Error
```
jphechter, line 65: If we're planning to cover this behavior in another ticket, that's not documented anywhere. I thought the plan was to implement all functionality.

content/src/fieldConfig/tax-clearance-certificate-download.json
```diff
@@ -0,0 +1,7 @@
+{
+  "taxClearanceCertificateDownload": {
+    "headerTwoLabel": "Your Certificate is Ready!",
+    "calloutText": "Download your certificate now, or you’ll have to request it again.",
+    "buttonText": "Tax Clearance Certificate (PDF)"
+  }
+}
```
jphechter, line 7: Is CMS search still broken? We really need to fix that because adding additional items to the CMS is only going to make fixing this in the future more difficult for the person who addresses this.

web/src/components/DocumentTile.tsx
```diff
@@ -0,0 +1,69 @@
...
+  downloadLink?: string;
+  isCentered?: boolean;
+  isRounded?: boolean;
+  hasExtraPadding?: boolean;
+  onClick?: () => void;
+  isWidthFull?: boolean;
+  hasDownloadIcon?: boolean;
+}
+
+export const DocumentTile = (props: Props): ReactElement => {
```
jphechter, line 19: [sand]
```typescript
export const DocumentDownloadTile = (props: Props): ReactElement => {
```

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -668,6 +674,34 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+    });
+  });
+
+  it("renders the download page when the api post request is successful", async () => {
+    mockApi.postTaxClearanceCertificate.mockResolvedValue({
+      certificatePdfArray: [],
+    });
+    renderComponent({});
+    fireEvent.click(screen.getByTestId("stepper-2"));
+    fireEvent.click(screen.getByTestId("next-button"));
```
jphechter, line 698: If we can avoid using `getByTestId` we should. Something we had [discussed briefly in Slack a while back](https://njcio.slack.com/archives/C04J8MTJNS2/p1726592403454379).

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
```diff
@@ -20,6 +21,7 @@ import { ReactElement, useEffect, useState } from "react";
 interface Props {
   anytimeAction: AnytimeActionLicenseReinstatement | AnytimeActionTask;
   CMS_ONLY_stepIndex?: number;
+  CMS_ONLY_downloadPage?: [];
```
jphechter, line 24: [dust]
```suggestion
  CMS_ONLY_certificatePdfArray?: [];
```

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
```diff
@@ -177,24 +182,35 @@ export const AnytimeActionTaxClearanceCertificateElement = (props: Props): React
...
+                      saveTaxClearanceCertificateData={saveTaxClearanceCertificateData}
+                    />
+                  )}
+                  {stepIndex === 2 && (
+                    <TaxClearanceStepThree
+                      setStepIndex={setStepIndex}
+                      setCertificatePdfArray={setCertificatePdfArray}
+                    />
+                  )}
+                </>
```
jphechter, line 212: [dust] I would probably extract this into a function that returns a TSX element.

web/src/lib/api-client/apiClient.ts
```diff
@@ -74,6 +75,10 @@ export const getCompletedFiling = (): Promise<UserData> => {
   return get(`/completed-filing`);
 };
 
+export const postTaxClearanceCertificate = (userData: UserData): Promise<TaxClearanceCertificateResponse> => {
+  return post(`/postTaxClearanceCertificate`, userData, false);
```
jphechter, line 79: Why do we need to include `false` here?

api/.env-template
FarazA22: Decision after chatting with Sam: we are storing the environment variables in BitWarden and the `.env-template.`

api/src/client/ApiTaxClearanceCertificateClient.ts
josephine-njooi: +1. Seeing this I was gonna find time to pair on what the errors look like (since I can't seem to repro an error response in the workspace), and fold it into https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7714. That ticket is currently ambiguous in whether it includes doing this work.

josephine-njooi: (But is an independently deployable unit if we want to break it into a different ticket)

web/src/lib/api-client/apiClient.test.ts
```diff
@@ -116,4 +117,11 @@ describe("apiClient", () => {
       { headers: { Authorization: "Bearer some-token" } }
     );
   });
+
+  it("posts tax clearance certificate request", async () => {
+    const userData = generateUserData({ user: generateUser({ id: "456" }) });
```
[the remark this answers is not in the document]
FarazA22: I'll make the update. Good catch.

web/src/components/DocumentTile.tsx
FarazA22: In formation, this component is used to render the entity id. It is 1 of 4 tiles that can render on the screen. That component is not clickable, it's purely static text. Given that is the case, I think we should keep the name as is.

web/src/lib/api-client/apiClient.ts
FarazA22: I'll set up some time to chat about this. I want to understand why the `authConfig` function is used, and when it is appropriate to use it. I noticed it's set to false for other api calls.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
FarazA22: To simplify this, I'll rename the prop and expect a boolean value `CMS_ONLY_renderDownloadPage?: boolean;` and create a new variable to manage whether is renders or not `const renderDownloadScreen = props.CMS_ONLY_renderDownloadPage || !!certificatePdfArray;`

FarazA22: Great suggestion, I'll make the update, the change is 100x more clear.

api/src/client/ApiTaxClearanceCertificateClient.test.ts
```diff
@@ -0,0 +1,164 @@
...
+            addressLine2: "apt 1",
+            addressCity: "atlanta",
+            addressState: { shortCode: "AK", name: "Alaska" },
+            addressZipCode: "12345",
+            taxId: "123456789101",
+            taxPin: "1234",
+          }),
+        }),
+      },
+    });
```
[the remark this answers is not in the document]
FarazA22: I extracted the variable and updated the assertions to reference it instead of the hard-coded values.

api/src/client/ApiTaxClearanceCertificateClient.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    const spyOnGetId = jest.spyOn(DummyLogWriter, "GetId").mockReturnValue("test");
+    const spyOnLogInfo = jest.spyOn(DummyLogWriter, "LogInfo");
+    client.postTaxClearanceCertificate(userData);
+    expect(spyOnGetId).toHaveBeenCalled();
+    expect(spyOnLogInfo).toHaveBeenCalledWith("Tax Clearance Certificate Client - Id:test");
+    spyOnGetId.mockRestore();
+    spyOnLogInfo.mockRestore();
+  });
+
+  it("logs post request to LogInfo", async () => {
```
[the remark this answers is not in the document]
FarazA22: I'll consolidate the tests

api/src/client/ApiTaxClearanceCertificateClient.test.ts
```diff
@@ -0,0 +1,164 @@
...
+          taxClearanceCertificateData: generateTaxClearanceCertificateData({
+            requestingAgencyId: "newJerseyBoardOfPublicUtilities",
+            businessName: "my business name",
+            addressLine1: "123 main st",
+            addressLine2: "apt 1",
+            addressCity: "atlanta",
+            addressState: { shortCode: "AK", name: "Alaska" },
+            addressZipCode: "12345",
+            taxId: "123456789101",
+            taxPin: "1234",
```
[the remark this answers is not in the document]
FarazA22: I'll makes similar updates

api/src/client/ApiTaxClearanceCertificateClient.test.ts
FarazA22: I decided to create some tests around logging to get more familiar with spying. I’ve consolidated them into a single test and updated the name.

api/src/client/ApiTaxClearanceCertificateClient.ts
FarazA22: I'll address errors as part of this ticket.

web/src/lib/api-client/apiClient.ts
FarazA22: Oversight, I'll make the updates. I need to dig into how Auth works in the backend.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
FarazA22: This work will be picked up immediately after merging this ticket.

web/src/lib/api-client/apiClient.test.ts
```diff
@@ -116,4 +117,13 @@ describe("apiClient", () => {
...
     );
   });
+
+  it("posts tax clearance certificate request", async () => {
+    const userData = generateUserData({ user: generateUser({}) });
+    mockAxios.post.mockResolvedValue({ data: userData });
+    expect(await postTaxClearanceCertificate(userData)).toEqual(userData);
+    expect(mockAxios.post).toHaveBeenCalledWith("/api/postTaxClearanceCertificate", userData, {
+      headers: { Authorization: "Bearer some-token" },
+    });
```
FarazA22, line 127: updated test to ensure auth is added to header

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
```diff
@@ -177,24 +179,18 @@ export const AnytimeActionTaxClearanceCertificateElement = (props: Props): React
...
-                }}
-              />
-              {stepIndex === 0 && <TaxClearanceStepOne setStepIndex={setStepIndex} />}
-              {stepIndex === 1 && (
-                <TaxClearanceStepTwo
-                  setStepIndex={setStepIndex}
+              {certificatePdfArray ? (
+                <TaxClearanceDownload />
+              ) : (
+                <TaxClearanceSteps
```
FarazA22, line 185: I’ve extracted the components that were being rendered as part of the stepper into a separate component. My main goal was to clearly separate the stepper-related logic from the download screen for better organization and maintainability. I chose to leave the download page as-is for now, since including it in the extraction would increase the number of props being passed down, which could add unnecessary complexity. Let me know if you’d prefer that I still pull the download page into the same component.

shared/src/taxClearanceCertificate.ts
```diff
@@ -2,6 +2,18 @@ import { orderBy } from "lodash";
 import taxClearanceCertificateAgenciesJSON from "../../content/src/mappings/taxClearanceCertificateIssuingAgencies.json";
 import { StateObject } from "./states";
 
+export type TaxClearanceCertificateResponse = {
+  error?: {
+    type:
+      | "INELIGIBLE_TAX_CLEARANCE_FORM"
+      | "FAILED_TAX_ID_AND_PIN_VALIDATION"
+      | "MISSING_FIELD"
+      | "SYSTEM_ERROR";
```
FarazA22, line 11: The API we’re integrating with returns a 400 status code along with a specific error message when a failure occurs in one of their internal systems (referred to as a Natural Program Error). Even though this comes back as a 400, we’re treating it as a system-level error on our end, and mapping it to a standardized SYSTEM_ERROR type for consistency.

This is distinct from other internal server errors that may occur, which will still throw and be handled as actual 500-level errors. The goal is to provide a consistent response shape while still recognizing and logging true unhandled exceptions appropriately.

api/src/client/ApiTaxClearanceCertificateClient.ts
```diff
@@ -0,0 +1,114 @@
...
+  password: string;
+};
+
+export const TAX_ID_MISSING_FIELD = "TaxpayerId  is required.";
+export const INELIGIBLE_TAX_CLEARANCE_FORM = "Clean Tax Verification Failed.";
+export const FAILED_TAX_ID_AND_PIN_VALIDATION =
+  "ADABase validation failed. Please verify the data submitted and retry.";
+export const MISSING_FIELD =
+  "Mandatory Field Missing. TaxpayerId, TaxpayerName, AddressLine1, City, State, Zip, Agency name, Rep Id and RepName are required.";
+export const SYSTEM_ERROR = "Error calling Natural Program. Please try again later.";
```
FarazA22, line 21: Here’s some context around the current API error handling:

- Natural Program Error:
This is essentially an internal system failure from one of the upstream systems the API is calling. While the API returns this as a 400 along with a specific error message, we are treating it as a server error on our end (SYSTEM_ERROR). 

Missing Fields Behavior:
- If the Tax ID is missing, the API returns a unique error message just for that field.
- If any other fields are missing, the API returns a generic error message listing all required fields.
This behavior is inconsistent but reflects how the upstream API currently operates.

- Ineligible Tax Clearance Form:
We’re receiving this error from the API, but it’s unclear what triggers it. I’ve documented this in the API docs, but we still need clarity on what validation is being applied (especially on our end via Bruno, if any).

api/src/client/ApiTaxClearanceCertificateClient.ts
```diff
@@ -0,0 +1,114 @@
...
+            return {
+              error: {
+                type: "FAILED_TAX_ID_AND_PIN_VALIDATION",
+                message: errorMessage,
+              },
+            };
+          }
+          if (errorMessage === SYSTEM_ERROR) {
+            return {
+              error: {
```
FarazA22, line 102: If a Natural Program Error is received, we treat it as a SYSTEM_ERROR and return a structured error object. For any other unexpected internal errors, we allow the error to propagate by throwing it.

api/src/client/ApiTaxClearanceCertificateClient.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    spyOnLogInfo.mockRestore();
+  });
+
+  it("returns certificate pdf array", async () => {
+    mockAxios.post.mockResolvedValue({ data: { certificate: [11, 22] } });
+    const result = await client.postTaxClearanceCertificate(userData);
+    expect(result).toEqual({ certificatePdfArray: [11, 22] });
+  });
+
+  it("fires error log when error response is returned", async () => {
```
FarazA22, line 138: error logging tests start here

josephine-njooi (commented):
Could you please update the PR description, and include steps to test and what to expect for both the success and various error modes? I'd like to make sure I understand the surface of various possible errors.

The PR currently says
> This ticket specifically focuses on handling successful responses, while error states will be addressed separately in a future ticket.

but I believe that, at least, is inaccurate.

api/src/client/ApiTaxClearanceCertificateClient.ts
```diff
@@ -0,0 +1,114 @@
...
+import axios, { AxiosError, AxiosResponse } from "axios";
+import { StatusCodes } from "http-status-codes";
+
+type Config = {
+  orgUrl: string;
+  userName: string;
+  password: string;
+};
+
+export const TAX_ID_MISSING_FIELD = "TaxpayerId  is required.";
```
josephine-njooi, line 15: Dust: am I understanding correctly that the error message that they return contains this extra space before `is`? I think it would be worth calling it out in a comment, lest someone change it to delete the space. (Also thinking about what might happen if/when they realize and correct the error on their end...)

josephine-njooi: From today's standup context that they might change the API without notifying us, I'm changing this to Sand: I think we should build to be robust to whether there's 1 or 2 spaces in there, whether that's via equality against either version/using `includes`/etc

And a question: Do we have front end error reporting at all? I.e. if it does drop all the way to the bottom, will we ever know of the `throw error.response?.status;` (or does it require the user to contact us about a generic error message they get)

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceStepThree.tsx
```diff
@@ -42,6 +45,18 @@ export const TaxClearanceStepThree = (props: Props): ReactElement => {
     });
   }
 
+  const handleButtonClick = async (): Promise<void> => {
+    if (!userData) return;
+    const taxClearanceResponse = await api.postTaxClearanceCertificate(userData);
+    if (taxClearanceResponse.certificatePdfArray)
+      props.setCertificatePdfArray(taxClearanceResponse.certificatePdfArray);
+
+    // TODO: Error Response will be addressed in a separate ticket
```
josephine-njooi, line 54: This separate ticket is https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7714, correct?

josephine-njooi (approved):
I'm still mushy on the surface area of errors and would love steps to reproduce them, especially in the context of wanting to create wiremocks for https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7714
That aside, lgtm!

jphechter (approved):

api/src/client/ApiTaxClearanceCertificateClient.ts
jphechter: My response was to your first comment. Didn't see the second one for some reason.

We do not have any front-end error reporting. If we get something unanticipated, we typically show the user a very general alert asking them to try again later. Ideally we will also be building in a health check for the integration, in which case if there is actually an outage we should be getting an alert in #bizx-alerts-bots.

Front-end error reporting is something we would love to have... just hasn't been a priority.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceStepThree.tsx
FarazA22: Yes that's the correct ticket. I do think it needs to be enhanced to include the other error states.

josephine-njooi: Thanks!

> I do think it needs to be enhanced to include the other error states.

To clarify, by this you mean that it needs to be expanded/new ticket to encompass front end work to handle the other error states you've added here and enumerated in the description, correct? And there are no more additional back end error states to add?

FarazA22: Yes, I believe all the backend work is done, and we need tickets to add functionality on the front end.

josephine-njooi: 👍 Thanks! I'll sync with Dillon

PR #10049, opened 2025-04-08
Remove entityId from tax clearance data

ashlsun:
## Description
We add a migration v159 to remove the field `entityId` from taxClearanceCertificateData. This PR also makes changes to the v158 migration because it was missing `taxClearanceCertificateData`, which was added in v157.

### Ticket

This pull request is a part of [#7713](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7713) (see Dev Notes section), though it does not resolve it.

### Approach

Ran the migration script to generate the v159 file, then filled in the function `migrate_v158Business_to_v159Business`. Used some intermediary types (prefixed with FOR_MIGRATION_ONLY) to avoid a TS error.

### Steps to Test

When running the app locally, run `yarn dynamo` and visit localhost:8001 to see inside your local DynamoDB tables. If you view the users table, you should see that your user is at `version: 159` and their `taxClearanceCertificateData` should not have an `entityId` field on it. This indicates the migration has been run successfully.

## Code author checklist

### DIFF
--- a/api/src/db/migrations/v159_remove_entity_id.ts
+++ b/api/src/db/migrations/v159_remove_entity_id.ts
@@ -1,156 +1,188 @@
-import { v157Business, v157UserData } from "@db/migrations/v157_add_tax_clearance_data";
+import { v158Business, v158UserData } from "@db/migrations/v158_login_email_to_lowercase";
 import { randomInt } from "@shared/intHelpers";
 
-export const migrate_v157_to_v158 = (v157Data: v157UserData): v158UserData => {
+export const migrate_v158_to_v159 = (v158Data: v158UserData): v159UserData => {
   return {
-    ...v157Data,
-    user: {
-      ...v157Data.user,
-      email: v157Data.user.email.toLowerCase(),
-    },
+    ...v158Data,
     businesses: Object.fromEntries(
-      Object.values(v157Data.businesses)
-        .map((business: v157Business) => migrate_v157Business_to_v158Business(business, v157Data))
-        .map((currBusiness: v158Business) => [currBusiness.id, currBusiness])
+      Object.values(v158Data.businesses)
+        .map((business: v158Business) => migrate_v158Business_to_v159Business(business, v158Data))
+        .map((currBusiness: v159Business) => [currBusiness.id, currBusiness])
     ),
-    version: 158,
-  } as v158UserData;
+    version: 159,
+  } as v159UserData;
 };
 
-export const migrate_v157Business_to_v158Business = (
-  business: v157Business,
-  userData: v157UserData
-): v158Business => {
-  return {
+export const migrate_v158Business_to_v159Business = (
+  business: v158Business,
+  userData: v158UserData
+): v159Business => {
+  type FOR_MIGRATION_159_ONLY_TaxClearanceCertificateData = {
+    requestingAgencyId: string | undefined;
+    businessName: string | undefined;
+    addressLine1: string | undefined;
+    addressLine2: string | undefined;
+    addressCity: string | undefined;
+    addressState?: v159StateObject;
+    addressZipCode: string | undefined;
+    entityId: string | undefined;
+    taxId: string | undefined;
+    taxPin: string | undefined;
+  };
+
+  type FOR_MIGRATION_159_ONLY_Business = {
+    id: string;
+    dateCreatedISO: string;
+    lastUpdatedISO: string;
+    profileData: v159ProfileData;
+    onboardingFormProgress: v159OnboardingFormProgress;
+    taskProgress: Record<string, v159TaskProgress>;
+    taskItemChecklist: Record<string, boolean>;
+    licenseData: v159LicenseData | undefined;
+    preferences: v159Preferences;
+    taxFilingData: v159TaxFilingData;
+    formationData: v159FormationData;
+    environmentData: v159EnvironmentData | undefined;
+    taxClearanceCertificateData: FOR_MIGRATION_159_ONLY_TaxClearanceCertificateData;
+    version: number;
+    versionWhenCreated: number;
+    userId: string;
+  };
+
+  const tempBusiness: FOR_MIGRATION_159_ONLY_Business = {
     ...business,
-    version: 158,
+    version: 159,
     userId: userData.user.id,
-  } as v158Business;
+  };
+
+  delete tempBusiness.taxClearanceCertificateData.entityId;
+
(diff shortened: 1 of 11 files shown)


FarazA22 (approved):

api/src/db/migrations/v159_remove_entity_id.ts
```diff
@@ -0,0 +1,992 @@
...
+    businesses: Object.fromEntries(
+      Object.values(v158Data.businesses)
+        .map((business: v158Business) => migrate_v158Business_to_v159Business(business, v158Data))
+        .map((currBusiness: v159Business) => [currBusiness.id, currBusiness])
+    ),
+    version: 159,
+  } as v159UserData;
+};
+
+type FOR_MIGRATION_ONLY_TaxClearanceCertificateData = {
```
SirSamuelJoseph, line 16: [pebble] Can we move these types into the migrate_business function?

My concern is that they may persist when using the migration generator script if people aren't looking closely at the whole file but only their changes. By moving the types into the migrate function, it's much more likely that it will be deleted there and not clutter up additional files

SirSamuelJoseph: @jphechter @FarazA22 would like your thoughts here

jphechter: I think that's a fair request. If we can do it, let's. If not, maybe we can rename to `FOR_MIGRATION_159_ONLY_TaxClearanceCertificateData` to be even more explicit to the next person.

api/src/db/migrations/v159_remove_entity_id.ts
```diff
@@ -0,0 +1,992 @@
...
+  requestingAgencyId?: string | undefined;
+  businessName: string | undefined;
+  addressLine1: string | undefined;
+  addressLine2: string | undefined;
+  addressCity: string | undefined;
+  addressState?: v159StateObject | undefined;
+  addressZipCode: string | undefined;
+  taxId: string | undefined;
+  taxPin: string | undefined;
+};
```
jphechter, line 658: [pebble] Given we're updating the type here in the migration file, we should also update the type definition used throughout the application in `taxClearanceCertificate.ts`.

When a new user signs up, we create an empty business record for them. That function `createEmptyBusiness` lives in `userData.ts`. On line 75 it sets `taxClearanceCertificateData` to `emptyTaxClearanceCertificateData`. We can also set some of the values in  `emptyTaxClearanceCertificateData` to `undefined`.

jphechter (commented):
1 minor note, but otherwise I'm good with this.

api/src/db/migrations/v159_remove_entity_id.ts
FarazA22: The entity id field should also be removed from emptyTaxClearanceCertificateData.

FarazA22:
I re-reviewed the comments and added one more of my own. The rest looks good.

shared/src/taxClearanceCertificate.ts
```diff
@@ -14,29 +14,27 @@ export const getTaxClearanceCertificateAgencies = (): TaxClearanceCertificateAge
   orderBy(taxClearanceCertificateAgencies, ["name"], ["asc"]);
 
 export const emptyTaxClearanceCertificateData: TaxClearanceCertificateData = {
-  requestingAgencyId: "",
-  businessName: "",
-  entityId: "",
-  addressLine1: "",
-  addressLine2: "",
-  addressCity: "",
+  requestingAgencyId: undefined,
```
FarazA22, line 17: We don’t need to set every value in emptyTaxClearanceCertificateData to undefined. If we’re doing that, we’re essentially just returning an empty object — and in that case, it would make more sense to return undefined altogether rather than an object filled with undefined values.

Separately, when we initialize empty data and the type is a string, our convention is to use an empty string (""). Fields like addressState, which expect a specific state object, are the exception and should remain undefined.

jphechter: @FarazA22 Good callout, but I don't totally agree.

Yes, it's something that does provide a slight convenience to us on the front-end, but it artificially inflates the size of our database. In the case of a no SQL database, we then are storing 9 additional entries with no data for every user (`9 * 190,000 = 1,710,000`) in the database. A better option for us there is to just initialize this as `undefined` and only store data that's absolutely necessary. The value could be an empty string in the future, but I don't want to initialize this value.

I've made some different changes from @ashlsun that I'm about to push up. I haven't modified this function out of an abundance of cation, but that is something I want to do as a fast follow to this.

jphechter: I know this is a pattern we have in some other places, but I don't think it's a good one.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
```diff
@@ -107,7 +104,6 @@ export const AnytimeActionTaxClearanceCertificateElement = (props: Props): React
       setProfileData({
         ...profileData,
         businessName: newBusinessName,
-        entityId: newEntityId,
```
ashlsun, line 110: [dust] Does this `entityId` that lives at the top level of `profileData` mean something different from the `entityId` in the tax clearance object? Or is it referring to the same thing? (It makes sense that we remove it here since without the new entityId we'd only overwrite it with the same value, just trying to make sense of what entity ID is in general!)

jphechter: TL;DR - Same value stored in 2 different locations.

An Entity ID is a unique identifier used by Treasury, and its publicly viewable. You can put in a business name on the [DORES Business Name Search](https://www.njportal.com/DOR/BusinessNameSearch/Search/BusinessName) website to see this in action.

The `entityId` stored in `profileData` is a value that we're getting back from the Formation API if the user completes business formation through our site. For simplicity sake when dealing with the Tax Clearance form context, we were also storing it in `taxClearanceCertificateData`. We could reference the value in `profileData`, but in the interest of moving forward quickly we were just seeding the value in `taxClearanceCertificateData` with what's in `profileData`.

We wanted to keep the context for this form as simple as possible for 1) quicker initial implementation and 2) because the spec for the API has changed a couple times. I think this was a good move because we just encountered an instance where that information is no longer required.

PR #10105, opened 2025-04-15
green check stepper

ashlsun:
### Ticket

## Code author checklist


web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -120,6 +129,115 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
       fireEvent.click(screen.getByText(Config.taxClearanceCertificateStep3.editButtonText));
       expect(screen.getByTestId("eligibility-tab")).toBeInTheDocument();
     });
+
+    it("renders the first step as complete only when on the eligibility or review tab", () => {
+      renderComponent({});
+      expect(screen.getByTestId(`stepper-0`).dataset.state).toEqual("INCOMPLETE-ACTIVE");
+      fireEvent.click(screen.getByTestId("stepper-1"));
```
FarazA22, line 136: Pebble - we are moving away from data-testid, here is an [RTL](https://testing-library.com/docs/queries/about/#priority) article. I recently added this [PR](https://github.com/newjersey/navigator.business.nj.gov/pull/10113) where I've removed data-testid from this file, it might help.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -120,6 +129,115 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
       fireEvent.click(screen.getByText(Config.taxClearanceCertificateStep3.editButtonText));
       expect(screen.getByTestId("eligibility-tab")).toBeInTheDocument();
     });
+
+    it("renders the first step as complete only when on the eligibility or review tab", () => {
```
FarazA22, line 133: boulder - usually our tests assert a single thing, in this instance the test asserts the first step as complete only when on the eligibility or review tab. You added several unrelated assertions in this test.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -120,6 +129,115 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+      "Zip Code",
+      "Tax ID",
+      "Tax PIN",
+    ])("renders the second step as incomplete when all fields are non empty except for %s", (emptyField) => {
+      renderComponent({ business: generateBusinessWithEmptyTaxClearanceData() });
+      expect(screen.getByTestId(`stepper-1`).dataset.state).toEqual("INCOMPLETE");
+
+      fireEvent.click(screen.getByTestId("stepper-1"));
+
+      if (emptyField !== "Requesting Agency") {
```
FarazA22, line 214: dust - We’re repeating the same if (emptyField !== ...) pattern, which clutters the test and makes it harder to maintain. Since selecting the inputs vary, it might makes sense to group the tests by similar inputs and simplify the test.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
ashlsun: Ah, which assertions did you find unrelated? I just checked and every assertion is about whether the first step is rendered complete or incomplete, based on the tab we should be on after clicking the corresponding step.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
ashlsun: Hmm, what do you mean by "selecting the inputs vary"?  This test is meant to fill out all required fields except one (whichever one is specified by `emptyField`) and make sure that one empty field is enough to trigger the error. I think grouping all required fields in this `it.each` is the only grouping possible, it can't be broken down further.

The other alternative I could see is removing the `it.each` entirely and having a separate test for each individual field, like so:

```
    it("renders the second step as incomplete if Requesting Agency is not filled in", () => {
      renderComponent({ business: generateBusinessWithEmptyTaxClearanceData() });
      fireEvent.click(screen.getByTestId("stepper-1"));

      // Fill in all fields except Requesting Agency
      fillText("Business name", "Test Name");
      fillText("Address line1", "123 Test Road");
      fillText("Address city", "Baltimore");
      selectValueByTestId("addressState", "MD");
      fillText("Address zip code", "21210");
      fillText("Tax id", "012345678901");
      fillText("Tax pin", "1234");

      expect(screen.getByTestId(`stepper-1`).dataset.state).toEqual("INCOMPLETE-ACTIVE");
    });


    it("renders the second step as incomplete if Business Name is not filled in", () => {
      renderComponent({ business: generateBusinessWithEmptyTaxClearanceData() });
      fireEvent.click(screen.getByTestId("stepper-1"));

      // Fill in all fields except Business Name
      selectValueByLabel("Tax clearance certificate requesting agency", "newJerseyBoardOfPublicUtilities");
      fillText("Address line1", "123 Test Road");
      fillText("Address city", "Baltimore");
      selectValueByTestId("addressState", "MD");
      fillText("Address zip code", "21210");
      fillText("Tax id", "012345678901");
      fillText("Tax pin", "1234");

      expect(screen.getByTestId(`stepper-1`).dataset.state).toEqual("INCOMPLETE-ACTIVE");
    });


    it("renders the second step as incomplete if Address line 1 is not filled in", () => {
      ....
    });

    .... and so on, for each field
  });

Comment (ashlsun) [reply web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx:214]: If we prefer separating the test so that there's one test for each individual field, I can do that. I can see how the repeated `if (emptyField !== ...)` pattern might be confusing!

File: web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElementAlert.tsx [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  if (!displayTaxClearanceErrorAlert()) return null;
+
+  return (
+    <Alert variant="error" dataTestid={"tax-clearance-error-alert"}>
+      <div>
+        <div className="text-bold float-left">{Config.taxClearanceCertificateShared.preHeaderErrorText}</div>
+        <div>
+          {props.fieldErrors.length === 1
```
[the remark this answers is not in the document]
josephine-njooi: I just happen to be looking at this code while working on https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7714 😛

Non-blocking fine-sand nit: If I'm understanding the code as is, it looks like "Missing or invalid information." is in one div so that it can be bold, and "Review the following field:" is in a second div so that it can separately not be bold. The first div has `float-left` so that the two pieces of text can be inline.

It seems to me it would be better to use a [span](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/span) here, since the difference is "`<div>` is a block-level element whereas a `<span>` is an inline-level element." Floats also tend to raise warnings in my brain, because they can mess up accessibility (and make it harder to debug) by modifying the visual order of elements to be different from their DOM order.

If you did want to put two divs inline, the `display: flex` [system](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) is the modern solution instead of floats; I personally like [this cheatsheet](https://css-tricks.com/snippets/css/a-guide-to-flexbox/).

Concretely, I'm thinking something like this (the second piece of text also doesn't need an enclosing tag):
```
      <div>
        <span className="text-bold">{Config.taxClearanceCertificateShared.preHeaderErrorText}</span>{" "}
        {props.fieldErrors.length === 1
          ? Config.taxClearanceCertificateShared.singularErrorText
          : Config.taxClearanceCertificateShared.pluralErrorText}
      </div>
```

ashlsun:
- [ ] remove the unnecessary awaits
- [ ] determined that the select___by___ functions are the same, combine them
- [ ] remove the rest of the getbytestids
- [ ] add the space in error alert box and use span in place of div

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
```diff
@@ -122,51 +138,61 @@ export const AnytimeActionTaxClearanceCertificateElement = (props: Props): React
   // TODO: intentionally left since we are picking up next error handling functionality
```
FarazA22, line 138: can we delete the TODO comment? is there more work todo?

ashlsun: I think we can delete this! Thanks

web/decap-config/collections/10-anytime-action.yml
```diff
@@ -257,6 +257,41 @@ collections:
                   required: true,
                   hint: Only applies to tab 2 and 3,
                 }
+              - {
+                  label: "Requesting Agency Error Text",
+                  name: "requestingAgencyErrorText",
+                  widget: "string",
+                  required: true,
+                  hint: Only applies to tab 2 and 3,
```
FarazA22, line 265: pebble - the hint texts need to be reviewed, I don't think the error text appear in tab 3

web/src/components/data-fields/address/AddressTextField.tsx
```diff
@@ -11,15 +12,25 @@ export interface Props extends Omit<GenericTextFieldProps, "value" | "error" | "
   label?: string;
   secondaryLabel?: string;
   errorBarType: "ALWAYS" | "MOBILE-ONLY" | "DESKTOP-ONLY" | "NEVER";
+  required?: boolean | false;
```
FarazA22, line 15: pebble, this is not a valid type, boolean already means true or false

web/src/components/data-fields/address/AddressTextField.tsx
```diff
@@ -11,15 +12,25 @@ export interface Props extends Omit<GenericTextFieldProps, "value" | "error" | "
...
   const { state, setAddressData } = useContext(AddressContext);
   const { doesFieldHaveError } = useAddressErrors();
+  const { state: taxClearanceCertificateData, setTaxClearanceCertificateData } = useContext(
+    TaxClearanceCertificateDataContext
+  );
 
   const handleChange = (value: string): void => {
     props.handleChange && props.handleChange(value);
     setAddressData((prevAddressData) => ({ ...prevAddressData, [props.fieldName]: value }));
+    if (taxClearanceCertificateData !== undefined) {
```
FarazA22, line 28: boulder - it's not clear to me why we are setting state in two different places. can you provide some context?

FarazA22: This component is a shared component, the AddressContext is used to store the value. We should have fewer moving parts in this component; TaxClearanceCertificateDataContext should be removed all together.

web/src/components/data-fields/address/UnitesStatesAddress.tsx
```diff
@@ -13,6 +13,7 @@ interface Props {
   onValidation: () => void;
   isFullWidth?: boolean;
   excludeNJ?: boolean | true;
+  required?: boolean | false;
```
FarazA22, line 16: your using 1 prop to drive the required behavior of most of the fields, this is not a good approach because you are coupling them and address line 2 is not using this prop. The prop should be broken up so the fields can be flexible. It's a bit confusing.

web/src/components/data-fields/ProfileDataField.tsx
```diff
@@ -13,6 +14,7 @@ export interface ProfileDataFieldProps<T = unknown> extends Omit<GenericTextFiel
   inputProps?: OutlinedInputProps;
   type?: HTMLInputTypeAttribute;
   inputWidth?: "full" | "default" | "reduced" | undefined;
+  required?: boolean | false;
```
FarazA22, line 17: fix

web/src/components/data-fields/ProfileDataField.tsx
```diff
@@ -21,6 +23,9 @@ export const ProfileDataField = <T,>({
   ...props
 }: ProfileDataFieldProps<T>): ReactElement => {
   const { state, setProfileData } = useContext(ProfileDataContext);
+  const { state: taxClearanceCertificateData, setTaxClearanceCertificateData } = useContext(
```
FarazA22, line 26: remove

web/src/components/data-fields/address/AddressTextField.tsx
FarazA22: The address components are not using DataFormErrorMapContext. You should look into how alerts and error are firing in other components.

web/src/components/data-fields/address/UnitesStatesAddress.tsx
```diff
@@ -22,7 +23,11 @@ export const UnitesStatesAddress = (props: Props): ReactElement => {
 
   return (
     <>
-      <AddressLines1And2 onValidation={props.onValidation} isFullWidth={props.isFullWidth} />
+      <AddressLines1And2
```
FarazA22, line 26: error handling is not firing for any address components when im tabbint through and all fields are emtpy

web/src/components/data-fields/TaxPin.tsx
```diff
@@ -9,6 +9,7 @@ interface Props {
   children?: ReactNode;
   handleChangeOverride?: (value: string) => void;
   inputWidth?: "full" | "default" | "reduced";
+  required?: boolean | false;
```
FarazA22, line 12: fix

web/src/components/profile/getErrorStateForAddressField.ts
```diff
@@ -78,18 +78,28 @@ export const getErrorStateForAddressField = ({
...
-        "addressMunicipality",
-        "addressZipCode",
-        "addressLine2",
-        "addressCity",
-        "addressProvince",
-        "addressCountry",
-      ],
-      label: (Config.formation.fields as any)[field].error,
-    });
+    let partialAddressError;
```
FarazA22, line 81: There is an issue here, if a user forms their business, then visits the tax clearance anytime action, then the error handling is handled differently than a person who forms then visit the tax clearance anytime action.

web/src/components/profile/getErrorStateForAddressField.ts
```diff
@@ -122,6 +132,7 @@ export const getErrorStateForAddressField = ({
     return fieldWithAssociatedFields({
       associatedFields: [
         "addressLine1",
+        "addressState",
```
FarazA22, line 135: This is causing a bug when forming as a Foreign Nexus / Intl Address screen.

web/src/components/profile/getErrorStateForAddressField.ts
```diff
@@ -167,10 +178,21 @@ export const getErrorStateForAddressField = ({
         ? (Config.formation.fields as any)[field].error
         : (Config.formation.fields as any)[field].foreign.errorUS;
 
-    const partialAddressError = fieldWithAssociatedFields({
-      associatedFields: ["addressMunicipality", "addressLine1", "addressLine2"],
-      label: zipCodeErrorLabel,
-    });
+    let partialAddressError;
```
FarazA22, line 181: this should be revisited, we are introducing two different behaviors based on whether they completed formation or not

web/src/components/profile/ProfileField.tsx
```diff
@@ -17,6 +17,9 @@ interface Props {
   hideHeader?: boolean;
   boldAltDescription?: boolean;
   boldDescription?: boolean;
+  hideLine?: boolean | false;
```
FarazA22, line 20: fix

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -65,6 +67,14 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
     );
   };
 
+  const generateBusinessWithEmptyTaxClearanceData = (): Business => {
+    return generateBusiness({
+      profileData: generateProfileData({ businessName: "", taxId: undefined, taxPin: undefined }),
+      taxClearanceCertificateData: emptyTaxClearanceCertificateData,
+      formationData: generateFormationData({ formationFormData: createEmptyFormationFormData() }),
```
FarazA22, line 74: you should remove formationData: generateFormationData({ formationFormData: createEmptyFormationFormData() }),` from your staged data, the behavior does not rely on formation data

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -123,6 +133,128 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+      fireEvent.click(thirdTab);
+      expect(thirdTab).toHaveAttribute("aria-selected", "true");
+      expect(firstTab).toHaveAttribute("aria-label", expect.stringContaining("State: Complete"));
+    });
+
+    it("renders tab two as complete when all required fields are non empty and valid", async () => {
+      renderComponent({ business: generateBusinessWithEmptyTaxClearanceData() });
+      const secondTab = screen.getAllByRole("tab")[1];
+
+      expect(secondTab).toHaveAttribute("aria-label", expect.stringContaining("State: Incomplete"));
```
FarazA22, line 163: consider placing the assertion under / after the click event.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -123,6 +133,128 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+    it.each(["Business name", "Address line1", "Address city", "Address zip code", "Tax id", "Tax pin"])(
+      "renders tab two as incomplete if the text field %s is empty",
+      (emptyField) => {
+        renderComponent({});
+        const secondTab = screen.getAllByRole("tab")[1];
+        expect(secondTab).toHaveAttribute("aria-label", expect.stringContaining("State: Complete"));
+        fireEvent.click(secondTab);
+
+        fillText(emptyField, "");
+        expect(secondTab).toHaveAttribute("aria-label", expect.stringContaining("State: Incomplete"));
```
FarazA22, line 191: We should be in a error state not incomplete

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -123,6 +133,128 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+        true
+      );
+      fillText("Business name", "Test Name");
+      fillText("Address line1", "123 Test Road");
+      fillText("Address city", "Baltimore");
+      fillText("Address zip code", "21210");
+      fillText("Tax id", "012345678901");
+      fillText("Tax pin", "1234");
+
+      expect(secondTab).toHaveAttribute("aria-label", expect.stringContaining("State: Incomplete"));
```
FarazA22, line 237: given how error handling works for the address component, the stepper should be in an error state

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -673,6 +812,91 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
     });
 
     describe("renders data when input is provided", () => {});
+
+    describe("renders errors on tax clearance step 2", () => {
```
FarazA22, line 816: we should have an onBlur test for the address components it feels odd they are not firing since they are required

web/test/helpers/helpers-testing-library-selectors.ts
```diff
@@ -25,6 +25,23 @@ export const selectDropdownByValue = (label: string, value: string): void => {
   fireEvent.click(listbox.getByTestId(value));
 };
 
+export const selectDropdownValueByText = (
```
FarazA22, line 28: I appreciate that you’re trying to make this utility flexible — that’s really thoughtful. That said, I think it might be better to split this into two separate functions: one for comboboxes and one for dropdowns.

Right now, the opensOnMouseDown flag adds a branching path, which could make it harder to reason about or maintain. Separating them would keep each function focused, clearer to read, and easier to debug if something breaks down the line.

Happy to chat if you’re thinking of another use case I might’ve missed — but I think this split would align better with single-responsibility principles.

web/src/components/data-fields/BusinessName.tsx
```diff
@@ -10,6 +10,8 @@ export const BusinessName = (props: Omit<ProfileDataFieldProps, "fieldName">): R
       {...props}
       fieldName="businessName"
       inputWidth={props.inputWidth ?? "default"}
+      validationText={props.validationText}
```
FarazA22, line 13: we may need a character limit on this

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
```diff
@@ -110,6 +125,7 @@ export const AnytimeActionTaxClearanceCertificateElement = (props: Props): React
 
       setAddressData({
         ...formationAddressData,
+        businessLocationType: business?.formationData.formationFormData.businessLocationType,
```
FarazA22, line 128: This is introducing a buggy behavior and should be removed

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.tsx
```diff
@@ -119,54 +135,61 @@ export const AnytimeActionTaxClearanceCertificateElement = (props: Props): React
...
-  }, [stepIndex]);
+  const taxClearanceCertificateSteps = [
+    {
+      name: Config.taxClearanceCertificateShared.stepperOneLabel,
+      hasError: false,
+      isComplete: stepIndex > 0,
+    },
+    {
+      name: Config.taxClearanceCertificateShared.stepperTwoLabel,
+      hasError: false,
```
FarazA22, line 165: how are hasErrors used? It's unclear why they are hard coded to false

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElementAlert.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { ReactElement } from "react";
+
+interface Props {
+  fieldErrors: string[];
+}
+
+export const AnytimeActionTaxClearanceCertificateElementAlert = (props: Props): ReactElement | null => {
+  const { Config } = useConfig();
+
+  const displayTaxClearanceErrorAlert = (): boolean => props.fieldErrors.length > 0;
```
FarazA22, line 14: we can remove the fn - `const displayTaxClearanceErrorAlert = props.fieldErrors.length > 0;`

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElementAlert.tsx
```diff
@@ -0,0 +1,71 @@
...
+    "addressZipCode",
+    "taxId",
+    "taxPin",
+  ];
+  const filteredOrder = customOrder.filter((item) => props.fieldErrors.includes(item));
+  props.fieldErrors.sort((a, b) => filteredOrder.indexOf(a) - filteredOrder.indexOf(b));
+
+  const getLabel = (field: string): string => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
```
FarazA22, line 33: can we fix the typescript issue without the eslint comment?

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElementAlert.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const formationLabel = Config.formation.fields[field]?.label;
+
+    let requestingAgencyLabel = "";
+
+    if (field === "requestingAgencyId") {
+      requestingAgencyLabel = "Requesting Agency";
```
FarazA22, line 39: this value should exist in the cms somewhere, we don't hard code text that is rendered in the UI because it is not editable by content.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceStepTwo.tsx
```diff
@@ -1,32 +1,59 @@
...
+import { useAddressErrors } from "@/lib/data-hooks/useAddressErrors";
 import { useConfig } from "@/lib/data-hooks/useConfig";
-import { ReactElement } from "react";
+import { useFormContextFieldHelpers } from "@/lib/data-hooks/useFormContextFieldHelpers";
+import { FormEvent, ReactElement } from "react";
 
 interface Props {
   setStepIndex: (step: number) => void;
   saveTaxClearanceCertificateData: () => void;
+  onSave: (event?: FormEvent<HTMLFormElement>) => void;
```
FarazA22, line 23: I noticed there is another on save?

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceStepTwo.tsx
```diff
@@ -1,32 +1,59 @@
...
+  onSubmit: (event?: FormEvent<HTMLFormElement>) => void;
 }
 
 export const TaxClearanceStepTwo = (props: Props): ReactElement => {
   const { Config } = useConfig();
 
+  const { doesFieldHaveError } = useAddressErrors();
+  const { setIsValid: setIsValidAddressLine1 } = useFormContextFieldHelpers(
```
FarazA22, line 31: DataFormErrorMapContext is nod being used by address fields

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceStepTwo.tsx
```diff
@@ -35,33 +62,37 @@ export const TaxClearanceStepTwo = (props: Props): ReactElement => {
...
         <div className="margin-top-3">
           <Heading level={2} styleVariant={"h3"}>
             {Config.taxClearanceCertificateStep2.businessInformationSectionHeader}
           </Heading>
         </div>
-        <div className="margin-bottom-2">
-          <FieldLabelProfile fieldName={"businessName"} />
-          <BusinessName inputWidth="full" />
+        <div className="margin-y-2">
+          <ProfileField fieldName={"businessName"} hideLine={true} hideTopSpace={true} isFullWidth={true}>
```
FarazA22, line 75: Why did we replace FieldLabelProfile with ProfileField? The Business Name field renders the input.

FarazA22: Same question for the other two instances? Previously we were using a label component and thats been replaced with the input component.

web/src/components/StateDropdown.tsx
```diff
@@ -24,9 +25,18 @@ interface Props {
 
 export const StateDropdown = (props: Props): ReactElement => {
   const [open, setOpen] = useState<boolean>(false);
+  const { state: taxClearanceCertificateData, setTaxClearanceCertificateData } = useContext(
```
FarazA22, line 28: remove

FarazA22 (changes requested):
Updated comments / removed anything relevant to Errors

web/src/components/data-fields/address/AddressLines1And2.tsx
```diff
@@ -6,6 +6,7 @@ import { ReactElement } from "react";
 interface Props {
   onValidation: () => void;
   isFullWidth?: boolean;
+  required?: boolean | false;
```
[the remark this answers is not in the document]
ashlsun: changed to just `boolean`

web/src/components/data-fields/tax-id/SingleTaxId.tsx
```diff
@@ -17,13 +18,17 @@ interface Props
   handleChangeOverride?: (value: string) => void;
   getShowHideToggleButton: () => ReactElement;
   taxIdDisplayStatus: TaxIdDisplayStatus;
+  required?: boolean | false;
```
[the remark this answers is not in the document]
ashlsun: changed to just `boolean`

web/src/components/data-fields/tax-id/SplitTaxId.tsx
```diff
@@ -21,6 +21,7 @@ interface Props
   handleChangeOverride?: (value: string) => void;
   getShowHideToggleButton: (toggleFunc?: (taxId: string) => void) => ReactElement;
   taxIdDisplayStatus: TaxIdDisplayStatus;
+  required?: boolean | false;
```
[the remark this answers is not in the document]
ashlsun: changed to just `boolean`

web/src/components/data-fields/ProfileDataField.tsx
ashlsun: changed to just `boolean`

web/src/components/data-fields/TaxPin.tsx
ashlsun: changed to just `boolean`

web/src/components/profile/ProfileField.tsx
ashlsun: changed to just `boolean`

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -123,6 +133,128 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+      fillText("Tax id", "012345678901");
+      fillText("Tax pin", "1234");
+
+      expect(secondTab).toHaveAttribute("aria-label", expect.stringContaining("State: Incomplete"));
+
+      selectDropdownValueByText("Address state", "MD", false);
+      expect(secondTab).toHaveAttribute("aria-label", expect.stringContaining("State: Complete"));
+    });
+
+    // TODO it.each(["Address zip code", "Tax id", "Tax pin"])(
```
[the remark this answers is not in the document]
ashlsun: This was my bad! Had merge conflict issues and left this in confusion, but I have uncommented and fixed this test.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElementAlert.tsx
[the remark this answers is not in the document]
ashlsun: Used the span tag and added a space, thank you for pointing this out!

web/src/components/data-fields/address/AddressTextField.tsx
ashlsun: changed to just `boolean`

jphechter:
@FarazA22 I'm going to be picking this work up while Ashley is OOO.

You've called out _a lot_ of items in here. Appreciate the feedback. I'm trying to get complete context of this work, and identify if all of these items are relevant per the modified scope of the ticket. I may reach out for clarifications, but I think I should be able to identify the necessary items.

web/decap-config/collections/10-anytime-action.yml
jphechter: fixed

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
jphechter: Because the tax clearance data is initial seeded by any values that might be present in the formation address or profile, I think this is here to just doubly ensure that's those fields are empty. When I remove this I'm not seeing any tests fail, but I'm going to leave it in unless there's a strong reason I'm not considering.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
```diff
@@ -123,6 +133,128 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+
+    it("renders tab one as complete when on tab two", () => {
+      renderComponent({});
+      const firstTab = screen.getAllByRole("tab")[0];
+      const secondTab = screen.getAllByRole("tab")[1];
+      expect(firstTab).toHaveAttribute("aria-label", expect.stringContaining("State: Incomplete"));
+
+      fireEvent.click(secondTab);
+      expect(secondTab).toHaveAttribute("aria-selected", "true");
+      expect(firstTab).toHaveAttribute("aria-label", expect.stringContaining("State: Complete"));
```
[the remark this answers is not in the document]
jphechter: I don't think this is true. The `secondTab` is the active tab after the click event.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
jphechter: We're going to incorporate the error states for the stepper in a future ticket.

web/test/helpers/helpers-testing-library-selectors.ts
jphechter: Done

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
```diff
@@ -24,21 +28,36 @@ export const TaxClearanceSteps = (props: Props): ReactElement => {
...
     <>
       <HorizontalStepper steps={props.steps} currentStep={props.currentStep} onStepClicked={onStepClick} />
       {props.currentStep === 0 && <TaxClearanceStepOne setStepIndex={props.stepIndex} />}
       {props.currentStep === 1 && (
-        <TaxClearanceStepTwo
-          setStepIndex={props.stepIndex}
-          saveTaxClearanceCertificateData={props.saveTaxClearanceCertificateData}
-        />
+        <>
+          <AnytimeActionTaxClearanceCertificateElementAlert fieldErrors={props.getInvalidFieldIds()} />
```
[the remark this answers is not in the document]
jphechter: Leaving for now given we're planning to handle this more specifically in a future ticket

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateElement.test.tsx
jphechter: I'm going to save this for the next ticket specifically related to the alert box

web/src/components/data-fields/BusinessName.tsx
jphechter: Adding a note in the ticket.

web/src/components/profile/ProfileField.tsx
```diff
@@ -7,18 +7,18 @@ import { ProfileContentField } from "@/lib/types/types";
...
-  fieldName: ProfileContentField;
+  boldAltDescription?: boolean;
+  boldDescription?: boolean;
   children: ReactNode;
+  displayAltDescription?: boolean;
+  fieldName: ProfileContentField;
+  fullWidth?: boolean;
+  hideHeader?: boolean;
+  hideLine?: boolean;
+  isVisible?: boolean;
```
jphechter, line 18: Consolidated some duplicative styling props and sorted.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateAlert.tsx
```diff
@@ -0,0 +1,72 @@
...
+    // @ts-expect-error
+    const formationLabel = Config.formation.fields[field]?.label;
+
+    let requestingAgencyLabel = "";
+
+    if (field === "requestingAgencyId") {
+      requestingAgencyLabel = Config.taxClearanceCertificateStep2.requestingAgencyLabel;
+    }
+
+    // TODO: Should this be referencing the tax clearance fields?
```
jphechter, line 40: Purposefully left. I'm not sure the answer yet.

web/src/components/tasks/anytime-action/tax-clearance-certificate/helpers.ts
```diff
@@ -0,0 +1,73 @@
+import { LookupMunicipalityByName } from "@businessnjgovnavigator/shared/municipality";
```
jphechter, line 1: Created this file just to move some logic out of the parent components that was making it difficult to track what was happening.

web/src/pages/profile.tsx
```diff
@@ -433,7 +433,7 @@ const ProfilePage = (props: Props): ReactElement => {
               fieldName="homeBasedBusiness"
               isVisible={displayHomedBaseBusinessQuestion()}
               displayAltDescription={displayAltHomeBasedBusinessDescription}
-              compact
+              hideLine
```
jphechter, line 436: There were multiple competing style props, I just consolidated these. See  `ProfileField.tsx` for full list.

web/src/lib/data-hooks/useFormContextFieldHelpers.ts
```diff
@@ -67,7 +67,7 @@ export const useFormContextFieldHelpers = <T, FieldError = FieldErrorType>(
   };
 
   // eslint-disable-next-line react-hooks/rules-of-hooks
-  const isFormFieldInValid = useMemo(
```
jphechter, line 70: Literally just a case change. I saw this and it infuriated me. Particularly because on `81` there were 2 competing names cased slightly differently.

jennyverdeyen (approved):
I think this refactor looks good! I pulled this down and tested out that the fields produce an error as expected, and I'm prohibited from entering certain characters into certain fields like Tax ID and Pin... I couldn't break it! I like how some logic was moved around to make this easier to track what was going on.

web/test/helpers/helpers-testing-library-selectors.ts
```diff
@@ -25,6 +25,32 @@ export const selectDropdownByValue = (label: string, value: string): void => {
   fireEvent.click(listbox.getByTestId(value));
 };
 
+export const selectDropdownValueByTextMouseDown = (
```
FarazA22, line 28: Pebble - I have a few suggestions:
1. I think `selectLocationByText` (line 4) can be refactored away, and we use `selectDropdownByValue` instead. This can be a separate refactor.
2. The two functions created are specifically for ComboBoxes, the name should reflect this. `selectDropdownByValue` uses a mouseDown event as well, but that is specifically used for MUI Select components (non-combo boxes). The name of that function should be updated to reflect this.

content/src/fieldConfig/tax-clearance-certificate-shared.json
```diff
@@ -4,6 +4,11 @@
     "backButtonText": "Back",
     "stepperOneLabel": "Requirements",
     "stepperTwoLabel": "Check Eligibility",
-    "stepperThreeLabel": "Review"
+    "stepperThreeLabel": "Review",
+    "preHeaderErrorText": "Missing or invalid information.",
+    "singularErrorText": "Review the following field:",
+    "pluralErrorText": "Review the following fields:",
+    "requestingAgencyErrorText": "Select a Requesting Agency",
+    "businessNameErrorText": "Enter a Business Name"
```
FarazA22, line 12: dust - should `businessNameErrorText` live with the business name field in the `profile.json`? There could be other cases where the business name input is required.

FarazA22: I noticed validationText it provided via props, I don't expect to see this content changing when used across different content fields. Could this be a ticket that is immediately picked up?

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { getFlow, scrollToTop, useMountEffectWhenDefined } from "@/lib/utils/helpers";
+import { emptyTaxClearanceCertificateData } from "@businessnjgovnavigator/shared";
+import {
+  emptyFormationAddressData,
+  FormationAddress,
+} from "@businessnjgovnavigator/shared/formationData";
+import { emptyProfileData, ProfileData } from "@businessnjgovnavigator/shared/profileData";
+import { Dispatch, ReactElement, SetStateAction, useState } from "react";
+
+const pickData = <T extends object, K extends keyof T>(obj: T, keys: K[]): Pick<T, K> => {
```
FarazA22, line 23: dust - I think this file should liev in `@/contexts/taxClearanceCertificateDataContext`, it would be easier to find from that file.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
```diff
@@ -1,45 +1,73 @@
...
   const onStepClick = (step: number): void => {
-    if (step === 2 && props.currentStep === 1) {
-      props.saveTaxClearanceCertificateData();
+    props.saveTaxClearanceCertificateData();
+    setStepIndex(step);
+  };
+
+  const onSave = (): void => {
+    props.saveTaxClearanceCertificateData();
+    if (props.isValid()) {
```
FarazA22, line 38: boulder -  This specific line of code is causing some unexpected behavior. If you’re on the Check Eligibility tab and click Save and Continue to reach the Review page, then navigate back to Check Eligibility using the stepper or back button, an alert appears — but it only includes the fields except for the address fields, which are missing.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
```diff
@@ -1,45 +1,73 @@
...
+      setStepIndex(2);
```
FarazA22, line 39: This line of code only gets invoked when the if condition is true, which is causing some odd behavior: I’m unable to use the primary button to move forward to the next step, but I can still navigate forward using the stepper when an error exist on the eligibility tab.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
```diff
@@ -1,45 +1,73 @@
...
     }
-    props.setStepIndex(step);
   };
 
-  const steps: { component: ReactElement }[] = [
-    { component: <Requirements setStepIndex={props.stepIndex} /> },
+  const stepperSteps: StepperStep[] = [
```
FarazA22, line 43: boulder -

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.test.tsx
```diff
@@ -129,6 +143,144 @@ describe("<AnyTimeActionTaxClearanceCertificateReviewElement />", () => {
...
+      const firstTab = screen.getAllByRole("tab")[0];
+      const thirdTab = screen.getAllByRole("tab")[2];
+      expect(firstTab).toHaveAttribute("aria-label", expect.stringContaining("State: Incomplete"));
+
+      fireEvent.click(thirdTab);
+      expect(thirdTab).toHaveAttribute("aria-selected", "true");
+      expect(firstTab).toHaveAttribute("aria-label", expect.stringContaining("State: Complete"));
+    });
+
+    it("renders tab one as incomplete when viewing tab one and not all required data is provided", () => {
```
FarazA22, line 171: test should be reevaluated based on changes made

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
FarazA22: Actually, it’s worth considering — since we don’t want to overcomplicate things — whether we should just stick to the existing pattern: showing errors only upon final submission. That would keep the logic simpler and more consistent with the business formation flow.

Oh, and another note: there’s some wonky behavior happening. If I go straight to the final screen, hit Save and Continue, then navigate back to the second tab where the alert is being generated, the icon is still marked as completed — even though the data is incomplete.

Again, not trying to over-engineer this, but the additional logic around the stepper icons is starting to feel pretty inconsistent and a bit funky overall.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateAlert.tsx
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export const AnytimeActionTaxClearanceCertificateAlert = (props: Props): ReactElement | null => {
+  const { Config } = useConfig();
+
+  const fieldErrorsOrder = [
+    "requestingAgencyId",
+    "businessName",
+    "entityId",
+    "addressLine1",
```
FarazA22, line 18: pebble - One thing we’re not currently addressing is the issue with empty address fields not appearing in the alert. Right now, if the tax clearance data is empty, all the relevant fields show up in the alert except for the address fields — those are missing and should probably be included for consistency.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificateAlert.tsx
FarazA22: The comment you left is an interesting one. I don’t think it makes sense to reference the tax clearance field names directly. For example, if you look at line 37, the field name is requestingAgencyId, but that’s not the label we want to display.

web/src/components/tasks/anytime-action/tax-clearance-certificate/StateAgencyDropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+  LookupTaxClearanceCertificateAgenciesById,
+} from "@businessnjgovnavigator/shared";
+import { FormControl, FormHelperText, MenuItem, Select, SelectChangeEvent } from "@mui/material";
+import { ReactElement, ReactNode, useContext } from "react";
+
+interface Props {
+  preventRefreshWhenUnmounted?: boolean;
+}
+
+export const StateAgencyDropdown = (props: Props): ReactElement => {
```
FarazA22, line 20: We have a data-fields folder that’s meant to house all of our input components — dropdowns, text fields, shared fields, etc. So it’s starting to feel a bit off that we’re now creating input components and storing them close to specific pages.

That kind of fragmentation doesn’t sit well long term. Even if a component is tied to a specific use case, I think it’s still valuable to keep it within a shared structure for consistency and discoverability. That’s actually why I named the component TaxClearanceCertificateAgencyDropdown — it’s hyper-specific, but it lives in the central data-fields area where similar components belong.

This might be a sign we need to start thinking about some light restructuring or setting clearer guidelines for how we handle these patterns moving forward. Curious to hear your thoughts.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
```diff
@@ -1,45 +1,73 @@
...
+    {
+      name: Config.taxClearanceCertificateShared.stepperOneLabel,
+      hasError: false,
+      isComplete: stepIndex > 0 || getAllFieldsNonEmpty(props.taxClearanceCertificateData),
```
FarazA22, line 47: I think we should remove the getAllFieldsNonEmpty function entirely.

The main issue with it is that it’s only checking whether fields are non-empty or if a state has been selected — but that completely ignores other validation rules we have in place, like minimum and maximum character limits. Because of that, it gives a false sense of completeness even when validation errors are present.

For step one, I think the most important check is whether the current index is greater than 0 — and that alone should determine whether the user has progressed past that step. That keeps things simple and aligned with how we’ve handled validation elsewhere.

If you look at the formation flow, we’ve intentionally kept validation logic very lightweight — the alerts and stepper icons only update after the final submit button is pressed. I think we should follow that same pattern here to avoid unnecessary complexity.

As for the second step, I believe the existing isValid function should be enough to handle what we need.

web/src/components/tasks/anytime-action/tax-clearance-certificate/helpers.ts
```diff
@@ -0,0 +1,73 @@
...
+    addressLine2,
+    addressCity,
+    addressState,
+    addressZipCode,
+    taxId,
+    taxPin,
+  };
+};
+
+export const getAllFieldsNonEmpty = (
```
FarazA22, line 60: This function is not correct, and the test we have for the green state, need to be revisited b/c once this goes away, the test fail.

web/test/helpers/helpers-testing-library-selectors.ts
jphechter: 1. I'm going to leave that alone. That's outside the scope of the work. If this is a change we want to implement, let's handle that separately.
2. I can change the function names to be `selectCombobox*`, but I initially made this distinction based on previous feedback you'd provided. If you have a better name for the functions in mind, let me know and I will incorporate.

content/src/fieldConfig/tax-clearance-certificate-shared.json
jphechter: If we move this to `profile.json` and use globally, that implies we'll want the same error text everywhere. I'm also doubtful that we're going to need to change this, but I think this would be a new pattern for us.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.tsx
jphechter: Sure. We can move this.

web/src/components/tasks/anytime-action/tax-clearance-certificate/StateAgencyDropdown.tsx
jphechter: Interesting, for a very similar reason I opted to put it in a location closer to where it will be used. This is a pattern we do follow other places in the application, business formation for example contains many components that are only used within that task.

Keeping this co-located with the task where it will be used also means that we can reasonably give the component a shorter name.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
jphechter: I'll take a look. Given we're primarily focused on the stepper states with this ticket, I'm tempted to leave this for now.

web/src/components/tasks/anytime-action/tax-clearance-certificate/helpers.ts
jphechter: Will take a look.

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.test.tsx
jphechter: Can you clarify what needs to be re-evaluated? I'm not I'm seeing any issues.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
jphechter: I opted to kill this. I don't think anyone's going to miss it.

PR #10131, opened 2025-04-18
chore: [AB#12942] notesMd validation to not include #

somabadri:
### Ticket

## Code author checklist

### DIFF
--- a/web/public/mgmt/config.yml
+++ b/web/public/mgmt/config.yml
@@ -38,7 +38,13 @@ collections:
     editor:
       preview: true
     fields:
-      - { label: "Internal Notes", name: "notesMd", widget: "markdown", required: false }
+      - {
+          label: "Internal Notes",
+          name: "notesMd",
+          widget: "markdown",
+          required: false,
+          pattern: ["^(?!.*#).*", "It cannot include a #"],
+        }
       - { label: "Summary Description", name: "summaryDescriptionMd", widget: "markdown" }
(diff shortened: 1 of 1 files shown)


web/public/mgmt/config.yml
```diff
@@ -38,7 +38,13 @@ collections:
     editor:
       preview: true
     fields:
-      - { label: "Internal Notes", name: "notesMd", widget: "markdown", required: false }
+      - {
+          label: "Internal Notes",
+          name: "notesMd",
+          widget: "markdown",
+          required: false,
+          pattern: ["^(?!.*#).*", "It cannot include a #"],
```
seidior, line 46: [dust] We may want to workshop the message to the user a bit, if you're open to it? What do you think about something like "Please remove the # in the URL and anything after it before saving."

Also, my read of this regex — and I could be wrong here — is that it'll match any string for this field that includes an octothorp rather than just URLs. Is that correct?

Finally, do we want to look through all the content in flight and remove the "#" and anything after it in advance of this, perhaps as a [sand]?

ZacharyRizer (approved):
LGTM

web/public/mgmt/config.yml
hinzed1127: > Please remove the # in the URL

It's not specifically URLs, though, right? That's just typically what this field is used for?

hinzed1127 (approved):
Looks good, deferring to you re any follow-up changes based on Spencer's comments

PR #10149, opened 2025-04-21
chore: [AB#0000] add warning to existing ts disable usages

josephine-njooi:
## Description

Existing usages of `ban-ts` without justifying comments seem to have facilitated unintentional propagation of disables in the codebase, see [this discussion](https://njcio.slack.com/archives/C04J8MTJNS2/p1745266548347089?thread_ts=1745266544.420649&cid=C04J8MTJNS2). Added comments to existing usages as a warning not to propagate.

### Ticket

## Code author checklist

### DIFF
--- a/web/src/components/profile/ProfileErrorAlert.tsx
+++ b/web/src/components/profile/ProfileErrorAlert.tsx
@@ -15,6 +15,7 @@ export const ProfileErrorAlert = (props: Props): ReactElement | null => {
   const displayProfileErrorAlert = (): boolean => props.fieldErrors.length > 0;
 
   const getLabel = (field: string): string => {
+    // We should try not to do this; if you do need to disable typescript please include a comment justifying why.
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-expect-error
     const formationLabel = Config.formation.fields[field]?.label;
--- a/web/test/pages/cms.test.tsx
+++ b/web/test/pages/cms.test.tsx
@@ -13,6 +13,7 @@ describe("cms", () => {
         "utf8"
       );
 
+      // We should try not to do this; if you do need to disable typescript please include a comment justifying why.
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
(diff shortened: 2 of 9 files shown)


web/src/components/profile/ProfileErrorAlert.tsx
```diff
@@ -15,6 +15,7 @@ export const ProfileErrorAlert = (props: Props): ReactElement | null => {
   const displayProfileErrorAlert = (): boolean => props.fieldErrors.length > 0;
 
   const getLabel = (field: string): string => {
+    // This is tech debt; if you do need to disable typescript please include a comment justifying why.
```
jphechter, line 18: I might modify this slightly in all instances. In some cases, I'm not sure that I would classify this as technical debt.

In this specific case, we're running into an issue where we cannot easily predict a type. We're not doing something "wrong", but it doesn't nicely conform to TypeScript's rules. I would say something similar for usages related to migrations. There's a pretty good reason for disabling this rule in those locations.

```suggestion
    // If you do need to disable typescript please include a comment justifying why.
```

web/test/pages/cms.test.tsx
```diff
@@ -13,6 +13,7 @@ describe("cms", () => {
         "utf8"
       );
 
+      // This is tech debt; if you do need to disable typescript please include a comment justifying why.
```
jphechter, line 16: I might modify this slightly in all instances. In most use cases, I'm not sure that I would classify this as technical debt.

In this specific case, we're running into an issue where we cannot easily predict a type. We're not doing something "wrong", but it doesn't nicely conform to TypeScript's rules. I would say something similar for usages related to migrations. There's a pretty good reason for disabling this rule in those locations.

```suggestion
    // If you do need to disable typescript please include a comment justifying why.
```

josephine-njooi: I like this note, thanks! I did wanted to convey that this is not preferable, not just "it's okay to do, just add a comment". How about:
> We should try not to do this; if you do need to disable typescript please include a comment justifying why.

?

jphechter: I like that better.

PR #10201, opened 2025-04-30
chore: [AB#7782] Tax Id Prefactor for Tax Pin

josephine-njooi:
## Description

#[7782](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7782) involves encrypting and building show/hide functionality for the tax pin component. 

Commit 1: refactors the tax id components to pull out and more generically name common functions that would be reused for the tax pin component. 
- Moves and renames `TaxIdDisplayStatus` to more generically be in `ShowHideToggleButton.tsx`
- Pulls out `getTaxIdEncryptionStatus` -> `isEncrypted`
- Pulls out `taxIdInitialDisplay` -> `getInitialShowHideStatus`

Commit 2: renames names around `AWSEncryptionDecryptionFactory`, the `/api/decrypt` endpoint, and the apiClient call to the endpoint to more generically refer to encryption and decryption without specifying `tax`. Does not touch anything higher-level than that, e.g. `getDecryptedTaxId` on the front or `getTaxId` on the back.

Commit 3: renames CMS `tax.json` -> `tax-id.json` in anticipation of one for tax-pin

### Ticket

This commit contributes to #[7782](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7782).

## Code author checklist

- [x] I have updated relevant `.env` values in both `.env-template` and in Bitwarde

### DIFF
--- a/web/src/components/ShowHideToggleButton.tsx
+++ b/web/src/components/ShowHideToggleButton.tsx
@@ -2,8 +2,10 @@ import { ButtonIcon } from "@/components/ButtonIcon";
 import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
 import { ReactElement } from "react";
 
+export type ShowHideStatus = "text-view" | "password-view";
+
 interface Props {
-  status: "text-view" | "password-view";
+  status: ShowHideStatus;
   toggle: () => Promise<void>;
   hideText: string;
   showText: string;
--- a/web/src/components/data-fields/tax-id/DisabledTaxId.tsx
+++ b/web/src/components/data-fields/tax-id/DisabledTaxId.tsx
@@ -1,13 +1,12 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { EncryptionStatus, TaxIdDisplayStatus } from "@/components/data-fields/tax-id/TaxIdHelpers";
-import { ShowHideToggleButton } from "@/components/ShowHideToggleButton";
+import { type ShowHideStatus, ShowHideToggleButton } from "@/components/ShowHideToggleButton";
 import { ProfileDataContext } from "@/contexts/profileDataContext";
 import { decryptTaxId } from "@/lib/api-client/apiClient";
 import { useConfig } from "@/lib/data-hooks/useConfig";
 import { formatTaxId } from "@/lib/domain-logic/formatTaxId";
 import { MediaQueries } from "@/lib/PageSizes";
-import { maskingCharacter, ProfileData } from "@businessnjgovnavigator/shared";
+import { getInitialShowHideStatus, isEncrypted } from "@/lib/utils/encryption";
 import { useMediaQuery } from "@mui/material";
 import { ReactElement, ReactNode, useContext, useEffect, useMemo, useState } from "react";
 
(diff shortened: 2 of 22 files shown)


web/src/components/data-fields/tax-id/DisabledTaxId.tsx
```diff
@@ -21,38 +20,20 @@ export const DisabledTaxId = (props: Props): ReactElement => {
...
-      return "password-view";
-    } else {
-      return "text-view";
-    }
-  };
-
-  const taxIdInitialDisplay = (): TaxIdDisplayStatus => {
-    return getTaxIdInitialStatus(getTaxIdEncryptionStatus(state.profileData));
-  };
+  }, [state.profileData.taxId, state.profileData.encryptedTaxId]);
```
josephine-njooi, line 32: Took the opportunity to add the other reactive value. Realistically, I don't think we've ever encountered the bug because if `state.profileData.encryptedTaxId` changes `state.profileData.taxId` should have as well

web/src/lib/utils/encryption.ts
```diff
@@ -0,0 +1,16 @@
+import { ShowHideStatus } from "@/components/ShowHideToggleButton";
+import { maskingCharacter } from "@businessnjgovnavigator/shared";
+
+export const isEncrypted = (
+  maskedValue: string | undefined,
+  encryptedValue: string | undefined
+): boolean | undefined => {
+  if (maskedValue === undefined || !encryptedValue) {
+    return;
+  }
```
josephine-njooi, line 10: I'm not sure there's much value in returning `undefined`, but I went with preserving fidelity to the original functions, which tested undefined for maskedValue but falsy-ness for encryptedValue.

web/src/components/data-fields/tax-id/DisabledTaxId.tsx
```diff
@@ -21,48 +20,32 @@ export const DisabledTaxId = (props: Props): ReactElement => {
...
-      status: taxIdDisplayStatus,
-      toggle: updateTaxIdDisplay,
-      showText: Config.tax.showButtonText,
-      hideText: Config.tax.hideButtonText,
-      useOverrideText: !isTabletAndUp,
-      showOverrideText: Config.tax.showButtonTextMobile,
-      hideOverrideText: Config.tax.hideButtonTextMobile,
-    });
+    return (
+      <ShowHideToggleButton
```
josephine-njooi, line 39: Changed the syntax so that React Devtools detects this as a component

web/src/components/data-fields/tax-id/DisabledTaxId.tsx
```diff
@@ -21,48 +20,32 @@ export const DisabledTaxId = (props: Props): ReactElement => {
   const { state } = useContext(ProfileDataContext);
   const { Config } = useConfig();
 
+  const taxIdIsEncrypted = isEncrypted(state.profileData.taxId, state.profileData.encryptedTaxId);
```
seidior, line 23: [dust] Does this need to be memoized, if it's based on a state variable? Should this variable be reactive, or do we only care about this on re-render?

web/src/components/ShowHideToggleButton.tsx
```diff
@@ -2,8 +2,10 @@ import { ButtonIcon } from "@/components/ButtonIcon";
 import { UnStyledButton } from "@/components/njwds-extended/UnStyledButton";
 import { ReactElement } from "react";
 
+export type ShowHideStatus = "text-view" | "password-view";
```
seidior, line 5: [dust] Could this name be more descriptive, in case we want to show/hide things that are not passwords in the future?

seidior (commented):
For the most part, seems like a fairly straightforward prefactor. Great work Josephine! Just a couple of questions for you.

FarazA22 (approved):

web/src/components/data-fields/tax-id/DisabledTaxId.tsx
josephine-njooi: > Does this need to be memoized, if it's based on a state variable?

I'm not really sure what you're getting at here; "it's based on a state variable" it isn't one of the [usage reasons for useMemo](https://react.dev/reference/react/useMemo#usage)? If the concern is that we're running this on every render, scroll down to the note and two deep dive boxes in the docs for why that isn't an issue?

> Should this variable be reactive, or do we only care about this on re-render?

I'm also not sure I'm tracking with this question 😅. By "Should this variable be reactive", do you mean should this be a state? I would say no, because it can be fully derived from other state (having unnecessary "intermediate" state is how one [ends up with unnecessary useEffects](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state)).
But your question says, "or do we only care about this on re-render?". I would say yes we only care on re-render, because this is fully derived from only things in state. Not I'm sure if I'm missing something, but I guess I can't think of what external force might caused this to change outside of a render cycle?

web/src/components/ShowHideToggleButton.tsx
josephine-njooi: `<input type="password">` is a spec for the hidden/masked dots, so I would say that `password` here generically refers to the input type, rather than the contents of the input. We're not actually using this type for any password fields 😛. So my 2c is to leave as is.

PR #10219, opened 2025-05-01
fix: [AB#14182] capitalization contextual info issue for content repo

adidner:
## Description

this is an impromptu bug that came as a result of a content person putting capitals in an ID, 

This broke our system because the ID had capital letters but the file name (which is suppose to match the ID) did not. 

### Ticket

This pull request resolves [#14182](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14182).

### Approach

I decided to make it so ID's from the content side are forced to be lower case as well. 

### Steps to Test

Go to the cms locally and try to type in capital letters into the ID field and see that you can't (it automatically lower cases them)

### Notes

I could have taken the opposite approach and tried to get the filenames to save with uppercase letters but I didn't for a few reasons. 

1) I think this is easier
2) I think decap might controll that so it may be harder to manipulate
3) our file names will be more consistent which is generally better

## Code author checklist


seidior (approved):
I'm guessing we might want to do more sanitization for this in the future, but I think this adequately covers the specific issue in this case.

PR #10273, opened 2025-05-06
feat: [AB#7782] Encrypt Profile Tax Pin

josephine-njooi:
## Description

Previously, tax id was masked and encrypted in the profile page, but not tax pin. This PR updates the profile tax pin field to also be masked and encrypted.

### Commit 1: Migration for users with existing values in profile tax pin

This commit adds a migration that
1. Adds an `encryptedTaxPin` field that is an encryption of the profile `taxPin` field if defined
2. Masks the the profile `taxPin` field with `*`

We are okay doing this encryption on a rolling per-user basis, because likely not many users have put in their tax pin, en we have ongoing work to run migrations for all users.

This is the first migration that makes an async call within the migration function. Wrapping a `Promise.resolve()` turns all migration function return values into a promise that is then awaited.

This also makes the assumption that no profile `taxPin`s are masked as of v160.

### Commit 2: Encrypt on user save and decrypt in frontend component

This commit
- Adds masking and encrypting the profile tax pin, if not yet masked and encrypted, to userRouter (this is the only backend change)
- Adds functionality to the `TaxPin` component to decrypt tax pin if encrypted and toggle show/hide
- Adds `encryptedTaxId` and `encryptedTaxPin` fields on `taxClearanceCertificateData`, so they can be populated by default by profile data to be provided to the component for decryption
    - Encrypting the taxId and taxPin fields on tax clearance certificate will be handled in a [separate ticket](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/12793)

It also
- Fixes a tiny bug in `TaxId` component show/hide, where if a user saves their tax ID, shows and deletes it, clicks to e.g. "Permits & Licenses" then clicks back, clicking "Show" does nothing and they cannot change the field without a refresh.

Commits are not independently deployable and will be squashed.

### Ticket

This PR resolves #[7782](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7782).

### Screenshots

<details>
<summary>Collapsible profile screenshots! 😄</summary>

Profile, before
image

Profile, after
image
image

</details>

<details>
<summary>Collapsible tax clearance screenshots! 😄</summary>
Tax clearance, before
image

Tax clearance, after
image
image
</details>

#### Profile page
- From the dashboard, click "Business Profile", then "Reference Numbers"
- Open the devtools network tab
- Input a tax pin. See that you can show/hide without triggering an API call
- Save the tax pin, then return back to the input
- Clicking show on the tax pin for the first time should reveal the same tax pin you input, and trigger a call to /decrypt
- Subsequent show/hides should not trigger a call to /decrypt
- Try to break it!

#### Tax clearance
- In `web/.env`, set `FEATURE_TAX_CLEARANCE_CERTIFICATE=true`
- Fill in tax pin from the profile (see above)
- On the dashboard page, open the anytime actions dropdown, select Get a Tax Clearance Certificate
- Click step two and see that the tax pin should pre-populate. Showing it should trigger a network call to /decrypt
- The value can be changed and saved. Afterwards, the new value should show up hidden, and be show/hideable. This new value should not call /decrypt, since the "new" value is saved to the tax clearance copy of tax pin which this PR it does not implement encrypting. The new value should also not propagate to the profile page.
- If you create a business and don't fill in tax pin from the profile, there should be no repopulation in tax clearance
- Try to break it!

#### The migration
- From main, create a v160 user, fill in tax pin in the profile. Copy the db data
- Check this branch out, insert the v160 data, do something and verify that the tax pin it is now encrypted and can be decrypted in profile

## Code author checklist


web/src/components/data-fields/tax-id/TaxId.tsx
```diff
@@ -43,15 +42,9 @@ export const TaxId = (props: Props): ReactElement => {
...
-    getInitialShowHideStatus(taxIdIsEncrypted),
+    getInitialShowHideStatus(business?.profileData.taxId),
   );
 
-  useEffect(() => {
-    if (business?.profileData.taxId?.includes(maskingCharacter)) {
-      setTaxIdDisplayStatus("password-view");
-    }
-  }, [business?.profileData.taxId]);
-
```
josephine-njooi, line 54: Previously (even pre-[prefactor](https://github.com/newjersey/navigator.business.nj.gov/pull/10201/files#diff-a76818ee1524cb1d4f0a3e7a0e637427fbde97ae33bc0f2933741861358c10bc)), this component would
1. Set initial `taxIdDisplayStatus` to whether the local state `taxId` is masked
2. Immediately override that, via this useEffect, with whether the database `taxId` is masked

Regardless of whether the two are the same, we're always clobbering (1) with the value of (2). Rather than initializing to one thing and then always clobbering with the other, I'm just initializing this to (2).

This previous code does have the now-unsupported functionality that if `business?.profileData.taxId` changes without a remount, we will not re-initialize. However, in our two usages one needs to click "save" in order to make the db `business` value change, and doing so unmounts the component.

(And if the db `business` value did change without unmount, the [recommendation](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes) is to specify a key instead of using a useEffect.)

web/src/components/data-fields/tax-id/TaxId.tsx
```diff
@@ -79,18 +72,14 @@ export const TaxId = (props: Props): ReactElement => {
   };
 
   const taxIdToggle = (toggleFunc?: (taxId: string) => void) => async (): Promise<void> => {
-    if (!state.profileData.taxId) {
-      return;
-    }
-
```
josephine-njooi, line 85: This caused a bug where if your local state taxid is empty (e.g. because you deleted a previously saved value) but your db state is populated because of that previously saved value (thereby initialing the view to password-view), you are locked in password view, and clicking "Show" does nothing so you can't change it.

Now, clicking "Show" with db masked and local state empty just toggles show/hide with an empty local state value.

web/src/components/data-fields/TaxPin.tsx
```diff
@@ -23,17 +37,59 @@ export const TaxPin = (props: Props): ReactElement => {
...
         handleChange={props.handleChangeOverride}
+        inputProps={{
+          endAdornment: (
+            <InputAdornment position="end">
+              {isTabletAndUp && getShowHideToggleButton()}
+            </InputAdornment>
+          ),
+        }}
       />
-      {props.children}
```
josephine-njooi, line 36: Removing because there are no existing usages, now or during the [initial introduction](https://github.com/newjersey/navigator.business.nj.gov/pull/1502/files#diff-1d81a59cd0060217367a29df40a58fee03d2a369a2aeb7a6a34da10e9a6898ad) of this line and component.

web/src/components/data-fields/TaxPin.tsx
```diff
@@ -23,17 +37,59 @@ export const TaxPin = (props: Props): ReactElement => {
...
+    } else {
+      setShowHideStatus("password-view");
+    }
+  };
+
   return (
     <>
       <NumericField
-        inputWidth={props.inputWidth ?? "default"}
+        inputWidth={props.inputWidth ?? "reduced"}
```
josephine-njooi, line 73: Since taxId also defaults to "reduced"

api/src/domain/user/encryptFieldsFactory.ts
```diff
@@ -0,0 +1,80 @@
...
+};
+
+export const encryptTaxIdForBatchLambdaFactory = (
+  encryptionDecryptionClient: EncryptionDecryptionClient,
+): ((userData: UserData) => Promise<UserData>) => {
+  return async (userData: UserData): Promise<UserData> => {
+    const userDataWithEncryptedTaxId = await encryptTaxId(userData, encryptionDecryptionClient);
+    return userDataWithEncryptedTaxId;
+  };
+};
```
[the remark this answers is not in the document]
aadedejifearless: I do not believe its still in use per conversations I have had with @somabadri  and @SirSamuelJoseph in the past. Can you guys confirm this ?

aadedejifearless (approved):
Looks good to me.

api/src/api/userRouter.ts
```diff
@@ -138,7 +138,7 @@ export const userRouterFactory = (
   logger: LogWriterType,
 ): Router => {
   const router = Router();
-  const encryptTaxId = encryptTaxIdFactory(encryptionDecryptionClient);
+  const encryptFields = encryptFieldsFactory(encryptionDecryptionClient);
```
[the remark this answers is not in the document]
jphechter: If you wanted to be slightly more generic, you could just say
```
  const encryptData = encryptionFactory(encryptionDecryptionClient);
```

`encryptData` may still imply _everything_ is encrypted, but I don't hate either option.

api/src/client/ApiFormationHealth.ts
```diff
@@ -41,6 +41,9 @@ const profileData: ProfileData = {
   entityId: "8206864808",
   employerId: "366451721",
   taxId: "752066817169",
+  encryptedTaxId: undefined,
```
jphechter, line 44: Not your fault, but I really, really, really hate this pattern. It's yet another instance where we need to manually update a field value.

I desperately want a refactor that generates this for us.

api/src/domain/user/encryptFieldsFactory.ts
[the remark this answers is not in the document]
jphechter: I believe it was run once, and hasn't been used since. We're probably ok, but glad we're considering backwards compatibility.

web/src/components/data-fields/TaxPin.test.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    setLargeScreen(true);
+    profileData = generateProfileData({
+      businessPersona: "STARTING",
+    });
+  });
+
+  describe("TaxPin", () => {
```
jphechter, line 62: [dust] I don't think we need this inner `describe`

josephine-njooi: Done, thanks!

web/src/components/data-fields/TaxPin.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+    });
+  });
+
+  it("pre-populates from profileData", () => {
+    renderComponent({
+      ...profileData,
+      taxPin: "****",
+    });
+    expect(screen.getByLabelText("Tax pin") as HTMLInputElement).toHaveValue("****");
+    expect((screen.getByLabelText("Tax pin") as HTMLInputElement).type).toEqual("password");
```
josephine-njooi, line 68: Re-posting comments removed by force push.

Per https://testing-library.com/docs/queries/byrole/, `<input type="password" />` has no implicit role and needs to be queried via `getByLabelText`.

web/src/components/data-fields/TaxPin.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+      ...profileData,
+      taxPin: "",
+    });
+    const taxPinInput = screen.getByRole("textbox", { name: "Tax pin" });
+    fireEvent.click(taxPinInput);
+    fireEvent.change(taxPinInput, {
+      target: { value: "123" },
+    });
+    fireEvent.blur(taxPinInput);
+    expect(taxPinInput).toHaveAccessibleDescription(configForField.errorTextRequired);
```
josephine-njooi, line 99: The taxId equivalent just queries if the error text is present on the screen, but here a stronger assertion is made that the error text is present as the accessible description on the input.

PR #10288, opened 2025-05-07
feat: [AB#7782] Prefactor pass encryption client to migration functions

josephine-njooi:
## Description

Commit 1: Define `MigrationClients` type and update the 160 migration functions to take in `MigrationClients`
Commit 2: Pass encryption client to migration functions

The migration to encrypt tax pin ([wip pr](https://github.com/newjersey/navigator.business.nj.gov/pull/10273)) requires passing an encryption client to the migration. This PR modifies migration function signatures to take in the encryption client, and passes them the client.

### Ticket

This PR contributes to #[7782](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/7782).

### Steps to Test

- Checkout an old version of main with migration version < 160, create a user, then copy their db data (`yarn dynamo`)
- Check out this branch. Replace the user data with the old version.
- Use the website, then check the db data to verify that the migration worked.

## Code author checklist

### DIFF
--- a/api/src/db/migrations/v100_add_updated_timestamp.ts
+++ b/api/src/db/migrations/v100_add_updated_timestamp.ts
@@ -1,3 +1,4 @@
+import { type MigrationClients } from "@db/migrations/types";
 import { v99UserData } from "@db/migrations/v99_added_encrypted_tax_id";
 import { randomInt } from "@shared/intHelpers";
 
@@ -15,7 +16,11 @@ export interface v100UserData {
   version: number;
 }
 
-export const migrate_v99_to_v100 = (v99Data: v99UserData): v100UserData => {
+export const migrate_v99_to_v100 = (
+  v99Data: v99UserData,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _?: MigrationClients,
+): v100UserData => {
   return {
     ...v99Data,
     licenseData: v99Data.licenseData
--- a/api/src/db/migrations/v101_change_error_field.ts
+++ b/api/src/db/migrations/v101_change_error_field.ts
@@ -1,3 +1,4 @@
+import { type MigrationClients } from "@db/migrations/types";
 import { v100UserData } from "@db/migrations/v100_add_updated_timestamp";
 import { randomInt } from "@shared/intHelpers";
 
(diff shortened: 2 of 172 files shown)


api/src/db/migrations/v101_change_error_field.ts
```diff
@@ -15,7 +16,11 @@ export interface v101UserData {
   version: number;
 }
 
-export const migrate_v100_to_v101 = (v100Data: v100UserData): v101UserData => {
+export const migrate_v100_to_v101 = (
+  v100Data: v100UserData,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _: MigrationClients | undefined = undefined,
```
josephine-njooi, line 22: Went with having a default so that the migration test files don't all need to mock out the encryption client.

FarazA22: Have you tried `_?: MigrationClients` as the type? If it works, eslint-disable-next-line should not be needed, nor a default value.

josephine-njooi: Thanks! Changed; it helps with the default value, but not with the no-unused-vars

api/src/db/migrations/v100_add_updated_timestamp.ts
```diff
@@ -15,7 +16,11 @@ export interface v100UserData {
   version: number;
 }
 
-export const migrate_v99_to_v100 = (v99Data: v99UserData): v100UserData => {
+export const migrate_v99_to_v100 = (
+  v99Data: v99UserData,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
SirSamuelJoseph, line 21: [dust] I still don't love this pattern of importing and ignoring this value in every migration but the only other thing I can thing of is using `Function.length` to determine how many parameters the migration function has and only supplying the clients if it's more than one which I also really don't like. So I'll defer to you on which you prefer

SirSamuelJoseph (approved):
Prefactor makes sense, good problem solving!

api/src/db/migrations/v100_add_updated_timestamp.ts
josephine-njooi: Agreed, I don't like function signature evaluation even more 😕. If anyone else has ideas we can switch!

Thanks!

PR #10292, opened 2025-05-09
feat: [AB#14239] Personalize My Tasks button

ZacharyRizer:
## Description

* added a `GradientButton` component
* added the `PersonalizeMyTasksButton` component which uses the new `GradientButton`
* added the `PersonalizeMyTasksButton` to the `DashboardOnDesktop` and `DashboardOnMobile`
* added rendering logic for when to show the button
* added query param option for `profile.tsx` to optionally load a specific tab when routed to
* added an analytic event for this button press
* updated tests

### Ticket

This pull request resolves [#14239](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14239).

### Steps to Test

* run the app locally
* the button should appear for the following conditions:
1. `STARTING` business after the Business Structure has been selected
2. `NEXUS` business after the Business Structure has been selected (pick foreign business and chose option 1)
3. All other cases will display the button
* For `OWNING`, `STARTING` with structure selected, and `NEXUS` with structure selected, the button will link to the `Permits & Licenses` tab
* For other cases, it will default to the `info` tab

### Notes

A refactor was made the `Dashboard` component and `dashboard` page. Originally this was split into separate components for `DashboardOnMobile` and `DashboardOnDesktop` along with separate test files for each. The two component files and test files were almost 100% repeated code. They have now been combined into a single `Dashboard.tsx` and `Dashboard.test.tsx` files. 

## Code author checklist


web/src/components/PersonalizeMyTasksButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+    analytics.event.roadmap_personlize_my_tasks_button.click.go_to_profile_tasks_screen();
+    router && router.push(`${ROUTES.profile}?tab=${tabValue}`);
+  };
+
+  return (
+    <GradientButton
+      text={Config.dashboardHeaderDefaults.personalizeMyTasksButtonText}
+      onClick={handleClick}
+      icon="star"
+      ariaLabel={`Link To ${Config.dashboardHeaderDefaults.personalizeMyTasksButtonText}`}
```
adidner, line 28: pebble - so instead we should create a prop (something like, linksToANewPage) then if thats true internally in the child, we should add a role="link" to the button tag. 

That way we get the "Link To" concept for free and it will show up in the various other accessibility important things (like if a screen reader users searches by link etc)

web/src/components/njwds-extended/GradientButton.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    const button = screen.getByRole("button");
+    expect(button).toHaveAttribute("aria-label", "Test aria label");
+  });
+
+  test("renders Icon component with correct props", () => {
+    render(<GradientButton text="Click me" />);
+
+    const icon = screen.getByTestId("mocked-icon");
+    expect(icon).toBeInTheDocument();
```
adidner, line 60: dust - I'm not exactly sure what this test is doing? / Meant to do? 

The name says renders Icon components with correct props, which would make me think that it was suppose to have Icon prop passed then verified that the icon existed but the test itself doesn't pass an icon then verified an icon exists for some reason? I'm a bit confused

adidner: Should be more specific as there are multiple icons, looking deeper at the code. I assume this refers to the chevron for navigate_next on the right side existing.

adidner: Also the Icon component there is one of ours and can Take a test-id if I'm not mistaken which means you wouldn't need to mock it

adidner (approved):
For easier review next time, might be good to have this as 2 commits? 1 where the refactor is done and then the other where the ticket work is done? Would make reviewing as separate commits a bit easier. 

2 minor comments to be addressed. 

Nice work, love to see some good refactoring done, love to see new components and scss written and good general logic considered. Also love seeing nice PR descriptions / testing instructions. 

Approved.

web/src/lib/utils/analytics.ts
```diff
@@ -792,6 +794,19 @@ export default {
...
+      click: {
+        go_to_profile_tasks_screen: () => {
+          eventRunner.track({
+            event: "navigation_clicks",
+            legacy_event_action: "click",
+            legacy_event_category: "roadmap_profile_edit_button",
+            legacy_event_label: "go_to_profile_screen",
+            clicked: "go_to_profile_tasks_screen",
+          });
+        },
```
cbrunefearless, line 807: DUST - As long as this is the same pattern as other analytics I am ok with it.

cbrunefearless (approved):
lgtm

web/src/components/PersonalizeMyTasksButton.tsx
ZacharyRizer: I think in the case of this component, it will always be linking to a new page, so maybe instead of a prop we just include the `role="link"` by default?

web/src/components/njwds-extended/GradientButton.test.tsx
ZacharyRizer: Looking at this test again, I'm just going to remove it. It isn't actually testing anything. That `IconComponent` will always render. I got confused with the other icon that was tested above.

web/src/components/PersonalizeMyTasksButton.tsx
adidner: That sounds good. 

I'm not sure how the gradient button is intended to be used but if thats the general use case at the moment a future dev can add said change if it needs to be toggleable in the future.

ZacharyRizer: Refactored the `GradientButton` to have an optional role parameter, and added a test for that. Should resolve this comment.

PR #10303, opened 2025-05-12
WIP: feat: [AB:13770] Add guest mode tax clearance access

jennyverdeyen:
## Description

Added guest user authentication flow for the Tax Clearance Certificate feature. This includes showing the NeedsAccountModal for unauthenticated (guest) users. Also added unit tests to verify this behavior.

The ticket notes that redirecting the user back to the dashboard is fine for now, rather than back to the tax clearance page. In working this ticket I learned that the returnToLink functionality in our application doesn't actually work. Instead of re-working it here which would increase the scope significantly, we can update that functionality in a follow-on ticket.

### Ticket

This pull request resolves [#13770](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/13770).

### Approach

No dependencies were added or modified

### Steps to Test

Check out this branch.

As a guest mode user:

- Navigate to the Tax Clearance Certificate task (/actions/tax-clearance-certificate-apply)
- Click Continue at the requirements step
- Verify the "Link with myNJ to save progress" appears
- Using the dev tools in your browser, under Application, verify the returnToLink is set in sessionStorage
- Click through the login flow
- Verify you are redirected back to the Tax Clearance Certificate page
- Using the dev tools in your browser, under Application, verify the returnToLink is cleared from sessionStorage

As an authenticated user:

- Navigate to the Tax Clearance Certificate task (/actions/tax-clearance-certificate-apply)
- Click Continue at the requirements step
- Verify you advance to the next step
- Verify no modal appears
- Verify no returnToLink is saved in sessionStorage

### Notes

~The handling of the "returnToLink" encounters an edge case here where the guest user signs in with an existing account, and their existing business preferences (where the returnToLink would normally be stored) are replaced by their user data. So I decided to use a separate variable, also kept in session storage in order to handle the redirect. The path stored in session is validated and should be cleared after the loading step is completed.~

After further investigation... I believe the returnToLink actually gets lost for both existing an new account log-ins. It just appears to work in other instances within the application because the default redirect is to /dashboard. We can address this in a future ticket, and for now this PR only adds the modal, redirecting the user to the dashboard as a default.

## Code author checklist


web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
```diff
@@ -26,16 +44,15 @@ export const Requirements = (props: Props): ReactElement => {
...
-            }}
+            onClick={handleContinue}
             dataTestId="cta-primary-1"
             isRightMarginRemoved={true}
           >
             {Config.taxClearanceCertificateStep1.continueButtonText}
           </PrimaryButton>
         </ActionBarLayout>
       </CtaContainer>
+      <NeedsAccountModal />
```
FarazA22, line 55: pebble - this component renders on _app.tsx, rendering it here is not necessary

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
```diff
@@ -1,17 +1,35 @@
...
+import { ReactElement, useContext } from "react";
 
 interface Props {
   setStepIndex: (step: number) => void;
 }
+
 export const Requirements = (props: Props): ReactElement => {
   const { Config } = useConfig();
+  const { state: authState } = useContext(AuthContext);
+  const { setShowNeedsAccountModal } = useContext(NeedsAccountContext);
```
FarazA22, line 22: pebble - NeedsAccountContext can return isAuthenticated, AuthContext is not needed

FarazA22: BusinessFormationPaginator.tsx is a great example to review

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
```diff
@@ -1,17 +1,35 @@
...
+
+  const handleContinue = async (): Promise<void> => {
```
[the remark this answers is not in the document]
FarazA22: This approach works, what do you think about the method I'm proposing?

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
```diff
@@ -1,17 +1,35 @@
...
+    if (authState.isAuthenticated === IsAuthenticated.FALSE) {
+      const returnToLink = ROUTES.taxClearanceCertificate;
+      sessionStorage.setItem("returnToLink", returnToLink);
+      setShowNeedsAccountModal(true);
+    } else {
+      props.setStepIndex(1);
```
FarazA22, line 30: The ticket mentions modal opens when stepper is clicked, it's not accounted for in this PR

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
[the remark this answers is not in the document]
FarazA22: It if doesn't work, is it possible to store the returnLink to an existing field? business.preferences.returnToLink

FarazA22 (changes requested):
Your solution works, but I’m wondering if we could move the redirect logic directly into the NeedsAccount modal or its context — basically manage it there. Could you take another look with that in mind?

I think sessions could work, but since we already store return links in userData, it might make more sense to follow that existing pattern instead of introducing a new one.

web/src/pages/loading.tsx
```diff
@@ -29,6 +29,19 @@ const LoadingPage = (): ReactElement => {
...
+        sessionStorage.removeItem("returnToLink");
+        if (Object.values(ROUTES).includes(returnToLink)) {
+          router.push(returnToLink);
+        } else {
+          router.push(ROUTES.dashboard);
+        }
+        return;
+      }
+    }
+
```
FarazA22, line 44: take a look at the useMountEffectWhenDefined code below, there is an else statement that routes the user to the dashboard, I don't think we need it here.

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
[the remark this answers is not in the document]
jennyverdeyen: Thanks for the suggestion! I addressed the `business.preferences.returnToLink` option in the description of the PR - this data gets cleared when a guest user logs in with an existing account, as it's part of userData. I can look into the other suggestion involving NeedsAccountContext.

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+      // The modal is automatically closed on mount for authenticated users
+      expect(mockSetShowNeedsAccountModal).toHaveBeenCalledWith(false);
+    });
+
+    it("advances to the next step when Continue is clicked", () => {
+      renderRequirements(IsAuthenticated.TRUE);
+      fireEvent.click(screen.getByText(Config.taxClearanceCertificateStep1.continueButtonText));
+      expect(mockSetStepIndex).toHaveBeenCalledWith(1);
+    });
+  });
```
jphechter, line 57: I appreciate you breaking out these cases into individual tests. Looks good!

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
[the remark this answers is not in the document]
jphechter: I'm not sure this constitutes a `boulder`. I definitely see the value in centralizing the logic, but I think the approach Jenny's taking here does actually account for more edge cases here, (as she's calling out) namely handling for a guest user logging into an existing account.

I could go either way on this.

jphechter (commented):
I think Faraz called out a few good points for improvement, but overall I'm happy with this.

web/src/components/tasks/anytime-action/tax-clearance-certificate/steps/Requirements.tsx
jennyverdeyen: Thanks for this pointer!!

jennyverdeyen:
I'm going to re-work this and simplify it a lot. I realized that my workaround solution to save the returnToLink separately in the tax clearance flow introduced a bug, which illuminated for me that to fix it I'm almost re-engineering the way our app handles redirect urls. I don't want to do that since we have an established pattern of using business preferences to save the returnToLink everywhere else. Our existing behavior already doesn't handle the edge case (user logs in with an existing account and doesn't get redirected to where they were), so I'm going to let it be consistent with that and revert to using that pattern. There is probably room for improvement in the way we do this in the future, that considers that edge case.

FarazA22 (approved):
Nice rework

PR #10404, opened 2025-05-26
Add PII hashing logic

jphechter:
TODO:
- Update `.env` values in Bitwarden and CircleCI. Because some of the variables are being modified, we are waiting on review to introduce those changes. 

## Description

This work adds a function to salt and hash a sensitive value to the existing encryption/decryption factory (now `AwsCryptoFactory`). Additionally, we have included a call to this function in `encryptFieldsFactory.ts` which will hash a user's tax ID if it is included in their `profileData`.

### Ticket

This pull request resolves [#14311](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14311).

### Approach

While this PR touches 37 files, the bulk of the logical changes can be found in `AwsCryptoFactory.ts`, `encryptFieldsFactory.ts`, `userRouter.ts`, and the associated unit test files. The majority of the remaining changes are limited to renaming variables, the inclusion of additional environment variables, and updating imports.

`AwsCryptoFactory.ts` – We've added a new function `hashValue()` which will salt and then hash a sensitive string. This is largely based off of the [exploratory work](https://github.com/newjersey/business-crypto-exploration/) conducted by @cbrunefearless and @seidior.

The function takes in a string value, sanitizes the value removing all non-alphanumeric characters, salts the value using an encrypted salt (which can be decrypted with a new KMS key specific to tax id hashing), and then hashes the salted value. We're using the Node.js crypto library to salt and hash.

`encryptFieldsFactory.ts` – This factory is used to encrypt a user's tax ID. We've added another step to hash the tax ID if it is present in `profileData`. 

`userRouter.ts` – We do not want to expose the hashed tax ID anywhere. While we have taken great effort to make this a secure value, it is in our best interest to prevent this value from being exposed anywhere to the user. In this file, we've added a function to remove the `hashedTaxId` before resolving requests to `GET` user data. The value will persist in the database, but it should never be available on the frontend.

### Steps to Test

- Create a new business
- Navigate to the business profile
- under the reference numbers section, add a numeric value to the tax ID field
- save
- open the database, find the relevant user, and check that a hashed tax ID exists within the profile data

Not necessary for testing, but the same tax ID should always return the same hashed value. You can validate this by creating an additional business, adding the same tax ID, and checking the returned hashed value in the database is the same.

Additionally, check the network requests and confirm that a `hashedTaxId` is not present in the user data returned from the API.

### Notes

This work will be built upon in order to deliver the complete tax clearance certificate work. Specifically, this will allow us to assert that each business requesting a tax ID is using a unique tax ID.

## Code author checklist


scripts/clear-ports.sh
```diff
@@ -4,5 +4,6 @@ npx kill-port 5002
 npx kill-port 5050
 npx kill-port 6006
 npx kill-port 8000
+npx kill-port 8001
```
jphechter, line 7: DynamoDBLocal and Jest both use `8001`

api/src/domain/user/encryptFieldsFactory.ts
```diff
@@ -1,21 +1,24 @@
...
 import { maskingCharacter } from "@shared/profileData";
 import { UserData } from "@shared/userData";
 
 export const encryptFieldsFactory = (
-  encryptionDecryptionClient: EncryptionDecryptionClient,
+  encryptionDecryptionClient: CryptoClient,
+  hashingClient: CryptoClient,
 ): ((userData: UserData) => Promise<UserData>) => {
   return async (userData: UserData): Promise<UserData> => {
+    const userDataWithHashedTaxId = await hashProfileTaxId(userData, hashingClient);
```
jphechter, line 13: Added this line ahead of the other `encryptionFunctions` for 2 reasons:
1. The hashing needs to occur before encryption because after encryption, we're left with a masked value such as `********1234`. We do not want to hash the masked value.
2. We're using 2 different `CryptoClient`s. We could have restructured this logic to tie the function to the associated CryptoClient, but that would have made this a bit harder to read.

api/src/api/userRouter.ts
```diff
@@ -180,8 +181,10 @@ export const userRouterFactory = (
           .then((userData) => updateRoadmapSidebarCards(userData))
           .then((userData) => asyncUpdateLicenseStatus(userData));
 
-        await dynamoDataClient.put(updatedUserData);
-        res.json(updatedUserData);
+        await databaseClient.put(updatedUserData);
+
+        const santizedUserData = removeSensitiveData(updatedUserData);
+        res.json(santizedUserData);
```
jphechter, line 187: Remove the hashed tax ID before resolving requests.

api/src/domain/types.ts
```diff
@@ -114,9 +114,10 @@ export interface TaxFilingInterface {
   }) => Promise<UserData>;
 }
 
-export interface EncryptionDecryptionClient {
+export interface CryptoClient {
```
SirSamuelJoseph, line 117: [dust] I feel like encryption/decryption client is a more descriptive name of what this interface does. I'd be fine with `EncryptionClient` as well, I think crypto as a keyword is a bit confusing here

.circleci/config.yml
```diff
@@ -209,9 +209,12 @@ commands:
             echo 'export AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID_DEV' >> $BASH_ENV
             echo 'export AWS_ACCOUNT_ID=$AWS_ACCOUNT_ID_DEV' >> $BASH_ENV
             echo 'export AWS_CRYPTO_CONTEXT_ORIGIN=$AWS_CRYPTO_CONTEXT_ORIGIN' >> $BASH_ENV
-            echo 'export AWS_CRYPTO_CONTEXT_PURPOSE=$AWS_CRYPTO_CONTEXT_PURPOSE' >> $BASH_ENV
             echo 'export AWS_CRYPTO_CONTEXT_STAGE=$AWS_CRYPTO_CONTEXT_STAGE_DEV' >> $BASH_ENV
-            echo 'export AWS_CRYPTO_KEY=$AWS_CRYPTO_KEY_DEV' >> $BASH_ENV
+            echo 'export AWS_CRYPTO_TAX_ID_ENCRYPTION_KEY=$AWS_CRYPTO_TAX_ID_ENCRYPTION_KEY_DEV' >> $BASH_ENV
+            echo 'export AWS_CRYPTO_CONTEXT_TAX_ID_ENCRYPTION_PURPOSE=$AWS_CRYPTO_CONTEXT_TAX_ID_ENCRYPTION_PURPOSE' >> $BASH_ENV
```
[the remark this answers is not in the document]
jphechter: The purpose is the same in every cloud environment – at least it has been up to this point.

SirSamuelJoseph (approved):
Makes sense to me aside from a quibble and a question around variable naming

cbrunefearless (approved):
lgtm, thanks for the comments and the PR description!!

PR #10406, opened 2025-05-27
8142 cms update allow=content team members to add anytime action cateogires

adidner:
## Description

Added the ability to add anytime action categories via the CMS

### Ticket

This pull request resolves [#8142](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/8142).

### Approach

To get this to work a refactor was necessary to get everything to pull data in the right ways (fs fileread vs try catch import things)

This is a seperate commit for the purposes of easy reviewal. 

### Steps to Test

Run the app

Go through to the up and running phase (via the cards on the side) 

See that anytime actions currently look as one would think

Go into the local CMS (via starting the app and running the proxy)

Change the name or wording of an anytime action category that you can see in app 

Refresh the page for the app dashboard, see that your change propagated. 

Add a new anytime action category

For a task that you can see in the dropdown, find it on the CMS side under anytime action tasks, change it's category to your new category

Refresh dashboard, see that new category shows up with corresponding task under it. 

## Code author checklist


web/src/lib/async-content-fetchers/fetchAnytimeActionByFilename.ts
```diff
@@ -1,26 +0,0 @@
...
-
-export const fetchAnytimeActionByFilename = async (id: string): Promise<AnytimeActionTask> => {
-  try {
-    const fileContent = await import(
-      `@businessnjgovnavigator/content/anytime-action-tasks/${id}.md`
-    );
-    return convertAnytimeActionTaskMd(fileContent.default, id);
-  } catch {
-    return {
-      callToActionLink: undefined,
```
adidner, line 12: dust - this was causing an import / build issue / testing issue. So I refactored the Government Contracting section to remove the need for this kind of call.

web/src/lib/static/anytimeActionCategoryIntegrity.test.ts
```diff
@@ -0,0 +1,20 @@
...
+    for (const slug of urlSlugs) {
+      const AnytimeAction = loadAnytimeActionTaskByUrlSlug(
+        slug.params.anytimeActionTaskUrlSlug,
+        true,
+      );
+      expect(AnytimeAction.category[0].categoryId).not.toEqual(
+        AnytimeAction.category[0].categoryName,
+      );
+    }
+    expect(true).toBe(true);
```
adidner, line 18: dust - this is my integrity test for anytime actions which makes sure that all anytime actions loaded have a valid category id that corresponds to an anytime action category element.

web/src/lib/static/loadAnytimeActionTasks.ts
```diff
@@ -1,50 +1,77 @@
...
 import path from "path";
 
 type PathParams<P> = { params: P; locale?: string };
-const anytimeActionsTaskDir = path.join(
+const anytimeActionsTaskDirApp = path.join(
   process.cwd(),
   "..",
   "content",
   "src",
   "anytime-action-tasks",
```
adidner, line 13: dust - so as far as I can tell, when we do a build the code base is compiled from the web and API directories as root. Thus to get to the file directory properly we have the `..` in the file path here to get to our content src files. 

The issue with that is that our tests are run from the root directory so if we run these functions to actually use our real data in our tests the extra `..` causes problems. 

I think testing with real data / testing the integrity of our src content data is important and more ideal especially when that content is sensitive or determines structure. 

The best way I could think of to allow us to get the best of both world was with an optional defaulted param to use the testing path. 

Open to other ideas though if we have better ones. (I considered something with constant variables that we could mock out but was struggling to get that working as desired given it would require some moving around of files / imports etc, but I am no expert in that approach.

web/decap-config/collections/10-anytime-action.yml
```diff
@@ -5,6 +5,24 @@ collections:
     fields:
       - { name: "" }
 
+  - name: "anytime-action-categories"
```
FarazA22, line 8: pebble - we should disable the ability to delete the category. if they delete one, we may run into a bug where it doesn't display.

web/src/lib/static/loadAnytimeActionLicenseReinstatements.test.ts
```diff
@@ -63,6 +80,12 @@ describe("loadAnytimeActionLicenseReinstatements", () => {
             contentMd: "Some content description 1",
             callToActionLink: "CallToActionLink1",
             callToActionText: "CallToActionText1",
+            category: [
```
FarazA22, line 83: dust - should categories be a list? is there a future goal to have multiple categories? If not, maybe update this to be an object

web/src/components/dashboard/AnytimeActionDropdown.tsx
```diff
@@ -20,7 +24,10 @@ interface Props {
 }
 
 type AnytimeAction = AnytimeActionTask | AnytimeActionLicenseReinstatement;
-type AnytimeActionWithTypeAndCategory = AnytimeAction & { type: string; category: string[] };
+type AnytimeActionWithTypeAndCategory = AnytimeAction & {
+  type: string;
```
FarazA22, line 28: dust - it's not clear why this type is needed? category is added to AnytimeActionTask and AnytimeActionLicenseReinstatement... so the only piece that is being added it type, what is that and where is it coming from?

web/src/lib/utils/tasksMarkdownReader.ts
```diff
@@ -0,0 +1,142 @@
+import {
```
FarazA22, line 1: dust - consider renaming the file, anytime actions are not tasks

web/src/pages/actions/government-contracting.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { GetStaticPropsResult } from "next";
+import { NextSeo } from "next-seo";
+import { ReactElement } from "react";
+
+interface Props {
+  governmentContractingStepAnytimeActions: AnytimeActionTask[];
+  governmentContractingTask: AnytimeActionTask;
+}
+
+const AnytimeActionTaskPage = (props: Props): ReactElement => {
```
FarazA22, line 17: dust - moving this to a separate page is clever, do you think it maes sense given the refactor work we need to do? move all the context to config in the fefactor ticket?

web/test/factories.ts
```diff
@@ -332,6 +332,12 @@ export const generateAnytimeActionLicenseReinstatement = (
     summaryDescriptionMd: `some-summary-description-md-${randomInt()}`,
     licenseName: randomElementFromArray(Object.values(taskIdLicenseNameMapping)),
     synonyms: [`some-synonym-${randomInt()}`],
+    category: [
+      {
+        categoryName: `Reactivate My Expired Permit, License or Registration`,
```
FarazA22, line 337: pebble, the values should be randomized?

web/src/lib/utils/tasksMarkdownReader.ts
FarazA22: does it hurt to keep this functionality within markdownReader.ts?

web/src/pages/actions/government-contracting.tsx
adidner: I still think it makes sense. 

If nothing else because it removes a try catch import block from our code and simplified the frontend code even if the content / backend will be simplified later.

web/src/lib/utils/tasksMarkdownReader.ts
adidner: Yes, it does. The issue is that part of markdownReader.ts are used in the code base outside of getStaticProps. 

When those functions are called in the main codebase their corresponding parts cannot try to import fs. 

Some of these import fs and are only used in getStaticProps so they need to be separated.

web/src/lib/static/loadAnytimeActionTasks.ts
```diff
@@ -1,50 +1,77 @@
...
 );
+const anytimeActionsTaskDirTest = path.join(
```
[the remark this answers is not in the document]
adidner: So I talked about this in the Approach section of my PR but perhaps it was unclear. 

So this is used right now in the anytimeActionCategoryIntegrity.test file. That file is making sure that no anytime action references a category that no longer exists. 

The new path is needed because the when the app builds it builds from inside of web as it's context so the .. is needed to get to the right place. However the test themselves run from the root directly so the .. is explicitly not wanted. 

But with this parameter our tests can actually grab our real underlying content rather than mocking it out which for the purposes of verifying that our data meets certain constrains is important. 

Also in some cases it would be better to test with real data so its nice to have the option to grab the real .md files if we want.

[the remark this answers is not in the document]
adidner: So you can't really mock the variable from the test and have things work correctly from what I understand as the variable is used interally to the file itself. 

You could pass the path as a variable into the function itself but you end up having to pass it down many layers and at one layer a different function needs to grab for different files without the .. but the path itself is different for that grab. So you would actually need to pass 2 different paths down multiple layers and it started to look real messy and unreadable in my opinion so this seemed a simpler path.

web/src/components/tasks/government-contracting/GovernmentContractingSteps.ts
```diff
@@ -17,10 +14,6 @@ export const GovernmentContractingSteps: {
   { name: "Prevailing Wages", fileName: "prevailing-wage" },
 ];
```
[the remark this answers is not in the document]
adidner: That seems like a good refactor to me. 

I added this additional context for removing the only-show-in-subtask category and related code as a comment on that ticket.

web/src/lib/static/loadAnytimeActionLicenseReinstatements.test.ts
adidner: There is intent there but it's waiting on design work

web/decap-config/collections/10-anytime-action.yml
```diff
@@ -29,26 +47,14 @@ collections:
       - {
```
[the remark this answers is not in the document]
adidner: so license reinstatements are formatted differently and passed down at a prop level to the dropdown component. 

In the dropdown component it handles assigning all of those the license reinstatement category. (would be felt weird to me to have a list of potential categories for reinstatments when we know by their structure and folder that they have to always be reinstatement category. Seems like its just a setup for a content person to have a weird bug.)

I have no idea how the Tax Clearance Anytime Action stuff is handled in the dropdown. Looking a bit at it it seems like the tax-clearance-certificate.md file is what determines the category for that specific situation.

web/decap-config/collections/10-anytime-action.yml
adidner: Fair, I was hoping my integrity test would prevent the need for this but it's still probably wise.

web/src/lib/static/loadAnytimeActionTasks.ts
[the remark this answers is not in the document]
FarazA22: As I mentioned in another comment, if we prevent Content from deleting categories—and we manage this ourselves—then the additional logic and tests become unnecessary. We already prevent Content from deleting files in certain collections to avoid breaking parts of the app.

If Content deletes a category (which they’re currently allowed to do), we end up having to investigate test failures and potentially spend time debugging the issue. This is more time-consuming than handling category removal through a dedicated ticket, where we can perform a quick spot check to ensure nothing breaks.

web/src/components/dashboard/AnytimeActionDropdown.tsx
FarazA22: Using the ID to determine the type might add unnecessary complexity.

I noticed the type is already defined in getApplicableAnytimeActions and anytimeActionLicenseReinstatementsWithType within the dropdown file. Maybe we can move this into the CMS config as a background field that gets auto-generated. That way, we wouldn’t need to manage a separate type manually.

We’d still need to add this field to each existing markdown file, but that could be handled as a one-time refactor ticket.

adidner: I think thats a good idea and I'll make a refactor ticket for that.

PR #10437, opened 2025-06-05
chore: [AB#14662] implement kill switch mechanism for migration logic

aadedejifearless:
## Description
This PR introduces a kill switch mechanism to the Migration Lambda to allow us to halt record processing in the event of unexpected cost overruns or operational risks.

As part of our migration effort, we are switching from provisioned to on-demand capacity for our DynamoDB tables, which gives us flexibility but also increases the risk of unbounded costs if something goes wrong (e.g., large scans, retries, or runaway processes).
Additionally, large-scale record migrations introduce the potential for:
- Unexpected cost spikes from unthrottled read/write activity.
- Lambda throttling or retries, especially under load.
- Infinite or runaway processing loops, particularly if logic errors lead to repeated record processing
To mitigate these risks, we are implementing a kill switch that allows us to quickly stop the migration process without a redeploy. 
**Terraform Additions**
As part of this effort, we have also added a CloudWatch alarm that monitors the SNS topic used to trigger the kill switch. This alarm will notify us when a message is published to the topic (i.e., when the kill switch is invoked), ensuring visibility into when and why migration processing is halted.

### Ticket

This pull request resolves [#14662](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14662).

### Approach

- A stage-specific kill switch flag, e.g `/testing/feature-flag/users-migration/kill-switch` is stored as a boolean value in AWS SSM Parameter Store.
- The Existing Migration Lambda: `migrateUsersVersion`  checks the value of this parameter at runtime before processing records.
- If the flag is set to true, the Lambda logs a warning and exits without processing any records.
- A separate Lambda function: `updateKillSwitchParameter` is introduced to toggle this flag (e.g., set it to true to stop processing).
- This update Lambda is triggered via an SNS topic, allowing automated systems or alerts to activate the kill switch programmatically.
- To further limit risk, a MAX_SAFE_MIGRATION_COUNT constant caps the number of records processed in a single run (e.g., 5,000).
This ensures that even if the kill switch fails or is slow to activate, we avoid runaway migrations and unbounded costs in a single Lambda invocation

This allows us to remotely and dynamically stop migrations without requiring a redeploy or manual code changes

## Code author checklist

### DIFF
--- a/api/src/libs/ssmUtils.ts
+++ b/api/src/libs/ssmUtils.ts
@@ -1,0 +1,53 @@
+import { GetParameterCommand, PutParameterCommand, SSMClient } from "@aws-sdk/client-ssm";
+
+const ssmClient = new SSMClient({});
+const parameterName = `/${process.env.STAGE}/feature-flag/users-migration/kill-switch`;
+
+let cache: {
+  value?: boolean;
+  expiresAt?: number;
+} = {};
+
+export const isKillSwitchOn = async (): Promise<boolean> => {
+  const now = Date.now();
+
+  if (cache.value !== undefined && cache.expiresAt && cache.expiresAt > now) {
+    return cache.value;
+  }
+  try {
+    const command = new GetParameterCommand({
+      Name: parameterName,
+      WithDecryption: true,
+    });
+
+    const response = await ssmClient.send(command);
+    const paramValue = response.Parameter?.Value ?? "false";
+    const boolValue = paramValue.toLowerCase() === "true";
+
+    cache = {
+      value: boolValue,
+      expiresAt: now + 30 * 1000,
+    };
+
+    return boolValue;
+  } catch (error) {
+    console.error("Failed to read kill switch parameter:", error);
+    return false;
+  }
+};
+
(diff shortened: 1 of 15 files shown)


somabadri (approved):
Great work Ade! This is going to be a great guard rail against any database queries that go wrong. I just had one question regarding what the process to toggle it back would be since the lambda fires off the update function to only turn the switch on. I will bring this up in Slack so others have visibility as well.

api/src/libs/ssmUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    return boolValue;
+  } catch (error) {
+    console.error("Failed to read kill switch parameter:", error);
+    return false;
+  }
+};
+
+export const updateKillSwitch = async (value: boolean): Promise<void> => {
```
somabadri, line 39: [pebble] this function is only being called with the value `true`, so it's only purpose is to turn on the kill switch, thoughts on changing it to represent that if that's the only way you anticipate using this function, so no parameter to pass in and the function renamed to something like `turnOnKillSwitch`

PR #10441, opened 2025-06-06
feat: [AB#14466] clear submission error after user updates tax clearance form

ashlsun:
## Description

After a user submits the tax clearance form on the Review step and an error alert box appears, that error alert should be cleared once the user goes back to Check Eligibility and makes a change to any of the form fields.

### Ticket

This pull request resolves [#14466](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14466).

### Approach
The simplest approach was using a useEffect to clear the error once any element of the tax clearance data changes. 
Noting that I was wary of using useEffect after reading [You Might Not Need An Effect](https://react.dev/learn/you-might-not-need-an-effect), and I think perhaps there is a way to do this without effects, but with the way state is organized across the tabs & parent components, clearing `responseErrorType` through individual event handlers for each of the form fields would be a lot messier, while using useEffect made for the smallest change.

Also added tests for each form field to make sure that it would trigger the error clearing. I tried to group all the text fields and made separate ones for the dropdowns, since making drop selections would require different function calls.

### Steps to Test

Create an up and running business, submit a completed tax clearance form with an invalid tax ID and PIN combination (anything that does not match what's in `api/wiremock/mappings/get-tax-clearance-success.json`) and see a submission error appear on the form. Go back to the Check Eligibility step and update any of the fields (make a dropdown selection, type in the text fields) and see that the submission error goes away.

Screen recording of clearing submission error on change
## Code author checklist

### DIFF
--- a/web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.test.tsx
+++ b/web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.test.tsx
@@ -25,6 +25,7 @@ import {
   getTaxClearanceCertificateAgencies,
   LookupTaxClearanceCertificateAgenciesById,
   randomElementFromArray,
+  StateObject,
   TaxClearanceCertificateData,
   TaxClearanceCertificateResponse,
   TaxClearanceCertificateResponseErrorType,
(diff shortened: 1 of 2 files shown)


web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.test.tsx
```diff
@@ -1534,15 +1535,119 @@ describe("<AnyTimeActionTaxClearanceCertificate />", () => {
       };
       mockApi.postTaxClearanceCertificate.mockResolvedValue(response);
       renderComponent({});
-      fireEvent.click(screen.getByTestId("stepper-2"));
-      fireEvent.click(screen.getByTestId("next-button"));
+      const thirdTab = screen.getAllByRole("tab")[2];
```
SirSamuelJoseph, line 1538: [dust] It should be possible to grab the tab by some attribute, which would make this test more resilient if future steps are added. You're already doing this for button though so it's possible you tried and couldn't in which case I'm fine with it this way

web/src/components/tasks/anytime-action/tax-clearance-certificate/AnytimeActionTaxClearanceCertificate.test.tsx
```diff
@@ -1534,15 +1535,119 @@ describe("<AnyTimeActionTaxClearanceCertificate />", () => {
...
+        screen.getByRole("button", { name: Config.taxClearanceCertificateShared.saveButtonText }),
+      );
 
       await waitFor(() => {
         expect(screen.getByTestId("tax-clearance-error-alert")).toBeInTheDocument();
       });
     },
   );
 
+  it.each([
```
SirSamuelJoseph, line 1550: [dust] I'm not sure we need this as an it.each - I think it would be fine to select one randomly each test. Especially with the logging of seeds we have now

SirSamuelJoseph (approved):
Great work!

PR #10448, opened 2025-06-09
fix: [AB#14779] fix flaking test by hardcoding current time

ashlsun:
## Description
Discovered that this test in `api/src/client/ApiTaxClearanceCertificateClient.test.ts` would occasionally fail due to a 0.001 second difference between a timestamp when the function was called and the timestamp when a result was mocked.

See this example: https://app.circleci.com/pipelines/github/newjersey/navigator.business.nj.gov/38111/workflows/7c1fb9b9-263a-4532-8784-07a11a18c19e/jobs/184521/parallel-runs/6/steps/6-109

Running it locally, I saw that it would fail sometimes but pass most of the time. By hardcoding the current time with jest, we ensure that this test will never fail due to a timing issue.

### Ticket

This pull request resolves [#14779](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14779).

## Code author checklist

### DIFF
--- a/api/src/client/ApiTaxClearanceCertificateClient.test.ts
+++ b/api/src/client/ApiTaxClearanceCertificateClient.test.ts
@@ -210,6 +210,10 @@ describe("TaxClearanceCertificateClient", () => {
   });
 
   it("returns certificate pdf array and updated user data", async () => {
+    const NOW = "2000-01-01T10:00:00.000Z";
+    jest.useFakeTimers();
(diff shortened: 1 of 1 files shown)


ZacharyRizer (commented):
I think we should actually move this to a beforeEach at the top of the test file. I think we should be doing this for all tests. What do you think?

hinzed1127 (approved):

PR #10474, opened 2025-06-10
fix: [AB#14795] fix test flaking due to invalid NJ zipcode

ashlsun:
## Description
A unit test related to #10441 was found failing [on CircleCI occasionally](https://app.circleci.com/pipelines/github/newjersey/navigator.business.nj.gov/38215/workflows/f2707c90-4185-4576-a9f1-452870358b79/jobs/184916). The test expects that there won't be an error alert box on the screen after firing all the actions, but because `generateTaxClearanceCertificateData` sometimes chooses the state NJ along with a non-NJ zipcode (not beginning in 07 or 08), the error alert box remains on the screen.

### Ticket

This pull request resolves [#14795](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14795).

### Approach

I think another way of fixing this flaky test could be using a more specific query on the page — instead of looking for an element with `"tax-clearance-error-alert"` as the id, we could check that there's no element with the error text `"This business is not currently eligible for this certificate."`, as that is more precisely what we want to test. If we think that is preferable I can add try to add that in here as well.

This approach fixes taxClearanceCertificateData so that it will always be valid to the frontend, which might be better in general, if any future tests have to do with error handling.

## Code author checklist

### DIFF
--- a/shared/src/test/factories.ts
+++ b/shared/src/test/factories.ts
@@ -411,14 +411,15 @@ export const generateTaxClearanceCertificateData = (
   overrides: Partial<TaxClearanceCertificateData>,
 ): TaxClearanceCertificateData => {
   const taxId = randomInt(12).toString();
+  const addressState = generateUnitedStatesStateDropdownOption({});
   return {
     requestingAgencyId: randomElementFromArray(taxClearanceCertificateAgencies).id,
     businessName: `some-business-name-${randomInt()}`,
     addressLine1: `some-address-1-${randomInt()}`,
     addressLine2: `some-address-2-${randomInt()}`,
     addressCity: `some-city-${randomInt()}`,
-    addressState: generateUnitedStatesStateDropdownOption({}),
-    addressZipCode: randomInt(5).toString(),
+    addressState,
+    addressZipCode: addressState.shortCode === "NJ" ? generateNjZipCode() : randomInt(5).toString(),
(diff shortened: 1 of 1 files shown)


shared/src/test/factories.ts
```diff
@@ -411,14 +411,16 @@ export const generateTaxClearanceCertificateData = (
...
     requestingAgencyId: randomElementFromArray(taxClearanceCertificateAgencies).id,
     businessName: `some-business-name-${randomInt()}`,
     addressLine1: `some-address-1-${randomInt()}`,
     addressLine2: `some-address-2-${randomInt()}`,
     addressCity: `some-city-${randomInt()}`,
-    addressState: generateUnitedStatesStateDropdownOption({}),
-    addressZipCode: randomInt(5).toString(),
+    addressState,
+    addressZipCode:
+      addressState.shortCode === "NJ" ? `07${randomInt(3).toString()}` : randomInt(5).toString(),
```
seidior, line 423: [sand] Do we have a New Jersey ZIP Code™ generator function? If we don't, should we refactor this so that other tests can use it to reduce flakiness? That way we can have a rock-solid implementation and change things across the app if our ZIP Codes™ ever change.

Long term, I think that we should have a generic state and postal code generator function that only generates valid info, but I think that's way out of scope here.

PR #10548, opened 2025-06-13
feat: [AB#14385] Add location based information to profile accordion

SirSamuelJoseph:
## Description

Adds location based section to profile accordion

### Ticket

This pull request resolves [#14385](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14385).

### Approach

I pulled out some functions from profile into a profileDisplayHelpers file so we could reuse the functions. Also added new tests for these functions that didn't exist before.

### Steps to Test

1. Make sure you have the feature flag for personalize my tasks tab enabled
2. Go through onboarding and go to the profile
3. Click on the "Personalize My Tasks" tab
4. The location accordion should have questions underneath it- one for home based question and 2 more if the answer to home based question is no
5. Navigating to the "Permits and Licenses" tab should reflect the same answers as in the location accordion

## Code author checklist


web/src/components/profile/profileDisplayLogicHelpers.ts
```diff
@@ -0,0 +1,32 @@
+import { isHomeBasedBusinessApplicable } from "@/lib/domain-logic/isHomeBasedBusinessApplicable";
+import { Business, isNexusBusiness, ProfileData } from "@businessnjgovnavigator/shared";
+import { nexusLocationInNewJersey } from "@businessnjgovnavigator/shared/domain-logic/nexusLocationInNewJersey";
+
+export const displayHomedBaseBusinessQuestion = (
```
SirSamuelJoseph, line 5: Moved these from `profile.tsx' to enable reuse in `PersonalizeYourTasksTab.tsx`

web/src/components/profile/PersonalizeYourTasksTab.tsx
```diff
@@ -60,7 +74,52 @@ export const PersonalizeYourTasksTab = (): ReactElement => {
...
+              <ProfileField
+                fieldName="homeBasedBusiness"
+                hideLine
+                fullWidth
+                hideHeader
+                boldAltDescription
+                boldDescription
+              >
+                <HomeBasedBusiness />
+              </ProfileField>
```
adidner, line 89: boulder - this code is repeated in 2 places nearly verbatum here and in profile in a number of places. 

More specifically the actually underlying components here are just radio questions with the same field name 

Also the profile field with these props and a dynamic field name is also repeated everywhere. 

So what I decided to do in my ticket was componentize the questions at a minumum. 

image

Ideally I think we also have some kind of general consolidation of these components into a larger comonent like we do with my PR for nonEssentailQuestions

web/src/components/profile/profileDisplayLogicHelpers.ts
adidner: I think thats a good call, for reuse purposes

web/src/components/profile/profileDisplayLogicHelpers.ts
```diff
@@ -0,0 +1,32 @@
...
+  profileData: ProfileData,
+  business?: Business,
+): boolean => {
+  if (!business) return false;
+  if (!profileData.industryId) {
+    return true;
```
adidner, line 11: pebble - kinda messy to pass both profileData and business when instead you could just pass business and then reference business.profileData

web/test/pages/profile/profile-shared.test.tsx
```diff
@@ -825,5 +826,141 @@ describe("profile - shared", () => {
         }),
       ).toBeInTheDocument();
```
adidner, line 827: pebble - is there a reason this isn't just part of personalizeMyTasks.test instead?

adidner (changes requested):

web/src/components/profile/PersonalizeYourTasksTab.tsx
adidner: I'm quite worried about not doing this because if someone in the future doesn't know they also need to change instances in both places we will have inconsistencies which is a bad code smell. 

Whether that's in the props of profileField for visual difference or needing to add new questions in both places or even just changing the order or something. Would be easy to miss.

web/src/components/profile/profileDisplayLogicHelpers.ts
SirSamuelJoseph: So the reason I don't do that is that these questions are relying on a live version of profileData (maintained in the context) and not the one under the business.

SirSamuelJoseph: If we just use `business.profileData`, the user would need to Save or refresh the page to have the dependent questions appear based on your answer to the first question

adidner: dust - Ah understood, thats interesting and makes sense. I wonder if there is a better way to signify that from a naming perspective? Or maybe it would be intuitive if I was thinking a little harder. I'll defer to you there.

adidner (approved):
That looks good to me now!

PR #10551, opened 2025-06-13
feat: [AB#14471] update returnToLink when showing sign-in/register modal on tax clearance steps

jennyverdeyen:
## Description

This adds a returnToLink to the business preferences when a guest user tries to move forward with Tax Clearance steps and are prompted with a sign-in modal.

Note that the returnToLink functionality only works when the guest user *registers* an account, not when they sign in with an existing account, because it is tied to their business data which gets replaced if they log in with an existing user.

This PR also addresses a bug with the returnToLink logic in the signinHelper, where the redirect won't work if the user passes through the account setup page - which occurs in this flow. Now, the returnToLink should be correctly passed to the postSelfReg api call.

### Ticket

This pull request resolves [#14471](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14471).

## Code author checklist

### DIFF
--- a/web/src/lib/auth/signinHelper.test.ts
+++ b/web/src/lib/auth/signinHelper.test.ts
@@ -109,7 +109,7 @@ describe("SigninHelper", () => {
       expect(mockSetRegistrationStatus).toHaveBeenCalledWith("IN_PROGRESS");
     });
 
-    it("does not use a returnToLink if path is account-setup", async () => {
+    it("posts userData to api self-reg with current pathname included when returnToLink is empty", async () => {
       const business = generateBusiness({
         preferences: generatePreferences({ returnToLink: "" }),
       });
@@ -137,7 +137,7 @@ describe("SigninHelper", () => {
       });
     });
 
-    it("posts userData to api self-reg with current pathname included when returnToLink is empty", async () => {
+    it("does not use a returnToLink if path is account-setup", async () => {
       fakeRouter = { replace: mockPush, asPath: ROUTES.accountSetup };
       const business = generateBusiness({
         preferences: generatePreferences({ returnToLink: "" }),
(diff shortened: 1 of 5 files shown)


web/src/lib/auth/signinHelper.test.ts
```diff
@@ -109,7 +109,7 @@ describe("SigninHelper", () => {
       expect(mockSetRegistrationStatus).toHaveBeenCalledWith("IN_PROGRESS");
     });
 
-    it("does not use a returnToLink if path is account-setup", async () => {
+    it("posts userData to api self-reg with current pathname included when returnToLink is empty", async () => {
```
jennyverdeyen, line 112: These test names were swapped for what they actually respectively test, so I just switched them.

web/src/lib/auth/signinHelper.ts
```diff
@@ -52,9 +52,9 @@ export const onSelfRegister = ({
   }
   setRegistrationStatus("IN_PROGRESS");
 
-  const route = router.asPath?.includes(ROUTES.accountSetup)
-    ? ""
-    : getCurrentBusiness(userData).preferences.returnToLink || router.asPath || "";
+  const { returnToLink } = getCurrentBusiness(userData).preferences;
```
jennyverdeyen, line 55: These updated lines are the key to getting the returnToLink to actually work for a user who goes through self registration. It was completely resetting the returnToLink whenever a user went through the account setup page (which I think always happens, before self registration).

web/src/lib/auth/signinHelper.ts
```diff
@@ -52,9 +52,9 @@ export const onSelfRegister = ({
...
+  const fallback = router.asPath?.includes(ROUTES.accountSetup) ? "" : router.asPath || "";
```
jennyverdeyen, line 56: Shoutout to @ZacharyRizer for the elegant and readable ternary logic!

ZacharyRizer (commented):
LGTM

PR #10576, opened 2025-06-16
chore: [AB#0000] update serverless to include API_BASE_URL

aadedejifearless:
## Description
This PR adds the `API_BASE_URL` environment variable to the Lambda configuration. Previously, the variable was missing, causing the Lambda to default to the fallback (production) API endpoint. With this change, each environment will now correctly use its respective API_BASE_URL, ensuring requests are routed appropriately in lower environments.

### Ticket

## Code author checklist

### DIFF
--- a/api/src/libs/healthCheck.ts
+++ b/api/src/libs/healthCheck.ts
@@ -15,9 +15,7 @@ const healthCheckEndPoints: Record<string, string> = {
   taxClearance: "tax-clearance",
 };
 
-const url = process.env.API_BASE_URL ?? "https://api.account.business.nj.gov";
-
-const healthCheck = async (type: string, logger: LogWriterType): Promise<Status> => {
+const healthCheck = async (type: string, url: string, logger: LogWriterType): Promise<Status> => {
   return axios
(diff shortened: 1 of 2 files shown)


api/src/libs/healthCheck.ts
```diff
@@ -36,9 +34,14 @@ const healthCheck = async (type: string, logger: LogWriterType): Promise<Status>
 };
 
 export const runHealthChecks = async (logger: LogWriterType): Promise<StatusResult> => {
+  const url = process.env.API_BASE_URL;
+  if (!url) {
+    logger.LogError("Missing required environment variable: API_BASE_URL");
+    throw new Error("Missing required environment variable: API_BASE_URL");
+  }
```
jphechter, line 41: [dust] Not a huge deal, but we don't tend to mention the environment variable names explicitly. If we're going to log here, I'd rather log the problem rather than specifically the environment variable name.
```suggestion
  if (!url) {
    logger.LogError("API URL is undefined");
    throw new Error(API URL is undefined");
  }
```

PR #10591, opened 2025-06-17
feat: [AB#14132] add manage business vehicles task page

FarazA22:
### Ticket

## Code author checklist


web/src/components/tasks/manage-business-vehicles/ManageBusinessVehicles.tsx
```diff
@@ -0,0 +1,64 @@
+import { Content } from "@/components/Content";
```
ZacharyRizer, line 1: Dust: 

I'm wondering if we could simplify the state management for this radio button. The current implementation with `formFuncWrapper` and `useFormContext` seems like it might be more complex than needed for a single input.

A few questions to consider:

Could we use a simpler approach with a basic radio button and manual submit/error handling?

Are we anticipating this form to expand with additional fields in the future?
If not, would a lighter-weight solution be more maintainable?

Happy to discuss the tradeoffs - just want to make sure we're not over-engineering if this is likely to stay as a single radio button.

ZacharyRizer (commented):
Overall this looks solid - no major issues spotted!

One piece of feedback for future PRs: this was quite substantial and covered two distinct types of changes, which made it challenging to review thoroughly. It might be helpful to consider splitting similar PRs into separate chunks - perhaps one focused on the new Managing Vehicles functionality and another for the UpdateQueue refactor work. This would make it easier to evaluate the new features independently from the refactoring changes.

I've also left a specific comment on ManageBusinessVehicleQuestion.tsx about potentially simplifying the implementation there.

Great work on this feature!

ZacharyRizer (approved):

web/src/components/tasks/manage-business-vehicles/ManageBusinessVehicles.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const {
+    getInvalidFieldIds,
+    FormFuncWrapper,
+    onSubmit,
+    state: formContextState,
+  } = useFormContextHelper(createDataFormErrorMap());
+
+  const isTaskCompleted =
+    business?.taskProgress[props.task.id] === "COMPLETED" &&
+    business?.roadmapTaskData?.manageBusinessVehicles !== undefined;
```
SirSamuelJoseph, line 34: [dust] If the `manageBusinessVehicles` field is filled out, do we care if the `taskProgress` is "COMPLETED"?

web/src/lib/types/types.ts
```diff
@@ -517,25 +509,6 @@ export type ProfileTabs = (typeof _profileTabs)[number];
 
 export const profileTabs = _profileTabs as unknown as ProfileTabs[];
 
-export interface UpdateQueue {
```
SirSamuelJoseph, line 520: I think this refactor should have been its own PR - it's hard to review a 91 file PR and the changes made here don't rely on the other changes for the business vehicles task page.

I like this change but think it would make both changes more easy to understand if they were done separately.

web/src/components/tasks/manage-business-vehicles/ManageBusinessVehicles.tsx
FarazA22: removed

web/src/components/tasks/manage-business-vehicles/ManageBusinessVehicles.tsx
FarazA22: @SirSamuelJoseph 
I also chatted with Zach about this request, and I do think it makes sense to remove the formFunction wrapper and formContext to simplify things. That said, given the timeline and our goal of getting this out the door, I’m leaning toward leaving the code as-is for now.

One reason is that we might end up adding something later, so keeping the current structure could save us some work. Plus, there’s already quite a bit of boilerplate in place, and undoing it would take time we may not have at the moment.

If there’s a strong push for refactoring, I’m definitely open to it — and in that case, I think creating a separate refactor ticket could be the right way to go.

What do you think — would that make sense to you?

api/src/db/migrations/v166_add_roadmap_task_data.ts
```diff
@@ -0,0 +1,1059 @@
...
+): Promise<v166Business> => {
+  const v165BuinessCopy = cloneDeep(business);
+  const newBusiness = { ...v165BuinessCopy, version: 166, roadmapTaskData: {} } as v166Business;
+
+  const nonEssentialRadioAnswers =
+    newBusiness.profileData.nonEssentialRadioAnswers["business-vehicle-title-reg"];
+  delete newBusiness.profileData.nonEssentialRadioAnswers["business-vehicle-title-reg"];
+
+  if (nonEssentialRadioAnswers === true) {
+    newBusiness.roadmapTaskData.manageBusinessVehicles = true;
```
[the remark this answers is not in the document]
SirSamuelJoseph: This all makes sense, this choice works in that context, and keeping profileData tied to profileData.

PR #10606, opened 2025-06-19
fix: [AB#14891] callout icon bullets will show without needing body text

adidner:
## Description

We were having an issue where the callout component required body text to use the bulleted icon text attributes. 

### Ticket

This pull request resolves [#14891](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/14891).

### Steps to Test

run the app
run the cms proxy
go into tasks, add a callout block to markdown field
image
Make it a quick reference callout

see that if you add header text and then add to any of the bottom attributes (the ones with Icon links)

that they render without needing body copy. 

## Code author checklist

### DIFF
--- a/web/src/components/njwds-extended/callout/CalloutLayout.tsx
+++ b/web/src/components/njwds-extended/callout/CalloutLayout.tsx
@@ -33,12 +33,10 @@ export const CalloutLayout = (props: CalloutLayoutProps): ReactElement => {
           )}
           <span className="text-bold">{props.headingText}</span>
         </div>
-        {props.children && (
-          <div className="margin-top-105">
-            <div>{props.children}</div>
-            <IconTextList items={props.iconItems} />
-          </div>
-        )}
+        <div className="margin-top-105">
+          <div>{props.children}</div>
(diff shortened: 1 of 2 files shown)


web/src/components/njwds-extended/callout/CalloutLayout.tsx
```diff
@@ -33,12 +33,10 @@ export const CalloutLayout = (props: CalloutLayoutProps): ReactElement => {
...
-        {props.children && (
-          <div className="margin-top-105">
-            <div>{props.children}</div>
-            <IconTextList items={props.iconItems} />
-          </div>
-        )}
+        <div className="margin-top-105">
+          <div>{props.children}</div>
+          <IconTextList items={props.iconItems} />
+        </div>
```
adidner, line 39: dust - this logic was unnecessary and causing our issue. Because the body copy is children, if children were undefined we wouldn't get the IconTextList even if it had values. 

The entire conditional rendering is unnecessary though because if children are undefined or empty they will render nothing and same thing with IconTextList, this is even the pattern used slightly further down in this file.

PR #11041, opened 2025-08-20
chore: [AB#13034] update CloudWatch alarm to track endpoint-level errors

aadedejifearless:
## Description

Previously, our health check alarms weren’t showing enough context in Slack notifications. The alerts did not clearly display the failing endpoint or the stage, which made it difficult to quickly identify the root cause and its source.

These changes update the health check flow so that metrics are pushed directly to CloudWatch with Endpoint and Stage as dimensions. This allows Slack notifications to clearly render which endpoint failed in which stage, making troubleshooting much faster.

- Added @aws-sdk/client-cloudwatch dependency
- Updated healthCheck method to publish metrics directly instead of relying on "FAIL" log filters
- Added rule in .dependency-cruiser for @aws-sdk/client-cloudwatch

### Ticket

This pull request resolves [#13034](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/13034).

## Code author checklist

### DIFF
--- a/api/src/libs/healthCheck.ts
+++ b/api/src/libs/healthCheck.ts
@@ -1,8 +1,11 @@
 import { LogWriterType } from "@libs/logWriter";
 import axios, { AxiosError, AxiosResponse } from "axios";
+import { CloudWatchClient, PutMetricDataCommand } from "@aws-sdk/client-cloudwatch";
 
 type Status = "PASS" | "FAIL" | "ERROR";
 type StatusResult = Record<string, Status>;
+
+const cw = new CloudWatchClient({ region: process.env.AWS_REGION ?? "us-east-1" });
 
 const healthCheckEndPoints: Record<string, string> = {
   self: "self",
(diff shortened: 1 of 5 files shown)


api/src/libs/healthCheck.ts
```diff
@@ -17,23 +20,50 @@ const healthCheckEndPoints: Record<string, string> = {
...
   return axios
-    .get(`${url}/health/${type}`)
+    .get(fullUrl)
     .then((response: AxiosResponse) => {
       if (response.data.success === true) {
         logger.LogInfo(`Health Check Status - Endpoint: ${type}: PASS`);
         return "PASS";
       } else {
         logger.LogError(`Health Check Status - Endpoint: ${type}: FAIL`, response.data);
+        sendMetric(type, logger);
```
FarazA22, line 33: sand - should there be an await here since this fn is async?

api/src/libs/healthCheck.ts
```diff
@@ -17,23 +20,50 @@ const healthCheckEndPoints: Record<string, string> = {
...
+              { Name: "Stage", Value: process.env.STAGE ?? "unknown" },
+            ],
+            Value: 1,
+            Unit: "Count",
+          },
+        ],
+      }),
+    );
+  } catch {
+    logger.LogError("Failed to push metric to CloudWatch");
```
FarazA22, line 63: sand - if we are not able to successfully send the metric should we log additional details?

aadedejifearless: Thanks for that suggestion. I’ve updated the sendMetric function and added the any type caster to the caught error to handle TypeScript constraints. To accommodate this, I also disabled the `@typescript-eslint/no-explicit-any` rule for the function. This ensures that any errors from CloudWatch are logged properly without breaking the healthCheck flow

PR #11148, opened 2025-09-03
Encrypt cigarette sales licenese application tax id

jphechter:
## Description

Adds functionality to encrypt the cigarette sales application's tax id. Including this as part of the clean up work to complete the task.

### Ticket

This pull request resolves [#15791](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/15791).

### Approach

Extends the current pattern for encrypting our fields, and implements this for the data associated with the cigarette sales license application.

### Steps to Test

- Go to cigarette sales license application
  - requires you are using a new business in the retail industry and have answered "yes" to the associated cigarette non-essential question
- enter a value for your tax id
- trigger a save
  - this will happen when you click the "next" CTA button
- view encrypted value in your network tab or in the database

### Notes

Given the similarities in our encryption logic, there may be an opportunity to consolidate.

## Code author checklist

- [X] I have rebased this branch from the latest main branch
- [X] I have performed a self-review of my code
- [X] My code follows the style guide
- [X] I have created and/or updated relevant documentation on the engineering documentation website
- [X] I have not used any relative imports
- [X] I have pruned any instances of unused code
- [X] I have not added any markdown to labels, titles and button text in config
- [X] If I added/updated any values in `userData` (including `profileData`, `formationData` etc), then I added a new migration file
- [X] I have checked for and removed instances of unused config from CMS
- [X] If I added any new collections to the CMS config, then I updated the search tool and `cmsCollections.ts` (see CMS Additions in Engineering Reference/FAQ on the engineering documentation site)
- [X] I have updated relevant `.env` values in both `.env-template` and in Bitwarden

PR #11163, opened 2025-09-04
feat: [AB#15381] added GA to env perm stepper

williamavasquez:
## Description

Added GA tracking to see how users interact with the env permit questionnaire  tracks things like where people drop off, if they save their progress, and completion rates. This will help us figure out what can be improved. 

### Ticket

This pull request resolves [#15381](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/15381).

### Approach

Continued using our current GA patterns by adding to the analytics JSON as well as use the legacy events structure. 

### Steps to Test

This is a google analytics ticket so the events can be see on GA tag manager page

## Code author checklist


jphechter:
@williamavasquez I'm not seeing a lot of information in this PR overview. Can you provide some more information here? It doesn't have to be extensive, but given our PRs are often times our only documentation for a piece of work it should provide context for a reviewer today and someone who might reference this in the future.

Here's an example of 2 PRs I put together recently: #10925, #11148

web/src/components/auth/NeedsAccountModal.tsx
```diff
@@ -30,6 +30,7 @@ export const NeedsAccountModal = (): ReactElement => {
     if (isAuthenticated === IsAuthenticated.TRUE) {
       setShowNeedsAccountModal(false);
     }
+    analytics.event.gen_guidance_stepper_save_modal_displayed.appears.general_guidance_save_modal_displayed();
```
jphechter, line 33: [boulder] The `NeedsAccountModal` is used in multiple places throughout the application. Making a change to this component directly means that any time it appears this analytics event will fire. Based on what I'm seeing in the AC, I don't think that's the intention. See if there's another way to get this to fire.

web/src/components/tasks/environment-questionnaire/EnvPermit.tsx
```diff
@@ -5,21 +5,25 @@ import { UnlockedBy } from "@/components/tasks/UnlockedBy";
...
 import { EnvPermitsResults } from "./EnvPermitsResults";
 
 interface Props {
   task: Task;
 }
 
 export const EnvPermit = (props: Props): ReactElement => {
+  useEffect(() => {
+    analytics.event.gen_guidance_stepper_initiated.arrive.general_guidance_initiated();
+  }, []);
```
ZacharyRizer, line 26: Pebble: is this useEffect necessary? We aren't defendant on any changes in the component, and it will happen when the component loads. Couldn't we just put the analytic event directly on line 24?

ZacharyRizer (commented):
Small comment, and agree with JP's comments. Otherwise looks good to me!

web/src/components/tasks/environment-questionnaire/EnvPermit.tsx
jphechter: @ZacharyRizer This component is a wrapper around a form. Without the `useEffect`, wouldn't a page re-render cause us to fire the event multiple times?

web/src/components/tasks/environment-questionnaire/EnvQuestionnaireStepper.tsx
```diff
@@ -143,6 +144,31 @@ export const EnvQuestionnaireStepper = (): ReactElement => {
...
+      userWantsToContinueWithoutSaving === false
+    ) {
+      updateQueue?.queuePreferences({ returnToLink: ROUTES.envPermit }).update();
+      setShowContinueWithoutSaving(true);
+      setShowNeedsAccountModal(true);
+      envContext.setStepIndex(newStep);
+      return;
+    }
+    envContext.setStepIndex(newStep);
+  };
```
jphechter, line 170: I know you didn't initially author this code, but I think this is a good opportunity for some cleanup and to leave things better than we found them. For that reason I'm going to categorize these as [dust], but I probably would have made them all [pebble]s if I reviewed this code the first go round.

jphechter: [dust]
I recommend pulling the condition out into a named variable that better articulates what we're actually checking for.
```suggestion

    const myConditionName = envContext.state.stepIndex === 0 && isAuthenticated === IsAuthenticated.FALSE && userWantsToContinueWithoutSaving === false;

    if (myConditionName) {
      updateQueue?.queuePreferences({ returnToLink: ROUTES.envPermit }).update();
      setShowContinueWithoutSaving(true);
      setShowNeedsAccountModal(true);
      envContext.setStepIndex(newStep);
      return;
    }
    envContext.setStepIndex(newStep);
  };
```

jphechter: [dust] It looks like both in both cases we're invoking `envContext.setStepIndex(newStep)`. You could simplify this slightly.

```suggestion
    if (
      envContext.state.stepIndex === 0 &&
      isAuthenticated === IsAuthenticated.FALSE &&
      userWantsToContinueWithoutSaving === false
    ) {
      updateQueue?.queuePreferences({ returnToLink: ROUTES.envPermit }).update();
      setShowContinueWithoutSaving(true);
      setShowNeedsAccountModal(true);
    }
    envContext.setStepIndex(newStep);
  };
```

web/src/lib/utils/analytics.ts
```diff
@@ -2594,6 +2620,150 @@ export default {
...
+          });
+        },
+      },
+    },
+    gen_guidance_stepper_contact_info_engagement: {
+      mouseover: {
+        general_guidance_contact_info_engage: () => {
+          eventRunner.track({
+            event: "env_contact_info_engaged",
+            legacy_event_action: "mouseover",
```
jphechter, line 2700: @williamavasquez @somabadri This is more of a question for product. Is mouseover the best choice here?

The only other place in the application we're firing events on mouseover is for the tool-tip. This seems like a new pattern for us, and I'm curious what we're trying to gather by using this as opposed to tracking when a user actually engages with a field via focus.

Can y'all confirm with this is really what we want to do?

web/src/components/tasks/environment-questionnaire/EnvQuestionnaireStepper.tsx
```diff
@@ -143,6 +144,31 @@ export const EnvQuestionnaireStepper = (): ReactElement => {
...
+
+    if (
+      envContext.state.stepIndex === 0 &&
+      isAuthenticated === IsAuthenticated.FALSE &&
+      userWantsToContinueWithoutSaving === false
+    ) {
+      updateQueue?.queuePreferences({ returnToLink: ROUTES.envPermit }).update();
+      setShowContinueWithoutSaving(true);
+      setShowNeedsAccountModal(true);
+      envContext.setStepIndex(newStep);
```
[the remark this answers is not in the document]
somabadri: Yep, we implemented a `continue without saving button` and when the user clicks it, we want them to continue hence this complicated logic.

somabadri (commented):
great work Will! I think it's good to merge after addressing the comments JP left!

web/src/components/tasks/environment-questionnaire/EnvQuestionnaireStepper.tsx
somabadri: This is on me. I was going to refactor it when implementing the next version of the feature, but feel free to hack away at it now.

williamavasquez:
Added more details to the PR itself. Also, going to bring up the issue with modal, I think its a good point and I want to make sure I am going down the correct path. I agree that if this modal pops up in multiple places we might get the wrong data, which might make it more noise than anything useful, in the meanwhile updated the other comments.

web/src/components/tasks/environment-questionnaire/EnvQuestionnaireStepper.tsx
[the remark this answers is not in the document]
jphechter: Unresolving to follow up. This doesn't block the work in this PR.

@somabadri I must have missed when we incorporated this. This feels like a design we should make sure the whole team is aware of.

PR #11747, opened 2025-10-28
feat: [AB#16473] Add Messaging Service AWS Client for Emails

jennyverdeyen:
## Description

This adds an AWS MessagingService Client to our backend, for the purposes of sending email communications through a separate standalone lambda. It also adds logic to the existing registration router to send an email when a user successfully registers their account.

### Ticket

This pull request resolves [#16473](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16473).

### Approach

This uses the AWS Lambda SDK to directly invoke the MessagingService Lambda that we currently have set up in dev. Ashley is working on a separate task to add this lambda to the code, and these changes account for the lambda when it exists in all environments, but right now this should at least work with the lambda in dev, for testing purposes.

This PR adds:
- New MessagingServiceClient using AWS Lambda SDK for direct Lambda invocation
- Welcome email trigger in selfRegRouter on successful user registration
- IAM permissions for Lambda invocation via serverless config
- Health check added at /health/messaging-service
- Unit tests for client and updated selfRegRouter tests for email sending behavior

### Steps to Test

Because our SES is still in sandbox mode we would need to verify your email with SES before you can manually test this. But I have been able to trigger an email successfully upon account registration, locally! You can take my word for it or let me know if you want to test the flow yourself and we can add you to the SES verified emails.

## Code author checklist

### DIFF
--- a/api/src/api/selfRegRouter.test.ts
+++ b/api/src/api/selfRegRouter.test.ts
@@ -1,5 +1,5 @@
 import { selfRegRouterFactory } from "@api/selfRegRouter";
-import { DatabaseClient, SelfRegClient } from "@domain/types";
+import { DatabaseClient, MessagingServiceClient, SelfRegClient } from "@domain/types";
 import { setupExpress } from "@libs/express";
 import { DummyLogWriter } from "@libs/logWriter";
 import { generateUser, generateUserData } from "@shared/test";
@@ -16,6 +16,7 @@ describe("selfRegRouter", () => {
 
   let stubDynamoDataClient: jest.Mocked<DatabaseClient>;
   let stubSelfRegClient: jest.Mocked<SelfRegClient>;
+  let stubMessagingServiceClient: jest.Mocked<MessagingServiceClient>;
 
   beforeEach(async () => {
     stubDynamoDataClient = {
@@ -33,8 +34,20 @@ describe("selfRegRouter", () => {
       resume: jest.fn(),
     };
 
+    stubMessagingServiceClient = {
+      sendMessage: jest.fn().mockResolvedValue({ success: true, messageId: "test-message-id" }),
+      health: jest.fn(),
+    };
+
     app = setupExpress(false);
-    app.use(selfRegRouterFactory(stubDynamoDataClient, stubSelfRegClient, DummyLogWriter));
+    app.use(
+      selfRegRouterFactory(
+        stubDynamoDataClient,
+        stubSelfRegClient,
+        stubMessagingServiceClient,
+        DummyLogWriter,
+      ),
+    );
   });
 
   afterAll(async () => {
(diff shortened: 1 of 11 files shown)


api/src/functions/express/index.ts
```diff
@@ -34,6 +34,13 @@ export default (cognitoArn: string, vpcConfig: FnType["vpc"]): FnType => {
           cors: true,
         },
       },
+      {
+        http: {
+          method: "ANY",
+          path: "/api/messages/{proxy+}",
+          cors: true,
+        },
+      },
```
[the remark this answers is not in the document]
jennyverdeyen: Not sure! I thought maybe this was necessary to define but you know more about this than I do - we can discuss today and I'll probably clean this up.

api/src/api/selfRegRouter.test.ts
```diff
@@ -70,6 +83,15 @@ describe("selfRegRouter", () => {
 
       expect(response.status).toEqual(StatusCodes.OK);
       expect(response.body).toEqual({ authRedirectURL: selfRegResponse.authRedirectURL });
+
+      // Verify welcome email is sent
```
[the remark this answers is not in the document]
jphechter: Something like 
```
calls auth resume with myNJ key, saves data, returns the auth redirect URL on success, and sends a welcome email
```
is a long name.

For brevity sake, and also to ensure each test is focused, I would also vote for moving this into it's own test case.

api/src/client/AwsMessagingServiceClient.ts
```diff
@@ -0,0 +1,111 @@
+import { InvokeCommand, LambdaClient } from "@aws-sdk/client-lambda";
+import { HealthCheckMetadata, MessagingServiceClient } from "@domain/types";
+import { LogWriterType } from "@libs/logWriter";
+import { ReasonPhrases } from "http-status-codes";
+
+export const AwsMessagingServiceClient = (config: {
+  functionName: string;
```
jphechter: If we're always going to invoke the same function, then this doesn't need to be a param we pass in. It could be hard-coded into the function `sendEmail`

api/src/client/AwsMessagingServiceClient.ts
```diff
@@ -0,0 +1,111 @@
...
+  logWriter: LogWriterType;
+}): MessagingServiceClient => {
+  const logId = config.logWriter.GetId();
+  const lambdaClient = new LambdaClient({});
+
+  const sendEmail = async (
```
jphechter: We can make this function name a little more generic to something like `sendMessage`. The Messaging Service will determine the channel (email, text, etc).

```suggestion
  const sendMessage = async (
```

api/src/client/AwsMessagingServiceClient.ts
```diff
@@ -0,0 +1,111 @@
...
+    toEmailAddress: string,
+    options?: { templateId?: string; userId?: string },
```
jphechter: The Messaging Service will use the user's id to determine their comms preferences and also pull out their email address.

I'd pass the type of message we want to send (welcome, reminder, etc.) and then let the Messaging Service determine the template to use. Otherwise, we'll have to update every invocation of the function to account for any template changes.

```suggestion
    userId: string,
    messageType: string,
```

PR #11770, opened 2025-11-03
feat: [AB#16471] create messaging service lambda

ashlsun:
## Description
This PR creates a MessagingService lambda that takes a `SimpleSendRequest`, sends an email, and records a log of it in a Messages DynamoDB table. In combination with #11747, this works as a prototype for sending transactional emails to new users. 

### Ticket

This pull request resolves [#16471](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16471).

### Approach
This PR adds the components in red in the below diagram.
Architecture diagram, backend calls the messaging service lambda, which writes to the messages table and calls SES

We use CDK to create & manage these new resources. `lambdaStack.ts` creates the MessagingService lambda, and `dataStack.ts` creates the MessagesTable. I also refactored `dataStack` to create all of our tables using a new helper function in `stackUtils`.

In order for all these resources to interact with each other, they need permissions. The lambda containing our app's backend needs permission to invoke the new MessagingService lambda, the MessagingService lambda needs permissions to send emails, and also to update the messagingTable. These live in `iamStack.ts`.

The code for the MessagingService lambda is contained in `api/src/functions/messagingService/app.ts`. When it receives a request to send a message, it looks up the user with the `userId` given in the request to get their email address and name, and then attempts to send the email with the user's name filled in and log the message to the MessagesTable.

### Steps to Test

Go to testing.account.business.nj.gov, onboard as a new business, and create a new account with myNJ using your email address. You should receive an email that looks like this!

Screenshot 2025-11-24 at 3 26 07 PM

### Notes
- [x] ~Pulling the email template from a file instead of hardcoded in `app.ts`~ Didn't work after deploying, reverted back to keeping the template in the file
- [x] ~Parsing and inserting name/business into the template~
- [x] Unit tests for new Messages DB client 
- [x] CDK IaaC to replace the changes in `serverless.ts`

## Code author checklist

### DIFF
--- a/api/.dependency-cruiser.js
+++ b/api/.dependency-cruiser.js
@@ -20,6 +20,10 @@ module.exports = {
     },
     {
       from: {},
+      to: { path: "@aws-sdk/client-ses" },
+    },
+    {
+      from: {},
       to: { path: "@aws-sdk/util-dynamodb" },
     },
     {
@@ -53,6 +57,10 @@ module.exports = {
     {
       from: {},
       to: { path: "util" },
+    },
+    {
+      from: {},
+      to: { path: "uuid" },
     },
     {
       from: {},
--- a/api/jest-dynalite-config.js
+++ b/api/jest-dynalite-config.js
@@ -168,6 +168,67 @@ module.exports = {
         WriteCapacityUnits: 1,
       },
     },
+    {
+      TableName: "messages-table-test",
+      AttributeDefinitions: [
+        {
+          AttributeName: "taskId",
+          AttributeType: "S",
+        },
+        {
+          AttributeName: "dueAt",
+          AttributeType: "S",
+        },
+        {
+          AttributeName: "userId",
+          AttributeType: "S",
+        },
+      ],
+      KeySchema: [
+        {
+          AttributeName: "taskId",
+          KeyType: "HASH",
+        },
+      ],
+      GlobalSecondaryIndexes: [
+        {
+          IndexName: "DueAtIndex",
+          KeySchema: [
+            {
+              AttributeName: "dueAt",
+              KeyType: "HASH",
+            },
+          ],
+          Projection: {
+            ProjectionType: "ALL",
+          },
+          ProvisionedThroughput: {
+            ReadCapacityUnits: 1,
+            WriteCapacityUnits: 1,
+          },
+        },
+        {
+          IndexName: "UserIdIndex",
+          KeySchema: [
+            {
+              AttributeName: "userId",
+              KeyType: "HASH",
+            },
+          ],
+          Projection: {
+            ProjectionType: "ALL",
+          },
+          ProvisionedThroughput: {
+            ReadCapacityUnits: 1,
+            WriteCapacityUnits: 1,
+          },
+        },
+      ],
+      ProvisionedThroughput: {
+        ReadCapacityUnits: 1,
+        WriteCapacityUnits: 1,
+      },
+    },
   ],
   port: 8002,
 };
--- a/api/package.json
+++ b/api/package.json
@@ -30,6 +30,7 @@
     "@aws-sdk/client-dynamodb": "3.896.0",
     "@aws-sdk/client-lambda": "3.896.0",
     "@aws-sdk/client-s3": "3.896.0",
+    "@aws-sdk/client-ses": "3.922.0",
     "@aws-sdk/client-sns": "3.896.0",
     "@aws-sdk/client-ssm": "3.896.0",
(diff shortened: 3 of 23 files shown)


api/src/db/DynamoMessagesDataClient.ts
```diff
@@ -0,0 +1,117 @@
...
+  // Whether to remove undefined values while marshalling.
+  removeUndefinedValues: true, // false, by default.
+  // Whether to convert typeof object to map attribute.
+  convertClassInstanceToMap: false, // false, by default.
+};
+
+const unmarshallOptions = {
+  // Whether to return numbers as a string instead of converting them to native JavaScript numbers.
+  wrapNumbers: false, // false, by default.
+};
```
ashlsun: Unclear how necessary these configs are?

api/.dependency-cruiser.js
```diff
@@ -75,11 +83,11 @@ module.exports = {
       to: { path: "winston-cloudwatch" },
     },
     {
-      from: { path: "src/db" },
+      from: { path: "src/db", path: "src/functions/messagingService" },
       to: { path: "fs" },
     },
     {
-      from: { path: "src/db" },
+      from: { path: "src/db", path: "src/functions/messagingService" },
```
aadedejifearless: The current dependency-cruiser rules aren’t doing what we intend.
They define path twice inside the same from object, which means only the last one is actually used:
```suggestion
     {
  from: { path: ["src/db", "src/functions/messagingService"] },
  to: { path: "fs" },
},
{
  from: { path: ["src/db", "src/functions/messagingService"] },
  to: { path: "path" },
},
```

api/package.json
```diff
@@ -30,6 +30,7 @@
     "@aws-sdk/client-dynamodb": "3.896.0",
     "@aws-sdk/client-lambda": "3.896.0",
     "@aws-sdk/client-s3": "3.896.0",
+    "@aws-sdk/client-ses": "^3.922.0",
```
aadedejifearless: we want to remove the carrot here `^` to match with the 
```suggestion
    "@aws-sdk/client-ses": "3.922.0",
```

api/jest-dynalite-config.js
```diff
@@ -168,6 +168,47 @@ module.exports = {
         WriteCapacityUnits: 1,
       },
     },
+    {
```
jphechter: Just noting that if you do add user ID as a GSI, it would be good to include here.

api/src/client/AwsMessagingServiceClient.ts
```diff
@@ -4,14 +4,16 @@ import { STAGE } from "@functions/config";
 import { LogWriterType } from "@libs/logWriter";
 import { ReasonPhrases } from "http-status-codes";
 
+const SERVICE_NAME = "businessnjgov-api-v2";
```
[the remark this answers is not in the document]
jphechter: I like the sound of that.

api/src/db/DynamoMessagesDataClient.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  describe("get", () => {
+    it("should throw an error when attempting to retrieve a non-existent message by taskId", async () => {
+      const randomTaskId = `task-id-${randomInt()}`;
+
+      await expect(dynamoMessagesDataClient.get(randomTaskId)).rejects.toEqual(
+        new Error("Not found"),
+      );
+    });
+
+    it("inserts and retrieves a message from the db", async () => {
```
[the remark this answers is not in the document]
jphechter: Minor nit - I like it when the intended behavior is listed first, and then the negative cases follow. But that's just personal preference.

api/src/domain/types.ts
```diff
@@ -44,6 +44,26 @@ import { AxiosResponse } from "axios";
...
+  templateId: MessageTemplateId;
+  topic: MessageTopic; // welcome | taxReminder - TBD as comms generated
+  //     "userId" and "topic" used for deduplication
+  templateData: // used to construct the message from the template
+  {
+    name: string;
+    business: string;
+  };
+  dueAt: string; // date to deliver
+  deliveredAt: string; // remains undefined until message is sent
```
jphechter: Should this be possibly undefined?
```suggestion
  deliveredAt: string | undefined;
```

api/.dependency-cruiser.js
ashlsun: Ah, thank you for catching that! I realized that these also no longer need to be modified because we are no longer using these libraries in messagingService/app.ts, so I'm gonna revert this!

PR #11786, opened 2025-11-13
chore: [AB#15850] employer rates no account match

FarazA22:
### Ticket

This pull request resolves [#15850](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/15850).

### Approach
I added the logic needed to handle the error case that comes back from the Employer Rates API. The API returns a 200 response, but inside the response body there can be an error key with a string message. That key-value pair only appears when the business name, email, and EIN do not match what the system expects. In other words, it is a validation mismatch on their side.

I updated our flow so that when this error message is present, we surface it to the user.

I also added logic around the initial radio button that controls when the user is allowed to submit the request. If an Employer Rates validation error is currently visible and the user interacts with that radio button again, the error clears out so they can try again with a clean state.

Finally, I updated the EIN field behavior. If the EIN input shows its own error due to not meeting the fifteen character requirement, that message should override and remove the APR validation error so there is never a stack of competing errors.

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/employer-rates.json
+++ b/content/src/fieldConfig/employer-rates.json
@@ -19,6 +19,7 @@
     "dolEinAlertLabelText": "DOL EIN",
     "dolEinLabelText": "***`DOL EIN|dol-ein`***",
     "dolEinErrorText": "DOL EIN must be 15 digits.",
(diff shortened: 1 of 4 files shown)


SirSamuelJoseph (approved):
Really clean, nice work!

PR #11793, opened 2025-11-17
fix: [AB#16705] return the entire error from Webservice license search

jphechter:
## Description

License search is currently broken because we are not handling errors from the Webservice license status client correctly. The Webservice tools are currently unreachable, which surfaced an issue in our logic which causes the entire license search feature to fail. This work should allow us to handle those errors more gracefully.

### Ticket

This pull request resolves [#16705](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16705).

### Approach

If the Webservice license search encounters an error, return the entire error, not just the message. Previously, we were just returning the error message which could be undefined in the case of a connection timeout. Because of this, our entire implementation of license search was failing.

### Steps to Test

**Previously**
- All license searches resulted in an error message regardless of license type or data source
- Any license search resulted in a more general error message or return license data
Screenshot 2025-11-17 at 3 31 34 PM

**Webservice**
- Create a business that can conduct a license search using the legacy system (e.g. cosmetology)
- While the service is unreachable, we should get a more general error message
Screenshot 2025-11-17 at 3 36 22 PM

**Regulated Businesses**
- Create a business that can conduct a license search using the Dynamics - Regulated Business system (e.g. public movers)
- If there is a matching license, that information should be returned.
Screenshot 2025-11-17 at 3 31 52 PM

### Notes

We're currently not getting warned via healthchecks that the Webservice is down. I don't know if this change will fix that issue or not. Post deployment, I will continue investigating to figure out if this change was enough or if that needs further focus.

## Code author checklist

### DIFF
--- a/api/src/client/webservice/WebserviceLicenseStatusClient.ts
+++ b/api/src/client/webservice/WebserviceLicenseStatusClient.ts
@@ -32,7 +32,7 @@ export const WebserviceLicenseStatusClient = (
           `Webservice License Status Client - Error. - Id:${logId} - Error:`,
           error,
         );
-        throw error.response?.status;
+        throw error;
       });
   };
 
--- a/api/src/client/dynamics/license-status/RegulatedBusinessDynamicsLicenseStatusClient.ts
+++ b/api/src/client/dynamics/license-status/RegulatedBusinessDynamicsLicenseStatusClient.ts
@@ -19,7 +19,7 @@ interface RgbApiConfiguration {
   rgbChecklistItemsClient: ChecklistItemsForAllApplicationsClient;
 }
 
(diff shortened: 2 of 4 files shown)


api/src/client/webservice/WebserviceLicenseStatusClient.ts
```diff
@@ -32,7 +32,7 @@ export const WebserviceLicenseStatusClient = (
           `Webservice License Status Client - Error. - Id:${logId} - Error:`,
           error,
         );
-        throw error.response?.status;
+        throw error;
```
jphechter: This is the main change. Everything else in here is mostly renaming variables.

PR #11863, opened 2025-12-08
feat: [AB#16676] Handle unsaved changes in profile with new modal and updated navigation interception logic

jennyverdeyen:
## Description
There was existing code to handle unsaved changes in the profile when a user tries to navigate away, but it didn't appear to be working.  A new modal design was provided, and this implements that as well as some new logic so that it works as expected.

**First, an implementation note:**

I am learning Next.js stuff as I go, so open to suggestions if my understanding is wrong, but basically:

- We’re using the Pages Router’s documented pattern to cancel navigation: listen on routeChangeStart and throw; Next aborts the transition and emits routeChangeError with err.cancelled = true (see https://nextjs.org/docs/pages/api-reference/functions/use-router#routerevents).
- This is the only reliable way to block navigation while using Pages Router; it’s non-ideal but standard. In production there’s no Next error visible to the user, but it is annoying in local/dev.
- I've added a guard to skip the throw when NODE_ENV === development so we can go on without the error overlay; basically this functionality isn't testable locally and instead navigation is allowed with unsaved changes. I've pushed this to testing so we can see the modal there.

With that said, here's what's changed in this PR:

- Added a 3-button escape modal (ModalThreeButton) and updated ProfileEscapeModal to use it.
- Implemented useUnsavedChangesGuard custom hook to block in-app navigation on unsaved profile changes (Next router routeChangeStart) and show the modal; also uses the browser-built-in beforeunload prompt for refresh/close.
- Integrated the guard into profile.tsx to control the modal with pending redirect logic, and save logic from the modal.
- Expanded mock router with events support; added navigation intercept test and save-and-redirect test in `profile-starting.test.tsx`.
- Added CMS/content fields for the new modal copy.

<sup>Ooh look a new disclaimer because I'm trying to be a responsible and transparent AI user:</sup> Portions drafted with AI assistance; all changes reviewed and verified by me, a human!

### Ticket

This pull request resolves [#16676](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16676).

### Steps to Test

To test this, head to the dashboard *in the testing environment* and make any change to your profile. Without saving those changes, try to navigate away. You can hit the "Go Back" link, the browser back buton, any other link on the page, you can try to close the tab, you can try to refresh: in all cases you should see the navigation be intercepted either by the custom modal or by the browser's dialog (in the cases of closing or refreshing)

Make sure the modal buttons do what you think they should do!

## Code author checklist


web/src/lib/data-hooks/useUnsavedChangesGuard.ts
```diff
@@ -0,0 +1,111 @@
...
+    };
+  }, [router]);
+
+  // Handle browser refresh/close
+  useEffect(() => {
+    const handleBeforeUnload = (e: BeforeUnloadEvent): string | undefined => {
+      if (hasUnsavedChanges) {
+        e.preventDefault();
+        // For cross-browser support of the built-in "Leave site?" dialog, set returnValue in addition to preventDefault.
+        // See MDN: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event
```
jennyverdeyen: I left this note here for whoever looks at this in the future because TypeScript shows e.returnValue as being deprecated, but apparently it's recommended for legacy cross-browser support

web/src/components/ModalThreeButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <ReverseOrderInMobile
+        className={`flex flex-column flex-align-stretch ${isMobileAndUp ? "" : "gap-2"}`}
+      >
+        <>
+          <UnStyledButton
+            onClick={props.tertiaryButtonOnClick ?? props.close}
+            dataTestid="modal-button-tertiary"
+            className={`${isMobileAndUp ? "padding-right-2" : ""}`}
+          >
+            <span className="text-bold" style={{ color: "#538200" }}>
```
jphechter: 2 different notes:
- `UnStyledButton` has a `isTextBold` prop. It looks like setting that would have the same outcome as applying the class name here.
- I only see that hex code 1 other place in our codebase (in a commit from 4 years ago!). If we intend to start using this color, we might want to more formally incorporate this into our styles.

web/src/lib/data-hooks/useUnsavedChangesGuard.ts
```diff
@@ -0,0 +1,113 @@
+import { useRouter } from "next/compat/router";
```
jphechter: It might be worthwhile adding a test file for this.

web/src/lib/data-hooks/useUnsavedChangesGuard.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      setPendingUrl(url);
+      setIsBlocked(true);
+
+      if (process.env.NODE_ENV === "development") return;
+
+      const err: Error & { cancelled?: boolean } = new Error(
+        "Navigation blocked due to unsaved changes",
+      );
```
jphechter: I might actually add a comment explaining this piece. It's not a pattern I'm familiar with. Especially depending on the outcomes of our conversations tomorrow at engineering sync.

web/src/components/ModalThreeButton.tsx
jennyverdeyen: It's funny because it's called... unstyled button... 🤔

PR #11862, opened 2025-12-08
feat: [AB#16118] implement simple welcome email in html

ashlsun:
## Description

This PR replaces the placeholder HTML email with [the design for Version B](https://www.figma.com/design/tRRnDHLBgQfZGjLddfaTBT/91-95-Business-Experience-Sprint-Designs?node-id=64891-22991&t=NNpqKk1tpOyCkdnX-4). The images in this email are hosted at https://github.com/newjersey/business-files.

### Ticket

This pull request resolves [#16118](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16118).

### Approach

HTML in emails looks a little different from HTML for webpages. Due to limited support for rendering CSS in email clients, we need to use inline CSS and a table layout (instead of flexbox or grid). I wrote out a skeleton and used AI to convert it into a layout that would be supported by most email clients. I did several passes of corrections based on what I saw from opening the HTML file in my browser and sending the email to my Gmail and Outlook inboxes.

- Moves the welcome email template into its own file
- Updates for Lambda
  - Add a type definition so that `html` files are viewed as a module
    - This allows us to import the file in the `MessagingService`
    - CDK will include any imports as part of the bundle deployed to the Lambda
  - Instruct CDK to process the `html` files as a text file
- Modularize some of the logic within `MessagingService`

Not a code change: We've updated intercom to allow users to start new chat conversations via contacting us at a specific email address. That email address, however, has been added to the `replyTo` param of the SES command.

### Steps to Test

- Create a new account
  - use a real email address you have access to
- Link with myNJ
- Upon return to the Dashboard of Business.NJ.gov, a welcome email should be sent to the address set on your account

**Desktop**
Screenshot 2025-12-11 at 9 30 32 AM

**Mobile (dark mode)**
IMG_8649

### Notes

This will eventually be replaced with a React Email implementation. 

## Code author checklist

- [X] I have rebased this branch from the latest main branch
- [X] I have performed a self-review of my code
- [X] My code follows the style guide
- [X] I have created and/or updated relevant documentation on the engineering documentation website
- [ ] I have not used any relative imports
  - I am using relative imports for loading the email template. This will change soon with the refactor to utilize React Email
- [X] I have pruned any instances of unused code
- [X] I have not added any markdown to labels, titles and button text in config
- [X] If I added/updated any values in `userData` (including `profileData`, `formationData` etc), then I added a new migration file
- [X] I have checked for and removed instances of unused config from CMS
[...]

PR #11866, opened 2025-12-09
feat [AB#16898]: reimplement simple welcome email in React Email

ashlsun:
## Description

This PR replaces the HTML email in #11862 with a React Email implementation.

### Ticket

This pull request resolves [#16898](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16898).

### Approach

I followed [this guide to getting started with React Email](https://react.email/docs/getting-started/monorepo-setup/yarn). It had me create a new workspace for React Email at `api/src/functions/messagingService/reactEmail`. 

The workflow for creating new emails is as follows: `cd` into the `reactEmail` directory and run `yarn dev` to start the React Email preview app. This app lets you preview email components in its `./emails` folder and scans for compatibility/accessibility issues.

For the MessagingService Lambda to use these components, we compile React Email components to static HTML at build time using `build.tsx`. 

This build script imports the email components, renders them to HTML using `@react-email/render`, and outputs the HTML files to `api/src/functions/messagingService/email-templates/`. The Lambda then reads these pre-built HTML files at runtime, which eliminates the need for React dependencies in the Lambda itself.

The build script currently only handles `WelcomeEmailB` but can be extended to auto-discover all email components when we add more emails in the future.

### Steps to Test
This is currently deployed in the testing site - sign up with a new myNJ account to receive the generated welcome email.

You can also locally test that the build script works by running `yarn build` from the root. This should re-run the build script inside react email and create the welcome email HTML in the `api/src/functions/messagingService/reactEmail/emails` folder!

### Notes

- The generated HTML files are ignored from git since CI will overwrite them when building.
- Since we plan on storing full emails in S3 for compliance, I don't think we need template versioning at this point.
- This PR doesn't remove the original 2 HTML emails in the `api/src/functions/messagingService/emails` folder for ease of rollback, but we'll plan to remove them once other emails are implemented in React Email.
  
## Code author checklist

### DIFF
--- a/api/.eslintrc.json
+++ b/api/.eslintrc.json
@@ -128,6 +128,12 @@
         "unicorn/switch-case-braces": "off",
         "unicorn/prefer-at": "off"
       }
+    },
+    {
+      "files": ["src/functions/messagingService/reactEmail/**"],
+      "rules": {
+        "no-restricted-imports": "off"
+      }
     }
   ]
 }
--- a/api/src/functions/messagingService/app.ts
+++ b/api/src/functions/messagingService/app.ts
@@ -24,7 +24,9 @@ import { LogWriter } from "@libs/logWriter";
 import { getConfigValue, USER_MESSAGING_CONFIG_VARS } from "@libs/ssmUtils";
 import { v4 as uuidv4 } from "uuid";
 // eslint-disable-next-line no-restricted-imports
-import welcomeHtmlTemplate from "./emails/welcomeEmail.html";
+import welcomeEmailShortVersionTemplate from "./email-templates/welcomeEmailShortVersion.html";
+// eslint-disable-next-line no-restricted-imports
+import welcomeEmailShortVersionPlaintext from "./email-templates/welcomeEmailShortVersion.txt";
 // eslint-disable-next-line no-restricted-imports
 import testReminderHtmlTemplate from "./emails/testReminderEmail.html";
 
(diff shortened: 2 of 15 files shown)


api/.eslintrc.json
```diff
@@ -128,6 +128,12 @@
         "unicorn/switch-case-braces": "off",
         "unicorn/prefer-at": "off"
       }
+    },
+    {
+      "files": ["src/functions/messagingService/reactEmail/*"],
+      "rules": {
+        "no-restricted-imports": "off"
+      }
```
ashlsun: Added this to get around the restricted import when I was doing `import { WelcomeEmailB} from "./react-email/emails/welcomeEmailB"` but this might not be doing anything anymore, should probably remove.

api/src/functions/messagingService/app.ts
```diff
@@ -24,7 +24,7 @@ import { LogWriter } from "@libs/logWriter";
 import { getConfigValue, USER_MESSAGING_CONFIG_VARS } from "@libs/ssmUtils";
 import { v4 as uuidv4 } from "uuid";
 // eslint-disable-next-line no-restricted-imports
-import welcomeHtmlTemplate from "./emails/welcomeEmail.html";
+import welcomeEmailBTemplate from "./email-templates/welcomeEmailB.html";
```
SirSamuelJoseph: [dust] Can we have a more descriptive name than Email B?

ashlsun: Makes sense, I renamed it to WelcomeEmailShortVersion since the next email will be a longer version!

PR #11876, opened 2025-12-16
feat: add login help page

hinzed1127:
## Description

This adds a support page for troubleshooting login issues. It includes a small amount of conditional rendering described in the steps to test

### Ticket

This pull request resolves [#16449](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16449).

### Approach

- Leveraging `getActiveUser` for the conditional bits
- Added a new section to the CMS. Since we aspire to potentially add other support-style pages is in the future, that felt cleaner than adding to `12-misc.yml`

### Steps to Test

Running the app locally:
1. Go to http://localhost:3000/support/login
2. If authenticated, verify that the top right of the nav bar _does not_ display the "Log in" button and _does_ display the message at the bottom of the page to "log out". If unauthenticated, verify the inverse is true (i.e. "log in" nav item renders, "log out" message at the bottom is hidden)
3. Run `yarn cms:proxy` in a separate terminal. Navigating to http://localhost:3000/mgmt/search and searching for "logging" should show a result for the login support page.
4. Everything else is static copy. Validate that nothing looks amiss or isn't in line with [design specs](https://www.figma.com/design/oA6upbAtUmk4L893vWO0jx/86-90-Business-Experience-Sprint-Designs?node-id=62663-4894)
## Code author checklist

- [x] I have rebased this branch from the latest main branch
- [x] I have performed a self-review of my code
- [x] My code follows the style guide
- [x] I have created and/or updated relevant documentation on the engineering documentation website
- [x] I have not used any relative imports
- [x] I have pruned any instances of unused code
- [x] I have not added any markdown to labels, titles and button text in config
- [x] If I added/updated any values in `userData` (including `profileData`, `formationData` etc), then I added a new migration file
- [x] I have checked for and removed instances of unused config from CMS
- [x] If I added any new collections to the CMS config, then I updated the search tool and `cmsCollections.ts` (see CMS Additions in Engineering Reference/FAQ on the engineering documentation site)
- [x] I have updated relevant `.env` values in both `.env-template` and in Bitwarden

PR #11972, opened 2025-12-22
feat: add login assistance message, support link on `/loading`

hinzed1127:
## Description

This one updates the loading page to better help users if the the loading page hangs for some unknown reason or, more likely, they're encountering the unlinked myNJ account error.
### Ticket

This pull request resolves [#16647](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16647)

### Approach

This one changed a little bit as I worked on it due to what error states I thought were or were not possible.

#### Some backstory

I had previously observed some behavior with unlinked accounts where the following was happening:
1. you log into myNJ with an unlinked account
2. A Cognito session *would* be established, but with an invalid state where the `id` from Cognito would be the same as the logged in user's email, e.g. something like this:
```json
{
	id: 'dan@example.com',
	email: 'dan@example.com'
	// other attributes in the JWT response from Cognito
}
```
3. The infinite loading state would occur because the code would detect both that a user was authenticated, but then fetching by that invalid ID would fail to find a user.

I'm not sure if something changed in myNJ more accurately sending back proper SAML failures, or something else in how I was testing things, but in building out this feature, the aforementioned state no longer appears to be replicated. Rather, we will encounter a flow previously expected for and (somewhat) handled:
1. you log into myNJ with an unlinked account
2. on the chained redirects of myNJ -> Cognito -> navigator, a Cognito session fails to be established , and we get redirected to the loading page with some type of url like this: https://account.business.nj.gov/loading?error_description=Invalid+SAML+response+received%3A+Name+ID+value+was+not+found+in+SAML+Assertion+&state=qJs69qH5i5Jfn7Jb3OWp4eyqYQ8r6mXm&error=server_error
3.  `Name+ID+value+was+not+found+in+SAML` is detected in the query params and we show the appropriate error copy
Where this previously would trigger a modal, this PR makes updates so that we're now showing an error message with links to the new static help page
### Steps to Test

1. Log in with a local account you know is unlinked (this will require either bypassing the interstitial login page or the unlinked account using an email you've already used for a linked account
2. The error message should appear on the redirected loading page

The other states are accounted for in the loading component and unit tests, but the only one that's easy to replicate naturally (to my knowledge is that one). The four possible states accounted for:

| Logged in user? | Unlinked myNJ? | Is this testable? |
| --- | --- | --- |
| ❌ | ✅ | Follow the aforementioned testing steps |
| ❌ | ❌ | Accounted for in code and unit tests but can't naturally replicate |
| ✅ | ✅ | Accounted for in code and unit tests but can't naturally replicate |
| ✅ | ❌ | Accounted for in code and unit tests but can't naturally replicate |

TL;DR we only expect users to naturally encounter hitting the loading page with an error state of "unauthenticated in navigator (i.e., no Cognito session) + an unlinked myNJ account", but we still account for the other error states _in the unlikely case that they occur_.

### Notes

For a cleaner diff, I've made my `dh-static-login-page` branch, which is currently unmerged, the base branch (see https://github.com/newjersey/navigator.business.nj.gov/pull/11876 for those changes, which are related but independent of these)

## Code author checklist

### DIFF
--- a/content/src/fieldConfig/login-support-page.json
+++ b/content/src/fieldConfig/login-support-page.json
@@ -1,5 +1,12 @@
 {
   "loginSupportPage": {
+    "unlinkedAccount": "It looks like you haven’t linked your myNewJersey account.",
+    "havingTrouble": "Having trouble logging in?",
+    "logoutButtonText": "Log out",
+    "logoutButtonTextUnlinked": "log out",
+    "logoutAndTryAgain": "~~logoutButton~~ and try again.",
+    "logoutAndTryAgainUnlinked": "Please ~~logoutButton~~ and try again.",
+    "forMoreAssistance": "For more assistance, visit our [Login Issues Help page](support/login).",
     "pageTitle": "Having trouble logging in to Business.NJ.gov?",
     "lastUpdated": "Nov 5, 2025",
     "multipleAccountsCallout": ":::miniCallout{ calloutType=\"informational\" }\n**If you have multiple myNewJersey accounts, use the email linked to your business.**\n:::",
--- a/web/decap-config/collections/13-support.yml
+++ b/web/decap-config/collections/13-support.yml
@@ -21,6 +21,21 @@ collections:
             collapsed: false
             widget: object
             fields:
+              - { label: "Unlinked Account", name: "unlinkedAccount", widget: "string" }
+              - { label: "Having Trouble", name: "havingTrouble", widget: "string" }
+              - { label: "Logout Button Text", name: "logoutButtonText", widget: "string" }
+              - {
+                  label: "Logout Button Text Unlinked",
+                  name: "logoutButtonTextUnlinked",
+                  widget: "string",
+                }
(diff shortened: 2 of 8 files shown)


content/src/fieldConfig/login-support-page.json
```diff
@@ -1,5 +1,12 @@
 {
   "loginSupportPage": {
+    "unlinkedAccount": "It looks like you haven’t linked your myNewJersey account.",
+    "havingTrouble": "Having trouble logging in?",
+    "logoutButtonText": "Log out",
+    "logoutButtonTextUnlinked": "log out",
+    "logoutAndTryAgain": "~~logoutButton~~ and try again.",
+    "logoutAndTryAgainUnlinked": "Please ~~logoutButton~~ and try again.",
```
jphechter: I'm not familiar with this syntax. I've seen us use something like
```
:::callout
:::
```

jphechter (approved):
I think these changes make sense.

PR #12395, opened 2026-01-11
chore: [AB#16176] upgrade to React 19

seidior:
## Description

This pull request upgrades the Navigator application from **React 18.3.1 to React 19.2.3**, along with significant updates to the MUI (Material-UI) ecosystem, Next.js build configuration, and Storybook. This is a major infrastructure upgrade that touches the core rendering engine of the application, the UI component library, and the build toolchain.

### What is React 19?

React 19 represents a significant evolution in how React handles rendering, state updates, and component composition. The most impactful changes for our codebase include:

1. **Automatic Batching Everywhere**: React 19 extends automatic batching to all state updates, including those in promises, timeouts, and native event handlers. Previously, only updates within React event handlers were batched. This means multiple `setState` calls in async contexts are now batched together, reducing re-renders but potentially changing the timing of when state updates become visible in the DOM.

2. **Concurrent Rendering by Default**: React 19 leans more heavily into concurrent rendering features, which means the framework may pause, restart, or abandon renders. This can cause issues with code that makes assumptions about render timing or expects synchronous behavior.

3. **Ref Handling Simplification**: The `forwardRef` API, while still supported, is now considered legacy. React 19 allows refs to be passed directly as props, simplifying component APIs and reducing boilerplate.

4. **Stricter Rules and Deprecations**: Several patterns that were merely discouraged in React 18 now produce warnings or errors in React 19, particularly around `findDOMNode` usage, test utilities, and certain hook patterns.

### Ticket

This pull request resolves [#16176](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16176).

### Approach

This upgrade required changes across multiple dimensions of the application. Here's a detailed breakdown:

#### 1. Dependency Updates

**Core React Ecosystem:**

- `react`: 18.3.1 → 19.2.3
- `react-dom`: 18.3.1 → 19.2.3
- `@types/react`: 18.3.27 → 19.2.7
- `@types/react-dom`: 18.3.7 → 19.2.3

**Material-UI (MUI) Components:**

- `@mui/material`: 5.18.0 → 7.3.6
- `@mui/icons-material`: 5.18.0 → 7.3.6
- `@mui/lab`: 5.0.0-alpha.177 → 7.0.1-beta.20
- `@mui/x-date-pickers`: 5.0.20 → 8.23.0
- `@mui/system`: 5.18.0 → 7.3.6
- `@mui/styles`: 5.18.0 → 6.4.8

**Markdown/Unified Ecosystem:**

- `unified`: 10.1.2 → 11.0.5
- `remark-parse`: 10.0.2 → 11.0.0
- `remark-gfm`: 3.0.1 → 4.0.1
- `remark-directive`: 2.0.1 → 4.0.0
- `remark-rehype`: 10.1.0 → 11.1.2
- `rehype-react`: 7.2.0 → 8.0.0
- `rehype-stringify`: 9.0.4 → 10.0.1
- `rehype-rewrite`: 4.0.3 → 4.0.4

**Testing & Tooling:**

- `@testing-library/react`: 16.3.0 → 16.3.1
- `eslint-plugin-react-hooks`: 4.6.2 → 7.0.1
- `swr`: 1.3.0 → 2.3.8
- `react-slick`: 0.30.0 → 0.31.0
- `decap-cms-app`: 3.0.12 → 3.9.0

**Storybook:**

- `storybook`: 7.6.21 → 10.1.11
- `@storybook/nextjs`: 7.6.20 → 10.1.11
- `@storybook/react`: 7.6.20 → 10.1.11
- `@storybook/addon-essentials`: 7.6.20 → 8.6.15
- Various other Storybook addons updated to v8-10 range

**Removed Dependencies:**

- `react-transition-group`: Replaced with a custom `CSSTransition` component for React 19 compatibility

**Added Dependencies:**

- `fs-extra`: 11.3.3 (for build scripts)
- `@types/fs-extra`: 11.0.4
- `tsx`: 4.21.0 (for running TypeScript scripts directly)

#### 2. Next.js Configuration Overhaul

The `next.config.js` file was converted to TypeScript (`next.config.ts`) and significantly refactored:

**Turbopack Migration:**

- Webpack configuration was replaced with Turbopack rules for markdown file handling
- Custom webpack plugins (CopyPlugin, CleanWebpackPlugin) were replaced with standalone build scripts
- The `--webpack` flag was removed from dev and build scripts

**New Build Scripts:**

- `copy-vendor-assets.ts`: Copies NJWDS (New Jersey Web Design System) images and JavaScript files from `node_modules` to `public/vendor/`
- `patch-njwds-css.ts`: Patches invalid CSS pseudo-element syntax (`::before--tile` → `::before`) in the NJWDS stylesheet that causes Turbopack build failures

**Standalone Output:**

- Added `output: "standalone"` configuration for optimized Docker deployments
- This produces a minimal production bundle that includes only the necessary `node_modules`

#### 3. MUI Date Picker API Migration

MUI X Date Pickers v8 introduced a completely new API that breaks backward compatibility. All date picker components were updated:

**Old API (v5):**

```tsx
<DatePicker
  inputFormat={"MM/YYYY"}
  disableMaskedInput={false}
  mask={"__/____"}
  onChange={(date: DateObject | null) => handleChange(date)}
  renderInput={(params) => <TextField {...params} />}
/>
```

**New API (v8):**

```tsx
<DatePicker
  enableAccessibleFieldDOMStructure={false}
  format={"MM/YYYY"}
  onChange={(date: unknown) => {
    const dateObj = date as DateObject | null;
    handleChange(dateObj);
  }}
  slotProps={{
    textField: { placeholder: "__/____", error: false, ... }
  }}
  slots={{
    textField: (params) => <TextField {...params} />
  }}
/>
```

Key changes:

- `inputFormat` → `format`
- `renderInput` → `slots.textField` with `slotProps.textField` for props
- `mask` and `disableMaskedInput` removed, placeholder in `slotProps`
- `enableAccessibleFieldDOMStructure={false}` required for backward-compatible behavior
- Type signatures changed (date picker onChange provides `unknown` instead of typed value)

Components affected:

- `DateOfFormation.tsx`
- `FormationDate.tsx`
- `EmergencyTripPermitDatePicker.tsx`
- `CigaretteSalesStartDate.tsx`

#### 4. MUI Popper anchorEl Fix

The MUI `Popper` component in v7 changed how it handles the `anchorEl` prop. It now requires a function that returns the element instead of the element directly:

**Old:**

```tsx
<Popper anchorEl={anchorRef.current} ... />
```

**New:**

```tsx
<Popper anchorEl={() => anchorRef.current as HTMLElement} ... />
```

This change was applied to `NavBarDesktopDropDown.tsx`.

#### 5. React 19 Ref Handling

React 19 deprecates `forwardRef` in favor of passing refs directly as props. Some components were updated accordingly:

**Old Pattern:**

```tsx
export const GenericButton = forwardRef(function GenericButton(
  props: GenericButtonProps,
  ref?: Ref<HTMLButtonElement>,
): ReactElement {
  return <button ref={ref} ... />
});
```

**New Pattern:**

```tsx
interface GenericButtonProps {
  // ... other props
  ref?: React.Ref<HTMLButtonElement>;
}

export function GenericButton(props: GenericButtonProps): ReactElement {
  /* eslint-disable react-hooks/refs */
  return <button ref={props.ref} ... />
  /* eslint-enable react-hooks/refs */
}
```

The ESLint disable comment is temporary while the `eslint-plugin-react-hooks` rule catches up with React 19's new patterns.

#### 6. Module-Level Config Initialization Fix

React 19's stricter initialization timing exposed a pattern where configuration objects were being read at module load time, before the configuration was fully loaded. This caused issues during SSG (Static Site Generation) and in test environments.

**Old Pattern (problematic):**

```tsx

export const statusObj: Record<CheckoffStatus, LicenseStatusMetadata> = {
  ACTIVE: { statusText: Config.licenseSearchTask.completedStatusText, ... },
  ...
};
```

**New Pattern (safe):**

```tsx
const getStatusObj = (): Record<CheckoffStatus, LicenseStatusMetadata> => {
  return {
    ACTIVE: { statusText: Config.licenseSearchTask.completedStatusText, ... },
    ...
  };
};

export const ChecklistTag = (props: Props): ReactElement => {
  const metadata = getStatusObj()[props.status];
  ...
};
```

This pattern was applied to:

- `ChecklistTag.tsx`
- `TaskProgressTagLookup.tsx`
- `XrayStatus.tsx`
- `dataFormErrorMapContext.ts`
- Various other files with module-level config reads

A new `defaultConfig.ts` file was added to `shared/src/contexts/` providing comprehensive fallback values for all configuration properties, ensuring `getMergedConfig()` always returns valid data even during SSG.

#### 7. React 19 State Update Timing

React 19's concurrent rendering and automatic batching can cause state updates in `useEffect` to run at different times than in React 18. Several components needed `setTimeout(..., 0)` wrappers to defer state updates to the next tick, preventing "Cannot update a component while rendering a different component" warnings:

```tsx
useEffect(() => {
  const timeoutId = setTimeout(() => {
    setFormValues(newValues);
  }, 0);
  return () => clearTimeout(timeoutId);
}, [dependencies]);
```

Components affected:

- `XrayStatus.tsx`
- `FilingsCalendarAsList.tsx`
- `LicenseTask.tsx`

#### 8. CSSTransition Replacement

The `react-transition-group` library uses `findDOMNode` internally, which is deprecated in React 19 and causes warnings. A new custom `CSSTransition` component (`web/src/components/transitions/CssTransition.tsx`) was created as a drop-in replacement:

```tsx
export const CSSTransition = (props: CSSTransitionProps): ReactElement | null => {
  // Pure CSS-based transitions using useState and useEffect
  // No findDOMNode usage
  // Same API as react-transition-group's CSSTransition
};
```

This component provides:

- Same `in`, `timeout`, `classNames`, `unmountOnExit` props
- Automatic class name transitions (`-enter`, `-enter-active`, `-exit`, `-exit-active`)
- Proper cleanup of timeouts on unmount

#### 9. rehype-react API Update

The `rehype-react` library v8 changed its API for creating React elements. Instead of passing `createElement` and `Fragment` directly, it now requires JSX runtime functions:

**Old:**

```tsx
.use(rehypeReact, {
  createElement: React.createElement,
  Fragment: React.Fragment,
  components: props.components,
})
```

**New:**

```tsx
import { Fragment, jsx, jsxs } from "react/jsx-runtime";

.use(rehypeReact, {
  jsx,
  jsxs,
  Fragment,
  components: props.components,
})
```

#### 10. Test Infrastructure Updates

React 19's concurrent rendering and automatic batching required significant test infrastructure changes:

**a. MessageChannel Polyfill:**
React 19 uses `MessageChannel` for scheduling state updates, which isn't available in jsdom. A polyfill was added to `setupTests.js`:

```javascript
global.MessageChannel = class MessageChannel {
  constructor() {
    this.port1 = { postMessage: ..., onmessage: null, close: () => {} };
    this.port2 = { postMessage: ..., onmessage: null, close: () => {} };
  }
};
```

**b. matchMedia Enhancement:**
The `matchMedia` mock was enhanced to include modern API methods required by MUI v7:

```javascript
window.matchMedia = function () {
  return {
    matches: false,
    media: "",
    onchange: null,
    addListener: function () {},      // Legacy
    removeListener: function () {},   // Legacy
    addEventListener: function () {}, // Modern
    removeEventListener: function () {}, // Modern
    dispatchEvent: function () { return true; },
  };
};
```

**c. Test Timeout Increases:**

- Global Jest timeout increased from 10s to 30s
- Individual `waitForElementToBeRemoved` calls increased to 120s for parallel CI execution
- `maxWorkers: "50%"` added to `jest.shared.ts` to reduce resource contention

**d. Async Test Pattern Migration:**
Tests were migrated from synchronous `fireEvent` patterns to async `userEvent` patterns:

**Old:**

```tsx
fireEvent.click(screen.getByTestId("next"));
```

**New:**

```tsx
await userEvent.click(screen.getByTestId("next"));
```

**Old:**

```tsx
const button = screen.getByRole("button");
```

**New:**

```tsx
const button = await screen.findByRole("button");
```

**e. New Test Helpers:**
A new `accessible-queries.ts` helper module was added with async-safe, accessible query functions:

```typescript
export const findButton = async (name: string | RegExp): Promise<HTMLElement>;
export const findLink = async (name: string | RegExp): Promise<HTMLElement>;
export const findInput = async (labelOrPlaceholder: string): Promise<HTMLElement>;
export const selectOptionByText = async (label: string, optionText: string): Promise<void>;
// ... and more
```

**f. Config Mocking:**
A new `useMockConfig()` function was added to ensure tests have access to configuration during the React 19 initialization phase:

```typescript
export const useMockConfig = (): void => {
  jest.spyOn(useConfigModule, "useConfig").mockReturnValue({ Config: getMergedConfig() });
};
```

**g. Timer Strategy:**
Tests were migrated from `jest.useFakeTimers()` to `jest.useRealTimers()` because React 19's async rendering conflicts with fake timers when using `waitFor` and `findBy*` queries.

#### 11. Docker Build Optimization

The `WebApp.Dockerfile` was completely restructured for better layer caching and optimized production builds:

**Multi-stage Build:**

1. `deps` stage: Installs all dependencies including devDependencies needed for build
2. `prod-deps` stage: Installs only production dependencies
3. `builder` stage: Compiles the Next.js application with `output: standalone`
4. `runner` stage: Minimal production image with standalone output

**Key Changes:**

- Uses Next.js standalone output for smaller production images
- Separates build dependencies from runtime dependencies
- Environment variables are now provided via `.env.production` file created during CI
- Uses `node server.js` directly instead of `yarn start` for faster startup

**New GitHub Action Step:**
The `package-webapp-docker` action now creates a `.env.production` file with all necessary environment variables before the Docker build.

#### 12. App-Level ConfigContext

The `_app.tsx` component now wraps the application with a `ConfigContext.Provider` to ensure configuration is available throughout the component tree during React 19's stricter initialization:

```tsx
return (
  <ConfigContext.Provider value={{ config, setOverrides: () => {} }}>
    {/* ... rest of app ... */}
  </ConfigContext.Provider>
);
```

This also required adding conditional checks when computing SEO metadata to avoid accessing undefined config during SSG.

#### Local Development

1. `yarn install` at the monorepo root
2. `yarn services:down` and `yarn start:dev`
3. Verify the application loads without console errors
4. Navigate through major user flows:
   - Landing page and starter kits
   - Onboarding flow (starting, owning, foreign business paths)
   - Profile editing (all tabs: info, permits, numbers, docs, notes)
   - Dashboard with roadmap and calendar
   - Task pages (license search, formations, cannabis priority status)
   - Tax clearance certificate flow
   - Emergency trip permit flow

#### Date Pickers

1. Navigate to business formation flow
2. Test date of formation picker (DateOfFormation.tsx)
3. Test business start date picker (FormationDate.tsx)
4. Navigate to ABC emergency trip permit flow
5. Test permit date picker (EmergencyTripPermitDatePicker.tsx)
6. Verify date selection, keyboard navigation, and error states

#### Dropdown Menus

1. Test navigation bar account dropdown (desktop)
2. Verify dropdown opens/closes properly
3. Test keyboard navigation within dropdown
4. Verify click-away closes dropdown

#### Transitions

1. Navigate to pages with animated sidebars/modals
2. Verify enter/exit animations play smoothly
3. Test rapid open/close interactions

#### Test Suite

1. `yarn test` at the monorepo root
2. All tests should pass (some may have slightly longer execution times due to async changes)
3. Run individual test files if needed: `yarn test web/test/pages/onboarding/onboarding-shared.test.tsx`

#### Build Verification

1. `cd web && yarn build`
2. Verify build completes without errors
3. `yarn start` and verify production build works

#### Docker Build (if applicable)

1. Build Docker image using the new Dockerfile
2. Verify container starts and serves the application
3. Test healthcheck endpoint at `/healthz`

#### Storybook

1. `cd web && yarn storybook`
2. Verify stories render without errors
3. Check component interactions in stories

#### Functional Testing

1. **Onboarding Flow**

- [x] Complete the full onboarding process for a new business
- [x] Test all business structure types (LLC, Corporation, Sole Proprietorship, etc.)
- [x] Verify industry selection and sector modal functionality
- [x] Test foreign business flow

1. **Dashboard & Navigation**

- [x] Navigate through all dashboard tabs (Roadmap, For You, Calendar)
- [x] Test mobile navigation menu and quick links slide-out
- [x] Verify desktop dropdown menus work correctly
- [x] Test search functionality

2. **Tasks & Forms**

- [x] Complete business formation flow (all steps)
- [ ] Test license applications (Cannabis, Cigarette, CRTK, etc.)
  - [X] Cannabis
  - [ ] Cigarette
    - Dependent upon Tax Filing API
  - [X] CRTK
  - [X] Cosmetology Shop
- [x] Verify date pickers work correctly (Formation Date, Sales Start Date, Emergency Trip Permit Date)
- [ ] Test tax-related tasks (Tax ID, Tax Clearance Certificate)
  - Tax Filing not working as of testing on 1/23
- [x] Test registration tasks (Elevator, Housing, Hotel/Motel, X-ray)
- [x] Verify NAICS code input and search
- [x] Test Cannabis Priority Status application

3. **Profile & Account**

- [x] Update profile information
- [x] Modify business details
- [x] Test "Personalize Your Tasks" tab
- [x] Verify contact information updates
- [x] Test account setup flow

4. **Forms & Validation**

- [x] Test all form validations (required fields, format validation, business rules)
- [x] Verify error messages display correctly
- [x] Test autosave functionality
- [x] Verify form field focusing and tab order
- [x] Test GenericTextField with various input types (numeric, text, masked)

5. **Animations & Transitions**

- [x] Verify modal open/close animations work smoothly
- [x] Test sidebar slide-in/slide-out animations
- [x] Check alert/snackbar transitions
- [ ] Verify tooltip animations

6. **CMS & Admin**

- [ ] Access CMS at `/mgmt/cms`
- [ ] Verify Decap CMS loads and functions correctly
- [ ] Test content preview functionality

7. **React Email**
- [x] Validate that running and previewing the React Email template still works

#### Visual & Accessibility Testing

- [x] Test responsiveness across desktop, tablet, and mobile viewports
- [x] Verify no visual regressions in MUI components (buttons, dropdowns, date pickers)
- [ ] Run accessibility audit (lighthouse, axe)
- [x] Test keyboard navigation throughout the application
- [x] Verify screen reader compatibility

#### Build & Performance Testing

- [x] Run `yarn build` and verify successful build
- [x] Check that vendor assets are copied correctly to `public/vendor`
- [x] Verify NJWDS CSS patch applied successfully (no pseudo-element errors)
- [x] Run full test suite: `./scripts/pre-commit.sh`
- [ ] Run Storybook: `yarn storybook`
- [ ] Check bundle size hasn't increased significantly
- [ ] Verify no console errors or warnings in production build

#### Storybook Testing

- [ ] Launch Storybook and verify all stories load
- [ ] Test addon functionality (a11y, designs, interactions)
- [ ] Verify component documentation renders correctly

#### Breaking Changes in Dependencies

**MUI v7:**

- The `@mui/styles` package is now on v6.x while other MUI packages are on v7.x. This is expected; `@mui/styles` (JSS styling) is being phased out in favor of Pigment CSS-based styling.
- If you encounter styling issues, check that component prop APIs haven't changed.

**Storybook v10:**

- Some story configurations may need updates for the new Storybook architecture.
- Addon APIs have changed; refer to Storybook migration guides if extending stories.

#### Known Deprecation Warnings

The following warnings may appear in development and are expected during the React 19 transition period:

1. `findDOMNode is deprecated` - From third-party libraries still using legacy patterns
2. `ReactDOMTestUtils.act is deprecated` - Testing library compatibility messages
3. `A props object containing a "key" prop is being spread` - From MUI components

These warnings are suppressed in the test environment via `setupTests.js` but may appear during development.

#### Performance Considerations

- React 19's concurrent rendering may cause components to render more or fewer times than expected
- The `setTimeout(..., 0)` pattern added to some components adds minimal latency (~0-4ms) but ensures correct behavior
- Test execution may be slightly slower due to increased timeouts and async patterns

#### Future Improvements

1. **Remove `forwardRef` migration:** Once all consumers are updated, remove remaining `forwardRef` wrappers
2. **Remove suppressions:** As libraries update, remove deprecation warning suppressions from `setupTests.js`
3. **Adopt React 19 features:** Consider adopting React Server Components, `use()` hook, and other React 19 features where beneficial
4. **MUI v7 migration completion:** Fully migrate away from `@mui/styles` to Emotion-based styling
5. **Update NJWDS:** Once the upstream package fixes the CSS pseudo-element syntax, the patch script can be removed

#### Key Files Changed

| Category | Files |
| ---------- | ------- |
| Dependencies | `package.json`, `web/package.json`, `yarn.lock` |
| Build Config | `web/next.config.ts`, `jest.shared.ts` |
| Build Scripts | `web/scripts/copy-vendor-assets.ts`, `web/scripts/patch-njwds-css.ts` |
| Docker | `WebApp.Dockerfile`, `.github/actions/package-webapp-docker/action.yml` |
| Test Setup | `web/setupTests.js`, `web/test/helpers/accessible-queries.ts`, `web/test/mock/mockUseConfig.ts` |
| Components | 50+ component files (date pickers, buttons, navigation, etc.) |
| Test Files | 30+ test files with async pattern updates |
| Contexts | `shared/src/contexts/configContext.ts`, `shared/src/contexts/defaultConfig.ts` |

## Code author checklist


web/scripts/patch-njwds-css.ts
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env tsx
+
+/**
+ * Patches @newjersey/njwds CSS to fix invalid pseudo-element syntax
+ * that causes Turbopack build failures.
+ *
+ * Issue: ::before--tile is invalid CSS (should be ::before)
+ * This is a workaround until the upstream package is fixed.
+ */
+
```
seidior: Open question: do we upgrade njwds as part of this? Unclear if the newer version of NJWDS fixes this issue with this peculiar pseudo-selector.

web/src/components/crtk/crtkSearchResult.tsx
```diff
@@ -21,9 +21,25 @@ interface Props {
...
+        <li>
+          Email:{" "}
+          <a href="mailto:rtk@dep.nj.gov" className="text-underline">
+            rtk@dep.nj.gov
+          </a>
+        </li>
+      </ul>
+    </>
+  );
+};
```
seidior: Linting error: no nested React components.

web/src/components/dashboard/BusinessStructurePrompt.test.tsx
```diff
@@ -55,7 +52,7 @@ describe("<BusinessStructurePrompt />", () => {
 
   it("routes user to business structure task page on button click", () => {
     render(<BusinessStructurePrompt />);
-    fireEvent.click(screen.getByText(Config.businessStructurePrompt.buttonText));
+    fireEvent.click(screen.getByTestId("business-structure-prompt-button"));
```
seidior: Because Config could be empty or it could have the value, it's better that we use `getByTestId`, although an async method would be better.

web/src/components/data-fields/tax-id/TaxId.tsx
```diff
@@ -35,7 +35,7 @@ export const TaxId = (props: Props): ReactElement => {
     return "SPLIT";
   };
 
-  const initialType = useRef<"FULL" | "SPLIT">(getFieldType());
+  const initialType = getFieldType();
```
seidior: I *think* this is just a simplification, that a ref wasn't needed here to make this component work, but this would need testing. I don't see anything in [the React 19 Upgrade Guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument) that says this behavior is discouraged.

web/src/components/navbar/desktop/NavBarDesktopVerticalLineDivider.tsx
```diff
@@ -1,4 +1,4 @@
-import { ReactElement } from "rehype-react/lib";
+import { ReactElement } from "react";
```
seidior: Campgrounding. Not sure how this accident got past code review, but regardless, it's fixed now.

web/src/components/tasks/anytime-action/tax-clearance-certificate/TaxClearanceSteps.tsx
```diff
@@ -119,7 +123,7 @@ export const TaxClearanceSteps = (props: Props): ReactElement => {
       {props.certificatePdfBlob ? (
         <Download
           certificatePdfBlob={props.certificatePdfBlob}
-          downloadFilename={`Tax Clearance Certificate - ${Date.now()}`}
+          downloadFilename={`Tax Clearance Certificate - ${downloadTimestamp}`}
```
seidior: I'm guessing this isn't really a React 19-relevant thing. Instead, `Date.now()` gets evaluated on most recent render rather than at the current time that the user clicks the download button.

Instead here, we're using a state variable to call a function that returns Date.now() when it's most relevant, which helps us two ways:
 
- If we ever move to Server Components, it won't cause errors
- This gives a predictable pattern that the value is only evaluated when needed and not evaluated multiple times.

web/src/components/PureMarkdownContent.tsx
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import React, { ReactElement } from "react";
+import { ReactElement } from "react";
+import { Fragment, jsx, jsxs } from "react/jsx-runtime";
```
seidior: The missing fixes that enable the `unified` upgrades.

web/next.config.ts
```diff
@@ -43,55 +53,10 @@ module.exports = withBundleAnalyzer({
...
-        ],
-      }),
-    );
-
-    return config;
-  },
   experimental: {
     largePageDataBytes: 1.28 * 1024 * 1024,
   },
+  output: "standalone",
```
seidior: Without [output type of "standalone"](https://nextjs.org/docs/pages/api-reference/config/next-config-js/output), `turbopack` wouldn't work. It's not expressly needed as part of the React 19 upgrade, but we wouldn't get the huge performance wins on `yarn start:dev` without it, and fixing these React issues was the primary reason holding us back from switching.

web/package.json
```diff
@@ -48,82 +50,84 @@
...
     "js-yaml": "4.1.1",
     "lodash": "4.17.21",
     "next": "16.1.1",
     "next-seo": "6.8.0",
-    "react": "18.3.1",
+    "react": "19.2.3",
     "react-clamp-lines": "3.0.3",
-    "react-dom": "18.3.1",
-    "react-slick": "0.30.0",
-    "react-transition-group": "4.4.5",
```
seidior: This library is not compatible with React 19, so it's been removed, and custom CSSTransition code has been written.

web/package.json
```diff
@@ -167,13 +171,13 @@
...
     "sass-loader": "14.2.1",
-    "storybook": "7.6.21",
+    "storybook": "10.1.11",
     "style-loader": "3.3.4",
     "ts-jest": "29.4.5",
     "ts-loader": "9.5.4",
     "ts-node": "10.9.2",
     "typescript": "5.9.3",
-    "webpack": "5.103.0"
+    "webpack": "5.104.1"
```
seidior: Strangely, we still need this as a dependency. Jest maybe?

seidior (commented):
Self-review.

api/src/functions/messagingService/reactEmail/package.json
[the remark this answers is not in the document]
hinzed1127: For the sake of thoroughness, I added it as a to-do at the bottom of the "steps to test" section

.github/actions/package-webapp-docker/action.yml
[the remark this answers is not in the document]
hinzed1127: Is there any way to add type security to this? If we add or remove envirment variables, it would be ideal if this somehow failed and shouted at us to update them here, as well.

shared/src/contexts/configContext.ts
```diff
@@ -80,273 +80,254 @@ import * as UnsupportedNavigatorUserPage from "../../../content/src/fieldConfig/
...
+  const keys = Object.keys(config as Record<string, unknown>);
+  if (keys.length === 1 && keys[0] === "default") {
+    return (config as Record<string, unknown>).default;
+  }
+  return config;
+};
+
+export type ConfigType = typeof AbcEmergencyTripPermit &
+  typeof AccountSetup &
+  typeof anytimeActionReinstatementAndLicenseCalendarEventStatusDefaults &
```
hinzed1127: this as the single lowercase config item is driving me crazy

web/test/helpers/helpers-renderers.tsx
```diff
@@ -123,3 +124,12 @@ export const withRoadmap = (params: {
...
 };
+
+export const withConfig = (subject: ReactElement): ReactElement => {
+  const Config = getMergedConfig();
+  return (
+    <ConfigContext.Provider value={{ config: Config, setOverrides: jest.fn() }}>
+      {subject}
+    </ConfigContext.Provider>
+  );
+};
```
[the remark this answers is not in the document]
seidior: To address your comments:

React 19 is stricter about requiring context providers when consumers exist. We should only wrap tests where the context is actually consumed to avoid unnecessary test setup.

  - `Industry.test.tsx`: The `Industry` component uses `useConfig()` hook
  - `taskUrlSlug.test.tsx`: The `TaskPage` component uses `useConfig()` hook 
  - `helpers-formation.tsx`: The `BusinessFormation` components use `useConfig()` in multiple places (paginator, success page, etc)

However:
  - `filingUrlSlug.test.tsx`: The `FilingPage` component does not use `useConfig()` or consume `ConfigContext` at all, so that's not needed.
  
 Where I was confused, not knowing the codebase super well, is the reason that both `useConfig()` and `getMergedConfig()` exist. Here's the explanation I got:

  - `getMergedConfig()`: Can be called anywhere (server, client, utils, tests)
  - `useConfig()`: React-specific, context-aware, potentially dynamic (like with the CMS preview pane)

The React 19 upgrade required the `withConfig` wrapper in tests because components using `useConfig()` now strictly require the `ConfigContext.Provider` in their component tree, whereas before React 19, React was more lenient about missing providers.

web/src/components/dashboard/SidebarCardFormationNudge.tsx
```diff
@@ -15,6 +15,8 @@ type Props = {
 
 export const SidebarCardFormationNudge = (props: Props): ReactElement => {
   const [modalOpen, setModalOpen] = useState<boolean>(false);
+  // React 19: Counter to force FormationDateModal remount when opened
+  const [formationModalResetKey, setFormationModalResetKey] = useState(0);
```
ashlsun: leaving a note to return, not sure what is happening here

shared/src/contexts/defaultConfig.ts
```diff
@@ -0,0 +1,2637 @@
+/**
+ * Default configuration object with safe fallback values
+ * This ensures getMergedConfig() always returns a valid ConfigType
+ * even during SSG when JSON imports may not be fully loaded
+ */
+export const getDefaultConfig = (): Record<string, unknown> => ({
```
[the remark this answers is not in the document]
hinzed1127: > every once in a while, we should set this to `ConfigType` to see if we're missing any fields, since in `configContext.ts` we're explicitly not checking it (`as unknown as ConfigType`).

[boulder] This feels like a smell to me. If that kind of check isn't automated in a way to stay on top of type safety, there's no accountability in making sure this file stays accurate. 

I'm still not entirely following this explanation:

> The way that both Turbopack and React 19 work, Config will load async. So in order not to get build-time errors or errors with our tests, we need to load default values for our Config object, then load in all of our actual values.

Is there not a way to have these take on fallback values of empty strings more generally without necessitating explicitly defining this huge object?

shared/src/static/loadAnytimeActionTasks.ts
```diff
@@ -5,28 +5,43 @@ import { convertAnytimeActionTaskMd } from "../utils/tasksMarkdownReader";
...
-  "anytime-action-tasks",
-);
-const anytimeActionsTaskDirectoryTest = path.join(
-  process.cwd(),
-  "content",
-  "src",
-  "anytime-action-tasks",
-);
+
+// Helper to find content directory from different test contexts
```
hinzed1127: superfluous comment

web/cypress/e2e/group_1_onboarding/onboarding-as-foreign-business.spec.ts
hinzed1127: The comments adding context for adding `cy.wait` + disabling the rule are helpful, but do we actually have context on what in React 19 make this necessary? I see the "Wait for state updates and validation to complete," but that doesn't explain to me how that's any different than previously

web/cypress/e2e/group_5/taxclearance.spec.ts
```diff
@@ -40,12 +50,26 @@ describe("Tax Clearance [feature] [all] [group5]", () => {
     cy.get('[data-testid="anytimeActionSearch"]').click();
     cy.get('[data-testid="option"]').contains("Get a Tax Clearance Certificate").click();
 
-    // Inside AA task
+    // Inside AA task - Requirements screen
     cy.get('[data-testid="cta-primary-1"]').click();
+
+    // Now on Check Eligibility screen
```
hinzed1127: The code from heres feels self-explanatory, doesn't need the comments

web/cypress/support/helpers/helpers.ts
hinzed1127: most of the React 19 comments here give a little more context, but some prompt the same questions I commented about previously

web/scripts/patch-njwds-css.ts
hinzed1127: I would favor not adding any more scope creep to the breadth of upgrades and take that on in the future

web/src/components/crtk/crtkForm.tsx
```diff
@@ -87,14 +86,17 @@ export const CRTKStatus = (props: Props): ReactElement => {
...
     if (business?.formationData?.formationFormData) {
-      setFormValues({
-        businessName: business?.profileData.businessName || "",
-        businessStreetAddress: business.formationData.formationFormData.addressLine1,
-        city: business.formationData.formationFormData.addressMunicipality?.displayName || "",
-        state: "NJ",
-        zip: business.formationData.formationFormData.addressZipCode,
-        ein: "",
-      });
+      const timeoutId = setTimeout(() => {
```
hinzed1127: Is this `setTimeout` actually needed? This doesn't even appear to be for updates, just initialization purposes

web/src/components/crtk/crtkPage.test.tsx
```diff
@@ -330,33 +333,35 @@ describe("<CRTKPage />", () => {
...
   describe("result display", () => {
-    it("displays FOUND status with facility details", () => {
+    it("displays FOUND status with facility details", async () => {
       renderWithBusinessData({
         crtkData: generateCRTKData(),
       });
 
-      expect(screen.getByText("Business Found")).toBeInTheDocument();
+      // React 19: Wait for async rendering to complete
+      expect(await screen.findByText("Business Found")).toBeInTheDocument();
```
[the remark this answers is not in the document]
hinzed1127: - [ ] create a chore or refactor to capture this update work

web/src/components/tasks/cigarette-license/LicenseeInfo.test.tsx
```diff
@@ -22,11 +22,23 @@ import {
...
+ * due to a known React 19 + MUI + @testing-library/user-event bug where userEvent.type() only
+ * enters the first character in controlled numeric inputs.
+ *
+ * See: https://github.com/testing-library/user-event/issues/1286
+ * "JavaScript heap out of memory" - userEvent.type() causes infinite loops with React 19 + MUI number inputs
+ *
+ * Once this upstream issue is fixed, these tests should be converted back to userEvent.type()
+ * for more realistic user interaction simulation.
+ */
+
```
hinzed1127: - [ ] create chore for tracking and eventually updating this

hinzed1127 (commented):
A few comments on things mentioned in the description:

> **c. Test Timeout Increases:**
> 
> - Global Jest timeout increased from 10s to 30s
>     
> - Individual `waitForElementToBeRemoved` calls increased to 120s for parallel CI execution
>     
> - `maxWorkers: "50%"` added to `jest.shared.ts` to reduce resource contention

I see `jest.setTimeout(30000);`, and `waitForElementToBeRemoved` calls removed, but I don't see the `maxWorkers` change. Am I missing something ?

> **d. Async Test Pattern Migration:**  
Tests were migrated from synchronous `fireEvent` patterns to async `userEvent` patterns:

This means we can link (and close once merged) [Chore 13019: Replace usages of fireEvent with userEvent in tests](https://dev.azure.com/njinnovation/Business%20First%20Stop/_workitems/edit/13019).

> 7. React 19 State Update Timing
> React 19's concurrent rendering and automatic batching can cause state updates in useEffect to run at different times 
> than in React 18. Several components needed setTimeout(..., 0) wrappers to defer state updates to the next tick, 
> preventing "Cannot update a component while rendering a different component" warnings:
>
> ```jsx
> useEffect(() => {
>  const timeoutId = setTimeout(() => {
>    setFormValues(newValues);
>  }, 0);
>  return () => clearTimeout(timeoutId);
> }, [dependencies]);
> ```


I need to research this more deeply, but my gut reaction is this feels like an antipattern and we should see if there's a better approach. This was the most pervasive change across many files that felt like a smell to me

---

My request for PRs of this scope in the future is, whenever possible, breaking this up into distinct commits for each chunk of the work. Whether they get squashed before merging is a separate discussion, but for the sake of review, it would be *much* easier to view things in chunks, such as:
- any results from running a codemod are their own commit
- most if not all of the 12 parts in the description breakdown could be their own commit
- any updates/fixes made in the process of review
- functionality changes such as the Tax ID input change
- etc

Given the scope of the work, that would make it much easier to review the various changes in more natural groupings, rather than a single diff of 176 files.

---

I left multiple comments on it within the review, as well as responded to your statement about it, but this has way too many extraneous comments right now. Since this is a known tendency of AI-generated code, we should reach a consensus on what the appropriate amount of comments are and share methods on the most effective techniques to limit those comments, the goal of that being it needs less attention in code review

seidior:
Closing in favor of #12920.

PR #12453, opened 2026-01-30
fix: [AB#17340] automatic biz name availability search should use formationFormData.businessName

jphechter:
## Description

Resolves an error caused by checking the business name availability of an empty string.

### Ticket

This pull request resolves [#17340](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17340).

### Approach

When you make a request to the NICUSA Business Name Search API with an empty string, the API responds with a 200 code and an empty body. The error message we're seeing on the backend is a result of us trying to reference a key inside this object that does not exist.

I addressed this in 2 steps:
1. Do not execute a search if the business name is an empty string.
1. Use the business name stored in `formationFormData` instead of `profileData`.
    - This is the field that would be used by the frontend, and is the one that would be used on the business formation form submission.

### Steps to Test

I listed repro steps in the linked ticket. Follow these. Assert that errors are not logged, and multiple GET requests do not trigger an alert in the #bizx-alerts-bot Slack channel.

### Notes

I documented the diagnostic process [in Slack](https://njcio.slack.com/archives/C04J8MTJNS2/p1769630480380619)

There's a case that I should have also handled for the NICUSA API returning an empty object, because that's what was triggering the error in the first place. I opted not to do this. Here's my reasoning why I didn't:
- The API should never do this if we're supplying a non-empty string
- If the API does do this, I'd like to be aware. The log messages surrounding this error may give us more information about a different edge case that we may want to handle differently.
- Handling for an empty object would have been more disruptive to the logic of the API client and would have been a larger lift.
  - I'd have to modify the function declaration and touch more files. This fix has a smaller scope and should be plenty adequate for us.

## Code author checklist

- [X] I have added the `request-reviewer` tag on github to request reviews

### DIFF
--- a/api/src/api/userRouter.test.ts
+++ b/api/src/api/userRouter.test.ts
@@ -6,6 +6,7 @@ import { getCurrentDate, parseDate } from "@shared/dateHelpers";
 import { getCurrentBusiness } from "@shared/domain-logic/getCurrentBusiness";
 import { modifyCurrentBusiness } from "@shared/domain-logic/modifyCurrentBusiness";
 import { createEmptyFormationFormData, emptyAddressData } from "@shared/formationData";
+import { randomInt } from "@shared/intHelpers";
 import { LicenseName } from "@shared/license";
 import {
   generateBusiness,
@@ -593,6 +594,51 @@ describe("userRouter", () => {
       });
 
       describe("when businessPersona is 'STARTING'", () => {
+        it("checks name availability using name from formationFormData", async () => {
+          const userData = generateUserDataForBusiness(
+            generateBusiness({
+              profileData: generateStartingProfileData({
+                businessName: "Business Name in Profile",
+              }),
+              formationData: generateFormationData({
+                businessNameAvailability: generateBusinessNameAvailability({
+                  status: "AVAILABLE",
+                  lastUpdatedTimeStamp: sixtyOneMinutesAgo,
+                }),
+                formationFormData: generateFormationFormData({
+                  businessName: "Business Name in Formation Form Data",
+                }),
+              }),
+            }),
+          );
+          stubUnifiedDataClient.get.mockResolvedValue(userData);
+
+          await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
+          expect(stubTimeStampBusinessSearch.search).toHaveBeenCalledWith(
+            "Business Name in Formation Form Data",
+          );
+        });
+
+        it("does not check name availability if formationFormData.businessName is an empty string", async () => {
+          const userData = generateUserDataForBusiness(
+            generateBusiness({
+              formationData: generateFormationData({
+                businessNameAvailability: generateBusinessNameAvailability({
+                  status: "AVAILABLE",
+                  lastUpdatedTimeStamp: sixtyOneMinutesAgo,
+                }),
+                formationFormData: generateFormationFormData({
+                  businessName: "",
+                }),
+              }),
+            }),
+          );
+          stubUnifiedDataClient.get.mockResolvedValue(userData);
+
+          await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
+          expect(stubTimeStampBusinessSearch.search).not.toHaveBeenCalled();
+        });
+
         it("does not update business name search if businessNameAvailability is undefined", async () => {
           const userData = generateUserDataForBusiness(
             generateBusiness({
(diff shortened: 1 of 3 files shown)


williamavasquez (approved):

seidior (approved):
# Code Review: Business Name Availability Search Fix (PR #12453 / AB#17340)

## Overview

This review analyzes the approach taken in commit `6df894212` which fixes the automatic business name availability search to use the correct data source.

## Summary of Changes

**The Fix**: Changed the automatic name availability search from using `profileData.businessName` to `formationData.formationFormData.businessName` for non-DBA cases.

**New Guard**: Added `if (nameToSearch === "") return userData;` to prevent searching with empty strings.

## Data Structure Analysis

Based on migration files and type definitions, business names are stored in **three distinct locations**:

### 1. `profileData.businessName` (line 102, v109)
- **Purpose**: Official/registered business name
- **Initial value**: `""` (empty string)
- **Usage**: Display, tax filing, general business metadata
- **Updated**: After formation completes or manual entry by user

### 2. `formationData.formationFormData.businessName` (line 260, v109)
- **Purpose**: Working business name during the formation process
- **Initial value**: `""` (empty string)
- **Usage**: Formation workflows, name availability searches, formation submissions
- **Updated**: As user types in formation form

### 3. `profileData.nexusDbaName` (line 121, v109)
- **Purpose**: DBA (Doing Business As) name for foreign/nexus businesses
- **Initial value**: `""` (empty string)
- **Usage**: Foreign business operations when main name is unavailable

## Migration History Shows Intent

Key migrations demonstrate the intended separation:

- **v108**: Added `businessNameSearch` to `formationData` (not profileData)
- **v109**: Refactored to `businessNameAvailability` with timestamp at formation level
- **v116**: Added separate `dbaBusinessNameAvailability` for nexus businesses
- **v177**: Added `businessNameConfirmation` field to formationFormData

The structure clearly indicates that **name availability checking belongs to the formation process**, not the profile.

## Current Approach Assessment

### Decision Logic (userRouter.ts:418-423)
```typescript
const needsDba = determineIfNexusDbaNameNeeded(currentBusiness);
const nameToSearch = needsDba
  ? currentBusiness.profileData.nexusDbaName           // For NEXUS businesses
  : currentBusiness.formationData.formationFormData.businessName; // For STARTING businesses

if (nameToSearch === "") return userData; // Skip search if empty
```

### Why This Approach is CORRECT

**1. Single Source of Truth Per Context**
- During formation: `formationFormData.businessName` is authoritative
- For nexus DBA: `nexusDbaName` is authoritative
- No ambiguity about which name is being validated

**2. Empty String Means "Not Ready"**
- If `formationFormData.businessName === ""`, user hasn't entered a name yet
- Searching `profileData.businessName` would validate the wrong name
- Better to skip the search than validate incorrect data

**3. Data Flow is Unidirectional**
- Formation data flows TO profile data (not the reverse)
- `formationFormData.businessName` → (on completion) → `profileData.businessName`
- Fallback logic would violate this flow

**4. Temporal Separation**
- `profileData.businessName` represents the past (what was registered)
- `formationFormData.businessName` represents the present (what's being worked on)
- Mixing them creates temporal confusion

## Edge Cases Considered

### Edge Case 1: User Had a Previous Business Name
**Scenario**:
- `profileData.businessName = "Old Company LLC"`
- User starts new formation
- `formationFormData.businessName = ""`
- Should we fall back to "Old Company LLC"?

**Answer**: NO
- They're working on a NEW formation
- Empty string means they haven't decided yet
- Validating the old name would be misleading

### Edge Case 2: Partial Formation Completion
**Scenario**:
- User fills `formationFormData.businessName = "New Name Inc"`
- Availability checked and stored
- User leaves and comes back, clears the field to `""`
- Should we fall back to `profileData.businessName`?

**Answer**: NO
- Clearing indicates intent to change
- Previous availability check is stale (will be >1 hour old eventually)
- System will re-check when they enter a new name

### Edge Case 3: Multiple Concurrent Businesses
**Scenario** (from v151+ multi-business support):
- User has multiple businesses
- Business A: `profileData.businessName = "Company A"`
- Business B: In formation with `formationFormData.businessName = ""`
- Should Business B fall back to its own `profileData.businessName`?

**Answer**: NO
- Same reasoning: empty means not ready
- Each business's formation data is independent

## What About Display Logic?

**Important Distinction**: The navbar display logic (`getNavBarBusinessTitle.ts`) DOES use fallback logic:

```typescript
// For display purposes (lines 22-39)
if (nexusDbaName needed and filled) → use nexusDbaName
else if (hasTradeName) → use tradeName
else → use businessName
```

This is appropriate because:
- Display needs to show SOMETHING
- Multiple display names (trade name, DBA, business name) are valid simultaneously
- This is about presentation, not validation

**Validation is different**:
- Availability checking should validate the SPECIFIC name being proposed
- Fallback would validate the wrong name
- Better to skip than to be wrong

## Potential Issues with Fallback Approach

If we were to implement fallback logic (NOT RECOMMENDED), problems would include:

1. **Misleading Results**
   - User enters nothing in formation
   - System searches old/different name from profile
   - User thinks their (non-existent) name is available/unavailable

2. **Race Conditions**
   - Which timestamp do we use? FormationData or ProfileData?
   - How do we know which availability status corresponds to which name?

3. **Complex Debugging**
   - "Why is it searching for X when I typed Y?"
   - Hard to trace which code path selected which name

4. **Test Complexity**
   - Need to test all permutations of empty/filled locations
   - Need to test priority ordering
   - Much higher cognitive load

## Empty String vs Undefined

**Current Behavior**:
- All three business name fields are typed as `string` with initial value `""`
- Empty string explicitly means "no value yet"
- The check `if (nameToSearch === "") return userData;` is appropriate

**Why Not Use Undefined?**:
- Would require type changes: `string | undefined`
- Would affect all downstream code
- Empty string is idiomatic for "unfilled form field"
- Existing codebase convention

## Test Coverage

The PR adds appropriate tests:

1. ✅ **Line 597**: Verifies correct source is used (formationFormData)
2. ✅ **Line 622**: Verifies empty string in formationFormData skips search
3. ✅ **Line 789**: Verifies empty string in nexusDbaName skips search

**Missing Test** (not critical but nice to have):
- Test with `formationFormData.businessName = ""` AND `profileData.businessName = "Something"`
- Verify it does NOT search for "Something"
- Would document that fallback is intentionally NOT implemented

## Recommendations

### ✅ APPROVE Current Approach
The implementation is correct and well-reasoned:
- Uses the right data source for each context
- Handles empty strings appropriately
- Aligns with data structure intent
- Prevents misleading search results

### 📝 Optional Enhancements (Not Blocking)

1. **Add Clarifying Comment**
   ```typescript
   // Don't fall back to profileData.businessName - if the formation form
   // is empty, it means the user hasn't entered a name yet. Searching
   // profileData would validate a different name than what they're working on.
   if (nameToSearch === "") return userData;
   ```

2. **Add Test for Non-Fallback Behavior**
   ```typescript
   it("does not fall back to profileData.businessName when formationFormData.businessName is empty", async () => {
     const userData = generateUserDataForBusiness(
       generateBusiness({
         profileData: generateProfileData({
           businessName: "Name in Profile Data",
         }),
         formationData: generateFormationData({
           formationFormData: generateFormationFormData({
             businessName: "", // Empty in formation
           }),
           businessNameAvailability: generateBusinessNameAvailability({
             lastUpdatedTimeStamp: sixtyOneMinutesAgo,
           }),
         }),
       }),
     );

     await request(app).get(`/users/123`).set("Authorization", "Bearer user-123-token");
     expect(stubTimeStampBusinessSearch.search).not.toHaveBeenCalled();
   });
   ```

3. **Consider Logging** (for debugging in production)
   ```typescript
   if (nameToSearch === "") {
     logWriter.LogInfo("Skipping name availability search - no name entered yet");
     return userData;
   }
   ```

## Conclusion

✅ **The approach in PR #12453 is sound and should be merged as-is** (with optional enhancements if desired).

The fix correctly identifies that:
1. Formation name availability should use formation data, not profile data
2. Empty strings should skip the search, not trigger fallbacks
3. Different business name locations serve different purposes

No architectural changes are needed. The data structure is well-designed and the fix aligns with its intent.

PR #12627, opened 2026-03-10
feat: [AB#16890] real-time GovDelivery Communications Cloud subscribe/unsubscribe

hinzed1127:
## Description

This integrates with the [GovDelivery Communications Cloud API](https://support.granicus.com/customersupport/s/article-group/a8C4w000000XcOdEAK/communications-govdelivery-api?language=en_US). When users make updates to their Contact & Notifications settings (via `/profile?tab=contact-notifications`), changes to a user's email or newsletter preferences are now propagated to and persisted with GovDelivery.

**Dev Disclaimer**: I found the organization and UX of the GovDelivery documentation pretty rough and leaned heavily on Claude to get things initially going, but everything in this PR has been updated, reviewed, and validated by me.

### Ticket

This pull request resolves [#16890](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16890) and [#16365](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16365).

### Approach

- A new client has been added specifically for the Communications Cloud API. While there is an existing `GovDeliveryNewsletterClient`, that uses the [Script Service API](https://support.granicus.com/s/article/Script-Service-API?language=en_US&c=&p=Communications%20(govDelivery)%20API) and authenticates with an API key, whereas the Communications Cloud API requires basic auth tied to a specific account. Given that difference, I opted to just create a new client. We should also be able to use this new client exclusively if we choose to deprecate `GovDeliveryNewsletterClient` in the future.
- **Backend:** `syncNewsletterSubscription` contains the core business logic. If calls to GovDelivery fail, `userRouter` returns a HTTP 502 response and the database call to update user data values for `email` and/or `receiveNewsletter` is skipped.
- **Frontend:** Any errors propagate through `post()` → `useUserData` → `profile.tsx` → `ContactTabPanel`, which renders specific error alerts. User edits to email/newsletter fields clear the error.

### Steps to Test

When `.env` values that hit GovDelivery's staging environment are omitted, Wiremock values default to mocking successful calls. However, these changes include the option to toggle these states to test failures.

#### Testing mocked failure modes

<details>
<summary>Testing subscribe failure:</summary>

1. Start WireMock and set `subscribe` to failure mode:
   ```bash
   curl -X PUT http://localhost:9000/__admin/scenarios/GovDeliverySubscribe/state -d '{"state": "Failure"}'
   ```
2. In `/profile?tab=contact-notifications`, check "Receive newsletter" and click the "Save" button.
3. Verify this error alert appears: `We were unable to subscribe you to the newsletter. Your changes were not saved. Please try again.`
4. Verify changes are NOT saved (uncheck the box and refresh - it should still be unchecked).
5. Edit the email field - the error alert should disappear.
6. Reset WireMock to success:
   ```bash
   curl -X PUT http://localhost:9000/__admin/scenarios/GovDeliverySubscribe/state -d '{"state": "Started"}'
   ```
7. Try subscribing again - it should now succeed.

</details>

<details>
<summary>Testing unsubscribe failure:</summary>

1. Subscribe successfully first (with WireMock in `Started` state).
2. Set `unsubscribe` to failure mode:
   ```bash
   curl -X PUT http://localhost:9000/__admin/scenarios/GovDeliveryUnsubscribe/state -d '{"state": "Failure"}'
   ```
3. In `/profile?tab=contact-notifications`, uncheck "Receive newsletter" and click Save.
4. Verify error alert: `We were unable to unsubscribe you from the newsletter...`
5. Verify newsletter checkbox is still checked after refresh.

</details>

<details>
<summary>Testing email update failure:</summary>

1. Subscribe to newsletter first.
2. Set `updateEmail` to failure mode:
   ```bash
   curl -X PUT http://localhost:9000/__admin/scenarios/GovDeliveryUpdateEmail/state -d '{"state": "Failure"}'
   ```
3. In `/profile?tab=contact-notifications`, change email address and click Save.
4. Verify error alert: `We were unable to update your newsletter subscription with your new email...`.
5. Verify the email change is NOT saved.

</details>

#### Using GovDelivery's staging environment to validate integration points

See me for a walkthrough of the staging environment and validating that changes propagate through when the correct `.env` values are set, since other engineers don't currently have admin access in the staging environment.

### Notes

- WireMock testing documentation is available in the new `api/wiremock/mappings/README.md`
- Right now we have a dedicated account to integrate with GovDelivery's staging environment. We'll need to have a account with similar permissions/access created for their prod environment before this can go live
   - **Update**: we have a similarly configured integration account in GovDelivery's prod.
- We're at our limit for repository secrets in GitHub ([there's a limit of 100 secrets per repository](https://docs.github.com/en/codespaces/managing-codespaces-for-your-organization/managing-development-environment-secrets-for-your-repository-or-organization#limits-for-secrets)). Given that these are all server side and we're already using it for other things, I've opted to add these secrets via AWS SSM.

## Code author checklist

    - ~~_TODO_: update this for deployed sites via GitHub secrets~~ Using AWS SSM instead of GitHub secrets for deployments, see notes above
- [x] I have added the `request-reviewer` tag on github to request reviews


api/wiremock/mappings/README.md
```diff
@@ -0,0 +1,56 @@
...
+   curl -X PUT http://localhost:9000/__admin/scenarios/GovDeliveryUpdateEmail/state -H "Content-Type: application/json" -d '{"state": "Failure"}'
+   ```
+
+2. Once finished, reset all scenario states back to "Started":
+
+   **GovDeliverySubscribe**
+
+   ```bash
+   curl -X PUT http://localhost:9000/__admin/scenarios/GovDeliverySubscribe/state -H "Content-Type: application/json" -d '{"state": "Started"}'
+   ```
```
hinzed1127: Claude generated the initial Wiremock mappings. I thought the state here being `Started` instead of `Success` was a little strange, but it turns out [this is just how Wiremock works](https://wiremock.org/docs/stateful-behaviour/#scenarios):
> WireMock supports state via the notion of scenarios. A scenario is essentially a state machine whose states can be arbitrarily assigned. Its starting state is always Scenario.STARTED.

This is the first time we've used Wiremock's scenarios feature in our mappings, but I think this is a good use case. Rather than filtering for specific arguments to an endpoint, default to these mocks all succeeding, and then someone can turn them off selectively if desired.

web/src/lib/data-hooks/useUserData.ts
```diff
@@ -145,9 +145,9 @@ export const useUserData = (): UseUserDataResponse => {
           mutate(response, false);
           localUpdateQueue.queue(response);
         })
-        .catch(() => {
+        .catch((error) => {
           setUserDataError("UPDATE_FAILED");
-          throw undefined;
+          throw error;
```
hinzed1127, line 154: updating this to throw an error lets us catch and display any govDelivery errors in `profile.tsx`. See an example in the unit test for this module's test file

api/.env-template
```diff
@@ -55,3 +55,8 @@ USE_WIREMOCK_FOR_FORMATION_AND_BUSINESS_SEARCH=
 XRAY_REGISTRATION_STATUS_BASE_URL=
 FEATURE_USERDATA_ENDPOINT=
 FEATURE_WELCOME_EMAIL_ENABLED=
+GOVDELIVERY_COMM_CLOUD_BASE_URL=
+GOVDELIVERY_COMM_CLOUD_ACCOUNT_CODE=
+GOVDELIVERY_COMM_CLOUD_USERNAME=
+GOVDELIVERY_COMM_CLOUD_PASSWORD=
+GOVDELIVERY_COMM_CLOUD_TOPIC_CODE=
```
hinzed1127: noted in the PR description, but calling out again here that these need to get added to the repo variables/secrets before merging and deploying

api/src/api/userRouter.test.ts
```diff
@@ -1226,6 +1254,121 @@ describe("userRouter", () => {
...
+            stubCryptoEncryptionClient,
+            stubCryptoHashingClient,
+            stubTimeStampBusinessSearch,
+            DummyLogWriter,
+            mockGovDeliveryClient,
+          ),
+        );
+      });
+
+      it("returns 502 with govDeliveryError body and does NOT call databaseClient.put when subscribe fails", async () => {
```
hinzed1127: Noting that the sync logic is the same across `subscribe`, `unsubscribe`, and `updateEmail`. Basically, while it's unlikely to occur, we don't want to be in a situation where a call to GovDelivery fails and a call to our db succeeds, resulting in user info getting out of sync between the two. As a result, we always try to update GovDelivery values first. If those calls succeed, and only succeed, then we update the user data internally.

api/src/api/userRouter.ts
```diff
@@ -274,11 +336,20 @@ export const userRouterFactory = (
...
+    }
+
+    if (existingUserData && industryHasChanged(userData, existingUserData)) {
       userData = clearTaskItemChecklists(userData);
     }
 
-    const userDataWithUpdatedLegalStructure = await updateLegalStructureIfNeeded(userData);
+    const userDataWithUpdatedLegalStructure = existingUserData
+      ? updateLegalStructureIfNeeded(userData, existingUserData)
+      : userData;
```
hinzed1127: This is a small campground improvement pointed out by claude that I iterated through. Prior to this PR, POST calls to `/users` were fetching user data twice. Initially, `syncNewsletterSubscription` was adding a third fetch. Refactors present here consolidated that down to a single fetch.

api/src/client/GovDeliveryCommCloudClient.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      expect(body).toContain("<email>bad&amp;actor@example.com</email>");
+    });
+
+    it("escapes & in newEmail for updateEmail", async () => {
+      mockAxios.put.mockResolvedValueOnce({ status: 200 });
+      await client.updateEmail("old@example.com", "bad&actor@example.com");
+      const body = mockAxios.put.mock.calls[0][1] as string;
+      expect(body).toContain("<email>bad&amp;actor@example.com</email>");
+    });
+  });
```
hinzed1127, line 102: some basic protection against XML injections

web/src/lib/api-client/apiClient.ts
```diff
@@ -207,7 +209,7 @@ export const post = async <T, R>(url: string, data: R, auth = true): Promise<T>
       return response.data;
     })
     .catch((error: AxiosError) => {
-      throw error.response?.status;
+      throw { status: error.response?.status, data: error.response?.data };
```
hinzed1127, line 210: Even though this is a shared utility function, the number of situations depending on the shape of this thrown error was pretty small. I think the only thing that needed to be updated was `web/src/components/LoginEmailCheck.tsx`. A unit test for that case was added, as well.

web/src/components/profile/ContactTabPanel.test.tsx
hinzed1127, line 1: Added by claude when adding unit tests for displaying the govDelivery errors. Since tests didn't exist for this component previously, there's some scaffolding/mocking and a few basic rendering smoke tests, as well.

web/src/pages/profile.tsx
hinzed1127, line 1: The changes here required more prop drilling from `Profile` -> `ContactTabPanel` -> `ContactInformationTab` than I would have liked, but using Context felt like overkill for this small of a use case, and that refactor wouldn't have really simplified the code enough to be worth it

web/test/pages/profile/profile-govdelivery.test.tsx
hinzed1127, line 1: the complexity of `/profile` and all of its dependencies makes this a bit of a mocking nightmare, hence the length of this test file 😞

hinzed1127 (commented):
The net lines changed in this PR are more than I'd like, but to add a little more context, this includes:
- almost 1,000 added lines are tests, including a handful of new test files entirely
- new markdown doc for wiremock mappings
- different config changes (wiremock mappings, CMS values, .env, etcz)

The meat of these changes are a few hundred lines of code. I would primarily direct your attention to:
- **`syncNewsletterSubscription.ts`** — this covers the 5 different combinations of subscribe/unsubscribe/email update that we're handling. The test file for this lays them out nicely, as well
-  **`GovDeliveryCommCloudClient.ts`** — The actual calls to the API. The only oddity here is it's a little antiquated with XML request bodies and added security checks to protect against XML injection
-  **`userRouter.ts`** — Where the actual sync happens. Also includes a small refactor I cover in my comments

api/src/functions/express/app.ts
```diff
@@ -335,6 +336,18 @@ const taxFilingClient = ApiTaxFilingClient(
   },
   logger,
 );
+const govDeliveryCommCloudClient = process.env.GOVDELIVERY_COMM_CLOUD_ACCOUNT_CODE
```
hinzed1127: `GOVDELIVERY_COMM_CLOUD_ACCOUNT_CODE` feels like an arbitrary value for whether or not this gets initialized. I just needed to have something. Would it be better if this is just a feature flag?

api/src/api/userRouter.ts
```diff
@@ -138,6 +142,63 @@ const businessHasFormed = (userData: UserData): boolean => {
   return getCurrentBusiness(userData).formationData.getFilingResponse?.success ?? false;
 };
 
+const industryHasChanged = (userData: UserData, oldUserData: UserData): boolean => {
```
hinzed1127: refactored to make a single get request for userData, which changes the arguments and moves up the scope of these helper functions out of the `.post` handler

api/src/api/userRouter.ts
```diff
@@ -138,6 +142,63 @@ const businessHasFormed = (userData: UserData): boolean => {
...
+      const oldBusiness = getCurrentBusiness(oldUserData);
+
+      return modifyCurrentBusiness(userData, (business) => ({
+        ...business,
+        profileData: {
+          ...business.profileData,
+          legalStructureId: oldBusiness.profileData.legalStructureId,
+        },
+      }));
+    }
```
jphechter: Unrelated to your work

I hadn't really occurred to me that we have checks like this in the backend, but it makes me think that this should be more robust. We're calling out a case that we don't want the legal structure to change if the business has formed, but I would think that we'd want to lock down ALL formation data in that case.

I guess we're assuming that the user is only going to be able to modify things that are available through the UI, but maybe that's an improvement we want to consider.

api/src/api/userRouter.ts
```diff
@@ -138,6 +142,63 @@ const businessHasFormed = (userData: UserData): boolean => {
...
+        formationFormData: {
+          ...createEmptyFormationFormData(),
+          ...address,
+        },
+        businessNameAvailability: undefined,
+        dbaBusinessNameAvailability: undefined,
+        lastVisitedPageIndex: 0,
+      },
+    }));
+  }
```
jphechter: Also, not directly related to your work, but this feels like something that might be benefited by a docstring-type comment. It took me a minute to consider why we'd intentionally reset formation data.

api/.env-template
jphechter: ~~I see the creds for the staging environment in Bitwarden. Do we have the creds for prod?~~

EDIT: Re-read your PR description

web/test/pages/profile/profile-govdelivery.test.tsx
```diff
@@ -0,0 +1,267 @@
...
+    "queueFormationData",
+    "queueFormationFormData",
+    "queueTaskItemChecklist",
+    "queueEnvironmentData",
+    "queueXrayRegistrationData",
+    "queueRoadmapTaskData",
+  ];
+  for (const method of chainableMethods) {
+    mockQueue[method] = jest.fn().mockReturnValue(mockQueue);
+  }
```
jphechter, line 53: aren't you clever

hinzed1127: Some real galaxy-brained claude work, I'll admit. But honestly, given the layers of components on this page, if I had to do all of this mocking by hand, this test file would have never gotten written

jphechter:
On the whole, this all works for me. 

The only things we're missing are `prod` creds, and setting the environment variables in a GHA workflow. See `deploy-dev-environment.yml` for example.

After we get those things, I have no issue merging this.

api/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "resolveJsonModule": true,
     "strict": true,
     "target": "ES2020",
+    "rootDir": "..",
```
hinzed1127, line 13: added due to some compiler complaints with importing some items from `shared/`.

PR #12677, opened 2026-03-30
feat: [AB#17584] add select industry task

ashlsun:
_Used Spencer's /create-pr-navigator Claude Code skill for this description. Many manual edits since._
## Description

Add a new "Select Industry" task to the generic industry roadmap that allows users to select their business industry inline, replacing a previous onboarding step.

### Ticket

This pull request resolves [#17584](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17584).

### Approach

Figma: https://www.figma.com/design/H2mER7gGeHKTM3TgySmSvm/101-105-Business-Experience-Sprint-Designs?node-id=71149-5110&t=GhbZCDdcbY9P1Ya2-0

- Created a new `select-industry` task md file and JSON for the body content.
- Added the `select-industry` step to the generic industry roadmap JSON (step 1, weight 2)
- Created `SelectIndustryTask` component that uses the industry dropdown.
- Added the new component to `TaskPageSwitchComponent` so that it gets displayed instead of default

### Steps to Test

1. In web/env, set `AB_TESTING_EXPERIENCE_B_PERCENTAGE=100`
2. Go through onboarding as a new business. You should see that you aren't asked to choose an industry before seeing your roadmap.
3. Verify the "Select Industry" task appears in step 1 of the roadmap
4. Open the task and confirm that selecting an industry and saving changes your industry.

### Notes
Will squash all the commits before rebasing and merging. I leave them here for ease of review in case it helps.

## Code author checklist

- [x] I have added the `request-reviewer` tag on github to request reviews


web/src/components/tasks/SelectIndustryTask.tsx
jphechter, line 1: More design feedback than anything else, but it feels odd to me that the industry dropdown aligns to the middle of the task container. I would expect it to sit closer to the top, similar to the way it does when the selected industry has additional questions, like the second image.
Image
Image

content/src/fieldConfig/select-industry-task.json
```diff
@@ -0,0 +1,8 @@
+{
+  "selectIndustryTask": {
+    "successAlert": "You’ve successfully saved your industry. Once you select your business structure, you may see changes to your guide.",
+    "content": "**What industry do you work in?**\nThis helps us show you licenses and permits you may need.",
+    "infoCallout": ":::miniCallout{ calloutType=\"informational\" }\n**To continue, select Save.**\n:::",
```
jphechter: I think we use language more like this throughout the app. We should defer to the content team's recommendation though.
```suggestion
    "infoCallout": ":::miniCallout{ calloutType=\"informational\" }\n**To continue, click Save.**\n:::",
```

web/decap-config/collections/06-tasks.yml
```diff
@@ -2674,3 +2674,23 @@ collections:
           - { label: CRTK Info Shared Pt One, name: crtkInfoSharedPtOne, widget: string }
           - { label: CRTK Info Shared Pt Two, name: crtkInfoSharedPtTwo, widget: string }
           - { label: CRTK Info Shared Pt Three, name: crtkInfoSharedPtThree, widget: string }
+  - name: "select-industry-task"
+    label: "Tasks - Select Industry"
+    delete: false
+    create: false
+    files:
+      - label: "Select Industry Task - Get Your Business Licenses and Permits"
```
jphechter: I might generify this, just because the task name "Get Your Business Licenses and Permits" could change and this label won't have parity with the task name. Check this with the content team, but I would lean towards this saying "Select Industry Task" or "Industry Selection Task" just to keep it aligned with the intent of the task.
```suggestion
      - label: "Select Industry Task"
```

content/src/roadmaps/tasks/select-industry.md
```diff
@@ -0,0 +1,11 @@
+---
+notesMd: On this CMS page, only changes to the Name and URL Slug will apply to this task. To edit the content, go to the Tasks - Select Industry collection.
+summaryDescriptionMd: ""
+urlSlug: select-industry
+displayname: SelectIndustry
+name: Get Your Business Licenses and Permits
```
jphechter: More content feedback, but I do not love this task name. I get the intent, but it doesn't really feel aligned with what we're offering the user.

content/src/fieldConfig/select-industry-task.json
jphechter: Right now, this whole section is being supplied as editable. This gives the content team the ability to supply anything here, not just updating information inside the callout.

I think I would either limit the `infoCallout` to just being the callout message and shove that into a `MiniCallout` component in the task, or rename something indicating that this is all of the content after the dropdown. This give you more flexibility if the nature of the content changes in the future.

Image

web/src/components/tasks/SelectIndustryTask.tsx
jphechter, line 1: Right now, if I select an industry on the profile this task does not auto-complete. Is that something that we want to do for the user?

web/src/pages/onboarding.tsx
```diff
@@ -149,11 +149,16 @@ const OnboardingPage = (props: Props): ReactElement => {
       }
     };
 
+    if (userData?.user.abExperience === "ExperienceB") {
```
jphechter, line 152: Not a blocking issue, but a thought I had looking at this...

Now that we're increasing our usage of A/B testing, it might be good for us to have globals or enums for `"EperienceA"`/`"ExperienceB"`

web/test/pages/onboarding/onboarding-additional-business.test.tsx
```diff
@@ -94,7 +98,9 @@ describe("onboarding - additional business", () => {
     const userId = "user-id";
     const emptyBusiness = createEmptyBusiness({ userId: userId });
     const initialBusiness = generateBusiness({ userId: userId });
-    const initialData = generateUserDataForBusiness(initialBusiness);
+    const initialData = generateUserDataForBusiness(initialBusiness, {
+      user: generateUser({ id: userId, abExperience: "ExperienceA" }),
```
jphechter, line 102: I'm assuming we're setting `abExperience` here because its randomized in our tests, is that correct?

I wonder if we can rely on defining the A/B split at the jest configuration level and then only define "ExperienceB" if we actually need it. Otherwise, I think we're going to end up needing to do this in lots of places as we continue to expand usage of A/B tests.

web/src/lib/roadmap/buildUserRoadmap.ts
```diff
@@ -45,7 +48,7 @@ export const buildUserRoadmap = async (
     addOns = [];
   }
 
-  let roadmap = await buildRoadmap({ industryId: industryId, addOns });
+  let roadmap = await buildRoadmap({ industryId: industryId ?? "generic", addOns });
```
hinzed1127, line 53: Idk if this is worth calling out in a comment above this line, but it could be helpful to communicate that this is needed because:
- If experience B is active, an industry is never selected, hence `undefined`
- An `industryId` of `generic` is equivalent to selecting "All Other Industries".

I know the team stance on comments is varied, but this is the kind of subtle business logic that it might be beneficial to document

web/test/pages/onboarding/onboarding-additional-business.test.tsx
hinzed1127: This is a good point, but something that I could see both sides. Randomizing means we're making sure functionality that's not directly tied to an A/B test is more thoroughly test on either side of that divide.

If we get to the point that we're running multiple A/B tests in parallel, which is something we're considering, this'll probably need to get considered in more depth from that angle, as well.

web/src/components/tasks/SelectIndustryTask.tsx
hinzed1127: I was curious and poked around, so here's a long answer to a relatively minor thing:

The real culprit here currently is actually coming from [a parent `div` in SidebarPageLayout](https://github.com/newjersey/navigator.business.nj.gov/blob/f0861228926c7cbb77f83a5a7d3bfb7fa7bf9d1e/web/src/components/njwds-layout/SidebarPageLayout.tsx#L60-L67), which hard-codes the entire container to a min-height of `40rem`. If the content wasn't centered, there would be a bunch of whitespace before the button row with the save button, which also isn't ideal.

IMO, it might be better to remove both that `40rem` style rule and the `min-height-38rem` class in this component. That combination would make the height of the whole container adapt to the content, but we'd probably need to check with design if that is preferred, or if a minimum height that's a little larger is intentional. The code for the min height of `40rem`  is pretty old; I'm guessing this is just the first time a task container has had such a small height, and so removing or lowering that minimum is probably not a big deal.

hinzed1127 (commented):
Overall looks good! Seems like a few things to clarify with content and design, but I'm feeling good about the more code-centered bits of this

jphechter:
Something I just thought about... because we're not assigning an industry to Experience B users, but rather assuming "generic" in the roadmap builder, we will need to account for this if we ever remove this functionality. Specifically, we will need a migration to update Experience B users with no industry to have an industry.

I don't think we'd catch this in a test, or in manual testing unless we were testing with a user that was in Experience B previous to the task being removed.

I also don't know what we would do to proactively catch this, but this is something we should be mindful of. I can envision this being a somewhat gnarly bug to figure out at some point in the future.

web/src/components/tasks/SelectIndustryTask.tsx
ashlsun: Thank you for finding that, I was struggling to track down where that minimum came from! Removed it and now the spacing doesn't look so odd.

content/src/fieldConfig/select-industry-task.json
ashlsun: Passing this feedback along to content team!

ashlsun: Since this callout is being conditionally shown, think it makes sense to just hardcode the component so there's no confusion if Content leads expect this to be all the content after the dropdown

web/src/components/tasks/SelectIndustryTask.tsx
ashlsun: Yes, good catch!

web/test/pages/onboarding/onboarding-additional-business.test.tsx
ashlsun: For now I think I will leave this as randomized except for the tests where we need to define it.

If we plan to run multiple parallel AB tests in the future, we'll have to significantly rework our AB assignment system, and I think that would the right place to reconsider our unit testing strategy. 

Right now, AB assignment is determined by a single string `abExperience` in the userData. If we rely on `abExperience` as is, all Experience B users will see all Experience B changes, and the tests won't be independent. A system equipped to handle multiple independent tests will likely look very different, and any jest config for AB tests we do here will be made obsolete.

content/src/roadmaps/tasks/select-industry.md
ashlsun: Content team members agreed to go with Select Your Industry!

PR #12695, opened 2026-04-10
chore: add a top-level and directory-specific `CLAUDE.md`/`AGENTS.md` files

hinzed1127:
## Description

This adds `CLAUDE.md`/`AGENTS.md` files to various levels of the repo, providing both top-level guidance to Claude/Codex/other agents, along with more specific guidance in the various directories of the mono-repo structure.

I also added a `./update-claude-md` as an easy way to periodically update those files, which utilizes a [public skill installable as a plugin](https://github.com/anthropics/claude-plugins-official/tree/main/plugins/claude-md-management)

#### My Initial Prompt

> Help me set up claude code for this repo. Consider the factors as you analyze and configure:
> 
> - Since this is a monorepo, I'm primarily interested in effective `CLAUDE.md` files that live in child directories. A top-level `CLAUDE.md` might be helpful for the essentials, but I'm guessing we'd benefit more from files scoped to the `web`, `api/`, `content/`, `shared`, and `scripts/` directories.
> 
> - Ghostty is a large open-source project that I think makes good use of this approach to granular `AGENTS.md`, the more generalized `CLAUDE.md`, files. Search for those via `gh api "search/code?q=filename:AGENTS.md+repo:ghostty-org/ghostty" --jq '.items[].path'` and look to them as one source of inspiration. Also follow [[these guidelines](https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md)](https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md) in the official Claude Code docs.
> 
> - Consider using [[rules](https://code.claude.com/docs/en/memory#organize-rules-with-claude%2Frules%2F)](https://code.claude.com/docs/en/memory#organize-rules-with-claude%2Frules%2F) as an alternative to scoped `CLAUDE.md` files. They seem to accomplish as similar goal.
> 
>   I'll defer to you on the best approach, but ask me any clarifying questions before you begin.

After a few small tweaks and corrections on my own, I ran it through an [Anthropic-authored skill](https://github.com/anthropics/claude-plugins-official/blob/main/plugins/claude-md-management/skills/claude-md-improver/SKILL.md) a few times, first with Sonnet 4.6 and then Opus 4.7. I added a command to prompt to run that same skill in the future (see review comments below).

#### Other considerations

Per the [Claude docs](https://code.claude.com/docs/en/memory#how-claude-md-files-load):

> Claude also discovers CLAUDE.md and CLAUDE.local.md files in subdirectories under your current working directory. Instead of loading them at launch, they are included when Claude reads files in those subdirectories.

Even though I asked about using [rules](https://code.claude.com/docs/en/memory#organize-rules-with-claude/rules/) instead of directory-scoped `CLAUDE.md` files, claude suggested this was overkill for our present needs. After skimming the rules docs, this seems true, since the main advantage is you can apply rules to various file paths using glob patterns, which we don't need at this point in time.

### Steps to Test

Unlike skills, which are explicitly shown to load when they get started, `CLAUDE.md` files seem to [load more silently](https://code.claude.com/docs/en/memory#how-claude-md-files-load). I'm not sure if there's an objective way to validate this vs just anecdotal experience.

## Code author checklist

- [ ] I have added the `pr-show` or `request-reviewer` tag on GitHub to show or request reviews


.claude/commands/update-claude-md.md
```diff
@@ -0,0 +1,7 @@
+Audit and update the CLAUDE.md files in this project to ensure they are accurate and up to date.
+
+First, check whether the `claude-md-management:claude-md-improver` skill appears in your available skills list. If it does not, stop and tell the user:
+
+> The `claude-md-management` plugin is required but not installed. Install it via the Claude Code plugin marketplace, then re-run `/update-claude-md`.
+
+If the skill is available, invoke it using the Skill tool.
```
hinzed1127, line 7: it looks like this plugin ships with a separate [`/claude-md-management:revise-claude-md` command](https://github.com/anthropics/claude-plugins-official/blob/main/plugins/claude-md-management/commands/revise-claude-md.md) that specifically "captures learnings from the current session". Maybe we should document that somewhere, but they serve slightly different use cases.

web/CLAUDE.md
```diff
@@ -0,0 +1,61 @@
...
+## `apiClient.ts` — Protected Contract
+
+`src/lib/api-client/apiClient.ts` is the **sole HTTP layer** between frontend and backend. It is used throughout the entire app.
+
+```typescript
+post<T, R>(url: string, data: R, auth?: boolean): Promise<T>
+get<T>(url: string, auth?: boolean): Promise<T>
+```
+
+**Do not change the signature or error-handling behavior of existing functions.** If new capability is needed, add a new function — do not alter `post` or `get`.
```
hinzed1127: In follow-ups to my initial prompt, I mentioned this as an area where I'd observed Claude changing the return value ([e.g. it's in this PR](https://github.com/newjersey/navigator.business.nj.gov/pull/12627/files#diff-4e1b184c75a7c641ea4c1b28a30b8f969c91a2b9031e52f02e622c19328fc24cR212)) without holistically checking about the downstream effects such a change might have. That exchange seemed to have left maybe _too_ strong an impression and resulted in this section. Do you think we should scrap it, or is the guidance valid?

hinzed1127 (commented):
Self-review. I reviewed all copy, but this is basically a one-shot of my initial prompt getting picked up by [this plugin skill](https://github.com/anthropics/claude-plugins-official/blob/main/plugins/claude-md-management/skills/claude-md-improver/SKILL.md). Only additional context on top of that added in my comments below

PR #12751, opened 2026-05-01
feat: [AB#17706] business structure task redesign

ashlsun:
## Description

Implements [this redesign](https://www.figma.com/design/H2mER7gGeHKTM3TgySmSvm/101-105-Business-Experience-Sprint-Designs?node-id=73091-4693&t=PwKiod0072DiuFOI-4) of the business structure task, which highlights the most common options for business structure and tucks the rest away in an accordion to make the page less overwhelming. 

We also replace the Learn More links with contextual info to give users info without kicking them off to another page.

| Before  | After |
| ------------- | ------------- |
| Screenshot 2026-05-05 at 2 00 18 PM | Redesigned Business Structure Task |
| Screenshot 2026-05-05 at 2 00 52 PM | Screenshot 2026-05-05 at 1 55 41 PM  |
### Ticket

This pull request resolves [#17706](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17706).

### Approach

- Added `isCommon: boolean` to the `LegalStructure` interface in `shared/src/legalStructure.ts` to categorize structures.
- Updated `LegalStructureRadio` to split structures into two collapsible MUI `Accordion` sections: **Common Business Structures** (expanded by default) and a second section for less common structures.
- Moved the save/submit button from its inline position into a standardized `ActionBarLayout`/`CtaContainer` at the bottom of the task (consistent with other task pages).
- Added a `LiveChatHelpButton` with an analytics event (`select_business_structure_task.click.open_live_chat`) alongside the save button.
- Added an informational `miniCallout` block to `business-structure.md` explaining the importance of picking a structure and when it can be changed, and added an option to hide the icon of a mini callout.

### Steps to Test

1. Complete onboarding as a new business in any industry.
2. Go to the "Select Your Business Structure".
3. Verify that the accordions contain all the options, and users can select only one business structure at a time (you should not be able to select one option under Common Business Structures and one under Other at the same time).
5. Select a structure and click Save — confirm the task completes, updates your business structure, and shows the success state and a link to check available names and form your business.

## Code author checklist


web/src/components/tasks/business-structure/BusinessStructureTask.tsx
```diff
@@ -122,9 +126,6 @@ export const BusinessStructureTask = (props: Props): ReactElement => {
     setProfileData(updatedProfileState);
   };
 
-  const preLookupContent = props.task.contentMd.split("${businessStructureSelectionComponent}")[0];
```
ashlsun, line 125: We originally pulled the content here from the body of `business-structure.md`, but given that we already have a lot of the pieces in fieldConfig's `business-structure-task.json`, I moved it there so that our content teammates don't have to switch between collections to edit this task.

This will also make it less confusing if we decide to display this content conditionally in the future.

web/src/components/njwds-extended/callout/calloutHelpers.ts
```diff
@@ -14,6 +14,7 @@ export interface LargeCalloutProps extends IconProps {
 
 export interface MiniCalloutProps {
   calloutType: CalloutTypes;
+  hideIcon?: string | boolean;
```
jphechter: Why is this accepting both a `string` and `boolean`? This feels odd to me.

web/src/components/tasks/business-structure/LegalStructureRadio.tsx
```diff
@@ -83,54 +107,126 @@ const LegalStructureRadio = forwardRef(
...
+                                (contentFromConfig.optionContent as Record<string, string>)[
+                                  legalStructure.id
+                                ]
+                              }
+                            </Content>
+                          </div>
+                        )}
+                      </div>
+                    );
+                  })}
```
jphechter: This block is almost identical to `189-218`. How do you feel about pulling this logic into a helper function and passing the legal structures as a param?

web/src/components/njwds-extended/callout/calloutHelpers.ts
ashlsun: That's fair! I followed the example right above this where where `showHeader` in `LargeCalloutProps` is `string | boolean`, but I think changing this to just `boolean` would work, so I will try that.

web/src/components/tasks/business-structure/LegalStructureRadio.tsx
```diff
@@ -83,54 +107,126 @@ const LegalStructureRadio = forwardRef(
...
-              <div className="text-error-dark text-bold" data-testid="business-structure-error">
-                {contentFromConfig.errorTextRequired}
-              </div>
-            )}
+                <Heading level={3} className={`flex flex-align-center margin-0-override`}>
+                  <div className="inline">Other Business Structures</div>
+                </Heading>
+              </AccordionSummary>
+              <AccordionDetails>
+                <WithErrorBar hasError={isFormFieldInvalid} type="ALWAYS">
```
jphechter: We might not need this `WithErrorBar`. I think the one on `110` is enough.

jphechter (approved):
Overall, this looks like a good change. My biggest piece of feedback here is probably with `calloutHelpers.ts` -> `MiniCalloutProps` -> `hideIcon`. I really don't like mixing the types. Understanding that this might not be easy given this is controlled by the CMS. But if that's the case, then maybe we rally around specific string vars like "true" and "false" instead of generic strings or potentially allowing the boolean value.

Aside from that. Minor nits. I tested this locally. I was initially concerned that it would be possible to assign a value in both radio groups, but it looks like MUI is handling that well enough.

PR #12756, opened 2026-05-04
feat: [AB#16176] replace react-transition-group with CssTransition

seidior:
### Summary

Replace `react-transition-group` with a lightweight, homemade `CssTransition` component. This removes the library as a direct dependency while preserving the same CSS class-based enter/exit transition behavior used on the onboarding page. This is the only page where we use `react-transition-group`, so we can safely remove it as a dependency once this page is validated to be just as usable and accessible as before.

This is step 2 of the React 19 transition strategy tracked in [AB#16176](https://dev.azure.com/NJInnovation/b378a1f5-9f4a-404b-b718-8e7bd4265ca2/_workitems/edit/16176): removing React-19-incompatible dependencies ahead of the eventual React 19 upgrade. It does not fully resolve the ticket; subsequent PRs will address MUI, unified/rehype, SWR, and React 19 itself.

### Ticket

This pull request is part of [#16176](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16176).

### Approach

- Created `CssTransition` component (`web/src/components/transitions/CssTransition.tsx`) that manages enter/exit CSS class transitions via `useState`/`useEffect`, with optional `unmountOnExit` support.
- Replaced `CSSTransition` from `react-transition-group` with the new `CssTransition` in `onboarding.tsx`.
- Converted `.then()` chains in onboarding page handlers to `async`/`await`.
- Removed `react-transition-group` and `@types/react-transition-group` from root and web `package.json`; updated `yarn.lock`.
- Added unit tests for the new component.

Note: `react-transition-group` remains an indirect dependency via MUI — this change only removes the direct dependency from the project's `package.json` files.

### Steps to Test

1. Navigate to the onboarding flow.
2. Step through each page — confirm slide transitions animate correctly (enter and exit).
3. Verify no console errors related to transitions.

### Notes

Acceptance criteria per the transition strategy: onboarding transitions work, no `react-transition-group` imports remain, current unit and targeted Cypress tests pass on React 18.

## Code author checklist


seidior:
I've pulled this out on its own because it's 100% a win regardless of whether we upgrade to React 19 now or later, since it's a prerequisite to a React 19 approach. It seems not to have a performance hit, but it remains to be seen if the accessibility is impacted.

web/src/pages/onboarding.tsx
```diff
@@ -424,20 +424,16 @@ const OnboardingPage = (props: Props): ReactElement => {
...
-            const nextCurrentPage = page.current + 1;
-            setPage({
-              current: nextCurrentPage,
-              previous: page.current,
-            });
-            routeToPage(nextCurrentPage);
-            headerRef.current?.focus();
-          });
+        await updateQueue.queueProfileData(newProfileData).update({ local: true });
+        const nextCurrentPage = page.current + 1;
```
SirSamuelJoseph: [dust] I know this was already here but I really don't like the name `nextCurrentPage`, I'd prefer just `nextPage`

web/src/components/transitions/CssTransition.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Exit duration in milliseconds before content is removed when unmounting is enabled.
+   */
+  readonly timeout: number;
+
+  /**
+   * Whether content should be removed from the DOM after the exit transition completes.
+   */
+  readonly unmountOnExit?: boolean;
```
SirSamuelJoseph, line 36: [dust] Maybe a test for the `unmountOnExit = false` case?

web/src/components/transitions/CssTransition.tsx
```diff
@@ -0,0 +1,126 @@
...
+type TransitionPhase = "enter" | "enter-active" | "exit";
+
+interface ParsedClassNames {
+  readonly baseClassName: string;
+  readonly transitionClassName: string;
+}
+
+const parseClassNames = (classNames: string): ParsedClassNames => {
+  const classNameParts = classNames.split(" ").filter(Boolean);
+  const transitionClassName = classNameParts[classNameParts.length - 1] ?? "";
```
SirSamuelJoseph: Can we rely on the transition class name always being the last in the list of class names?

SirSamuelJoseph (approved):
Approved with a small thing and a question around how we will need to use this going forward

web/src/pages/onboarding.tsx
seidior: Good call, renamed to nextPage

web/src/components/transitions/CssTransition.tsx
seidior: Added coverage for the unmountOnExit={false} case. Also a good catch.

web/src/components/transitions/CssTransition.tsx
seidior: I've now split the API into `className` and `transitionClassName` so the transition base is explicit. I like that approach better!

PR #12761, opened 2026-05-07
feat: add CMS content retrieval to static-site

jphechter:
## Description

Enables the static-site application to consume CMS content from the existing content pipeline. Introduces a dedicated `@businessnjgovnavigator/content-types` package as a single source of truth for content type definitions shared between the main app and the static-site.

### Ticket

This pull request resolves [#17621](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17621).

### Approach

**Other Updates**
- Run the app on port 4000
  - Navigator, and by extension the CMS, runs on 3000
- Added a build step to the `static-site-check` GHA
- Added `public/assets/njwds` to static-site `.gitignore` (generated build artifact)

**Static-Site Content Ingestion**
- Created `packages/content-types`
  - This is a zero-dependency, types-only package containing all CMS content types and interfaces
- Added `packages/content-types` to the root yarn workspace and as a dependency of `shared`
- Static-site reads pre-built JSON from `content/lib/` and raw markdown from `content/src/` via `domain/content/loadContent.ts`
- Static-site consumes types from `@businessnjgovnavigator/content-types` via pnpm `file:` reference
- Added `gray-matter` to static-site for parsing markdown frontmatter

Tested the implementation by adding content in 2 locations: a list of fundings w/in NJWDS card components on the main landing page, and a skeleton Business Names page that pulls from the markdown content.
Screenshot 2026-05-08 at 4 14 02 PM
Screenshot 2026-05-08 at 4 13 43 PM

### Steps to Test

1. Run `yarn build` at root (builds content-types as part of workspace)
2. `cd packages/static-site && pnpm install && pnpm typecheck` — should pass
3. `cd packages/static-site && pnpm build` — should succeed
4. `cd packages/static-site && pnpm dev` — visit `http://localhost:4000/en-US`, confirm "Featured Funding Opportunities" section renders with CMS content

#### Short-term follow-ups

- **Update `shared` to re-export from `content-types`** — Replace inline type definitions in `shared/src/types/types.ts` with `export type { Funding, Task, ... } from "@businessnjgovnavigator/content-types"`. This closes the loop so any content type change only happens in one place.
- **Docker deployment for static-site** — Update the Dockerfile to copy `content/lib/` and `content/src/` into the build context and set `CONTENT_LIB_DIR` / `CONTENT_SRC_DIR` env vars. Ensure `packages/content-types` is built before the static-site build.

#### Longer-term considerations

- **Add source-based collections to `content/build.ts`** — Pages, Recents, FAQs, Categories, and Sub-Categories are currently read as raw markdown/JSON from `content/src/` at request time using `gray-matter`. Adding them to the content build pipeline (like fundings and tasks) would pre-process them into `content/lib/` JSON, eliminating the `gray-matter` dependency and `CONTENT_SRC_DIR` requirement from the static-site.
- **CI pipeline for static-site** — The `static-site-ci.yml` and `static-site-deploy-ecs.yml` workflows will need to account for building `content-types` and making content files available during the static-site build.

## Code author checklist


aloverso (approved):
I might have more comments once I actually start working within the code here, but on reading through, it seems to make sense! Let's merge it and start loadtesting it in practice.

Wanted to call out that this followup seems high priority, how to best make sure it happens?:
> Update shared to re-export from content-types — Replace inline type definitions in shared/src/types/types.ts with export type { Funding, Task, ... } from "@businessnjgovnavigator/content-types". This closes the loop so any content type change only happens in one place.

packages/static-site/domain/content/loadContent.ts
```diff
@@ -0,0 +1,114 @@
...
+  LicenseEventType,
+  PageItem,
+  RecentItem,
+  Sector,
+  SubCategoryItem,
+  Task,
+} from "./types";
+
+const CONTENT_LIB_DIR =
+  // biome-ignore lint/style/noProcessEnv: configurable content paths for deployment
```
seidior: [dust] This was a deliberate decision on my part; by not relying on process.env, we break the pattern that everything is a process.env variable and instead move to configuration objects themselves, which not only can be typed, they reduce our overall need for so many GHA vars. However, strong opinion weakly held; if there's a good reason you'd rather process.env it, I'd love to know!

packages/static-site/domain/content/loadContent.ts
```diff
@@ -0,0 +1,114 @@
...
+    .readdirSync(dir)
+    .filter((f) => f.endsWith(".md"))
+    .map((fileName) => {
+      const raw = fs.readFileSync(path.join(dir, fileName), "utf-8");
+      const { data, content } = matter(raw);
+      return { ...data, body: content } as T;
+    });
+};
+
+const readMarkdownFile = <T>(subdir: string, slug: string): T | undefined => {
```
seidior: [sand] If at all possible, I'd love some JSDoc here that explains these functions and what people are getting out of it, especially documenting an example of how to use <T> since not everyone knows how to use generics.

packages/static-site/domain/content/loadContent.ts
```diff
@@ -0,0 +1,114 @@
...
+  const filePath = path.join(CONTENT_SRC_DIR, subdir, `${slug}.md`);
+  if (!fs.existsSync(filePath)) return undefined;
```
seidior: [pebble] Shouldn't we error with an explicit error type here (something like a MissingMarkdownFileError), hinting to the developer where the error might be with why their content isn't loading rather than silently failing?

seidior (commented):
I'm so proud of this PR. For the most part, it looks great, and I'm glad that some of the healthy patterns are being preserved here. I know you were a bit concerned with the generated nature of this, but I actually am really glad that you were able to deliver so fast with this and that you are very forward-looking for reinforcing these patterns, especially with the use of generics.

packages/content-types/src/anytime-action.ts
```diff
@@ -0,0 +1,41 @@
+export interface AnytimeActionCategory {
+  categoryId: string;
```
seidior: [sand] one other pattern I'd like us to get into, similar to @aloverso's suggestion never to destructure props: readonly interface members so we're not accidentally editing the original objects and causing state issues

packages/content-types/src/industry.ts
```diff
@@ -0,0 +1,23 @@
...
+  defaultSectorId?: string;
+  canHavePermanentLocation?: boolean;
+  roadmapSteps: RoadmapStep[];
+}
+
+export interface Sector {
+  id: string;
+  name: string;
+  nonEssentialQuestionsIds: string[];
+  industries: { id: string; name: string; naicsCodes?: string }[];
```
seidior: [dust] Aren't these `Industry[]`?

packages/content-types/src/county.ts
```diff
@@ -0,0 +1,48 @@
...
+  "Ocean",
+  "Passaic",
+  "Salem",
+  "Somerset",
+  "Sussex",
+  "Union",
+  "Warren",
+] as const;
+
+export type County =
```
seidior: [sand] This seems duplicative? Also I'd like types to be documented like interfaces.

packages/static-site/domain/content/loadContent.ts
```diff
@@ -0,0 +1,114 @@
...
+export const loadAnytimeActions = (): AnytimeActionTask[] =>
+  (readContentJson("actions.json") as { actions: AnytimeActionTask[] }).actions;
+
+export const loadCertifications = (): Certification[] =>
+  (readContentJson("certifications.json") as { certifications: Certification[] }).certifications;
+
+export const loadFilings = (): Filing[] =>
+  (readContentJson("filings.json") as { filings: Filing[] }).filings;
+
+export const loadFundings = (): Funding[] =>
```
seidior, line 87: [sand] Should we not be filtering on `publishStageArchive`?

packages/static-site/app/[locale]/business-names/page.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (!page) {
+    notFound();
+  }
+
+  return (
+    <main className="grid-container usa-section">
+      <h1>{page.name}</h1>
+      <p className="usa-intro">{page["sub-heading-text"]}</p>
+      <section>
+        <p>{page["main-text-1"]}</p>
```
Comment (seidior) [inline packages/static-site/app/[locale]/business-names/page.tsx:32]: [dust] Should these go through a Markdown rendering step for bold, etc support?

Comment (jphechter) [reply packages/static-site/app/[locale]/business-names/page.tsx:32]: They can, but the purpose of adding this wasn't to finalize the page, but rather just prove a pattern for pulling content.

I'm not going to modify this further just now.

packages/static-site/domain/content/loadContent.ts
jphechter: How strongly do we feel about JSDoc? I can do it. Just wondering how much overhead this adds.

packages/content-types/src/industry.ts
jphechter: I need to look at this more. This was `SectorIndustry` in `shared/src/sector.ts` which is these props. I assume you're correct, but I need to validate that.

packages/content-types/src/county.ts
jphechter: It does. And I don't have a good reason for it to be this way, other than that's how it was in `shared`.

Happy to pare it down if this doesn't impact other areas of the app.

packages/static-site/app/[locale]/business-names/page.tsx
Comment (aloverso) [reply packages/static-site/app/[locale]/business-names/page.tsx:32]: I think this can be deferred - this page is just a proof of concept as far as I'm aware

PR #12771, opened 2026-05-11
feat: [#17711] add learn page with typed content

aloverso:
## Description

- Adds a `/learn` route that renders category cards from localized message content. 
- Simplifies the i18n message validation layer from ~600 lines of per-field checks down to a single recursive key-structure comparison.

image

### Ticket

This pull request DOES NOT YET resolve [#17755](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17755). It is in service of getting to the next step (building the Plan page)

### Approach
- Added `learn` field to `ApplicationMessages` and the `en-US.json` / `es-US.json` message files
- Renamed `domain/landing/types.ts` to `domain/content/messageTypes.ts` since it now covers multiple sections. **Decision**: I think it'll be easier to manage / update the static typing for the JSON messages in a single file (rather than a file for `landing`, a file for `learn`, etc) - let me know what we think about this.
- Replaced all per-field validators in `domain/i18n/messages.ts` with a single recursive `hasSameStructure` check that validates non-en-US locales share the same key tree as en-US. **Decision**: I think this will be far, far easier to use. It ensures the validation of structural same keys between the different message JSONs, but does not require updating as we add new keys. In the old approach, the new `learn` messages were not getting validated because we didn't have validator functions in here for them. And adding them (and all the other future content) would have made this file 1000s of lines long, which is pretty unmaintainable.
- Replaced `loadLandingContentFromMessages` calls with direct `getApplicationMessages` usage. I don't see a good reason for the indirection layer here. I think we can call the messages loader directly.

### Steps to Test

- Navigate to `/en-US/learn` — should render the learn page with four category cards (Planning, Starting, Operating, Growing)
- Navigate to `/es-US/learn` — should render the same page. Still in English, spanish translations not yet added.

To test the structure validation:
- Add a new key to one of the JSONs, reload the dev site. Get an error:
Screenshot 2026-05-11 at 15 15 54

### Notes

The `domain/landing/types.ts` rename to `domain/content/messageTypes.ts` is a purely structural move — all type names are unchanged and imports were updated across the codebase.

Note that this is branched off [JP's unmerged commit](https://github.com/newjersey/navigator.business.nj.gov/pull/12761)

Also, I did try to break up my PR into small commits - reviewing commit-by-commit may help a bit. Sorry it's kind of big but there's a lot of boilerplate. Let me know if we want to break it into smaller PRs.

## Code author checklist


packages/static-site/domain/i18n/messages.ts
```diff
@@ -2,631 +2,73 @@
...
-const isLandingLink = (value: unknown): boolean => {
-  if (!isRecord(value)) {
-    return false;
+  if (typeof base === "object" && base !== null) {
+    if (typeof other !== "object" || other === null || Array.isArray(other)) return false;
+    const baseRecord = base as Record<string, unknown>;
+    const otherRecord = other as Record<string, unknown>;
+    return Object.keys(baseRecord).every(
+      (key) => key in otherRecord && hasSameStructure(baseRecord[key], otherRecord[key]),
+    );
```
seidior, line 36: [dust] I could be wrong with this, but while this enforces non-en-US locales have the minimum number of keys but that there can be keys in non-en-US locales that aren't in en-US. Not sure if you wanna check bidirectionally.

packages/static-site/app/[locale]/learn/page.tsx
```diff
@@ -0,0 +1,51 @@
...
+          <li key={category.title} className="usa-card tablet:grid-col-3">
+            <div className="usa-card__container">
+              <div className="usa-card__header">
+                <h3 className="usa-card__heading">{category.title}</h3>
+              </div>
+              <div className="usa-card__body">
+                <p>{category.description}</p>
+              </div>
+              <div className="usa-card__footer">
+                <a href={category.link.href} className="usa-button">
```
Comment (seidior) [inline packages/static-site/app/[locale]/learn/page.tsx:39]: [dust] Unfamiliar with how next-intl works but is this a location-aware way to link to content?

packages/static-site/domain/content/messageTypes.ts
```diff
@@ -320,6 +320,36 @@ export interface LandingMetadataContent {
...
+  /** Navigation name for the learn section. */
+  readonly name: string;
+  /** Subheading text rendered below the page title. */
+  readonly subHeadingText: string;
+  /** Secondary heading text rendered above the categories grid. */
+  readonly heading2: string;
+  /** Category cards rendered in the learn grid. */
+  readonly categories: readonly LearnCategory[];
+}
+
```
seidior, line 352: [sand] This file is getting pretty long, and given that you're thinking about simplifying this types setup, I'm wondering what you think about starting to split this up?

seidior (commented):
Overall, fantastic work. I think this new approach makes sense and strikes a good compromise between the initial approach and how we've loaded content in the past in order to speed up adoption of the content.

I can't wait to see how we finalize our approach to location-aware content, but I think this is a step in the right direction.

## Opportunities

I would like us to eventually move away from keys like `heading2` for something a bit more semantic, as we've found that tying the heading level to the content over time has become problematic. I know that's not necessarily tied to the PR, but something I'm beginning to notice.

Also, I'm wondering if we can add some unit tests here, focused on the message validation? Right now I think we only have stuff for the landing page, so the new validation behavior isn't checked. Things like:

```typescript
  it("rejects a locale payload missing a nested key", () => {
    // candidate.learn.categories[0].link.href removed
  });

  it("rejects a locale payload with extra keys", () => {
    // candidate.learn.unexpected = "extra"
  });

  it("rejects malformed later array entries", () => {
    // categories[0] valid, categories[1].link.href missing
  });

  it("rejects invalid literal/discriminator values", () => {
    // header.primaryItems[0].kind = "not-a-real-kind"
  });

  it("accepts a payload whose nested object and array structure exactly matches
  the base locale", () => {
    // happy path
  });
```

packages/static-site/domain/content/messageTypes.ts
aloverso: Good note! I called this out in the PR notes as an intentional choice.

> Decision: I think it'll be easier to manage / update the static typing for the JSON messages in a single file (rather than a file for landing, a file for learn, etc) - let me know what we think about this.

Rationale is that these static types are primarily for the computer to be reading, looking at, and managing so that humans can *use* the types in other places. So sticking a bunch of messy type declarations in a single file seems easier to ask the computer to update and whatnot, than spreading the mess out. Very loose opinion though, I could be swayed

packages/static-site/app/[locale]/learn/page.tsx
Comment (aloverso) [reply packages/static-site/app/[locale]/learn/page.tsx:39]: End of day, I might be missing something obvious, but which part might not work for location-specific content?

Comment (aloverso) [reply packages/static-site/app/[locale]/learn/page.tsx:39]: Per conversation: if we want to support localized pathnames, we should make it a separate feature ticket

packages/static-site/domain/content/messageTypes.ts
```diff
@@ -320,6 +320,36 @@ export interface LandingMetadataContent {
...
+ *
+ * This type defines a stable shape for related data.
+ */
+export interface LearnCategory {
+  /** Heading text for the category card. */
+  readonly title: string;
+  /** Supporting text rendered in the category card. */
+  readonly description: string;
+  /** Link metadata for the category card CTA. */
+  readonly link: LandingLink;
```
jphechter, line 334: I would rename `LandingLink`. I'm assuming it was originally named this because it was used on the Landing Page, but if we're going to use this type elsewhere, I'd rather it just be `Link` or something else that's more generic.

jphechter:
1 small comment. I think everything else in here makes sense to me. No additional feedback.

packages/static-site/domain/content/messageTypes.ts
aloverso: I actually already did this in the next PR that I'm working on right now. Are you okay with that refactor taking place then? Might be easier from the merge conflicts perspective.

PR #12775, opened 2026-05-12
feat: [AB#17713] add `ProgressBar` component for roadmap steps

hinzed1127:
## Description

Adds a progress bar to the "Plan" and "Start" section of the roadmap. Percentage a simple calculation of tasks completed/total tasks. Examples of the different visual phases:

**No Progress**

zero-progress

**Low Progress (<50%)**

low-progress

**Mid Progress (< 90%)**
 
mid-progress

**Complete (100%)**

complete-progress

### Ticket

This pull request resolves part of [#17713](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17713).

- For reference: [Figma designs](https://www.figma.com/design/H2mER7gGeHKTM3TgySmSvm/101-105-Business-Experience-Sprint-Designs?node-id=71388-101608&t=G35r5ymXh1INghh9-4)

### Approach

I basically copied this accepted Stack overflow answer, where it's 2 layers of gradients, but the "unfinished" portion of the progress bar is a single colored gradient (the CSS syntactically won't just allow a solid color, so that's the sort of hacky workaround).

### Steps to Test

Go through onboarding like normal and validate the progress bars update correctly

### Notes

This will get paired with ongoing work by @ashlsun , so it might ultimately get merged into her branch before going to `main`. TBD

## Code author checklist


web/src/styles/main.scss
hinzed1127, line 1: campground rules: just alphabetizing imports (assuming because it's scss/`@use` that it wasn't picked up by a linter)

web/src/styles/sass-files/variables.scss
```diff
@@ -108,6 +108,7 @@ $accent-semi-cool-extra-light: #effffb;
 $accent-semi-cool-lightest: #defff8;
 $accent-semi-cool-light: #bdfff2;
 $accent-semi-cool: #009f7f;
+$accent-semi-cool-400: #51d8c8;
```
hinzed1127, line 111: this was the only color in the designs that didn't exist as a variable, so I tried to slot it into a sensible spot within the existing palette

web/src/components/dashboard/SectionAccordion.tsx
```diff
@@ -58,13 +67,31 @@ export const SectionAccordion = (props: Props): ReactElement => {
             aria-controls={`${sectionName}-content`}
             id={`${sectionName}-header`}
             data-testid={`${sectionName}-header`}
+            sx={{ position: "relative" }}
```
hinzed1127, line 75: Our hands are tied a little. Since this is working within the MUI Accordion component, achieving the design requires some relative + absolutely positioned child trickery to get things right visually. This was my first time needing to leverage the [`sx` prop](https://mui.com/system/getting-started/the-sx-prop/), so TIL

web/src/styles/components/progress-bar.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.section-progress-bar-row {
+  margin-top: 1.5rem;
+}
+
+.section-progress-bar-label {
+  font-size: 0.9rem;
+  position: absolute;
+  bottom: 0;
+  right: $progress-bar-label-offset;
```
hinzed1127, line 13: the other half of the relative + absolute positioning

hinzed1127 (commented):
self-review

web/src/components/dashboard/SectionAccordion.tsx
seidior: Yeah, `sx` is the fourth of the four design systems I call out in Navigator as a frustration for working on front-end components.

web/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export function ProgressBar({
+  percentage = 0,
+  "data-testid": testId,
+}: {
+  percentage?: number;
+  "data-testid"?: string;
+}): ReactElement {
+  return (
+    <div
```
seidior, line 15: [dust/pebble] Any particular reason to use a `<div>` instead of `<progress>` so you get the semantic meaning and the appropriate ARIA `progressbar` role?

web/src/components/dashboard/Roadmap.tsx
```diff
@@ -22,8 +22,25 @@ export const Roadmap = (): ReactElement => {
...
+                task.stepNumber !== undefined && stepNumbersInSection.includes(task.stepNumber),
+            )
+          : [];
+        const completedCount = tasksInSection.filter(
+          (task) => business?.taskProgress[task.id] === "COMPLETED",
+        ).length;
+        const percentage =
+          tasksInSection.length > 0
+            ? Math.round((completedCount / tasksInSection.length) * 100)
+            : 0;
```
seidior: [sand] I'd love to see these pulled out into functions for readability, and so that you can save some of the interim values into semantic variable names in case this needs to be refactored later.

web/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,17 @@
+import type { CSSProperties, ReactElement } from "react";
+
+export function ProgressBar({
```
seidior: [sand] Why `export function` and not an arrow function?

web/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,17 @@
...
+  percentage = 0,
+  "data-testid": testId,
+}: {
+  percentage?: number;
+  "data-testid"?: string;
+}): ReactElement {
```
seidior: [sand] We generally extract these out into a named `interface` in other components right? It's been a while since I audited so I could be wrong on that.

web/src/components/dashboard/SectionAccordion.tsx
```diff
@@ -13,8 +14,16 @@ interface Props {
...
   isDividerDisabled?: boolean;
+  progressPercentage?: number;
 }
 
+const progressLabelStage = (percentage: number): string => {
+  if (percentage === 0) return "zero";
+  if (percentage < 50) return "low";
+  if (percentage < 90) return "mid";
+  return "high";
+};
```
seidior, line 30: [sand] Super nit-picky so feel free to avoid, but technically you're not returning a string, you're returning a union type here. This might help with tests or anything else that consumes it in the future.

seidior: (I would also extract out those numbers into variables so they can be named and changed later in case someone changes their mind. They don't need to be surfaced in the CMS though.)

seidior (commented):
Mostly an accessibility-related question, but it looks like you're otherwise matching some existing patterns.

I know this is for "plan" and "start" but how does this affect our more interesting roadmaps like domestic employer?

Also, will these still appear when the register section shows too, as implemented?

web/src/components/ProgressBar.tsx
ashlsun: I was just looking into this, and i wonder if `progressbar` is meant to communicate loading progress, as opposed to the user's progress.  (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/progressbar_role)

MUI's progress components are also about loading / waiting: https://mui.com/material-ui/react-progress/

Maybe what we're really going for is a `meter` role.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/meter_role

web/src/components/dashboard/SectionAccordion.tsx
ashlsun, line 1: I know that we use MiniRoadmap for the sidebar of our task pages, and MiniRoadmap uses this SectionAccordion component. The sidebar doesn't show the ProgressBar because progressPercentage is not defined, but I did notice that the line wrapping looks subtly different. I can't tell if it's due to some change in this file or how this would have been affected?
| dev  | local |
| ------------- | ------------- |
| Image | Image  |

web/src/components/ProgressBar.tsx
hinzed1127: actually, looking at the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress):

> The `<progress>` HTML element displays an indicator showing the completion progress of a task, typically displayed as a progress bar.

seems like a good fit, will update

web/src/styles/main.scss
hinzed1127: So doing this uncovered a bug that showed these styles that are not entirely isolated/modular. specifically, `base/typography` has styles overriding some header styles in `base/flex`. I'll have a more detailed breakdown of this later for a separate bug fix, but everything else appears to be ok here

web/src/components/dashboard/SectionAccordion.tsx
hinzed1127: I believe this is related to the [imports reordering discovery](https://github.com/newjersey/navigator.business.nj.gov/pull/12775#discussion_r3291001818), so resolving for now

PR #12776, opened 2026-05-12
feat: [AB#17525] transfer bucket ownership to CDK

aadedejifearless:
## Description

This PR transfers ownership and policy management of the relevant S3 bucket(s) to CDK and updates the bucket access model to align with Cognito-based roles instead of allowing access via AnyPrincipal().

1. Bucket Ownership Moved to CDK:  The bucket configuration is now fully managed in CDK, ensuring:
- Infrastructure is defined as code
- Policies are versioned and auditable
- Environment parity across dev/content/testing/staging/prod
- No drift between Terraform version and CDK

2. Updated the Bucket Policy To Use Cognito Roles: The previous policy was effectively open to AnyPrincipal (*) and relied entirely on conditions (Cognito identity prefix) to restrict access. The new policy explicitly grants access to the two roles  
   - `navigator_authRole`
   - `navigator_unauthRole`
This removes reliance on AnyPrincipal and scopes permissions to known IAM roles.
3. This improves the security as it replaced the broadly scoped principal with explicity role based access
   - Maintains Cognito sub-based object isolation.
   - Aligns with least-privilege principles.

### Ticket

This pull request resolves [#17525](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17525).

## Code author checklist

- [X] I have added the `pr-show` or `pr-review` label on GitHub to show or request reviews

### DIFF
--- a/api/cdk/lib/storageStack.ts
+++ b/api/cdk/lib/storageStack.ts
@@ -1,5 +1,6 @@
 import { RemovalPolicy, Stack, StackProps } from "aws-cdk-lib";
 import * as s3 from "aws-cdk-lib/aws-s3";
+import * as iam from "aws-cdk-lib/aws-iam";
 import { Construct } from "constructs";
 import { DEV_STAGE, LOWER_STAGES } from "./constants";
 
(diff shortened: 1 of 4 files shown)


api/cdk/lib/storageStack.ts
```diff
@@ -10,12 +11,62 @@ export interface StorageStackProps extends StackProps {
...
   public readonly intercomMacrosBucket?: s3.Bucket;
+  public readonly userDocumentsBucket: s3.Bucket;
+  public readonly userDocumentsLocalBucket?: s3.Bucket;
 
   constructor(scope: Construct, id: string, props: StorageStackProps) {
     super(scope, id, props);
 
     const isLowerEnv = LOWER_STAGES.includes(props.stage);
     const removalPolicy = isLowerEnv ? RemovalPolicy.DESTROY : RemovalPolicy.RETAIN;
+    const cognitoPrefix = "${cognito-identity.amazonaws.com:sub}"; // eslint-disable-line no-useless-escape
```
ashlsun, line 22: `no-useless-escape` is for "\\" characters right? can we safely remove this comment?

api/cdk/lib/storageStack.ts
```diff
@@ -10,12 +11,62 @@ export interface StorageStackProps extends StackProps {
...
+
+    this.userDocumentsBucket.addToResourcePolicy(
+      new iam.PolicyStatement({
+        sid: "ListYourObjects",
+        principals: [authRole, unauthRole],
+        actions: ["s3:ListBucket"],
+        resources: [this.userDocumentsBucket.bucketArn],
+        conditions: {
+          StringLike: {
+            "s3:prefix": [`/${cognitoPrefix}`, `/${cognitoPrefix}/*`],
```
ashlsun: are these leading slashes supposed to be here?

ashlsun (approved):
Are you manually running `cdk import` or importing through the console? I saw [on a AWS how-to guide](https://aws.amazon.com/blogs/devops/how-to-import-existing-resources-into-aws-cdk-stacks/) that this is a necessary step but not something that can happen purely in the code.

api/cdk/lib/storageStack.ts
```diff
@@ -10,12 +11,62 @@ export interface StorageStackProps extends StackProps {
...
+          },
+        },
+      }),
+    );
+
+    this.userDocumentsBucket.addToResourcePolicy(
+      new iam.PolicyStatement({
+        sid: "ReadDeleteYourObjects",
+        principals: [authRole, unauthRole],
+        actions: ["s3:GetObject", "s3:DeleteObject"],
```
seidior, line 63: [dust] Have we previously been granting `DeleteObject` permissions? While it's still restricted to the roles so I'm not overly worried, I'm unaware of any code that leverages DeleteObject permissions.

api/cdk/lib/storageStack.ts
seidior: [dust] Same question as Ashley. If the app writes keys under `{identityID}/formationDoc-...pdf` and we're using the path `/${cognito-identity.amazonaws.com:sub}`, then wouldn't this break the S3 paths, since S3 list prefixes are key prefixes and not absolute paths? Not that we leverage `ListBucket` behaviors within our app, but this may impact our ability to query for objects we have access to, no?

seidior (commented):
Directionally, I like the way you're thinking, grouping these policies around roles. I have a few questions about the implementation though.

api/cdk/lib/storageStack.ts
aadedejifearless: Good catch, S3 prefixes are indeed key prefixes, not absolute paths, so the leading / was incorrect and has been removed. The prefix now correctly matches the object structure used in the app ({identityID}/formationDoc-...pdf).

We still do include s3:ListBucket in the policy in this change

api/cdk/lib/storageStack.ts
aadedejifearless: Yes, we’ve historically been granting s3:DeleteObject in this policy, alongside GetObject, scoped to the identity prefix.

Like you said @seidior  don’t see any current application code that explicitly relies on object deletion from S3, so it’s likely not actively used in practice today. That said, it was included in the original Terraform policy and has been carried forward to preserve parity with the existing behavior.

We can remove DeleteObject if we want to tighten permissions further, but I’d treat that as a deliberate decision in a future PR

PR #12785, opened 2026-05-14
chore: add OIDC and deploy role for warehouse

aadedejifearless:
## Description
This change configures GitHub OIDC authentication for the Business Data Warehouse deployment by importing the existing OIDC provider in the AWS accounts and introducing a dedicated GitHub role that assumes the warehouseDeployRole.

**What This Does**
- Imports the existing GitHub OIDC provider in accounts where it already exists.
- Avoids recreating account-level OIDC infrastructure.
- Creates a new GitHub OIDC entry role used by GitHub Actions.
- Grants that role permission to assume the warehouseDeployRole.
- The warehouseDeployRole contains the actual deployment permissions for the Business Data Warehouse stack.

**Deployment Flow**

The GitHub Actions Pipeline will:
1. Authenticate using the existing OIDC provider.
2. Assume the GitHub OIDC role.
3. That role will assume the warehouseDeployRole.
4. The warehouseDeployRole performs infrastructure deployments.

### Ticket

## Code author checklist

### DIFF
--- a/api/cdk/lib/iamStack.ts
+++ b/api/cdk/lib/iamStack.ts
@@ -25,6 +25,7 @@ export class IamStack extends Stack {
   public readonly taxKMSRole?: iam.Role;
   public readonly authRole?: iam.Role;
   public readonly unauthRole?: iam.Role;
+  public readonly githubOidcRole?: iam.Role;
 
   constructor(scope: Construct, id: string, props: IamStackProps) {
     super(scope, id, props);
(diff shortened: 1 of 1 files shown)


api/cdk/lib/iamStack.ts
```diff
@@ -34,6 +34,71 @@ export class IamStack extends Stack {
...
+          },
+          "sts:AssumeRoleWithWebIdentity",
+        ),
+      });
+
+      githubOidcRole.addToPolicy(
+        new iam.PolicyStatement({
+          actions: ["sts:AssumeRole"],
+          resources: [
+            `arn:aws:iam::${this.account}:role/njbfs-business-data-warehouse-deploy-${props.stage}`,
```
jtabron1: The role ARN here is stage-suffixed (-${props.stage}), but the deploy role created below has roleName: njbfs-business-data-warehouse-deploy with no stage suffix. The OIDC role's sts:AssumeRole will fail because the named role won't match.

I suggest a change to the deploy role's roleName to match, e.g.: "roleName: `njbfs-business-data-warehouse-deploy-${props.stage}`,"

api/cdk/lib/iamStack.ts
```diff
@@ -34,6 +34,71 @@ export class IamStack extends Stack {
...
+            "athena:*",
+            "sns:*",
+            "events:*",
+            "cloudwatch:*",
+            "logs:*",
+            "kms:*",
+            "lambda:*",
+            "cloudformation:*",
+            "states:*",
+            "iam:PassRole",
```
jtabron1: Our warehouse CDK stacks create IAM roles (Glue service role, Lambda execution role, state machine role) so the deploy will fail at the first stack creating a new role.

If possible, would you mind adding:
iam:CreateRole, iam:DeleteRole, iam:AttachRolePolicy, iam:DetachRolePolicy, iam:PutRolePolicy, iam:DeleteRolePolicy, iam:UpdateRole, iam:UpdateAssumeRolePolicy, iam:GetRole, iam:GetRolePolicy, iam:ListRolePolicies, iam:ListAttachedRolePolicies, iam:TagRole, iam:UntagRole.

jtabron1 (changes requested):

api/cdk/lib/iamStack.ts
aadedejifearless: Added these permissions. In a future PR, we should tighten the resource scope to reduce the breadth of access currently granted to this role.

api/cdk/lib/iamStack.ts
aadedejifearless: Good catch @jtabron1 , I've  updated this to include the stage suffix.
Each account will have its own OIDC role, which is used to assume the corresponding warehouse deploy roles. Those deploy roles are scoped per stage, so access is explicitly segmented by environmen

jtabron1 (approved):
Both fixes look good. Thanks for the cleanup. Two notes:

The deployRoleAllowlist pattern is a nice touch, accommodates the dev sub-environments without us needing extra config. Appreciate that.
The added iam:CreateServiceLinkedRole is a good catch; Step Functions needs that on first deploy.

PR #12790, opened 2026-05-15
fix: remove cross-stack ECR export causing static-site deploy rollback

seidior:
## Description

The static-site deploy was rolling back on every run due to a CloudFormation cross-stack export dependency that CDK introduced automatically when both stacks were synthesized together.

### Approach

The static-site deploy was failing with `UPDATE_ROLLBACK_COMPLETE` because CDK was generating an automatic CloudFormation cross-stack export (`ExportsOutputRefStaticSiteRepository...`) in `StaticSiteRepositoryStack` when both stacks were synthesized together (i.e. when `STATIC_SITE_IMAGE_TAG` is set). On the next run, when only the repository stack is synthesized (without `STATIC_SITE_IMAGE_TAG`), that export disappears from the template — and CloudFormation rejects the update because `StaticSiteServiceStack` still imports it.

Fix: `StaticSiteServiceStack` now looks up the ECR repository by its deterministic name via `ecr.Repository.fromRepositoryName` instead of receiving it as a CDK object reference from `app.ts`. This eliminates the cross-stack dependency entirely. The unused `CfnOutput` with `exportName` in `StaticSiteRepositoryStack` was also removed.

Two regression tests were added to prevent reintroduction:
- `StaticSiteRepositoryStack` asserts it has no CloudFormation `Outputs`.
- `StaticSiteServiceStack` asserts it contains no `Fn::ImportValue` references.

### Steps to Test

Deploy to dev and confirm the `Deploy Static Site ECR Repository` step completes without rollback.

### Notes

The root cause was that the deploy action runs two separate CDK synth/deploy passes with different env contexts (step 1: no `STATIC_SITE_IMAGE_TAG`; step 3: with `STATIC_SITE_IMAGE_TAG`), so any cross-stack reference that only appears in the combined synthesis will orphan itself on the next isolated deploy.

## Code author checklist

### DIFF
--- a/api/cdk/bin/app.ts
+++ b/api/cdk/bin/app.ts
@@ -35,20 +35,15 @@ const env = {
   region: region,
 };
 
-const staticSiteRepositoryStack = new StaticSiteRepositoryStack(
-  app,
-  `StaticSiteRepositoryStack-${stage}`,
-  {
-    stage,
-    env,
-  },
-);
+new StaticSiteRepositoryStack(app, `StaticSiteRepositoryStack-${stage}`, {
+  stage,
+  env,
+});
 
 if (isStaticSiteDeploy) {
   new StaticSiteServiceStack(app, `StaticSiteServiceStack-${stage}`, {
     stage,
     env,
-    repository: staticSiteRepositoryStack.repository,
     imageTag: staticSiteImageTag,
   });
 }
--- a/api/cdk/lib/staticSiteRepositoryStack.ts
+++ b/api/cdk/lib/staticSiteRepositoryStack.ts
@@ -1,4 +1,4 @@
-import { CfnOutput, Duration, RemovalPolicy, Stack, StackProps } from "aws-cdk-lib";
+import { Duration, RemovalPolicy, Stack, StackProps } from "aws-cdk-lib";
 import * as ecr from "aws-cdk-lib/aws-ecr";
 import { Construct } from "constructs";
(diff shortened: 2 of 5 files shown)


seidior (commented):
Superseded by the corrected self-review context note below.

seidior (commented):
Self-review context notes; no blocker findings from the current diff.

- `StaticSiteRepositoryStack` is still instantiated in `app.ts` during the service deploy path. That is intentional: the deploy action still needs the repository stack as its own deploy target, but the service stack no longer receives the repository construct, so CDK no longer emits repository-stack exports/imports when both stacks are in the app.
- `ecr.Repository.fromRepositoryName(...)` is not an AWS lookup. CDK derives the ECR ARN/URI from the target account, region, and deterministic repository name, so the task definition and execution-role pull policy still synthesize without `Fn::ImportValue`.
- Removing `StaticSiteRepositoryUri` should not affect the static-site deploy action because the action already computes the image URI from `AWS_ACCOUNT_ID`, `AWS_REGION`, `STAGE`, and the image tag before pushing. I also checked CloudFormation in Innov-Dev earlier: `bfs-static-site-dev-repository-uri` had no importing stacks.
- The service-stack regression test now synthesizes the real repository stack and service stack in the same app on purpose. That recreates the deploy context that originally triggered CDK automatic cross-stack exports, so asserting the repository template still has no `Outputs` protects against this exact rollback path coming back.

PR #12792, opened 2026-05-18
feat: add API gateway load balancers

seidior:
## Description

Adds internal load balancers in front of the CDK-managed API Gateway REST APIs while intentionally keeping the existing API Gateway endpoint configuration unchanged. This gives us a stable load-balancer boundary now, without combining that infrastructure change with the larger behavior change of moving the APIs from their current public endpoint setup to private REST API endpoints.

The proxy Lambda is needed for the interim state because an Application Load Balancer forwards to registered targets, not directly to the service-managed public API Gateway hostname. Registering public API Gateway IPs would be brittle because those addresses are owned and changed by AWS, and switching directly to private API Gateway endpoints in this PR would create more deployment churn than necessary. Instead, the ALB forwards to a narrow Lambda target that preserves the incoming request shape and calls the deployed REST API stage URL. Later, when we intentionally switch the REST APIs to private, the ALB routing can be moved to the private REST API/VPC endpoint pattern.

### Ticket

This pull request partially resolves [#17509](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17509). The other changes needed in this ticket are infrastructure tickets with our partners, then there will be a follow-up ticket to resolve this work by switching these API Gateways private and these load balancers to those API Gateways directly.

### Approach

- Adds an internal HTTPS Application Load Balancer to non-local API stacks using the existing imported VPC, subnet, security group, and API Gateway certificate environment values.
- Adds a dedicated `apiGatewayLoadBalancerProxy` Lambda target that forwards ALB requests to the deployed REST API stage URL.
- Exports the API load balancer DNS name, target group ARN, and listener ARN for future DNS/WAF/private endpoint wiring.
- Adds CDK assertions for REST API usage, ALB creation, Lambda target routing, and local-stage exclusion.
- Adds unit tests for proxy request forwarding, base64 body handling, multi-value query strings, and upstream failure handling.

### Steps to Test

- `yarn workspace @businessnjgovnavigator/api-cdk test apiStack.test.ts --runInBand`
- `yarn test api/src/functions/apiGatewayLoadBalancerProxy/app.test.ts --selectProjects api --runInBand`
- `yarn workspace @businessnjgovnavigator/api-cdk build`
- `yarn workspace @businessnjgovnavigator/api typecheck`

## Code author checklist

### DIFF
--- a/api/cdk/lib/apiStack.ts
+++ b/api/cdk/lib/apiStack.ts
@@ -3,22 +3,108 @@ import {
   REMINDER_EMAIL_CONFIG_SET_BASE,
   WELCOME_EMAIL_CONFIG_SET_BASE,
 } from "@businessnjgovnavigator/api/src/libs/constants";
-import { CfnOutput, Stack, StackProps } from "aws-cdk-lib";
+import { CfnOutput, Duration, Stack, StackProps } from "aws-cdk-lib";
 import * as acm from "aws-cdk-lib/aws-certificatemanager";
 import * as apigateway from "aws-cdk-lib/aws-apigateway";
 import * as cognito from "aws-cdk-lib/aws-cognito";
-import { IFunction } from "aws-cdk-lib/aws-lambda";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
+import * as elbv2 from "aws-cdk-lib/aws-elasticloadbalancingv2";
+import * as elbv2Targets from "aws-cdk-lib/aws-elasticloadbalancingv2-targets";
+import * as iam from "aws-cdk-lib/aws-iam";
+import { Runtime } from "aws-cdk-lib/aws-lambda";
+import type { IFunction } from "aws-cdk-lib/aws-lambda";
 import { Construct } from "constructs";
-import { applyStandardTags, attachLambdaToResource, createSesConfigSet } from "./stackUtils";
+import path from "node:path";
+import {
+  applyStandardTags,
+  attachLambdaToResource,
+  createLambda,
+  createSesConfigSet,
+} from "./stackUtils";
 
-interface LambdaConfig {
-  lambda?: IFunction;
+/** Lambda function configuration consumed by API Gateway routes. */
+export interface LambdaConfig {
+  /** Lambda function to attach to API Gateway routes; optional for stage-specific routes. */
+  readonly lambda?: IFunction;
 }
 
-interface ApiStackProps extends StackProps {
-  stage: string;
-  lambdas: Record<string, LambdaConfig>;
+/** Properties required to create the API Gateway stack. */
+export interface ApiStackProps extends StackProps {
+  /** Deployment stage used for API names, tags, and stage-specific routing. */
+  readonly stage: string;
+
+  /** Shared Lambda IAM role used by API-owned Lambda functions. */
+  readonly lambdaRole: iam.Role;
+
+  /** Lambda functions used by API Gateway integrations. */
+  readonly lambdas: Record<string, LambdaConfig>;
 }
+
+interface RequiredEnvironmentValue {
+  /** Environment variable name to validate. */
+  readonly name: string;
+
+  /** Environment variable value to validate. */
+  readonly value: string | undefined;
+}
+
+interface ApiLoadBalancerNetwork {
+  /** Imported VPC that contains the internal API load balancer. */
+  readonly vpc: ec2.IVpc;
+
+  /** Imported private subnets used by the internal API load balancer. */
+  readonly subnets: ec2.ISubnet[];
+
+  /** Existing security group attached to the internal API load balancer. */
+  readonly securityGroup: ec2.ISecurityGroup;
+}
+
+interface CustomDomainConfig {
+  /** Custom domain name assigned to the API Gateway stage. */
+  readonly domainName: string;
+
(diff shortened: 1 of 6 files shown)


api/cdk/test/apiStack.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  });
+
+  test("creates a REST API without HTTP API resources", () => {
+    const { template } = createApiStackTemplate("dev");
+
+    expect(template.toJSON()).toBeDefined();
+    template.hasResourceProperties("AWS::ApiGateway::RestApi", {
+      Name: `${API_SERVICE_NAME}-dev`,
+    });
+    template.resourceCountIs("AWS::ApiGatewayV2::Api", 0);
```
seidior, line 93: Self-review context: this assertion is intentionally about API shape, not endpoint type. The private endpoint migration path depends on these staying REST APIs rather than HTTP APIs, but this PR avoids changing the existing edge/regional endpoint behavior to keep deployment churn low and availability maximized in the interim.

api/cdk/lib/apiStack.ts
```diff
@@ -94,6 +180,7 @@ export class ApiStack extends Stack {
     this.setupApiRoutes(expressLambda!, githubLambda);
 
     this.configureCustomDomain(props.stage);
+    this.configureApiLoadBalancer(props.stage, props.lambdaRole);
```
seidior, line 183: Self-review context: the ALB is wired after the existing API Gateway routes and custom domain so it is an additional access path, not a replacement for the current API Gateway exposure. That lets us deploy and validate the load-balancer boundary before the later private REST API cutover.

api/cdk/lib/apiStack.ts
```diff
@@ -237,4 +324,142 @@ export class ApiStack extends Stack {
...
+      role: lambdaRole,
+      id: `${API_SERVICE_NAME}-${stage}-apiGatewayLoadBalancerProxy`,
+      stage,
+      functionName: `${API_SERVICE_NAME}-${stage}-apiGatewayLoadBalancerProxy`,
+      entry: path.join(__dirname, "../../src/functions/apiGatewayLoadBalancerProxy/app.ts"),
+      handler: "handler",
+      runtime: Runtime.NODEJS_22_X,
+      timeout: Duration.seconds(API_LOAD_BALANCER_PROXY_TIMEOUT_SECONDS),
+      environment: {
+        API_GATEWAY_PROXY_BASE_URL: this.restApi.url,
```
seidior, line 413: Self-review context: using `this.restApi.url` keeps the proxy pointed at the synthesized REST API stage URL instead of introducing another manually-maintained environment value. When the REST API changes stage/API ID during deploys, the proxy follows the CDK reference.

api/cdk/lib/apiStack.ts
```diff
@@ -237,4 +324,142 @@ export class ApiStack extends Stack {
...
+    });
+  }
+
+  private createLoadBalancerTargetGroup(
+    stage: string,
+    proxyLambda: IFunction,
+  ): elbv2.ApplicationTargetGroup {
+    const targetGroup = new elbv2.ApplicationTargetGroup(this, "ApiLoadBalancerTargetGroup", {
+      targetGroupName: createApiLoadBalancerTargetGroupName(stage),
+      targetType: elbv2.TargetType.LAMBDA,
```
seidior, line 424: Self-review context: the Lambda target is the interim bridge because ALBs forward to registered targets, and the public API Gateway hostname is AWS-managed rather than a stable target group member. Registering resolved API Gateway IPs would be brittle; later private REST API work can replace this with the VPC endpoint target pattern.

api/src/functions/apiGatewayLoadBalancerProxy/app.ts
```diff
@@ -0,0 +1,276 @@
...
+  readonly apiGatewayBaseUrl: string;
+
+  /** Fetch-compatible HTTP client used to send the upstream request. */
+  readonly fetchRequest: FetchRequest;
+}
+
+/**
+ * Header names that cannot be safely forwarded through a reverse proxy.
+ */
+const NON_FORWARDABLE_HEADER_NAMES = new Set([
```
seidior, line 72: Self-review context: stripping hop-by-hop and authority headers keeps the proxy from forwarding ALB-specific connection state to API Gateway. In particular, preserving `host` or `content-length` here can create incorrect upstream routing or body framing once the request is rebuilt.

api/src/functions/apiGatewayLoadBalancerProxy/app.ts
```diff
@@ -0,0 +1,276 @@
...
+      targetUrl.searchParams.append(name, value);
+    }
+  }
+};
+
+const createTargetUrl = (apiGatewayBaseUrl: string, event: ApiGatewayLoadBalancerEvent): URL => {
+  const normalizedBaseUrl = normalizeApiGatewayBaseUrl(apiGatewayBaseUrl);
+  const targetUrl = new URL(normalizeRequestPath(event.path), normalizedBaseUrl);
+
+  if (event.multiValueQueryStringParameters !== undefined) {
```
seidior, line 138: Self-review context: when ALB provides multi-value query parameters, those are the more precise representation of the original request. Preferring them avoids collapsing repeated keys before the request reaches the REST API.

seidior (commented):
self-review

PR #12843, opened 2026-06-15
feat: add language picker + RTL foundation

seidior:
## Description

Adds a USWDS/NJWDS-based language picker to the locale-aware static site, then deliberately reduces the shipped locale set to English (US) and Spanish (US) while leaving the multilingual machinery — including right-to-left (RTL) support — intact and ready to switch back on.

The picker follows the USWDS language-selector "3+ languages" variant: a "Languages" dropdown where each option shows native + English name (`Español (Spanish)`), ordered alphabetically by native name, with no flags and a scrolling submenu past six languages. Accessibility was a primary constraint throughout (WCAG 2.1 AA): per-option `lang` (H58), `aria-current` for the active locale, `aria-expanded`/`aria-controls` on the trigger, and bidi-correct rendering of Latin proper nouns in either direction.

### Ticket

This pull request resolves [#17824](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17824).

### Approach

- **One locale source of truth.** `domain/i18n/locales.ts` declares `APP_LOCALES` (`as const`); `AppLocale` derives from it, so adding/removing a locale ripples through types, routing, sitemap, and message loading. `domain/i18n/languages.ts` holds the `LanguageDescriptor` records and **derives** display order by sorting on native name, so the "alphabetical" rule can't drift.
- **Stateless picker.** `LanguageSwitcher.tsx` is pure presentation rendering the `usa-language` markup; open/close/focus is driven by the bundled NJWDS `uswds.min.js` (no React state). A `descriptors` prop (defaulting to the real list) decouples it from the app's locale set so tests/harness can feed synthetic lists; a `submenuId` prop lets multiple instances coexist with unique `aria-controls`.
- **Clean URLs.** `next-intl` routing uses `localePrefix: "as-needed"` — the default locale is unprefixed (`/learn`), others prefixed (`/es-US/learn`). Pure, idempotent path helpers (`localePath.ts`) are reused by routing, the sitemap, and `hreflang`/`x-default` metadata. Switching language re-targets the same page rather than the home page.
- **RTL foundation.** The synced NJWDS stylesheet ships zero RTL support, so setting `dir="rtl"` flips text but repositions nothing. The fix is layered by code ownership: document `dir` from the descriptor; our own CSS converted to logical properties; a small composable `nj-rtl-utilities.css` (logical properties, loaded after NJWDS so source order wins without `!important`) for NJWDS's positional bugs; and `<bdi dir="ltr">` isolation for inherently-Latin strings. `docs/njwds-rtl-feedback.md` catalogues the upstream gaps.
- **RTL kept dormant, not deleted.** The locale set is reduced to en/es, but the RTL infra (utilities, `<bdi>`, placeholder `ar-US.json`, feedback doc) stays in the tree; `/ar-US`-dependent e2e tests are `.skip`'d with re-enable notes so reintroducing an RTL language is turnkey.

### Steps to Test

1. Run the static site (`pnpm dev` in `packages/static-site`).
2. Open the language picker in the header — confirm it's labeled "Languages", lists `English` and `Español (Spanish)`, and toggles open/closed via keyboard and mouse.
3. Select Español and confirm the URL gains an `/es-US` prefix, the page content switches, and you land on the same page (not the home page).
4. Confirm `<html lang>` updates and the language-prompt modal behaves.
5. Run `pnpm lint`, `pnpm typecheck`, `pnpm test`, `pnpm test:e2e`, `pnpm test:accessibility`, and `pnpm build` — all green.

### Notes

- A test-only harness route (`app/[locale]/language-switcher-harness`, `notFound()` in production) renders N synthetic options so the scroll-past-six behavior can be measured in a real browser, since jsdom has no layout engine.
- Playwright browsers install via `scripts/_install.sh` (dev/CI only), not a package `postinstall`, to keep ~150MB out of production Docker builds.
- Pre-existing, locale-independent a11y defects were fixed along the way: inline-link contrast (restored prose underlines), a duplicate unnamed `role="dialog"`, and an orphaned `aria-modal` on the NJWDS modal wrapper.
- To re-enable RTL later: restore an RTL locale to `APP_LOCALES` + a descriptor (`textDirection: "rtl"`), wire its messages, and remove the `.skip`s.

## Code author checklist

### DIFF
--- a/packages/static-site/app/[locale]/language-switcher-harness/page.tsx
+++ b/packages/static-site/app/[locale]/language-switcher-harness/page.tsx
@@ -1,0 +1,109 @@
+/**
+ * Test-only harness that renders the language switcher with a controlled
+ * number of options so e2e tests can assert scroll behavior at specific counts.
+ *
+ * This route is gated out of production builds via `notFound()`, so it never
+ * ships to real users. It exists solely to bridge the "inject a specific option
+ * count" need with a real browser, where scrollHeight/clientHeight are
+ * meaningful (jsdom has no layout engine).
+ */
+
+import { notFound } from "next/navigation";
+
+import { LanguageSwitcher } from "@/components/landing/LanguageSwitcher";
+import type { LanguageDescriptor } from "@/domain/i18n/languages";
+import { type AppLocale, hasAppLocale } from "@/domain/i18n/locales";
+import { getApplicationMessages } from "@/domain/i18n/messages";
+
+/** Upper bound on synthetic options, generous enough to exercise scrolling. */
+const MAX_SYNTHETIC_OPTIONS = 20;
+
+/**
+ * Describes route params and query for the harness page.
+ */
+interface HarnessPageProps {
+  /** Async locale route params provided by Next.js. */
+  readonly params: Promise<{ readonly locale: string }>;
+  /** Async query params; `count` selects how many languages to render. */
+  readonly searchParams: Promise<{ readonly count?: string }>;
+}
+
+/**
+ * Clamps the requested option count into the supported synthetic range.
+ *
+ * @param rawCount Raw `count` query value, possibly undefined or invalid.
+ * @returns A count between 1 and {@link MAX_SYNTHETIC_OPTIONS}.
+ */
+const resolveCount = (rawCount: string | undefined): number => {
+  const parsed = Number.parseInt(rawCount ?? "", 10);
+
+  if (Number.isNaN(parsed) || parsed < 1) {
+    return MAX_SYNTHETIC_OPTIONS;
+  }
+
+  return Math.min(parsed, MAX_SYNTHETIC_OPTIONS);
+};
+
+/**
+ * Builds a list of synthetic language descriptors for scroll measurement.
+ *
+ * The switcher is stateless, so its scroll behavior depends only on how many
+ * options it renders, not on the app's real language set. Generating options
+ * here keeps the scroll e2e tests independent of the supported locales. The
+ * synthetic `locale` values are display-only (the harness never navigates), so
+ * they are cast to {@link AppLocale} to satisfy the descriptor type.
+ *
+ * @param count How many synthetic options to produce.
+ * @returns A descriptor list of the requested length.
+ */
+const buildSyntheticDescriptors = (count: number): readonly LanguageDescriptor[] => {
+  return Array.from({ length: count }, (_unused, index) => {
+    return {
+      locale: `synthetic-${index}` as AppLocale,
+      nativeName: `Language ${index + 1}`,
+      englishName: `Language ${index + 1}`,
+      regionName: "United States",
+      textDirection: "ltr",
+    } satisfies LanguageDescriptor;
+  });
+};
+
+/**
+ * Renders the language switcher with a controlled number of options.
+ *
+ * @param props Route props with locale params and `count` query.
+ * @returns The harness markup, or `notFound()` in production / for bad locales.
+ */
+const LanguageSwitcherHarnessPage = async ({ params, searchParams }: HarnessPageProps) => {
+  // biome-ignore lint/style/noProcessEnv: gate this test-only route out of production builds.
+  if (process.env.NODE_ENV === "production") {
+    notFound();
+  }
+
+  const { locale } = await params;
+
+  if (!hasAppLocale(locale)) {
+    notFound();
+  }
+
+  const { count } = await searchParams;
+  const descriptors = buildSyntheticDescriptors(resolveCount(count));
+  const messages = getApplicationMessages({ locale });
+
+  return (
+    <div className="grid-container usa-section">
+      <h1>Language switcher harness</h1>
+      <LanguageSwitcher
+        buttonLabel={messages.layout.languageSwitcher.buttonLabel}
+        currentLanguageLabel={messages.layout.languageSwitcher.currentLanguageLabel}
+        currentLocale={locale}
+        descriptors={descriptors}
+        navigationAriaLabel={messages.layout.languageSwitcher.navigationAriaLabel}
+        pathname="/language-switcher-harness"
+        submenuId="harness-language-switcher-submenu"
+      />
+    </div>
+  );
+};
+
+export default LanguageSwitcherHarnessPage;
--- a/packages/static-site/app/[locale]/layout.tsx
+++ b/packages/static-site/app/[locale]/layout.tsx
@@ -17,11 +17,14 @@ import { GoogleTagManager } from "@/components/analytics/GoogleTagManager";
 import { Intercom } from "@/components/analytics/Intercom";
 import { GovBanner } from "@/components/landing/GovBanner";
 import { IdentifierSection } from "@/components/landing/IdentifierSection";
+import { LanguagePromptModal } from "@/components/landing/LanguagePromptModal";
 import { SiteFooter } from "@/components/landing/SiteFooter";
 import { SiteHeader } from "@/components/landing/SiteHeader";
 import { SkipNav } from "@/components/landing/SkipNav";
+import { textDirectionForLocale } from "@/domain/i18n/languages";
 import { APP_LOCALES, hasAppLocale } from "@/domain/i18n/locales";
 import { getApplicationMessages } from "@/domain/i18n/messages";
+import { SITE_BASE_URL } from "@/domain/siteConfig";
 
 /**
  * Stores the logo path used in metadata icons and social previews.
@@ -29,14 +32,18 @@ import { getApplicationMessages } from "@/domain/i18n/messages";
 const NJ_LOGO_IMAGE_PATH = "/assets/njwds/dist/img/nj-logo-gray-20.png";
 
 /**
- * Stores the canonical production URL for metadata URL resolution.
- */
-const METADATA_BASE_URL = "https://next.business.nj.gov";
-
-/**
  * Public path for the synced NJWDS stylesheet.
  */
 const NJWDS_STYLESHEET_PATH = "/assets/njwds/dist/css/styles.css";
+
+/**
+ * Public path for our flow-relative (RTL-aware) utility classes.
+ *
+ * Loaded immediately after the NJWDS stylesheet so these utilities win on
+ * source order at equal specificity, letting them flip NJWDS's physical
+ * directional properties under `dir="rtl"` without `!important`.
+ */
+const NJ_RTL_UTILITIES_STYLESHEET_PATH = "/styles/nj-rtl-utilities.css";
 
 /**
  * Public path for the synced NJWDS runtime script.
@@ -116,7 +123,7 @@ export const generateMetadata = async ({ params }: GenerateMetadataProps): Promi
   return {
     title: messages.metadata.title,
     description: messages.metadata.description,
-    metadataBase: new URL(METADATA_BASE_URL),
+    metadataBase: new URL(SITE_BASE_URL),
     icons: {
       icon: NJ_LOGO_IMAGE_PATH,
       shortcut: NJ_LOGO_IMAGE_PATH,
@@ -170,9 +177,10 @@ const LocaleLayout = async ({ children, params }: LocaleLayoutProps) => {
   const messages = getApplicationMessages({ locale });
 
   return (
-    <html lang={locale}>
+    <html dir={textDirectionForLocale(locale)} lang={locale}>
       <head>
         <link href={NJWDS_STYLESHEET_PATH} rel="stylesheet" />
+        <link href={NJ_RTL_UTILITIES_STYLESHEET_PATH} rel="stylesheet" />
       </head>
       <body>
         <GoogleTagManager />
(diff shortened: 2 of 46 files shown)


packages/static-site/domain/i18n/routing.ts
```diff
@@ -17,5 +17,10 @@ import { APP_LOCALES, DEFAULT_LOCALE } from "./locales";
 export const routing = defineRouting({
   locales: APP_LOCALES,
   defaultLocale: DEFAULT_LOCALE,
-  localePrefix: "always",
+  // `as-needed` leaves the default locale (en-US) unprefixed (`/page/x`) while
+  // prefixing every other locale (`/es-US/page/x`).
+  localePrefix: "as-needed",
```
seidior, line 22: Switched from `always` to `as-needed` so en-US stays unprefixed (`/learn` instead of `/en-US/learn`). Keeps the canonical URLs clean for the locale most of our visitors use. The prefix helpers in `localePath.ts` are the one place that actually knows about this convention — sitemap and hreflang both lean on them, so there's no second source of truth to drift.

packages/static-site/domain/i18n/routing.ts
```diff
@@ -17,5 +17,10 @@ import { APP_LOCALES, DEFAULT_LOCALE } from "./locales";
...
+  // The preferred-language modal handles locale suggestion client-side, so the
+  // middleware must never silently redirect based on cookie or accept-language.
+  localeDetection: false,
```
seidior, line 25: Turned `localeDetection` off on purpose. With it on, next-intl redirects based on the accept-language header / cookie before our preferred-language modal ever gets a say, and the two end up fighting over the same decision. The modal owns locale suggestion now, so I wanted the middleware to stay dumb about it.

packages/static-site/domain/i18n/languages.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Every selectable language, ordered alphabetically by native name.
+ *
+ * NJWDS specifies that a 3+ language selector orders options "alphabetically by
+ * the common, native language name", so the display order is derived here
+ * rather than hand-maintained. The order controls how options appear in the
+ * language switcher.
+ */
+export const LANGUAGE_DESCRIPTORS: readonly LanguageDescriptor[] = [
+  ...LANGUAGE_DESCRIPTORS_BY_LOCALE,
+].sort((a, b) => a.nativeName.localeCompare(b.nativeName));
```
seidior, line 62: NJWDS wants the 3+ language selector ordered alphabetically by native name, so I'm deriving the order here rather than trusting array order. I checked every consumer first — they all `.find` or iterate, nothing assumed "English is index 0" — so sorting was safe. Means we can't accidentally break the ordering rule by dropping a new language in the wrong spot.

packages/static-site/domain/i18n/languages.ts
```diff
@@ -0,0 +1,84 @@
...
+ * textDirectionForLocale("ar-US"); // "rtl"
+ * textDirectionForLocale("en-US"); // "ltr"
+ * ```
+ */
+export const textDirectionForLocale = (locale: AppLocale): "ltr" | "rtl" => {
+  const descriptor = LANGUAGE_DESCRIPTORS.find((entry) => entry.locale === locale);
+
+  return (
+    descriptor?.textDirection ??
+    LANGUAGE_DESCRIPTORS.find((entry) => entry.locale === DEFAULT_LOCALE)?.textDirection ??
```
seidior, line 81: The double fallback looks paranoid but it's deliberate: unknown locale → fall back to the default locale's direction → `ltr` as a last resort. Since `locale` is already an `AppLocale` the first `.find` basically always hits, but I really didn't want this to ever return `undefined` and leave `<html dir>` unset.

packages/static-site/components/landing/LanguageSwitcher.tsx
```diff
@@ -0,0 +1,169 @@
...
+  readonly currentLanguageLabel: string;
+  /** Locale currently shown to the visitor. */
+  readonly currentLocale: AppLocale;
+  /** Current page pathname without a locale prefix, used as the link target. */
+  readonly pathname: string;
+  /**
+   * Languages to offer, in display order. Defaults to the full app list;
+   * accepts an override so tests can exercise specific option counts.
+   */
+  readonly descriptors?: readonly LanguageDescriptor[];
```
seidior, line 48: Defaulting `descriptors` to the real list but letting it be overridden is what keeps the scroll tests honest — they feed synthetic lists of 2/6/7/8 so the "scrollbar past six" coverage doesn't break every time we add or drop a real locale. The component genuinely doesn't care what the app's locale set is, which felt like the right seam.

packages/static-site/components/landing/LanguageSwitcher.tsx
```diff
@@ -0,0 +1,169 @@
...
+    <li className="usa-language__submenu-item" key={descriptor.locale}>
+      <Link
+        aria-current={isCurrent ? "true" : undefined}
+        href={pathname}
+        hrefLang={descriptor.locale}
+        locale={descriptor.locale}
+      >
+        {/* Per WCAG H58, only the native-name text is in a foreign language, so
+            the lang attribute scopes to it rather than the whole link. */}
+        <strong lang={descriptor.locale}>{descriptor.nativeName}</strong>
```
seidior, line 105: Per H58, only the native word is actually foreign here — `Español (Spanish)` is half Spanish, half English. Tagging the whole anchor `lang="es-US"` would tell a screen reader to read "Spanish" with a Spanish voice, which is wrong. Scoping `lang` to just the `<strong>` keeps each run labeled correctly.

packages/static-site/app/globals.css
```diff
@@ -134,3 +156,16 @@ li {
     margin: 0;
   }
 }
+
+/* Show at most six language options before scrolling, per USWDS language-selector spec. */
+.usa-language__submenu {
+  max-height: 13.5rem;
```
seidior, line 162: 13.5rem ≈ six rows plus a sliver, which is the USWDS rule (show six, scroll past that). It's a magic number, I'll grant you — happy to pull it into a variable if we'd rather, but this is the one spot that encodes "six".

packages/static-site/app/[locale]/layout.tsx
```diff
@@ -170,9 +177,10 @@ const LocaleLayout = async ({ children, params }: LocaleLayoutProps) => {
   const messages = getApplicationMessages({ locale });
 
   return (
-    <html lang={locale}>
+    <html dir={textDirectionForLocale(locale)} lang={locale}>
       <head>
         <link href={NJWDS_STYLESHEET_PATH} rel="stylesheet" />
+        <link href={NJ_RTL_UTILITIES_STYLESHEET_PATH} rel="stylesheet" />
```
Comment (seidior) [inline packages/static-site/app/[locale]/layout.tsx:183]: Load order matters here: this has to come *after* the NJWDS sheet. Our RTL utilities win at equal specificity purely on source order, which is how we avoid an `!important` arms race. If anyone reorders these two `<link>`s, RTL silently breaks — worth a comment so nobody "tidies" them.

packages/static-site/components/landing/GovBanner.tsx
```diff
@@ -55,9 +55,11 @@ export const GovBanner = ({ content }: GovBannerProps) => {
             <div className="grid-col-fill">
               <LocalizedLink link={content.stateSiteLink} />
             </div>
-            <ul className="grid-col-auto display-flex flex-align-center">
+            <ul className="grid-col-auto display-flex flex-align-center nj-inline-separators">
               <li>
-                <LocalizedLink link={content.governorIdentityLink} />
+                <LocalizedLink link={content.governorIdentityLink}>
+                  <bdi dir="ltr">{content.governorIdentityLink.label}</bdi>
```
seidior, line 61: Wrapping the governor name in `<bdi dir="ltr">` so it doesn't get reordered when it sits inside an RTL banner — Latin-script proper nouns should read left-to-right regardless of document direction. Nice part is `LocalizedLink` already takes children, so no new prop was needed; same trick on the brand title and the identifier domain.

packages/static-site/public/styles/nj-rtl-utilities.css
```diff
@@ -0,0 +1,57 @@
...
+  border-inline-end: 0;
+}
+
+/* Opens an absolutely-positioned dropdown from the inline-end edge (flips under
+   RTL). Composes onto the language submenu, whose NJWDS rule hard-codes
+   right: 0. NJWDS targets it with a (0,3,0) selector
+   (.usa-language__primary-item:last-of-type .usa-language__submenu), so this
+   utility is qualified with the item + submenu classes to reach equal (0,3,0)
+   specificity and win on source order rather than fighting with !important. */
+.usa-language__primary-item .usa-language__submenu.nj-dropdown-inline-end {
```
seidior, line 56: This is the one ugly rule. NJWDS pins the submenu with a (0,3,0) selector, so a single utility class loses to it no matter the load order. I qualified this utility with the same item + submenu classes to *match* (0,3,0) and then win on source order, rather than reaching for `!important`. Everything else in this file is a plain single-class utility — this is the exception and I wanted to flag why.

scripts/_install.sh
```diff
@@ -387,6 +387,12 @@ yarn install && yarn build || { echo "Root install or build failed."; exit 1; }
 (
     cd packages/static-site
     pnpm install
+    # Playwright drives the static-site accessibility and e2e suites; install the
+    # browsers here (dev/CI only) rather than via a package postinstall, which
+    # would also pull Chromium into production Docker image builds. The headless
+    # shell is a separate download from full Chromium and is what the headless
+    # test runs actually launch.
+    pnpm exec playwright install chromium chromium-headless-shell
```
seidior, line 395: Putting the browser install in setup instead of a package `postinstall` on purpose — a postinstall would drag ~150MB of Chromium into the production Docker build for no reason. `chromium-headless-shell` is a separate, smaller download and it's what the headless runs actually launch, so I install both.

packages/static-site/tests/accessibility/homepage.a11y.spec.ts
```diff
@@ -25,9 +28,19 @@ interface CreateLocaleAccessibilityTestParams {
  * Creates a Playwright accessibility test for one locale.
  */
 const createLocaleAccessibilityTest = ({ locale }: CreateLocaleAccessibilityTestParams) => {
-  return async ({ page }: AccessibilityTestContext) => {
+  return async ({ page, context }: AccessibilityTestContext) => {
     const messages = await getApplicationMessages({ locale });
 
+    // Suppress the preferred-language prompt modal so its open/animation state
+    // cannot race with the axe scan, which made this audit flaky.
+    await context.addCookies([
```
seidior, line 36: The language-prompt modal was making this audit flaky — axe would sometimes scan mid open-animation. Seeding its dismissed cookie before navigation makes it deterministic. Side note: chasing this down is how I found and fixed a few pre-existing a11y issues that had nothing to do with the picker (a duplicate unnamed `role="dialog"`, an orphaned `aria-modal`, and inline-link contrast in prose).

packages/static-site/tests/e2e/rtlLayout.e2e.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+const RTL_ROUTE = "/ar-US";
+const TRIGGER = "nav.usa-language button.usa-language__link";
+const SUBMENU = "#language-switcher-submenu";
+
+// Skipped while no RTL locale ships: these assertions navigate to /ar-US, which
+// is not a route until an RTL language is added back to APP_LOCALES. The RTL
+// implementation they cover (the nj-rtl-utilities classes, bidi isolation, and
+// the placeholder messages/ar-US.json) is intentionally kept in the tree, so
+// re-enabling is just removing this `.skip` once ar-US is restored.
+const describeRtlLayout = test.describe.skip;
```
seidior, line 22: Skipping rather than deleting on purpose. These all navigate to `/ar-US`, which isn't a route now that we've trimmed to en/es — but the RTL implementation they cover is still in the tree and was verified working at `/ar-US` on this branch. When we bring an RTL locale back, re-enabling is literally removing this `.skip`. Felt wrong to throw away passing coverage we know we'll want again.

packages/static-site/app/[locale]/language-switcher-harness/page.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Renders the language switcher with a controlled number of options.
+ *
+ * @param props Route props with locale params and `count` query.
+ * @returns The harness markup, or `notFound()` in production / for bad locales.
+ */
+const LanguageSwitcherHarnessPage = async ({ params, searchParams }: HarnessPageProps) => {
+  // biome-ignore lint/style/noProcessEnv: gate this test-only route out of production builds.
+  if (process.env.NODE_ENV === "production") {
```
Comment (seidior) [inline packages/static-site/app/[locale]/language-switcher-harness/page.tsx:79]: This route exists only to measure scroll in a real browser — jsdom has no layout engine, so `scrollHeight > clientHeight` is meaningless in unit tests. Gating it behind `notFound()` in production keeps it out of users' hands while still letting the e2e suite hit it. The synthetic descriptors are display-only (it never navigates), hence the `as AppLocale` cast.

packages/static-site/messages/ar-US.json
```diff
@@ -0,0 +1,584 @@
+{
```
seidior, line 1: Placeholder, not real translation — the strings are filler Arabic so we had genuine RTL content to lay out and test against. Kept in the tree even though ar-US isn't a shipped locale right now, so bringing RTL back is just re-adding the locale, not re-doing this file.

seidior (commented):
self review

PR #12850, opened 2026-06-17
feat: [AB#17828] starter kits page

jphechter:
## Description

Add a Starter Kits page to the static site with an industry selector combo-box and Top Industries card grid. Replace USWDS JavaScript with React-managed state for all interactive components (header accordion, language switcher, modal, combo-box) using `@trussworks/react-uswds`.

### Ticket

This pull request resolves [#17828](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17828).

### Approach

**Starter Kits Page:**
- Added `PageSwitchComponent` to route specific page slugs to custom components instead of the default markdown renderer
- Created `StarterKitsPage` (async server component) that loads industries and localized messages
- Created `IndustrySelector` (client component) using `@trussworks/react-uswds` `ComboBox`
- Onboarding URLs built with the `URL` API for safe encoding, sourced from `ACCOUNT_APP_URL` in `domain/env.ts`
- Slimmed industry data to `{ id, name }` before passing across the RSC boundary
- All user-facing strings (headings, CTA text, top industries) stored in `messages/*.json` for translation

**USWDS JS Removal:**
- Removed `uswds.min.js` script from the layout — all interactive behavior now owned by React
- `HeaderPrimaryNav`: submenu accordion toggle managed via `useState`
- `LanguageSwitcher`: dropdown visibility, Escape key, and click-outside handled in React
- `LanguagePromptModalView`: replaced `data-open-modal`/`data-close-modal` with an `isOpen` prop pattern; manages focus and Escape key in React
- Added `.usa-nav` CSS override for desktop visibility (previously set by USWDS JS `is-visible` class)

**Other:**
- Added `isEnabled` to the `Industry` type in `content-types`
- `loadIndustries` filters out disabled industries
- Test factories use a seeded PRNG with a single seed per run (logged for reproducibility)
- Updated README with test seed documentation

### Steps to Test

1. Navigate to `/en-US/pages/starter-kits`
2. Verify the heading, sub-heading, h2, and description render correctly
3. Open the industry dropdown — all enabled industries should appear (not "Generic")
4. Select an industry and click "Go to Starter Kit" — should navigate to the onboarding URL with `?industry={id}`
5. Verify "Top Industries" cards render with correct links
6. Navigate to the page via in-app link — combo-box should work without a full page reload
7. Verify the desktop header submenu (Learn/Updates accordion) toggles on click
8. Verify the language switcher dropdown opens/closes and responds to Escape key
9. Test on mobile: hamburger menu, account drawer, and language picker should function normally

### Notes

- `@trussworks/react-uswds` is used only for the `ComboBox` component. The header accordion, language switcher, and modal use custom React state since they have NJ-specific markup that differs from upstream USWDS.
- NJWDS CSS remains for all styling. Only the JavaScript was removed.
- The `uswds-init.js` script is no longer loaded either (it only set/cleared `usa-js-loading` for the main script).

## Code author checklist

- [ ] I have updated relevant `.env` values in `.env-template`, in Bitwarden, and in our workspaces

### DIFF
--- a/packages/static-site/components/learn/IndustrySelector.test.tsx
+++ b/packages/static-site/components/learn/IndustrySelector.test.tsx
@@ -1,0 +1,70 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { getApplicationMessages } from "@/domain/i18n/messages";
+import { generateIndustry } from "@/tests/factories";
+import { IndustrySelector } from "./IndustrySelector";
+
+const messages = await getApplicationMessages({ locale: "en-US" });
+const starterKits = messages.learn.starterKits;
+
+const industries = [generateIndustry(), generateIndustry(), generateIndustry()];
+
+const baseUrl = "https://example.com/onboarding";
+
+describe("IndustrySelector", () => {
+  it("renders all industries as listbox options", () => {
+    render(
+      <IndustrySelector
+        industries={industries}
+        baseUrl={baseUrl}
+        heading={starterKits.industrySelectorHeading}
+        ctaText={starterKits.industrySelectorCta}
+      />,
+    );
+
+    const input = screen.getByRole("combobox");
+    fireEvent.click(input);
+
+    const options = screen.getAllByRole("option");
+    expect(options).toHaveLength(3);
+    const industriesNames = industries.map((industry) => industry.name);
+    expect(options.map((o) => o.textContent)).toEqual(industriesNames);
+  });
+
+  it("renders the CTA as disabled when no industry is selected", () => {
+    render(
+      <IndustrySelector
+        industries={industries}
+        baseUrl={baseUrl}
+        heading={starterKits.industrySelectorHeading}
+        ctaText={starterKits.industrySelectorCta}
+      />,
+    );
+
+    const cta = screen.getByText(starterKits.industrySelectorCta);
+    expect(cta).toHaveAttribute("aria-disabled", "true");
+    expect(cta).not.toHaveAttribute("href");
+  });
+
+  it("enables the CTA with correct href when an industry is selected", () => {
+    render(
+      <IndustrySelector
+        industries={industries}
+        baseUrl={baseUrl}
+        heading={starterKits.industrySelectorHeading}
+        ctaText={starterKits.industrySelectorCta}
+      />,
+    );
+
+    const input = screen.getByRole("combobox");
+    fireEvent.click(input);
+
+    const targetIndustry = industries[0];
+    const option = screen.getByRole("option", { name: targetIndustry.name });
+    fireEvent.click(option);
+
+    const cta = screen.getByRole("link", { name: starterKits.industrySelectorCta });
+    expect(cta).toHaveAttribute("href", `${baseUrl}?industry=${targetIndustry.id}`);
+    expect(cta).toHaveAttribute("aria-disabled", "false");
+  });
+});
--- a/packages/static-site/components/learn/IndustrySelector.tsx
+++ b/packages/static-site/components/learn/IndustrySelector.tsx
@@ -1,0 +1,61 @@
+"use client";
+
+import { ComboBox } from "@trussworks/react-uswds";
+import { useState } from "react";
+
+interface IndustryOption {
+  readonly id: string;
+  readonly name: string;
+}
+
+interface Props {
+  readonly industries: readonly IndustryOption[];
+  readonly baseUrl: string;
+  readonly heading: string;
+  readonly ctaText: string;
+}
+
+const buildOnboardingHref = (baseUrl: string, industryId: string): string => {
+  const url = new URL(baseUrl);
+  url.searchParams.set("industry", industryId);
+  return url.toString();
+};
+
+export const IndustrySelector = ({ industries, baseUrl, heading, ctaText }: Props) => {
+  const [selectedIndustryId, setSelectedIndustryId] = useState<string | undefined>(undefined);
+
+  const href = selectedIndustryId ? buildOnboardingHref(baseUrl, selectedIndustryId) : undefined;
+
+  const options = industries.map((industry) => ({
+    value: industry.id,
+    label: industry.name,
+  }));
+
+  return (
+    <div className="border-1px border-base-light radius-md padding-3 margin-top-3">
+      <label className="usa-label font-heading-lg text-bold" htmlFor="industry-select">
+        {heading}
+      </label>
+      <div className="grid-row grid-gap flex-align-end">
+        <div className="grid-col-8">
+          <ComboBox
+            id="industry-select"
+            name="industry"
+            options={options}
+            onChange={(val) => setSelectedIndustryId(val)}
+          />
+        </div>
+        <div className="grid-col-4">
+          <a
+            href={href}
+            className={`usa-button width-full${!selectedIndustryId ? " usa-button--disabled" : ""}`}
+            aria-disabled={!selectedIndustryId}
+            style={{ height: "2.5rem" }}
+          >
+            {ctaText}
+          </a>
+        </div>
+      </div>
+    </div>
+  );
+};
--- a/packages/static-site/components/learn/PageSwitchComponent.tsx
+++ b/packages/static-site/components/learn/PageSwitchComponent.tsx
@@ -1,13 +1,18 @@
 import type { ReactElement } from "react";
 import PageContent from "@/components/learn/PageContent";
+import { StarterKitsPage } from "@/components/learn/StarterKitsPage";
 import type { PageItem } from "@/domain/content/types";
+import type { AppLocale } from "@/domain/i18n/locales";
 
 interface Props {
   readonly page: PageItem;
+  readonly locale: AppLocale;
 }
 
-export const PageSwitchComponent = ({ page }: Props): ReactElement => {
+export const PageSwitchComponent = ({ page, locale }: Props): ReactElement => {
   switch (page.slug) {
+    case "starter-kits":
+      return <StarterKitsPage page={page} locale={locale} />;
     default:
       return <PageContent page={page} />;
   }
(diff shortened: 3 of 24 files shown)


packages/static-site/components/USWDSInitializer.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * USWDS enhances DOM elements (combo-boxes, accordions, etc.) on initial page
+ * load. On client-side route transitions, new elements miss that initialization.
+ * This component detects route changes and re-appends the USWDS script, which
+ * triggers initialization for any un-enhanced elements in the document.
+ */
+export const USWDSInitializer = () => {
+  const pathname = usePathname();
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: re-run USWDS init on every route change
+  useEffect(() => {
```
seidior: [boulder] I see the problem this is trying to solve: App Router navigation can render fresh NJWDS markup after the original script has already initialized the page. I do not think re-appending the whole `uswds.min.js` script on every pathname change is the safe version of that fix, though. The NJWDS bundle calls each component's `.on(document.body)` when it runs, so each route transition can add another set of delegated body listeners for unrelated components like header, accordion, banner, and modal. I was able to reproduce the class of issue in browser: after navigating to `/pages/starter-kits`, the mobile header `Menu` button stayed closed after click, which is consistent with duplicate listeners toggling the same state twice. Is there a way to make this bridge scoped/idempotent instead, maybe by initializing only newly rendered combo boxes, using a documented component-level NJWDS API, or owning this combo-box interaction in React?

packages/static-site/components/learn/IndustrySelector.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    render(
+      <IndustrySelector
+        industries={industries}
+        baseUrl={baseUrl}
+        heading="Find Your Industry"
+        ctaText="Go to Starter Kit"
+      />,
+    );
+
+    const select = screen.getByRole("combobox");
```
seidior: [pebble] These tests cover the native `<select>` state well, but they do not exercise the behavior this PR is most worried about: NJWDS-enhanced combo box behavior after client-side navigation. Could we add a small Playwright path that navigates in-app to `/pages/starter-kits`, selects an industry through the enhanced combo box, and asserts the CTA href? That would catch both the initialization path and regressions from duplicate global listeners.

packages/static-site/components/learn/IndustrySelector.tsx
```diff
@@ -0,0 +1,46 @@
...
+  readonly industries: readonly Industry[];
+  readonly baseUrl: string;
+  readonly heading: string;
+  readonly ctaText: string;
+}
+
+export const IndustrySelector = ({ industries, baseUrl, heading, ctaText }: Props) => {
+  const [selectedIndustryId, setSelectedIndustryId] = useState("");
+
+  const href = selectedIndustryId ? `${baseUrl}?industry=${selectedIndustryId}` : undefined;
```
seidior: [pebble] I get why the URL construction is this direct; the IDs are slug-like today and this is the only query param. Since the value still comes from content/config, could we use `new URL(baseUrl)` plus `searchParams.set("industry", selectedIndustryId)` so encoding stays correct if an ID ever includes `&`, `?`, `#`, `=`, or whitespace? A shared helper would also let the card links use the same safer path.

packages/static-site/components/learn/IndustrySelector.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const options = industries.map((industry) => ({
+    value: industry.id,
+    label: industry.name,
+  }));
+
+  return (
+    <div className="border-1px border-base-light radius-md padding-3 margin-top-3">
+      <label className="usa-label" htmlFor="industry-select">
+        <h3 className="margin-top-0">{heading}</h3>
```
seidior: [sand] The explicit `htmlFor`/`id` pairing is good, especially because NJWDS expects a real label for combo-box enhancement. The part I would adjust is the `<h3>` inside the `<label>`; headings are not valid phrasing content inside labels, and it mixes heading navigation semantics with form-label semantics. Could this be `<h3><label ...>{heading}</label></h3>` or a styled label instead?

packages/static-site/components/learn/PageSwitchComponent.tsx
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export const PageSwitchComponent = ({ page, locale }: Props): ReactElement => {
+  switch (page.slug) {
+    case "starter-kits":
+      return <StarterKitsPage page={page} locale={locale} />;
+    default:
+      return <PageContent page={page} />;
+  }
+};
```
seidior, line 19: [sand] This switch is a reasonable first extension point for one custom CMS-backed page. If we expect more pages like this, I would consider moving the slug-to-renderer mapping into a some sort of object/dictionary. That would make the list of custom pages scan better and avoid a long switch as the static site grows.

packages/static-site/components/learn/StarterKitsPage.tsx
```diff
@@ -0,0 +1,51 @@
+import { IndustrySelector } from "@/components/learn/IndustrySelector";
+import { loadIndustries } from "@/domain/content/loadContent";
+import type { PageItem } from "@/domain/content/types";
+import type { AppLocale } from "@/domain/i18n/locales";
+import { getApplicationMessages } from "@/domain/i18n/messages";
+
+// TODO: This should be controlled by stage
+const ONBOARDING_BASE_URL = "https://account.business.nj.gov/onboarding";
```
seidior: [sand] I would still avoid merging the TODO with a hardcoded production account URL unless we are intentionally saying every environment should send users to prod onboarding. Static site has non-prod deployments, and this would make dev/stage validation cross an account-app environment boundary. Could this come from `siteConfig` or a typed environment setting, or should the TODO be replaced with an explicit comment documenting the intended cross-environment behavior?

packages/static-site/components/learn/StarterKitsPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface Props {
+  readonly page: PageItem;
+  readonly locale: AppLocale;
+}
+
+export const StarterKitsPage = async ({ page, locale }: Props) => {
+  const industries = loadIndustries().filter((i) => i.id !== "generic");
```
seidior: [dust] Good edge-case catch excluding `generic` from the selector. Because `generic` is enabled content, though, this page-level filter is also the only thing preventing it from becoming a public option. Could we cover the assembled starter-kits page with a small test that asserts `generic` is not offered, or extract an `isStarterKitIndustry` helper so the domain rule has a name?

packages/static-site/components/learn/StarterKitsPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+      <h2>{page["heading-2"]}</h2>
+      {page["main-text-2"] && <p>{page["main-text-2"]}</p>}
+
+      <IndustrySelector
+        industries={industries}
+        baseUrl={ONBOARDING_BASE_URL}
+        heading={starterKitsContent.industrySelectorHeading}
+        ctaText={starterKitsContent.industrySelectorCta}
+      />
```
seidior, line 41: [sand] The server/client split is the right instinct here, but I'm worried about the size of the data we're passing around here.

`IndustrySelector` only needs `id` and `name`, while the enabled non-generic industry payload is significantly larger as full `Industry` objects versus something much smaller with `{ id, name }`.

Could `StarterKitsPage` map to a selector-specific shape before passing data to the client component so we do not ship roadmap/internal fields through the RSC boundary?

packages/static-site/components/learn/StarterKitsPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+      <h2 className="margin-top-6">Top Industries</h2>
```
seidior: [pebble] Hard-coded string: Top Industries

packages/static-site/components/learn/StarterKitsPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <ul className="usa-card-group">
+        {starterKitsContent.topIndustries.map((industry) => (
+          <li key={industry.id} className="usa-card tablet:grid-col-6">
```
seidior, line 46: [dust] we may want to add a small test that every `messages.learn.starterKits.topIndustries[].id` exists in the enabled industry list before we loop through it in case of malformed data (it's happened before)

packages/static-site/tests/factories.ts
```diff
@@ -0,0 +1,16 @@
+import type { Industry } from "@/domain/content/types";
+
+export const randomInt = (length = 8): number => {
+  return Math.floor(10 ** (length - 1) + Math.random() * (10 ** length - 10 ** (length - 1) - 1));
+};
```
seidior, line 10: [dust] Oh no. The return of random tests.

If we're going to re-introduce this pattern (I'd argue we don't need it quite yet), we should be logging these seed values for later.

seidior (commented):
The server/client split is great: load content and messages server-side, keep the selected industry state in a small client component, and avoid making the whole page client-rendered.

I also like a few of the domain catches here. Adding `isEnabled` to the shared industry type and filtering disabled industries before they hit the selector is a thoughtful edge case.

The blocker for me is the NJWDS initialization strategy. Re-appending the full NJWDS runtime from the root layout is broader than the combo-box problem, and checks found the same class of issue: duplicate global delegated listeners can double-toggle site-wide components. In a browser check on this branch, the mobile header menu did not open on the starter-kits page.

packages/static-site/components/learn/PageSwitchComponent.tsx
hinzed1127: +1 to the switch component. I'm cherry-picking these changes so the funding page is rendered in the same way. However, I'd say let's wait for the switch to grow more before abstracting it in any way

hinzed1127 (approved):

PR #12883, opened 2026-07-11
feat: [AB#17846] handle static-site legacy redirects

hinzed1127:
## Description

Adds Webflow → static-site legacy 308 redirects so old business.nj.gov URLs (English and Spanish) land on real new-site pages instead of 404ing, preserving inbound links, bookmarks, and SEO equity.

Route mappings based on [this exported spreadsheet from Webflow](https://docs.google.com/spreadsheets/d/1N1JiEPuIc1m74tCjJ-eh6bbi8Y2dGoS9vf_6DE9OO0w/edit?gid=500027024#gid=500027024).

### Ticket

This pull request resolves [#17846](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/17846).

### Approach

Added a standalone script, `domain/redirects/legacyRedirects.ts`, that generates the redirect mapping from a small rule set and wires it into `next.config.ts` via `async redirects()`:

- **Prefix rules** collapse the high-volume detail sections:
	- `/license/*` to the new licensing and certification page
	- `/funding/*` to the new funding page
	- `/starter-kits/*` to the new starter-kits page
	- `/recent/*` preserves the slug to `/updates/*` (two legacy article slugs are pinned to topic pages ahead of the prefix).
- **One-off rules** cover the rest (category hubs, COVID, plastic-ban, contracting, exporting, business-structure synonyms, `/impact` → `/impact-report`, `/page/*` singular-prefix URLs, etc.).
- Every internal rule gets a lowercase `/es-us` **Spanish route** whose destination is flag-aware of `NEXT_PUBLIC_MULTILINGUAL_ENABLED`:
	- the English page when the flag is off
	- the `/es-US/…` page when on.
	- External destinations are never locale-rewritten. All rules are `permanent: true`, so Next.js emits a **308**.
- **`/es-us` fallback (flag off):** any unlisted `/es-us/<path>` is stripped to its English equivalent, and the bare `/es-us` root → `/`. When the flag is **on**, no catch-all is emitted — `next-intl` resolves the prefix case-insensitively and renders `/es-US` directly (a normalizing redirect would self-loop, since Next's `redirects()` also matches case-insensitively).
- **Unknown `/pages/<slug>`** now renders the 404 page (via `loadPages().find()` + `notFound()`) instead of throwing a 500, matching the `updates/[slug]` route.
- Legacy URLs with no verified new-site destination are intentionally left to 404 so they surface as signals. A triage doc (shared with product) tracks these and the grey-zone URLs still needing a decision.

No dependency, migration, or env changes.

### Steps to Test

From `packages/static-site`:

```bash
NEXT_PUBLIC_MULTILINGUAL_ENABLED=false pnpm build && pnpm start
```

Then check status codes (do not follow redirects; default port 3000):

```bash
curl -sI http://localhost:3000/recent                             # 308 → /updates
curl -sI http://localhost:3000/license/acupuncture-license        # 308 → /pages/licensing-and-certification-guide
curl -sI http://localhost:3000/es-us/license/acupuncture-license  # 308 → /pages/licensing-and-certification-guide (English, flag off)
curl -sI http://localhost:3000/category/grow                      # 308 → /grow
curl -sI http://localhost:3000/starter-kits/food-truck            # 308 → /pages/starter-kits (on-site, single hop)
curl -sI http://localhost:3000/es-us/plan                         # 308 → /plan (unlisted es-us stripped to English, flag off)
curl -sI http://localhost:3000/pages/does-not-exist               # 404 (unknown slug renders the 404 page, not a 500)
curl -sI http://localhost:3000/search                             # 404 (intentionally not redirected)
```

Follow a redirect end-to-end and confirm the final page is 200:

```bash
curl -sIL http://localhost:3000/license/acupuncture-license | grep -iE "^HTTP"
curl -sIL http://localhost:3000/starter-kits/food-truck   | grep -iE "^HTTP"   # single hop → /pages/starter-kits → 200
```

Flag-on Spanish behavior (rebuild with `NEXT_PUBLIC_MULTILINGUAL_ENABLED=true`):

```bash
curl -sI http://localhost:3000/es-us/license/acupuncture-license  # 308 → /es-US/pages/licensing-and-certification-guide
curl -sIL http://localhost:3000/es-US/plan | grep -iE "^HTTP"     # renders (no loop) — next-intl handles the prefix directly
```

Automated coverage: `pnpm test` (unit tests for the builder) and `pnpm test:e2e:english-only` (flag-off redirect e2e).

> Note: a fresh checkout needs `content` built first (`cd content && yarn build`) or the `/pages/*` redirect targets 500 on missing `content/lib/*.json`.

### Notes

- `/license/*` and `/funding/*` aggregate to their guide pages because individual detail pages aren't built on the new site. TBD on if we are actually building the individual pages or just keeping the top-level redirects as the new pattern moving forward. Deep-linking to individual licenses is a follow-up if those pages get built (requires reconciling old Webflow slugs against new-site slugs).
- Starter-kit slugs collapse onto the single `/pages/starter-kits` page (the old per-kit `account.business.nj.gov` external URLs are no longer used — all kit traffic stays on-site).
- With the flag **off**, `/es-us/…` routes send Spanish users to the English page; with it **on**, they go to the real `/es-US/…` page. `next-intl` resolves `/es-us` vs `/es-US` case-insensitively, so a flag-on normalizing redirect would self-loop — hence no catch-all in that mode.
- Redirects use `permanent: true`, so Next.js emits a `308` — the canonical permanent redirect, which preserves the request method (unlike the method-coercing `301`) and matches the ticket and SEO expectations. They run before the `proxy.ts`/next-intl middleware, so lowercase `/es-us/…` paths match before locale routing.
- Cross-checked the full 1,704-row compiled redirect spreadsheet against the shipped table: **1,617 exact matches, 0 mismatches** (remaining rows are unchanged paths that correctly don't redirect, or intentional 404s). Triage of the undecided rows is shared with product.

## Code author checklist

- [ ] ~~I have created and/or updated relevant documentation on the engineering documentation website~~ N/A
- [ ] ~~If I added/updated any values in `userData` (including `profileData`, `formationData` etc), then I added a new migration file~~ N/A
- [ ] ~~I have checked for and removed instances of unused config from CMS~~ N/A
- [ ] ~~If I added any new collections to the CMS config, then I updated the search tool and `cmsCollections.ts` (see CMS Additions in Engineering Reference/FAQ on the engineering documentation site)~~ N/A
- [ ] ~~I have updated relevant `.env` values in `.env-template`, in Bitwarden, and in our workspaces~~ N/A

### DIFF
--- a/packages/static-site/app/[locale]/(learn)/pages/[slug]/page.tsx
+++ b/packages/static-site/app/[locale]/(learn)/pages/[slug]/page.tsx
@@ -2,7 +2,7 @@ import type { Metadata } from "next";
 import { notFound } from "next/navigation";
 import { PageSwitchComponent } from "@/components/learn/PageSwitchComponent";
 import { CATEGORY_HIERARCHY } from "@/domain/categories";
-import { loadPageBySlug } from "@/domain/content/loadContent";
+import { loadPages } from "@/domain/content/loadContent";
 import { buildAlternateLanguages } from "@/domain/i18n/alternateLanguages";
 import { type AppLocale, hasAppLocale } from "@/domain/i18n/locales";
 
@@ -40,7 +40,10 @@ const ContentPage = async ({ params }: Props) => {
     notFound();
   }
 
-  const page = loadPageBySlug(slug);
+  const page = loadPages().find((item) => item.slug === slug);
+  if (!page) {
+    notFound();
+  }
 
   return <PageSwitchComponent page={page} locale={locale} />;
 };
--- a/packages/static-site/domain/redirects/legacyRedirects.ts
+++ b/packages/static-site/domain/redirects/legacyRedirects.ts
@@ -1,0 +1,217 @@
+/**
+ * Builds the Webflow → static-site legacy redirect table.
+ *
+ * Every internal rule is emitted twice: once for the English (unprefixed) path
+ * and once for the lowercase `/es-us` Spanish path Webflow served. The Spanish
+ * route's destination is flag-aware — English path when multilingual is off,
+ * `/es-US`-prefixed when on. External (`http`) destinations are never rewritten.
+ * When the flag is off, a trailing catch-all strips any remaining `/es-us` path
+ * (and the bare `/es-us` root) to its English equivalent; when on, `next-intl`
+ * already renders `/es-US` so no catch-all is emitted. All rules are permanent
+ * (Next.js emits a 308 HTTP status code) to match the migration's SEO
+ * expectations.
+ */
+
+/** A single legacy redirect entry consumed by `next.config.ts` `redirects()`. */
+export interface LegacyRedirect {
+  readonly source: string;
+  readonly destination: string;
+  /** `permanent: true` makes Next.js emit a 308 HTTP status code. */
+  readonly permanent: true;
+}
+
+/** An internal rule before Spanish-route expansion. `destination` is an app path. */
+interface InternalRule {
+  readonly source: string;
+  readonly destination: string;
+}
+
+const LEGACY_SPANISH_PREFIX = "/es-us";
+const APP_SPANISH_LOCALE_PREFIX = "/es-US";
+
+/**
+ * Prefix and section rules. Aggregate rules drop the slug; /recent preserves it.
+ * A few /recent slugs are legacy article URLs whose content now lives on a topic
+ * page, so they are pinned to that page BEFORE the /recent/:slug* prefix — order
+ * matters because `redirects()` is first-match-wins.
+ */
+const PREFIX_RULES: readonly InternalRule[] = [
+  { source: "/recent", destination: "/updates" },
+  { source: "/recent/disposable-bag-ban", destination: "/pages/plastic-ban-law" },
+  { source: "/recent/state-financial-assistance-programs", destination: "/pages/covid19" },
+  { source: "/recent/:slug*", destination: "/updates/:slug*" },
+  { source: "/license", destination: "/pages/licensing-and-certification-guide" },
+  { source: "/license/:slug*", destination: "/pages/licensing-and-certification-guide" },
+  { source: "/funding", destination: "/pages/funding" },
+  { source: "/funding/:slug*", destination: "/pages/funding" },
+  { source: "/starter-kits/:slug*", destination: "/pages/starter-kits" },
+];
+
+/** Verified one-off rules with a real INTERNAL destination. Given an /es-us route. */
+const ONE_OFF_RULES: readonly InternalRule[] = [
+  // impact report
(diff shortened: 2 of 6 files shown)


packages/static-site/app/[locale]/(learn)/pages/[slug]/page.tsx
```diff
@@ -40,7 +40,10 @@ const ContentPage = async ({ params }: Props) => {
     notFound();
   }
 
-  const page = loadPageBySlug(slug);
+  const page = loadPages().find((item) => item.slug === slug);
+  if (!page) {
+    notFound();
+  }
```
Comment (hinzed1127) [inline packages/static-site/app/[locale]/(learn)/pages/[slug]/page.tsx:46]: this serves as a catch-all 404 against pages nested under `/learn`

hinzed1127 (commented):
**Self Review**: Not a lot of individual lines to call out, since it's essentially a giant config file in `packages/static-site/domain/redirects/legacyRedirects.ts`. The "Approach" section of the description is basically a guide to that file, but I'd recommend just looking at the code directly, since each of the grouped sections of redirects have comments. My self review was more these 2 documents:
- [Triage doc](https://gist.github.com/hinzed1127/d6f13fad2071d2d7c3e6ac476ca117b1) referenced in the description for reviewing with product
- [Claude artifact](https://claude.ai/code/artifact/9af3c663-9b54-4fab-84ef-2d4eba1c1fac) tracking that we were at parity with the routes covered in the spreadsheet

packages/static-site/domain/redirects/legacyRedirects.ts
```diff
@@ -0,0 +1,213 @@
...
+];
+
+/** One-off rules with an EXTERNAL destination. Given an /es-us route, but never locale-rewritten. */
+const EXTERNAL_ONE_OFF_RULES: readonly InternalRule[] = [
+  { source: "/panel", destination: "https://forms.business.nj.gov/panel/" },
+  { source: "/survey", destination: "https://www.surveymonkey.com/r/ZHQZP25" },
+  { source: "/navigator", destination: "https://account.business.nj.gov/" },
+];
+
+const isExternal = (destination: string): boolean => destination.startsWith("http");
```
seidior, line 152: [dust] Surely there's a less-brittle way of doing this, or something more-canonical to the spec?

hinzed1127: This only gets utilized as a helper function for `spanishDestination` to determine whether or not to prepend translated `/es-US/` for routes. If a url is external, we obviously don't wanna do that. This whole step is skipped to start, anyway, since we're launching without multilingual support, but it's future-proofed for when we flip that switch.

PR #12897, opened 2026-07-16
prefactor: centralize SWR user data lifecycle

seidior:
## Description

This PR centralizes Navigator's user-data lifecycle in one app-level `UserDataProvider`.

The provider now owns authenticated fetches, guest hydration, optimistic updates, canonical save responses, the shared `UpdateQueue`, persisted fallback data, error mapping, and profile-derived side effects. Existing consumers continue to use the same `useUserData()` API, but they now read one shared context instead of creating their own SWR lifecycle.

This PR also separates SWR's runtime cache from Navigator's persisted domain data, adds a versioned user-data storage format with one-way legacy migration, prevents late fetch/save completions from crossing authenticated identities, and expands multi-business and provider-integration coverage.

The shared `UserData` domain shape and backend API contract are unchanged. Browser session persistence does change: the storage key, serialization envelope, validation rules, and migration behavior are all owned by this PR.

### Ticket

This pull request partially resolves [AB#16176](https://dev.azure.com/NJInnovation/Business%20First%20Stop/_workitems/edit/16176). It's the first implementation slice of the React 19 / SWR upgrade strategy and is intentionally delivered on SWR 1.3 before the dependency upgrade.

### Why this change

Before this PR, every mounted `useUserData()` consumer participated in the SWR lifecycle. SWR also used session storage as its cache provider, which coupled:

- transient SWR cache entries;
- persisted Navigator `UserData`;
- authenticated fetch and save behavior;
- shared update-queue initialization; and
- consumer-level reconciliation side effects.

That shape made duplicate lifecycle work possible and made identity changes, storage migration, and the later SWR 2 upgrade harder to reason about.

```mermaid
flowchart LR
  subgraph Before["Before: lifecycle distributed across consumers"]
    C1["Consumer A"] --> H1["useUserData + useSWR"]
    C2["Consumer B"] --> H2["useUserData + useSWR"]
    H1 --> API["User data API"]
    H2 --> API
    H1 --> Q["Shared UpdateQueueContext"]
    H2 --> Q
    H1 <--> S["Session storage as SWR cache"]
    H2 <--> S
    S --> L["Legacy raw records<br/>$swrUserData$&lt;userId&gt;"]
  end

  subgraph After["After: one lifecycle owner"]
    CA["Consumers"] --> UH["useUserData"]
    UH --> UC["UserDataContext"]
    UC --> P["UserDataProvider"]
    P <--> M["SWR in-memory Map"]
    P --> API2["User data API"]
    P --> Q2["Shared UpdateQueue"]
    P --> D["Profile, roadmap, analytics,<br/>Intercom, and error state"]
    P <--> R["UserDataStorage"]
    R --> V["Versioned session record<br/>navigator:user-data:&lt;userId&gt;"]
    LG["Valid legacy record"] -->|"migrate once"| R
  end
```

#### Centralize ownership without changing consumers

- Add `UserDataProvider` and `UserDataContext` at the application level.
- Reduce `useUserData()` to a required-context consumer while preserving its public return contract.
- Make provider placement an explicit render-tree contract. Application routes receive it from `_app`; isolated renders and tests use provider-aware helpers.
- Give all mounted consumers one provider-owned SWR request per active key, one update queue, one set of persistence writes, and one set of profile-derived side effects. The pre-existing sign-in bootstrap fetch is unchanged.
- Keep guest/NJEDA initialization local and prevent guest flows from calling the authenticated user endpoint.
- Keep the public starter-kit navbar provider-free. In that variant, "My Account" is static text rather than a dashboard link; the adjacent sign-in action remains available.
- Add provider-aware render helpers and migrate affected component/page tests to the real context boundary.

#### Separate runtime cache from persisted domain data

- Replace session storage as SWR's cache provider with a normal in-memory `Map`.
- Persist Navigator-owned data through `UserDataStorage`, independent of SWR's internal cache representation.
- Store validated, versioned records under `navigator:user-data:<userId>`.
- Validate the envelope, stored identity, and current business before returning persisted data.
- Keep existing session-scoped retention; this PR does not move complete `UserData` to longer-lived storage.

#### Migrate legacy storage forward once

- Prefer a valid versioned record and remove any leftover legacy copy.
- When only `$swrUserData$<userId>` exists, validate it, write the versioned envelope, and delete the legacy key only after the new write succeeds.
- Retain a valid legacy record when migration cannot be persisted so the current session remains recoverable.
- Continue recognizing both namespaces for user discovery, logout deletion, and namespace-scoped clearing.
- Do not dual-write. Rollback to an older frontend artifact is unsupported; corrective changes deploy forward.

#### Make identity transitions explicit

- Key authenticated fetch completion to the user ID whose request settled.
- Keep `hasCompletedFetch` false when user A has completed but user B is still loading.
- Ignore stale fetch completions when their request key is no longer active.
- Persist successful fetch data only when the payload identity matches the request key.
- Ignore both failed and successful save completions when ownership changed while the request was in flight.
- Clear the prior identity's queue, local owner, and user-data error state during an authenticated identity change.

```mermaid
sequenceDiagram
  participant A as User A
  participant P as UserDataProvider
  participant API as User data API
  participant B as User B

  A->>P: Save optimistic update
  P->>API: POST UserData for A
  Note over P,B: Active identity changes from A to B
  P->>API: GET UserData for B
  API-->>P: Late canonical save response for A
  P-->>P: Ignore response because A is no longer the owner
  Note over P: B remains incomplete while B's GET is pending
  API-->>P: B's GET settles
  P->>P: Mark fetch complete for B and initialize B's queue
```

#### Keep persistence and derived-state failure boundaries separate

- Apply the canonical API response to SWR, storage, and the update queue before running profile-derived synchronization.
- If roadmap or profile-derived synchronization fails after a successful POST, log that failure without setting `UPDATE_FAILED` or asking the user to retry a write the server already committed.

#### Preserve and expand user-facing coverage

- Add a multi-business Cypress flow that saves distinct profile data, switches businesses, reloads, and verifies each business retains its own values.
- Add deterministic Jest coverage for authenticated fetch success/failure, persisted fallback, guest hydration, identity changes, stale refreshes, late save failures, and late successful saves.
- Add storage coverage for malformed values, identity mismatches, unavailable/full browser storage, one-way migration, failed migration recovery, and scoped deletion.
- Add focused provider-boundary coverage for NJEDA and public navbar variants.

### Before and after behavior

| Surface | Before | After |
| --- | --- | --- |
| Fetch ownership | Each `useUserData()` consumer mounted SWR | One app-level provider owns SWR |
| Consumer API | Hook returned user data and update operations | Same public hook contract |
| SWR cache | Session-storage-backed cache | In-memory `Map` |
| Domain persistence | Raw legacy SWR-shaped key | Versioned Navigator-owned record |
| Legacy compatibility | Legacy key remained part of normal writes | Legacy key is one-way migration input only |
| Identity completion | Completion could outlive the identity that fetched | Completion is keyed to the active user |
| Late save response | Could target state after an identity change | Ignored after ownership changes |
| Multi-consumer side effects | Reconciliation lived with each hook instance | Provider runs reconciliation once |

#### Consuming user data in application code

The consumer API is intentionally unchanged. Existing components keep using the hook the same way:

Before:

```tsx
const { business, updateQueue, error, refresh } = useUserData();
```

After:

```tsx
const { business, updateQueue, error, refresh } = useUserData();
```

The difference is ownership. Before, calling the hook also mounted that consumer's SWR and reconciliation lifecycle. After, the hook reads the shared `UserDataContext` owned by the nearest `UserDataProvider`.

#### Rendering the application tree

Before, SWR used `UserDataStorage` as its cache provider and components invoked the full hook lifecycle. Simplified provider tree:

```tsx
<SWRConfig value={{ provider: UserDataStorageFactory }}>
  <AuthContext.Provider value={{ state, dispatch }}>
    <UpdateQueueContext.Provider value={{ updateQueue, setUpdateQueue }}>
      <Component {...pageProps} />
    </UpdateQueueContext.Provider>
  </AuthContext.Provider>
</SWRConfig>
```

After, SWR has a transient in-memory cache and the app mounts one user-data owner inside the contexts it depends on. Simplified provider tree:

```tsx
<SWRConfig value={{ provider: createSWRCache }}>
  <AuthContext.Provider value={{ state, dispatch }}>
    <UpdateQueueContext.Provider value={{ updateQueue, setUpdateQueue }}>
      <UserDataProvider>
        <Component {...pageProps} />
      </UserDataProvider>
    </UpdateQueueContext.Provider>
  </AuthContext.Provider>
</SWRConfig>
```

Production components should continue calling `useUserData()` rather than mounting their own provider. Provider-free routes must not render components that consume user data.

#### Rendering components in focused tests

Before, a component test could render a user-data consumer directly because the hook created its own lifecycle:

```tsx
render(<ExampleComponent />);
```

After, focused tests use the shared helper, which supplies a default `UserDataContext`:

```tsx
renderWithUserData(<ExampleComponent />);
```

Tests that need an explicit state can pass the same `UseUserDataResponse` shape that production consumers receive:

```tsx
renderWithUserData(<ExampleComponent />, {
  userData,
  business: userData.businesses[userData.currentBusinessId],
  isLoading: false,
  hasCompletedFetch: true,
  error: undefined,
  updateQueue,
  refresh: jest.fn(),
  createUpdateQueue: jest.fn().mockResolvedValue(updateQueue),
  clearUserDataError: jest.fn(),
});
```

Tests that exercise edits over time continue using the stateful wrapper; it now publishes that changing state through `UserDataContext`:

```tsx
render(
  <WithStatefulUserData initialUserData={userData}>
    <ExampleComponent />
  </WithStatefulUserData>,
);
```

Provider lifecycle tests should mount the real `UserDataProvider` with fresh SWR and storage instances so request deduplication, identity changes, migration, and side effects are exercised rather than mocked.

These helpers are intentionally opt-in. A package-wide default context previously allowed component tests to pass while production composition omitted the required provider, so tests should declare this dependency explicitly.

#### Authenticated load and save

1. Sign in with an existing account.
2. Confirm the dashboard/onboarding flow loads normally.
3. Change profile data and save.
4. Reload the page and confirm the saved data is restored.
5. Sign out and confirm the session's Navigator user-data records are removed.

#### Multiple businesses

1. Save a distinct business name for the first business.
2. Add a second business and save a different business name.
3. Switch between the two businesses and confirm each profile displays its own name.
4. Reload each profile and confirm the selected business and its data remain correct.

#### Provider-free and guest routes

1. Open a public starter-kit route while signed out and confirm the navbar renders, "My Account" is static text, and the adjacent sign-in action remains available.
2. Open the NJEDA route while signed out and confirm the guest experience initializes without an authenticated user-data request.

### Local verification

Run with the repository's required Node version through `~/.nvm/nvm-exec`:

- `yarn verify:node`
- Focused Jest regression suites
- Full root Jest suite
- Web TypeScript typecheck
- Cypress TypeScript typecheck
- Targeted ESLint and Prettier checks for changed code
- `git diff --check`
- Production web build

The expanded multi-business Cypress spec was authored and typechecked but was not run locally because it requires the repository services, dev server, and authenticated browser state.

### Scope and follow-up

This PR intentionally includes:

- centralized user-data ownership;
- versioned persistence and one-way legacy migration;
- cross-identity fetch/save safety;
- provider integration and regression coverage; and
- no consumer-facing `useUserData` or `UserDataStorage` API changes.

This PR intentionally does **not** include:

- the SWR 2 dependency upgrade;
- SWR 2 mutation APIs or named mutation options;
- same-user overlapping-save ordering;
- broader SWR 2 read/write race coordination;
- changes to the backend `UserData` contract;
- user-data migrations; or
- longer-lived client storage.

Same-user save ordering and the broader SWR 2 race model remain part of the follow-up SWR 2.4.2 upgrade slice.

### Deployment and compatibility notes

- Deployment is forward-only. Restoring an older frontend artifact is not a supported rollback strategy.
- An older frontend artifact cannot read a successfully migrated guest record. Guest data has no server source of truth, so recovery is a corrective forward deployment rather than restoring an older artifact.
- Existing valid session data migrates on read. No API or database migration is required.
- Failed client-storage writes do not prevent the in-memory application from continuing.
- Invalid or mismatched persisted records are discarded rather than exposed to the wrong identity.
- The verification listed above was run locally. The Checks tab remains the source of truth for current CI status.

## Code author checklist


web/cypress/e2e/multiple-businesses.spec.ts
```diff
@@ -1,8 +1,36 @@
...
+    onProfilePage.getBusinessName().should("be.visible").clear().type(businessName);
+    onProfilePage.clickSaveButton();
+    cy.wait(`@${requestAlias}`).its("response.statusCode").should("eq", 200);
+    cy.url().should("contain", "/dashboard");
+  };
+
+  const completeAdditionalBusinessOnboarding = (): void => {
+    assertAdditionalBusinessOnboardingPage(1);
+    onOnboardingPage.selectBusinessPersona("OWNING");
+    onOnboardingPage.selectIndustrySector("construction");
```
aadedejifearless, line 29: Nit: Should we keep the casing consistent between `OWNING` and `construction`?. I'm guessing this is prevalent across the codebase but might be an opportunity in the future for us to maintain consistency.

web/test/render/renderWithUserData.tsx
```diff
@@ -0,0 +1,28 @@
...
+    createUpdateQueue: async (userData: UserData): Promise<UpdateQueue> => {
+      return new UpdateQueueFactory(userData, async (): Promise<void> => {});
+    },
+    clearUserDataError: (): void => {},
+  };
+};
+
+export const renderWithUserData = (
+  subject: ReactElement,
+  value: UseUserDataResponse = createDefaultUserDataContext(),
```
aadedejifearless, line 25: Just confirming my understanding: `renderWithUserData` is a test helper that wraps the component with `UserDataContext.Provider`, allowing the test to provide user data to the component if needed. This make sense

web/src/pages/njeda.tsx
```diff
@@ -41,7 +41,8 @@ interface Props {
 const NJEDAFundingsOnboardingPaage = (props: Props): ReactElement => {
```
aadedejifearless: we have a typo here. `NJEDAFundingsOnboardingPaage` and also on line 356

web/src/lib/data-hooks/UserDataProvider.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+  activeUserIdRef.current = activeUserId;
+  ownerUserIdRef.current = ownerUserId;
+  authenticationRef.current = state.isAuthenticated;
+
+  const { data, error, mutate } = useSWR<UserData | undefined>(
+    activeUserId || null,
+    api.getUserData,
+    {
+      fallbackData: persistedUserData,
```
seidior, line 75: Self-review: Concretely: a returning user with session-storage data now sees their real data on the very first render — not a spinner — while SWR quietly revalidates behind it. That's the stale-while-revalidate pattern actually reaching the UI. The loading flag itself, though, is still a manual !error && !activeData && !activeUpdateQueue check, not SWR's own isLoading, because SWR 1.3 has nothing to read yet.

web/src/lib/storage/UserDataStorage.ts
```diff
@@ -13,96 +17,250 @@ interface UserDataStorage {
   getRegistrationStatus: () => RegistrationStatus | undefined;
 }
 
-export const userDataPrefix = "$swrUserData$";
+interface PersistedUserData {
+  readonly version: 1;
```
aadedejifearless, line 21: Just confirming my understanding: if the persisted user data schema changes in the future, we would increment both version and persistedUserDataVersion, and existing data with the previous version would be considered incompatible and discarded rather than migrated. Is that correct?

web/cypress/e2e/multiple-businesses.spec.ts
seidior: Good catch, but these are two separate domains rather than one inconsistency: OWNING matches the BusinessPersona union type's convention (all-caps, STARTING/OWNING/FOREIGN), while construction matches how industry sector ids are defined elsewhere (e.g. isConstructionSector.ts). Each is consistent with its own domain's existing convention, so I'd rather not change either in this PR, although I agree it'd be a nice broader cleanup someday if we ever wanted to unify sector-id casing, but that's a separate, unrelated effort from this prefactor.

web/test/render/renderWithUserData.tsx
seidior: That's exactly right: `createDefaultUserDataContext()` supplies a sensible default (loading, no data) so most tests don't need to pass anything, and any test that cares about a specific state can override value directly.

web/src/pages/njeda.tsx
seidior: Good catch, though that typo predates this PR (it's already on main); this PR just happens to touch nearby lines. That said, I've campground the fix into this PR.

web/src/lib/storage/UserDataStorage.ts
seidior: Correct. `isPersistedUserData` checks `version === persistedUserDataVersion` exactly, so any mismatch (including an old version after a future bump) fails validation and the stored key gets deleted via  `safelyDeleteStorageKey`. It's discard-and-refetch, not migrate-in-place, same pattern as the legacy-key handling already in this file. If we ever need real migration (e.g. the persisted shape changes but we don't want to force a refetch), we'd need to add an explicit migration step keyed off the version number. I don't think that's a requirement for this work, so I kept it simple. Open to tackling that if you think it's an issue.

PR #12898, opened 2026-07-16
feat: add independent static site release pipeline

hinzed1127:
## Description

Today the static site can only reach production as the final step of `release-pipeline.yml`, which includes a full release of navigator (i.e. deploying `web/` and `api/`). This PR adds a workflow that lets us deploy the static site independently.

### Ticket

This pull request resolves [#17934](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17934).

### Approach

- This mostly strips out unnecessary navigator deploy elements and minimizes new logic. It reuses the existing `deploy-static-site` action and re-reaches the existing `StaticSiteServiceStack-{staging,prod}` CDK stacks.

- Uses separate tagging logic (`<YYYYMMDD>-<short-sha>` here vs. the `package.json` version in the full deploy). The ECS task definition pins whichever tag the most recent deploy set. See my inline review comments for the tag-format rationale, the `prod-hold` gate, and the `release-*` trigger contrast.

- Both jobs build `content` before deploying. The Dockerfile copies in `content/lib`, which is generated/gitignored, so a checkout alone can't produce it. `release-pipeline.yml` gets it for free from its full monorepo build, but this pipeline explicitly generates it.

- The two deploy jobs differ only by `_STAGING`/`_PROD` suffixes, matching how `release-pipeline.yml` already does it. A reusable workflow taking `stage` is a larger refactor we could consider in follow-up work.

### Steps to Test

The only testing is verification of the pipeline itself. I pre-verified the config to make sure all the `vars` and `secrets` references and the `prod-hold` environment are good, but **this has not been run against real infrastructure yet**, so the first dispatch is the real test.

1. Run the `Static Site Release Pipeline` workflow from this branch (or from `main` post-merge)
2. Watch `deploy-staging`: content build succeeds, image pushes to `bfs-static-site`, `StaticSiteServiceStack-staging` deploys. Confirm CDK touches **only** static site stacks.
3. Verify staging renders.
4. Approve at `production-hold`, confirm `deploy-production` promotes the same `<date>-<short-sha>` tag that staging deployed.
5. Confirm navigator prod is untouched.

Make sure to do a thorough staging-only trial before approving through to prod.

### Notes

Deliberately left for follow-ups:

- There's no automated verification between staging and prod. The gate is a human approval, same as the current release pipeline.
- `release-pipeline.yml` still deploys the static site, too. I think we'll want to continue that, but something to consider if this independent deploy seems to be reliable enough on its own.
- There's a redundant "Build Content Types" step in all three static-site workflows. The Dockerfile compiles content-types from source itself, so the host step is a no-op. Dropping it should be a separate PR touching all three together.

### DIFF
--- a/.github/workflows/static-site-release-pipeline.yml
+++ b/.github/workflows/static-site-release-pipeline.yml
@@ -1,0 +1,145 @@
+name: Static Site Release Pipeline
+
+# Deploys only the static site, independently of the navigator monorepo release.
+# The static site builds to its own image, ECR repository, and ECS service, so it
+# can ship on its own cadence. Both stages deploy the same image tag, so the
+# artifact approved on staging is the exact artifact promoted to production.
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: static-site-release-pipeline
+  cancel-in-progress: false
+
+jobs:
+  # The tag is resolved once and reused by both stages. Resolving it per job would
+  # let staging and production diverge when an approval lands on a later date.
+  resolve-image-tag:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      image_tag: ${{ steps.resolve.outputs.image_tag }}
+    steps:
+      - name: Resolve Image Tag
+        id: resolve
+        shell: bash
+        run: |
+          # Intentionally uses a date + short SHA instead of the application version.
+          # This release pipeline runs independently from the main release pipeline,
+          # so using a distinct tagging strategy helps prevent image tag collisions.
+          image_tag="$(date -u +%Y%m%d)-${GITHUB_SHA:0:7}"
+          echo "image_tag=$image_tag" >> "$GITHUB_OUTPUT"
+          echo "Static site image tag: $image_tag"
+
+  deploy-staging:
+    needs: resolve-image-tag
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    environment: staging
+    env:
+      AWS_ACCOUNT_ID: ${{ vars.AWS_ACCOUNT_ID_STAGING }}
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_STAGING }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_STAGING }}
+      AWS_REGION: ${{ vars.AWS_CRYPTO_CONTEXT_ORIGIN }}
+      BASIC_AUTH_PASSWORD: ${{ secrets.BASIC_AUTH_PASSWORD }}
+      BASIC_AUTH_USERNAME: ${{ vars.BASIC_AUTH_USERNAME }}
+      NEXT_PUBLIC_MULTILINGUAL_ENABLED: ${{ vars.NEXT_PUBLIC_MULTILINGUAL_ENABLED }}
+      NEXT_PUBLIC_WEB_BASE_URL: ${{ vars.WEB_BASE_URL_STAGING }}
+      SG_ID: ${{ vars.SG_ID_AWS_STAGING }}
+      STAGE: ${{ vars.STAGE_STAGING }}
+      SUBNET_ID_01: ${{ vars.SUBNET_ID_01_AWS_STAGING }}
+      SUBNET_ID_02: ${{ vars.SUBNET_ID_02_AWS_STAGING }}
+      VPC_ID: ${{ vars.VPC_ID_AWS_STAGING }}
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/static-site-release-pipeline.yml
```diff
@@ -0,0 +1,142 @@
...
+    outputs:
+      image_tag: ${{ steps.resolve.outputs.image_tag }}
+    steps:
+      - name: Resolve Image Tag
+        id: resolve
+        shell: bash
+        run: |
+          image_tag="$(date -u +%Y%m%d)-${GITHUB_SHA:0:7}"
+          echo "image_tag=$image_tag" >> "$GITHUB_OUTPUT"
+          echo "Static site image tag: $image_tag"
```
hinzed1127, line 34: this is a different approach than what `release-pipeline.yml` takes for the static site, with a tagging format of `-<date>-<short-sha>` instead of `-<version>` from `package.json`. I didn't want to take the exact same tagging approach, to avoid collisions, but I'm open to other suggestions here. The main goal was just make sure that the two release pipelines can co-exist neatly.

.github/workflows/static-site-release-pipeline.yml
```diff
@@ -0,0 +1,142 @@
+name: Static Site Release Pipeline
+
+# Deploys only the static site, independently of the navigator monorepo release.
+# The static site builds to its own image, ECR repository, and ECS service, so it
+# can ship on its own cadence. Both stages deploy the same image tag, so the
+# artifact approved on staging is the exact artifact promoted to production.
+
+on:
+  workflow_dispatch:
```
hinzed1127, line 9: contrast this with the full repo deploy, which includes the trigger with `release-*` branches getting merged

.github/workflows/static-site-release-pipeline.yml
```diff
@@ -0,0 +1,142 @@
...
+          image_tag: ${{ needs.resolve-image-tag.outputs.image_tag }}
+
+  production-hold:
+    needs: deploy-staging
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    environment: prod-hold
+    steps:
+      - run: echo "Awaiting approval for static site production release."
```
hinzed1127, line 94: this is the same type of manual hold gate we currently use with navigator's deploy

hinzed1127 (commented):
self-review

.github/workflows/static-site-release-pipeline.yml
jphechter: Can't believe I'm saying this, but this might make for a good comment. I can easily see someone attempting some "cleanup" at a point in the future and creating a collision without realizing it.

.github/workflows/static-site-release-pipeline.yml
jphechter: I'm torn with this being a manual trigger. In theory, we could run this off any branch. That gives us flexibility, but doesn't cement us to an easily identifiable point in time if we needed to run a hotfix. I think the reality is that this probably won't get used terribly frequently after the first couple months of the site launch, and this probably becomes less of a concern.

.github/workflows/static-site-release-pipeline.yml
aadedejifearless, line 34:
```suggestion
    steps:
      - name: Resolve Image Tag
        id: resolve
        shell: bash
        run: |
          # Intentionally uses a date + short SHA instead of the application version.
          # This release pipeline runs independently from the main release pipeline,
          # so using a distinct tagging strategy helps prevent image tag collisions.
          image_tag="$(date -u +%Y%m%d)-${GITHUB_SHA:0:7}"
          echo "image_tag=$image_tag" >> "$GITHUB_OUTPUT"
          echo "Static site image tag: $image_tag"
```

aadedejifearless (approved):
Aside from the additional comment suggested by JP, everything else looks good. Approving. Nice work, Dan!

seidior (changes requested):
[boulder] Because this PR omits the checklist, it omits the reminder that changes like these should include documentation. Please create a documentation PR for this work before considering this for merger.

seidior:
Took a first pass at that documentation here if you want to take a look: https://github.com/newjersey/business-dev-docs/pull/140/changes

.github/workflows/static-site-release-pipeline.yml
hinzed1127: A valid callout. It might be worth following this up with more rigid constraints around what's deployable, but to your point, this pipeline is likely to be more a short-term need anyway. Something we can modify in the future if those needs change

hinzed1127:
https://github.com/newjersey/business-dev-docs/pull/140 is merged, so this should be good to go now. Merging with some follow-ups considered down the line

PR #12928, opened 2026-07-28
fix: [AB#17957] repair stranded KMS ciphertext

seidior:
## Description

`POST /api/decrypt` can return HTTP 500 with `{"error":"unencryptedDataKey has not been set"}` when a user reveals a tax ID or another protected identifier that was encrypted before the stage-specific KMS key rotation.

A prior KMS rotation introduced a current key per stage. Some ciphertext remained wrapped by the legacy key. `AWSCryptoFactory` currently uses one generator keyring for encryption and decryption. A strict single-key keyring
discards encrypted data keys wrapped by any other key before calling KMS, so a current-only client cannot decrypt legacy ciphertext and a legacy-only client cannot decrypt ciphertext already rotated to the current key.

The migration runner compounds the incident. It catches individual migration errors, continues the chain, forces `CURRENT_VERSION`, and persists the result. This can remove a partially migrated record from future version scans.

User data is duplicated across one users-table item containing embedded businesses and separate businesses-table projections. There are no DynamoDB Streams synchronizing these copies. Existing migration writes can therefore
advance the users-table version while leaving one or more business projections unchanged.

The production `migrateUsersVersion` Lambda is enabled every five minutes during its daily window. Recent CloudWatch metrics show every invocation failing. Its deployed environment omits the hashing key, hashing purpose, and hashing salt read by the handler.

This PR restores access to those values and provides a safe path for the scheduled user migration to converge every affected record onto the current key. It also closes a separate migration failure mode that could mark a partially migrated record as current and prevent the scheduler from ever finding it again.

### Ticket

This pull request resolves [#17957](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17957).

### What failed beyond `/api/decrypt`

The HTTP 500 was the most visible symptom, but it was only one point in a larger failure chain:

1. **Some protected fields became unreadable while others still worked.** The failure was field-specific, not necessarily user-specific. A user could have one value wrapped by the legacy key and another wrapped by the current key, so revealing one tax identifier could succeed while another returned `unencryptedDataKey has not been set`. The error did not by itself mean that KMS rejected access or that the ciphertext had been corrupted. It meant the local keyring discarded every encrypted data key in that ciphertext before making a KMS request.
2. **The v191 and v192 migrations could disagree with the ciphertext they received.** v191 decrypted with a legacy-only client and re-encrypted with the current key. v192 repeated the crypto work but still decrypted with the legacy-only client. If v191 succeeded, v192 could immediately fail because it was being asked to decrypt newly current-wrapped ciphertext with the wrong strict keyring. Records created on a newer version could begin with current-wrapped ciphertext and encounter the same legacy-only mismatch.
3. **A migration error could become permanent record state.** The old migration runner caught an individual migration exception, logged it, continued the chain, and returned the record with `CURRENT_VERSION` forced onto it. The subsequent write could therefore persist an incompletely transformed record as current. Scheduled scans select records by version, so that record would no longer be selected for another attempt even though one or more protected fields still required repair.
4. **The failure log exposed too much data without reliably identifying the failing operation.** The old catch block serialized the full in-memory user record. That could put email addresses, ciphertext, and unrelated user data into a migration error while still leaving operators to infer which field and crypto phase had failed. This PR replaces that with migration, field, attempt, version, and persistence-phase context without logging plaintext, ciphertext, tax identifiers, email addresses, or complete records.
5. **The users table and businesses table could diverge.** The users-table item was written first, followed by separate business projection writes. A later business write failure was logged and swallowed. That could leave the canonical user stamped current while one or more business-table projections remained on an older schema or retained old ciphertext.
6. **The scheduled repair path had an independent configuration failure.** The Lambda's CDK environment omitted the hashing key, hashing purpose, and hashing salt required by v191 and v192. A record that reached the hashing step could fail with `Salt must be provided`, even after its KMS decryption issue was fixed.
7. **Scheduled failures did not reliably become Lambda failures.** Per-user work used `Promise.allSettled`, rejected users were logged, the outer migration method converted thrown errors into `{ success: false }`, and the handler did not rethrow that result. CloudWatch could therefore see a successful Lambda invocation even when no affected user was repaired.
8. **The existing alarm watched for a message that was never emitted.** Its metric filter matched `"MigrateUserVersions Failed"`, while the relevant code logged messages such as `"MigrateData Failed"` or `"Failed to migrate user"`. A persistent migration failure could therefore avoid both the operator alert and the automatic kill-switch path.
9. **Testing had two authenticated encryption contexts.** Before the CircleCI-to-GitHub Actions deployment migration, testing inherited the dev encryption context: `stage=dev`, `purpose=tax_id_encryption`, and `origin=us-east-1`. The GitHub Actions workflow kept the testing application stage but omitted `AWS_CRYPTO_CONTEXT_STAGE` and `AWS_CRYPTO_CONTEXT_ORIGIN`, so the deployed Lambdas expected empty values and newer testing ciphertext was written with that empty context. Once the compatible keyring found the affected legacy encrypted data key, the existing post-decryption validation correctly rejected its historical dev context.
10. **The request-time fallback retried migration through the GET route.** The data client caught the v193 failure and returned the unchanged v192 record as designed. The user GET route then ran its ordinary derived updates and unconditionally called `put()`, which attempted v193 a second time and surfaced the same failure as HTTP 500. Returning an unchanged record only preserves availability if the caller also avoids mutating and persisting that outdated fallback.

Together, those behaviors created a feedback loop: a key or authenticated-context mismatch could prevent decryption, the migration framework could hide that failure by advancing the version anyway, the two table representations could disagree, the request fallback could retry the failed migration, and the scheduled process could continue reporting successful invocations without making the record eligible for repair again.

### KMS background

Navigator uses the [AWS Encryption SDK's envelope-encryption model](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/how-it-works.html):

1. The SDK generates a one-time data key and uses it to encrypt the sensitive value.
2. KMS encrypts, or "wraps," that data key with a customer managed KMS key (CMK).
3. The resulting ciphertext stores the encrypted value, the wrapped encrypted data key, the wrapping key's identifier, and the encryption context.
4. During decryption, the [AWS KMS keyring](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/use-kms-keyring.html) first selects an encrypted data key whose wrapping key is allowed. Only after that selection does it ask KMS to unwrap the data key.

Before this change, Navigator used one strict keyring configured with only the current key for both encryption and decryption. Ciphertext written with the legacy key still contained a valid encrypted data key, but the keyring rejected it because its wrapping key was not configured. With no acceptable encrypted data key left, the Encryption SDK raised `unencryptedDataKey has not been set` before making a KMS request.

This distinction matters operationally:

- `unencryptedDataKey has not been set` means the local keyring could not find an allowed encrypted data key in the ciphertext.
- An IAM, disabled-key, throttling, or KMS availability error means the keyring found a configured key and the subsequent KMS operation failed.
- Malformed ciphertext or an unexpected [encryption context](https://docs.aws.amazon.com/kms/latest/developerguide/encrypt_context.html) must fail closed; it is not a signal to skip that field and mark the record migrated. Deployment verification showed one exact, explainable exception in testing: a historical workflow omission created a second known context. This PR allowlists that complete historical context rather than weakening validation generally.

#### Terms used in this PR

- **Plaintext:** the original tax ID, tax PIN, or employer identifier. Plaintext must never be written to logs or stored directly.
- **Data key:** a short-lived symmetric key generated for encrypting a value. AWS explains the distinction between plaintext and encrypted copies in its [data key documentation](https://docs.aws.amazon.com/kms/latest/developerguide/data-keys.html).
- **Encrypted data key (EDK):** the data key after a KMS key wraps it. The EDK can be stored alongside ciphertext because KMS authorization is still required to unwrap it.
- **Wrapping key or KMS key:** the stage-specific customer managed key that encrypts the data key. This PR calls the previous wrapping key the legacy key and the replacement the current key.
- **Generator key:** the keyring's KMS key used when creating new ciphertext. Only the current key may be a generator after this change.
- **Keyring:** the Encryption SDK component that decides which keys may create EDKs during encryption and which EDKs may be considered during decryption.
- **Encryption context:** authenticated, non-secret metadata describing the stage, purpose, and origin of the ciphertext. Navigator verifies those values after decrypting so ciphertext created for a different context cannot be silently accepted.
- **Primary encryption context:** the exact stage, purpose, and origin used for new ciphertext. Encryption always uses this context.
- **Decrypt-only encryption context:** a complete historical stage, purpose, and origin combination that may be accepted during convergence but is never used for new encryption.
- **Ciphertext message:** the serialized Encryption SDK output containing the encrypted value, EDK metadata, encryption context, and algorithm information. AWS documents those components in the [Encryption SDK message format](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/message-format.html).

### A concrete mixed-key example

Consider a hypothetical v191 user with two businesses after the original rotation:

- Business A's profile tax ID is still wrapped by the legacy key.
- Business A's tax PIN was updated later and is wrapped by the current key.
- Business B's tax-clearance tax ID is still wrapped by the legacy key because that field was omitted from the earlier migration type tree.

Before this PR:

1. The current-only runtime client could decrypt the tax PIN but rejected both legacy-wrapped tax IDs before calling KMS.
2. The legacy-only migration client could decrypt the tax IDs but rejected the current-wrapped tax PIN.
3. Because migration operates on the whole user, either single-key client eventually encountered a field it could not accept.
4. The old runner could catch that failure, force the root record to the latest version, and persist it. A later scheduled scan then saw the version as current and skipped the user even though the fields remained mixed.
5. If the users-table write succeeded and a later business projection write failed, the two tables could disagree about both schema version and ciphertext.

After this PR, the decryption allowlist accepts both expected wrapping keys, every affected field is rotated in memory, and the users-table item plus all business projections commit in one transaction. The record moves forward only when the complete migration unit succeeds.

#### Keep encryption current-only while allowing compatible decryption

`AWSCryptoFactory` now creates separate keyrings for the two directions:

- Encryption uses only the current stage CMK as the generator key. New ciphertext cannot be wrapped by the legacy key.
- Decryption uses a strict allowlist containing the full ARNs of the current and legacy stage CMKs. It does not accept arbitrary KMS keys.

The same crypto client can therefore decrypt a record containing either generation of ciphertext while continuing to produce current-key-only ciphertext. Encryption-context validation remains in place after decryption.

This intentionally does not configure one encryption keyring with both keys. The Encryption SDK can use [multi-keyrings](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/use-multi-keyring.html) to wrap a data key under a generator key and additional child keys during encryption. That is useful when new ciphertext should carry multiple EDKs, but it is the wrong contract here because the legacy key must be decrypt-only. Separate direction-specific keyrings make the asymmetry explicit: write with exactly one current key, read with exactly two allowlisted keys during convergence.

The decrypt keyring is also deliberately not a discovery keyring. Restricting it to full current and legacy CMK ARNs prevents Navigator from accepting ciphertext merely because the Lambda role happens to have access to some other KMS key.

The Encryption SDK reuses its regional KMS clients, and the AWS SDK performs its [standard retry behavior with exponential backoff](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html) with three total attempts. This PR does not add a second retry layer around those network calls. The local tax-ID hashing step is retried up to three total attempts and logs each failed attempt with migration and field context.

Ciphertext already wrapped by the current key may be decrypted and re-encrypted once by the catch-up migration. Avoiding those one-time KMS calls would require additional ciphertext inspection and branching without changing the final state, so it is intentionally outside this repair.

#### Preserve exact encryption-context compatibility in testing

The testing workflow now restores `AWS_CRYPTO_CONTEXT_STAGE` from the dev context variable and explicitly supplies `AWS_CRYPTO_CONTEXT_ORIGIN`. New testing ciphertext therefore returns to the historical primary context: `dev`, `tax_id_encryption`, and `us-east-1`.

`AWSCryptoFactory` can also receive explicit decrypt-only contexts. After KMS decrypts a message, the complete header context must match either the primary context or one complete decrypt-only context. Values cannot be mixed across allowed contexts, and there are no wildcard stage, purpose, or origin values.

Only `STAGE=testing` receives the historical GitHub Actions context with empty stage and origin. This covers both legacy-key/dev-context records and current-key/empty-context records created during the workflow gap. v193 re-encrypts either form with the current key and corrected primary context, so the exception converges existing data instead of creating more empty-context ciphertext. Staging and production do not receive this compatibility context.

This remains distinct from a discovery keyring or disabled context validation. A ciphertext must use an allowlisted full KMS key ARN and one exact accepted context combination.

#### Repair the migration chain before adding v193

Adding only a v193 migration would not repair users whose stored version is older than v192. Navigator runs migrations in order, so a v190 record must successfully complete v191 and v192 before v193 can execute.

| Stored version | Required path after deployment | Reason |
| --- | --- | --- |
| v190 or earlier | Existing migrations through v190, then v191, v192, and v193 | Migrations cannot be skipped. v190 itself does not use KMS, but v191 and v192 require the repaired clients and fail-fast runner. |
| v191 | v192, then v193 | The record may already contain a mixture of legacy- and current-wrapped fields. |
| v192 | v193 | The catch-up migration re-enrolls records already stamped current by the previous runner and rotates fields omitted earlier. |
| v193 | No migration | New writes already use the current key, and the record is outside the scheduler's outdated-version scan. |

This PR updates the existing migration path so that:

- the compatible crypto client and hashing client are available to the migrations that need them;
- the scheduled Lambda receives the hashing key, purpose, and salt that were missing from its environment;
- migrations stop immediately at the first failure;
- the runner no longer catches an error, continues later migrations, and force-stamps `CURRENT_VERSION`; and
- migration work occurs on a copy, so a failed migration leaves the stored source record unchanged.

This makes older records able to reach v193 and keeps failed records eligible for the next scheduled attempt.

#### Use v193 as a complete catch-up rotation

The v193 migration decrypts with the current-plus-legacy allowlist and re-encrypts with the current key for every persisted field affected by the rotation:

- `profileData.encryptedTaxId`
- `profileData.encryptedTaxPin`
- `profileData.deptOfLaborEin`
- `cigaretteLicenseData.encryptedTaxId`
- `taxClearanceCertificateData.encryptedTaxId`
- `taxClearanceCertificateData.encryptedTaxPin`

The tax-clearance fields are included because runtime code can persist them, even though the v191 and v192 versioned type trees did not declare them.

All affected fields in all of a user's businesses must succeed in memory. A missing optional field remains absent. A malformed value, unsupported wrapping key, exhausted KMS error, encryption error, or exhausted hash error fails the whole user migration; the code does not preserve the bad field while advancing the record's version.

#### Commit the user and all business projections atomically

Navigator stores each user's businesses in two places:

- embedded in the users-table item; and
- as separate items in the businesses table.

There is no DynamoDB Stream synchronizing those copies. Updating the users item first and then writing businesses individually could therefore leave the user at v193 while one or more business projections remained on an older version.

Successful request-time and scheduled migrations now use one `TransactWriteItems` call containing the users-table put and every corresponding businesses-table put. DynamoDB commits all of those writes or none of them.

The users-table put is conditional on the exact source snapshot read before migration. If another request changes that record first, DynamoDB cancels the transaction, no migration writes commit, and the newer record remains authoritative. That concurrency conflict is expected contention and is retried on a later pass; it is not treated as a terminal data or infrastructure failure.

Existing drift in a businesses-table projection does not block repair. The users-table snapshot remains canonical and overwrites each projection as part of the successful transaction.

DynamoDB limits a transaction to 100 items, so one user and at most 99 businesses can be migrated as one unit. A record above that limit is a terminal failure requiring operator attention rather than a partial write.

#### Preserve request-time availability

Reading a user record may attempt a migration, but a migration problem must not turn the application read itself into an outage.

Request-time and scheduled migration share the same transformation and atomic transaction, but they have different responsibilities at the outer boundary. The request path protects application availability, so failure returns the unchanged stored record. The scheduled path owns convergence and operational detection, so a persistent terminal failure must stop the backfill, alert the team, and activate the kill switch instead of quietly moving to the next user.

If the migration kill switch is enabled, request-time migration is skipped and the unchanged stored record is returned. If a request-time migration fails:

- no migration data is written;
- the failure is logged with source version, target version, failed field or persistence phase, and the root error; and
- the unchanged stored record is returned to the caller.

Request-time logs do not serialize the user record or include an email address, plaintext, ciphertext, tax identifier, or user identifier. Request-time failures also do not directly activate the kill switch. The scheduled Lambda is the circuit-breaker owner and will encounter persistent failures independently.

Deployment verification exposed one additional caller responsibility. When `databaseClient.get()` returns an unchanged record below `CURRENT_VERSION`, the user GET route now recognizes that version as a migration fallback, sanitizes it, and returns HTTP 200 without running derived business-name, operating-phase, sidebar, license, or x-ray updates and without calling `put()`. A successfully migrated current record continues through the ordinary update-and-save path.

#### Stop and alert on scheduled terminal failures

The scheduled migration distinguishes retryable contention from a terminal failure:

- A transaction conflict is logged and skipped so a later scheduled pass can retry the newer record.
- A terminal field, record, KMS, hashing, configuration, permission, or persistence failure stops processing and fails the Lambda invocation.

The terminal scheduled log includes the opaque DynamoDB user key and non-sensitive migration context so an operator can locate the record. It does not include the email address, plaintext, ciphertext, tax identifiers, or full record.

The Lambda timeout is increased to four minutes. It stops starting another user when less than 30 seconds remain, exits successfully, and leaves the remaining users for the next scheduled invocation. This prevents normal KMS workload duration from looking like a structural migration failure.

Monitoring now alarms on the native `AWS/Lambda` `Errors` metric for the migration function instead of matching a log phrase the function did not emit. The first failed invocation publishes to both the existing operator-alert topic and the existing migration kill-switch topic. That path alerts the team and enables the SSM migration kill switch automatically.

The Lambda role retains its existing access to both configured KMS keys and receives the DynamoDB transaction permission required by the new atomic write path.

### Why this is the smallest deployable vertical slice

The runtime keyring change is the smallest code change that restores `POST /api/decrypt` for legacy ciphertext, but it is not independently sufficient to resolve the incident for existing users:

1. **Records must converge off the legacy key.** Leaving permanent dual-key decryption in place would make the legacy key an indefinite runtime dependency and leave the affected data unmigrated.
2. **Older records cannot jump directly to v193.** The ordered migration chain must be able to complete v191 and v192 first, including their crypto and hashing work.
3. **The old runner could permanently strand failures.** A catch-up migration is unsafe if an earlier failure can still be swallowed and the record stamped current.
4. **A user is duplicated across two tables.** Rotating sensitive fields while advancing only one copy can create a partially migrated record, so the persistence boundary must be atomic.
5. **The scheduled Lambda is the recovery mechanism.** It needs the same crypto and hashing configuration, enough execution time, the necessary IAM permissions, and a reliable terminal-failure signal.
6. **A terminal failure must stop the backfill.** Continuing after a malformed record or structural KMS problem risks repeating partial or invalid work across the database. The existing alarm and kill-switch path must therefore be connected to real Lambda failures.
7. **Known testing contexts must converge without weakening context authentication.** Correcting the workflow alone would make empty-context ciphertext unreadable, while keeping the empty primary context would continue producing incorrectly scoped ciphertext. The exact decrypt-only context provides the compatibility window needed to rewrite both populations correctly.
8. **Read fallback must survive the full HTTP request path.** Catching migration inside the data client is insufficient if the route immediately writes the outdated record and retries the same migration. The route-level version guard is part of the availability boundary.

These concerns span crypto configuration, migration logic, persistence, versioned schemas, Lambda infrastructure, IAM, monitoring, and tests, but they form one operational path: read legacy ciphertext, safely transform the whole migration unit, commit it once, and stop visibly if that cannot be done. Removing any one of those pieces would either fail to reach affected records, permit partial migration, or leave the scheduled repair unable to run safely.

Much of the apparent diff is also required by the repository's migration architecture rather than by new product behavior. Each migration carries a complete versioned TypeScript type tree, and the version bump must remain consistent with the migration registry, shared `CURRENT_VERSION`, Zod schemas, and schema-generation tooling. The runtime behavior remains narrowly limited to the KMS compatibility and migration-safety path described above.

### Failure behavior after this PR

| Condition | Request-time behavior | Scheduled behavior |
| --- | --- | --- |
| Value is wrapped by the current key | Decrypt and continue | Rotate once and continue |
| Value is wrapped by the configured legacy key | Decrypt and continue | Rotate to the current key |
| Testing value uses the corrected `dev` / `tax_id_encryption` / `us-east-1` context | Decrypt and continue | Rotate once with the corrected primary context |
| Testing value uses the known empty-stage and empty-origin context | Decrypt and continue through the testing-only compatibility context | Rotate to the corrected primary context |
| Wrapping key is not configured | Return the unchanged record and log context | Fail the invocation and activate the alert/kill-switch path |
| KMS permission, key-state, throttling, timeout, or availability failure after SDK retries | Return the unchanged record and log context | Fail the invocation and activate the alert/kill-switch path |
| Malformed ciphertext or any non-allowlisted encryption context | Return the sanitized unchanged record without a follow-up write and log context | Fail the invocation and activate the alert/kill-switch path |
| Re-encryption fails after SDK retries | Return the unchanged record and log context | Fail the invocation and activate the alert/kill-switch path |
| Hashing fails after three total attempts | Return the unchanged record and log each attempt | Fail the invocation and activate the alert/kill-switch path |
| Source record changes before the transaction commits | Return the unchanged source read | Skip as retryable contention; reconsider the newer record later |
| Lambda approaches its timeout between users | Not applicable | Stop cleanly; resume remaining users on the next invocation |

### Steps to Test

Automated verification completed for this branch:

1. Run `CI=true ./scripts/pre-commit.sh`.
   - Formatting, linting, type checking, builds, spellcheck, and configured unit suites pass.
2. Start the repository services and development application.
   - The local web and API applications start successfully.
3. Run the configured Cypress end-to-end suite against the development stack.
   - The affected user workflows pass.
   - One unrelated dashboard interaction timed out in the full run and then passed repeatedly in an isolated seeded rerun.
4. Run the API and CDK suites independently.
   - Crypto/keyring, migration, transaction, Lambda, IAM, and monitoring coverage passes.
5. Run `git diff --check`.
   - No whitespace errors are reported.
6. Run the lead PR review workflow.
   - No pebbles or boulders remain.

The focused tests cover:

- current-only encryption and strict current-plus-legacy decryption;
- encryption-context mismatch and malformed/decryption failure propagation;
- exact decrypt-only context acceptance and rejection when any stage, purpose, or origin value differs;
- failures representing unconfigured keys, IAM/KMS errors, exhausted encryption errors, and exhausted hashing errors;
- hashing success after retry and failure after the final attempt;
- fail-fast migration ordering and unchanged source data on failure;
- rotation of every affected field across multiple businesses;
- one conditional transaction for the user and all business projections;
- no write when any field fails;
- retryable transaction conflicts and terminal transaction failures;
- request-time fallback to the sanitized unchanged stored record, including a route-level assertion that no derived updater or follow-up write runs;
- scheduled Lambda failure propagation and timeout-aware clean stopping; and
- native Lambda error alarming to the operator and kill-switch topics.

Follow-up verification after the testing deployment ran 176 API/CDK suites with 971 passing tests, then reran the 93 tests covering the crypto factory, data client, and user route after the fallback correction. API type checking, linting, formatting, documentation spellcheck, and `git diff --check` also pass.

### Deployment verification

The following checks require deployed AWS resources and should be completed in dev before production rollout:

1. Confirm the current and legacy key settings are full stage-specific CMK ARNs and that the Lambda role has the intended least-privilege KMS access.
2. Encrypt a new value and verify its ciphertext contains only the current wrapping key and the intended primary encryption context.
3. Verify `POST /api/decrypt` can decrypt representative ciphertext wrapped by both the current and legacy keys with their expected contexts.
4. Exercise representative v190, v191, and v192 records through v193.
5. Run a real DynamoDB transaction for a user with multiple business projections, including an intentionally drifted projection.
6. Confirm a failed field leaves both tables unchanged.
7. Confirm a conditional transaction conflict commits nothing and does not activate the kill switch.
8. Confirm a scheduled terminal failure raises the Lambda `Errors` metric, alerts operators, and enables the migration kill switch.

The testing deployment has now verified the failure and corrective crypto path against deployed AWS resources:

1. The deployed express and migration Lambdas both had empty `AWS_CRYPTO_CONTEXT_STAGE` and `AWS_CRYPTO_CONTEXT_ORIGIN` values.
2. The affected v192 record's ciphertext header used the legacy testing key with `dev`, `tax_id_encryption`, and `us-east-1`.
3. The updated client decrypted that real ciphertext, re-encrypted the plaintext with the current testing key and corrected primary context, and decrypted the result successfully. The check did not persist plaintext or ciphertext, and the source record remained at v192.
4. CloudWatch logs confirmed the availability bug's two phases in one request: `get()` logged and caught the first v193 context failure, then the GET route's follow-up `put()` retried v193 and produced the HTTP 500. The route regression test now covers the no-update, no-write fallback behavior.

DynamoDB Local in this repository does not implement `TransactWriteItems`, so the command shape and behavior are covered by unit tests while the real transaction remains a dev deployment check. The affected legacy-key ciphertext and a generated current-key ciphertext have now been exercised with the deployed testing keys. Equivalent staging and production key/context verification remains a rollout check and cannot be proven by the local test environment.

### Rollout and rollback

This migration is forward-only after any v193 record is persisted. Rolling the entire application back to the previous v192 implementation would restore the migration runner that can swallow failures and force the version forward.

If a rollback is required after migration begins:

- enable the migration kill switch to stop additional work; and
- retain the compatible decrypt keyring, fail-fast migration runner, and request-time fallback while rolling forward with a corrective deployment.

The legacy key must remain enabled and authorized for decryption until a verified scan confirms no legacy-wrapped ciphertext remains.

### Notes and follow-up work

The following improvements are intentionally deferred because they are broader than this KMS repair:

- Reject ordinary writes when a stale browser submits a schema version older than the stored users-table version. The migration transaction protects against writes that happen before it commits, but a browser that loaded v192 can still save after v193 and downgrade the record.
- Redesign users-table and businesses-table projection synchronization around DynamoDB Streams.
- Remove legacy-key decryption only after a verified zero-legacy-ciphertext scan.
- Generalize migration-framework cleanup beyond the paths required to reach v193.
- Improve migration throughput, dashboards, and structured tracing beyond the contextual non-PII logs required for this repair.
- Replace the legacy Serverless-style hashing-salt placeholder with CDK-native SSM retrieval across all hash producers, followed by a coordinated rehash of existing records.

## Code author checklist

### DIFF
--- a/api/cdk/lib/lambdaStack.ts
+++ b/api/cdk/lib/lambdaStack.ts
@@ -339,7 +339,7 @@ export class LambdaStack extends Stack {
       entry: path.join(__dirname, "../../src/functions/migrateUsersVersion/app.ts"),
       handler: "handler",
       runtime: Runtime.NODEJS_22_X,
-      timeout: Duration.seconds(30),
+      timeout: Duration.minutes(4),
       memorySize: 1024,
       ephemeralStorageSize: Size.mebibytes(512),
       ...vpcProps,
@@ -352,6 +352,9 @@ export class LambdaStack extends Stack {
         AWS_CRYPTO_TAX_ID_ENCRYPTION_KEY: awsCryptoTaxIdEncryptionKey,
         LEGACY_AWS_CRYPTO_TAX_ID_ENCRYPTION_KEY: legacyAwsCryptoTaxIdEncryptionKey,
         AWS_CRYPTO_CONTEXT_TAX_ID_ENCRYPTION_PURPOSE: awsCryptoContextTaxIdEncryptionPurpose,
+        AWS_CRYPTO_TAX_ID_HASHING_KEY: awsCryptoTaxIdHashingKey,
+        AWS_CRYPTO_CONTEXT_TAX_ID_HASHING_PURPOSE: awsCryptoContextTaxIdHashingPurpose,
+        AWS_CRYPTO_TAX_ID_ENCRYPTED_HASHING_SALT: awsCryptoTaxIdHashingSalt,
       },
     });
 
--- a/api/cdk/lib/monitoringStack.ts
+++ b/api/cdk/lib/monitoringStack.ts
@@ -49,12 +49,6 @@ export class MonitoringStack extends Stack {
       `/aws/lambda/businessnjgov-api-v2-${props.stage}-express`,
     );
 
-    const migrateUsersVersionLogGroup = logs.LogGroup.fromLogGroupName(
-      this,
-      "MigrateUsersVersionLogGroup",
-      `/aws/lambda/businessnjgov-api-v2-${props.stage}-migrateUsersVersion`,
-    );
-
     if (props.stage === "dev") {
       new logs.LogGroup(this, "ApiLogLocalGroup", {
         logGroupName: `/${NAVIGATOR_WEBSERVICE}/local/ApiLogs`,
@@ -115,23 +109,14 @@ export class MonitoringStack extends Stack {
       },
     );
 
-    const migrationFailureMetricFilter = new logs.MetricFilter(
-      this,
-      "MigrationFailureMetricFilter",
-      {
-        logGroup: migrateUsersVersionLogGroup,
-        filterName: `MigrationFailureCount-${props.stage}`,
-        metricNamespace: "BFS/Navigator/Lambda",
-        metricName: "MigrateUserVersionsFailures",
-        filterPattern: logs.FilterPattern.literal('"MigrateUserVersions Failed"'),
-        metricValue: "1",
-        defaultValue: 0,
-      },
-    );
-
-    const migrationFailureMetric = migrationFailureMetricFilter.metric({
+    const migrationFailureMetric = new cloudwatch.Metric({
+      namespace: "AWS/Lambda",
+      metricName: "Errors",
       statistic: "Sum",
       period: Duration.seconds(300),
+      dimensionsMap: {
+        FunctionName: `businessnjgov-api-v2-${props.stage}-migrateUsersVersion`,
+      },
     });
 
     this.migrationLambdaTopic = new sns.Topic(this, "migrationLambdaTopic", {
@@ -166,13 +151,14 @@ export class MonitoringStack extends Stack {
       metric: migrationFailureMetric,
       threshold: 1,
       evaluationPeriods: 1,
-      comparisonOperator: cloudwatch.ComparisonOperator.GREATER_THAN_THRESHOLD,
+      comparisonOperator: cloudwatch.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
       treatMissingData: cloudwatch.TreatMissingData.NOT_BREACHING,
-      alarmDescription: `Alarm when the MigrateUserVersions failure count exceeds the threshold for ${props.stage}`,
+      alarmDescription: `Alarm when MigrateUserVersions reports an invocation error for ${props.stage}`,
     });
 
     migrationFailureAlarm.addAlarmAction(
       new cloudwatch_actions.SnsAction(migrateUserVersionErrorTopic),
+      new cloudwatch_actions.SnsAction(this.migrationLambdaTopic),
     );
     migrationFailureAlarm.applyRemovalPolicy(RemovalPolicy.RETAIN);
 
--- a/api/src/client/AwsCryptoFactory.ts
+++ b/api/src/client/AwsCryptoFactory.ts
@@ -1,16 +1,21 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const AWSCrypto = require("@aws-crypto/client-node");
 import { fromBase64, toBase64 } from "@aws-sdk/util-base64-node";
-import { CryptoClient } from "@domain/types";
-import { NonSharedBuffer } from "node:buffer";
+import { type CryptoClient } from "@domain/types";
+import { type NonSharedBuffer } from "node:buffer";
 import * as crypto from "node:crypto";
 import { TextDecoder } from "node:util";
 
-type Context = {
-  stage: string;
-  purpose: string;
-  origin: string;
-};
+export interface EncryptionContext {
+  readonly stage: string;
+  readonly purpose: string;
+  readonly origin: string;
+}
+
+const contextMatches = (
+  actualContext: Record<string, string>,
+  expectedContext: EncryptionContext,
+): boolean => Object.entries(expectedContext).every(([key, value]) => actualContext[key] === value);
 
 export const cryptoUtils: { pbkdf2: typeof crypto.pbkdf2 } = {
   pbkdf2: (
@@ -27,18 +32,23 @@ export const cryptoUtils: { pbkdf2: typeof crypto.pbkdf2 } = {
 
 export const AWSCryptoFactory = (
   generatorKeyId: string,
-  context: Context,
+  context: EncryptionContext,
   encryptedHashingSalt?: string,
+  decryptOnlyKeyIds: readonly string[] = [],
+  decryptOnlyContexts: readonly EncryptionContext[] = [],
 ): CryptoClient => {
   const { encrypt, decrypt } = AWSCrypto.buildClient(
     AWSCrypto.CommitmentPolicy.REQUIRE_ENCRYPT_REQUIRE_DECRYPT,
   );
-  const keyring = new AWSCrypto.KmsKeyringNode({ generatorKeyId });
+  const encryptKeyring = new AWSCrypto.KmsKeyringNode({ generatorKeyId });
+  const decryptKeyring = new AWSCrypto.KmsKeyringNode({
+    keyIds: [...new Set([generatorKeyId, ...decryptOnlyKeyIds].filter(Boolean))],
+  });
 
   const decoder = new TextDecoder();
 
   const encryptValue = async (plainTextValue: string): Promise<string> => {
-    const { result } = await encrypt(keyring, plainTextValue, {
+    const { result } = await encrypt(encryptKeyring, plainTextValue, {
       encryptionContext: context,
     });
     const base64Value = toBase64(result);
@@ -48,14 +58,17 @@ export const AWSCryptoFactory = (
   const decryptValue = async (encryptedValue: string): Promise<string> => {
     const bufferedValue = fromBase64(encryptedValue);
 
-    const { plaintext, messageHeader } = await decrypt(keyring, bufferedValue);
+    const { plaintext, messageHeader } = await decrypt(decryptKeyring, bufferedValue);
 
     const { encryptionContext } = messageHeader;
 
-    for (const [key, value] of Object.entries(context)) {
-      if (encryptionContext[key] !== value) {
-        throw new Error("Encryption Context does not match expected values");
-      }
+    const acceptedContexts = [context, ...decryptOnlyContexts];
+    if (
+      !acceptedContexts.some((acceptedContext) =>
+        contextMatches(encryptionContext, acceptedContext),
+      )
+    ) {
+      throw new Error("Encryption Context does not match expected values");
     }
 
     const decodedValue = decoder.decode(plaintext);
--- a/api/src/db/DynamoDataClient.ts
+++ b/api/src/db/DynamoDataClient.ts
@@ -1,6 +1,13 @@
-import { BusinessesDataClient, DatabaseClient, UserDataClient } from "@domain/types";
-import { LogWriterType } from "@libs/logWriter";
-import { Business, CURRENT_VERSION, UserData } from "@shared/userData";
+import {
+  type BusinessesDataClient,
+  type DatabaseClient,
+  MigrationConflictError,
+  type MigrationDataClient,
+  type MigrationRunOptions,
+  type UserDataClient,
+} from "@domain/types";
+import { type LogWriterType } from "@libs/logWriter";
+import { type Business, CURRENT_VERSION, type UserData } from "@shared/userData";
 import { chunk } from "lodash";
 import { parseUserData } from "@db/zodSchema/zodSchemas";
 import { getConfigValue } from "@libs/ssmUtils";
@@ -10,13 +17,18 @@ export const DynamoDataClient = (
   businessesDataClient: BusinessesDataClient,
   logger: LogWriterType,
   isKillSwitchOn: () => Promise<boolean>,
+  migrationDataClient?: MigrationDataClient,
 ): DatabaseClient => {
   const MAX_SAFE_MIGRATION_COUNT = 5000;
-  const migrateOutdatedVersionUsers = async (): Promise<{
+  const migrateOutdatedVersionUsers = async (
+    options?: MigrationRunOptions,
+  ): Promise<{
     success: boolean;
     migratedCount?: number;
     error?: string;
   }> => {
+    const canStartNextUser = options?.canStartNextUser ?? ((): boolean => true);
+
     try {
       let nextToken: string | undefined = undefined;
       let migratedCount = 0;
@@ -46,11 +58,14 @@ export const DynamoDataClient = (
             return { success: true, migratedCount };
           }
 
-          await processBatch(batch, isZodParsingOn);
-          migratedCount += batch.length;
+          const batchResult = await processBatch(batch, isZodParsingOn, canStartNextUser);
+          migratedCount += batchResult.migratedCount;
           logger.LogInfo(
-            `Processed batch of ${batch.length} users. Total migrated so far: ${migratedCount}`,
+            `Completed ${batchResult.migratedCount} migrations in this batch. Total migrated so far: ${migratedCount}`,
           );
+          if (batchResult.stoppedForTime) {
+            return { success: true, migratedCount };
+          }
         }
         nextToken = newNextToken;
       } while (nextToken);
@@ -68,27 +83,71 @@ export const DynamoDataClient = (
   const processBatch = async (
     usersToMigrate: UserData[],
     isZodParsingOn: string,
-  ): Promise<void> => {
-    const results = await Promise.allSettled(
-      usersToMigrate.map(async (user) => {
-        await updateUserAndBusinesses(user);
-        logger.LogInfo(`Migrated user ${user.user.id} to version ${CURRENT_VERSION}`);
-        if (isZodParsingOn === "true") {
-          parseUserData(logger, user);
-        }
-      }),
-    );
+    canStartNextUser: () => boolean,
+  ): Promise<{ migratedCount: number; stoppedForTime: boolean }> => {
+    if (!migrationDataClient) {
+      throw new Error("Migration data client is required for coordinated migrations");
+    }
 
-    for (const [index, result] of results.entries()) {
-      const user = usersToMigrate[index];
-      if (result.status === "rejected") {
-        logger.LogError(`Failed to migrate user ${user.user.id}: ${result.reason}`);
+    let migratedCount = 0;
+    for (const user of usersToMigrate) {
+      if (!canStartNextUser()) {
+        logger.LogInfo("Migration paused before Lambda timeout; remaining users will retry");
+        return { migratedCount, stoppedForTime: true };
       }
+
+      try {
+        const migratedUser = await migrationDataClient.migrateAndPut(user);
+        migratedCount += 1;
+        logger.LogInfo(`Migrated user to version ${CURRENT_VERSION}`);
+        if (isZodParsingOn === "true") {
+          parseUserData(logger, migratedUser);
+        }
+      } catch (error) {
+        if (error instanceof MigrationConflictError) {
+          logger.LogInfo("Skipped migration because the user record changed concurrently");
+          continue;
+        }
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        logger.LogError(
+          `Scheduled migration failed for user ${user.user.id} from version ${user.version} to ${CURRENT_VERSION}: ${errorMessage}`,
+        );
+        throw error;
+      }
+    }
+
+    return { migratedCount, stoppedForTime: false };
+  };
+
+  const migrateForRequest = async (sourceUserData: UserData): Promise<UserData> => {
+    if (sourceUserData.version >= CURRENT_VERSION) {
+      return sourceUserData;
+    }
+
+    if (await isKillSwitchOn()) {
+      logger.LogInfo("Request-time migration skipped: kill switch is ON");
+      return sourceUserData;
+    }
+
+    if (!migrationDataClient) {
+      logger.LogError("Request-time migration failed: migration data client is not configured");
+      return sourceUserData;
+    }
+
+    try {
+      return await migrationDataClient.migrateAndPut(sourceUserData);
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      logger.LogError(
+        `Request-time migration failed from version ${sourceUserData.version} to ${CURRENT_VERSION}: ${errorMessage}`,
+      );
+      return sourceUserData;
     }
   };
 
   const get = async (userId: string): Promise<UserData> => {
-    return await userDataClient.get(userId);
+    const sourceUserData = await userDataClient.get(userId);
+    return await migrateForRequest(sourceUserData);
   };
 
   const updateUserAndBusinesses = async (userData: UserData): Promise<void> => {
@@ -126,6 +185,12 @@ export const DynamoDataClient = (
 
   const put = async (userData: UserData): Promise<UserData> => {
     try {
+      if (userData.version < CURRENT_VERSION) {
+        if (!migrationDataClient) {
+          throw new Error("Migration data client is required for coordinated migrations");
+        }
+        return await migrationDataClient.migrateAndPut(userData);
+      }
       await updateUserAndBusinesses(userData);
       return userData;
     } catch (error) {
@@ -136,7 +201,8 @@ export const DynamoDataClient = (
   };
 
   const findByEmail = async (email: string): Promise<UserData | undefined> => {
-    return await userDataClient.findByEmail(email);
+    const sourceUserData = await userDataClient.findByEmail(email);
+    return sourceUserData ? await migrateForRequest(sourceUserData) : undefined;
   };
   const findUserByBusinessName = async (businessName: string): Promise<UserData | undefined> => {
     try {
@@ -146,7 +212,7 @@ export const DynamoDataClient = (
         return undefined;
       }
       const userId = business.userId;
-      const user = await userDataClient.get(userId);
+      const user = await get(userId);
       if (!user) {
         logger.LogInfo(`No user found for business: ${businessName}`);
         return undefined;
@@ -167,9 +233,7 @@ export const DynamoDataClient = (
         logger.LogInfo(`No Businesses Found with prefix: ${prefix}`);
         return [];
       }
-      const users = await Promise.all(
-        businesses.map((business) => userDataClient.get(business.userId)),
-      );
+      const users = await Promise.all(businesses.map((business) => get(business.userId)));
 
       return users as UserData[];
     } catch (error) {
--- a/api/src/db/DynamoMigrationDataClient.test.ts
+++ b/api/src/db/DynamoMigrationDataClient.test.ts
@@ -1,0 +1,144 @@
+import { type DynamoDBDocumentClient, TransactWriteCommand } from "@aws-sdk/lib-dynamodb";
+import { DynamoMigrationDataClient } from "@db/DynamoMigrationDataClient";
+import {
+  MigrationConflictError,
+  type MigrationDataClient,
+  type UserDataClient,
+} from "@domain/types";
+import {
+  generateBusiness,
+  generateProfileData,
+  generateUser,
+  generateUserData,
+} from "@shared/test";
+import { CURRENT_VERSION, type UserData } from "@shared/userData";
+
+const usersTableName = "users-table";
+const businessesTableName = "businesses-table";
+
+const generateOutdatedUser = (businessCount = 2): UserData => {
+  const userId = "migration-user";
+  const businesses = Object.fromEntries(
+    Array.from({ length: businessCount }, (_, index) => {
+      const businessId = `${userId}-business-${index}`;
+      return [
+        businessId,
+        generateBusiness({
+          id: businessId,
+          userId,
+          version: CURRENT_VERSION - 1,
+          profileData: generateProfileData({ businessName: `Business ${index}` }),
+        }),
+      ];
+    }),
+  );
+
+  return generateUserData({
+    user: generateUser({ id: userId }),
+    version: CURRENT_VERSION - 1,
+    currentBusinessId: Object.keys(businesses)[0],
+    businesses,
+  });
+};
+
+const migrateUser = (source: UserData): UserData => ({
+  ...source,
+  version: CURRENT_VERSION,
+  businesses: Object.fromEntries(
+    Object.entries(source.businesses).map(([id, business]) => [
+      id,
+      { ...business, version: CURRENT_VERSION },
+    ]),
+  ),
+});
+
+describe("DynamoMigrationDataClient", () => {
+  let send: jest.Mock;
+  let userDataClient: jest.Mocked<UserDataClient>;
+
+  beforeEach(() => {
+    send = jest.fn().mockResolvedValue({});
+    userDataClient = {
+      get: jest.fn(),
+      findByEmail: jest.fn(),
+      put: jest.fn(),
+      migrateToLatest: jest.fn(async (source) => migrateUser(source)),
+      getNeedNewsletterUsers: jest.fn(),
+      getNeedTaxIdEncryptionUsers: jest.fn(),
+      getUsersWithOutdatedVersion: jest.fn(),
+    };
+  });
+
+  const makeClient = (): MigrationDataClient =>
+    DynamoMigrationDataClient({
+      db: { send } as unknown as DynamoDBDocumentClient,
+      userDataClient,
+      usersTableName,
+      businessesTableName,
+    });
+
+  it("writes the user and every business in one conditional transaction", async () => {
+    const source = generateOutdatedUser();
+
+    const result = await makeClient().migrateAndPut(source);
+
+    expect(result.version).toBe(CURRENT_VERSION);
+    expect(send).toHaveBeenCalledTimes(1);
+    const command = send.mock.calls[0][0] as TransactWriteCommand;
+    expect(command.input.TransactItems).toHaveLength(3);
+    expect(command.input.TransactItems?.[0].Put).toMatchObject({
+      TableName: usersTableName,
+      ConditionExpression: "#data = :sourceData",
+      ExpressionAttributeValues: { ":sourceData": source },
+      Item: { userId: source.user.id, version: CURRENT_VERSION },
+    });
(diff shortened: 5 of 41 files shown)


api/src/client/AwsCryptoFactory.ts
```diff
@@ -29,16 +29,20 @@ export const AWSCryptoFactory = (
...
   context: Context,
   encryptedHashingSalt?: string,
+  decryptOnlyKeyIds: readonly string[] = [],
 ): CryptoClient => {
   const { encrypt, decrypt } = AWSCrypto.buildClient(
     AWSCrypto.CommitmentPolicy.REQUIRE_ENCRYPT_REQUIRE_DECRYPT,
   );
-  const keyring = new AWSCrypto.KmsKeyringNode({ generatorKeyId });
+  const encryptKeyring = new AWSCrypto.KmsKeyringNode({ generatorKeyId });
+  const decryptKeyring = new AWSCrypto.KmsKeyringNode({
```
seidior, line 44: I split encryption and decryption into different keyrings intentionally. If the legacy key were configured as an additional encryption key, new ciphertext would keep carrying a legacy-wrapped data key. The contract here is asymmetric: encrypt with only the current key, but decrypt with the current _or legacy key_ while the backfill converges.

I also kept this as a strict allowlist instead of a discovery keyring so IAM access alone cannot make an unrelated key acceptable.

api/src/db/DynamoUserDataClient.ts
```diff
@@ -24,42 +28,45 @@ const unmarshallOptions = {
...
   logger: LogWriterType,
   migrationClients?: Omit<MigrationClients, "cryptoClient">,
 ): UserDataClient => {
-  const migrateData = async (data: UserData, logger: LogWriterType): Promise<any> => {
+  const migrateToLatest = async (data: UserData): Promise<UserData> => {
     const logId = logger.GetId();
     const dataVersion = data.version ?? CURRENT_VERSION;
     const migrationsToRun = Migrations.slice(dataVersion);
-    let migratedData = data;
+    let migratedData = structuredClone(data);
```
seidior, line 50: The clone is doing more than protecting the caller from mutation. I needed migration to become a pure transformation so every field and business can finish before the first DynamoDB write. The old path caught a failure, kept working on the same object, and then force-stamped `CURRENT_VERSION`; that is how a transient KMS failure could become a permanently stranded record.

api/src/db/DynamoUserDataClient.ts
```diff
@@ -24,42 +28,45 @@ const unmarshallOptions = {
...
         migratedData = await Promise.resolve(
-          migration(migratedData, { cryptoClient, ...migrationClients }),
+          migration(migratedData, { cryptoClient, ...migrationClients, logger }),
         );
       } catch (error) {
         logger.LogError(
-          `Database Migration Error - Id:${logId} - Error: ${error} - Data: ${JSON.stringify(
-            migratedData,
-          )}`,
+          `Database Migration Error - Id:${logId} - Failed upgrading to ${
```
seidior, line 63: I removed the serialized user record from this error intentionally. It was noisy and could put email addresses, ciphertext, and unrelated user data into CloudWatch without saying which operation failed. The migration and field wrappers now add the useful context, while this layer reports the failed version transition and rethrows.

api/src/db/DynamoMigrationDataClient.ts
```diff
@@ -0,0 +1,104 @@
...
+import { createBusinessDataItem } from "@db/DynamoBusinessDataClient";
+import { createUserDataItem } from "@db/DynamoUserDataClient";
+import {
+  MigrationConflictError,
+  type MigrationDataClient,
+  type UserDataClient,
+} from "@domain/types";
+import { type UserData } from "@shared/userData";
+
+const MAX_TRANSACTION_BUSINESSES = 99;
```
seidior, line 15: DynamoDB transactions allow 100 items, so 99 businesses is the hard ceiling after reserving one write for the user. I chose to stop and alert above that limit rather than split the migration because splitting would reintroduce the partial-write state this client exists to prevent.

api/src/db/DynamoMigrationDataClient.ts
```diff
@@ -0,0 +1,104 @@
...
+        `Cannot atomically migrate user with more than ${MAX_TRANSACTION_BUSINESSES} businesses`,
+      );
+    }
+
+    const transactItems: NonNullable<TransactWriteCommandInput["TransactItems"]> = [
+      {
+        Put: {
+          TableName: usersTableName,
+          Item: createUserDataItem(migratedUserData),
+          ConditionExpression: "#data = :sourceData",
```
seidior, line 78: The condition is deliberately only on the users-table snapshot. That item is canonical and protects us from overwriting a concurrent user update. I did not condition on the business projections because existing projection drift is one of the states this migration needs to repair; a successful transaction overwrites those mirrors from the canonical snapshot. This does not prevent a stale browser from writing an older version after the transaction commits, which is real but broader write-contract work that I left as a follow-up.

api/src/db/DynamoMigrationDataClient.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      migrateToLatest: jest.fn(async (source) => migrateUser(source)),
+      getNeedNewsletterUsers: jest.fn(),
+      getNeedTaxIdEncryptionUsers: jest.fn(),
+      getUsersWithOutdatedVersion: jest.fn(),
+    };
+  });
+
+  const makeClient = (): MigrationDataClient =>
+    DynamoMigrationDataClient({
+      db: { send } as unknown as DynamoDBDocumentClient,
```
seidior, line 74: I am deliberately mocking the document client in this suite despite our normal preference for DynamoDB Local. The local engine used by this repo does not implement `TransactWriteItems`, so these tests lock the transaction shape and failure classification. The actual transaction still needs a deployed dev DynamoDB check.

api/src/db/DynamoDataClient.ts
```diff
@@ -68,27 +83,71 @@ export const DynamoDataClient = (
...
-        await updateUserAndBusinesses(user);
-        logger.LogInfo(`Migrated user ${user.user.id} to version ${CURRENT_VERSION}`);
+    canStartNextUser: () => boolean,
+  ): Promise<{ migratedCount: number; stoppedForTime: boolean }> => {
+    if (!migrationDataClient) {
+      throw new Error("Migration data client is required for coordinated migrations");
+    }
+
+    let migratedCount = 0;
+    for (const user of usersToMigrate) {
```
seidior, line 93: I changed this from `Promise.allSettled` concurrency to serial processing on purpose. A malformed record or structural KMS/configuration failure should stop the scheduled backfill immediately, and the remaining-time check needs a safe boundary between users. Transaction conflicts are the exception: no writes committed, so the newer record can be considered again later.

api/src/db/DynamoDataClient.ts
```diff
@@ -68,27 +83,71 @@ export const DynamoDataClient = (
...
+    }
+
+    try {
+      return await migrationDataClient.migrateAndPut(sourceUserData);
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      logger.LogError(
+        `Request-time migration failed from version ${sourceUserData.version} to ${CURRENT_VERSION}: ${errorMessage}`,
+      );
+      return sourceUserData;
```
seidior, line 144: This request-time catch is intentionally different from scheduled behavior. A user-data read should still return the unchanged stored record if migration fails, so this logs non-sensitive context and relies on the scheduled Lambda to encounter persistent failures and own the kill switch.

api/src/db/migrations/v191_rotate_new_kms_keys.ts
```diff
@@ -6,34 +6,81 @@ import {
 import { randomInt } from "@shared/intHelpers";
 import { type MigrationClients } from "@db/migrations/types";
 
+const decryptField = async (
+  fieldName: string,
+  encryptedValue: string,
+  clients: MigrationClients,
+): Promise<string> => {
+  try {
+    return await clients.cryptoClient.decryptValue(encryptedValue);
```
seidior, line 15: Editing an already-deployed migration is intentional here. A v190 record cannot jump directly to v193, and v191/v192 can encounter either legacy- or current-wrapped ciphertext depending on how far an earlier attempt got. Both existing migrations need the compatible decrypt client or the catch-up migration remains unreachable for the records that need it most.

api/src/db/migrations/v191_rotate_new_kms_keys.ts
```diff
@@ -6,34 +6,81 @@ import {
...
+
+const hashTaxId = async (plaintext: string, clients: MigrationClients): Promise<string> => {
+  if (!clients.newHashingClient) {
+    throw new Error("newHashingClient is required");
   }
 
-  const plaintext = await clients.legacyTaxIdCryptoClient.decryptValue(encryptedValue);
-  return await clients.cryptoClient.encryptValue(plaintext);
+  let lastError: unknown;
+  for (let attempt = 1; attempt <= 3; attempt += 1) {
```
seidior, line 69: I only added an explicit retry loop around hashing. KMS calls already get AWS SDK retry and backoff; wrapping those calls again would multiply attempts and make the Lambda duration less predictable. Hashing is local and previously had no retry behavior, so this is capped at three attempts and logs migration, field, and attempt before failing the whole user.

api/src/db/migrations/v192_fix_confirmation_email_sent_typo.ts
```diff
@@ -7,34 +7,81 @@ import {
 import { randomInt } from "@shared/intHelpers";
 import { type MigrationClients } from "@db/migrations/types";
 
+const decryptField = async (
+  fieldName: string,
+  encryptedValue: string,
+  clients: MigrationClients,
+): Promise<string> => {
+  try {
+    return await clients.cryptoClient.decryptValue(encryptedValue);
```
seidior, line 16: Yes, v192 still repeats crypto work even though its named schema change is the confirmation-email typo. I considered removing that historical behavior, but v192 is already deployed and a broader rewrite would be harder to reason about during this repair. Making its decryption compatible lets stranded records pass through it, while v193 remains the explicit catch-up. The extra current-key round trip is a one-time cost I accepted here.

api/src/db/migrations/v193_rotate_stranded_legacy_kms_fields.ts
```diff
@@ -0,0 +1,1368 @@
...
+    clients,
+  );
+
+  const encryptedDeptOfLaborEin = await rotateField(
+    "profileData.deptOfLaborEin",
+    business.profileData.deptOfLaborEin,
+    clients,
+  );
+
+  // These fields exist in stored records but were omitted from the v191/v192 type tree.
```
seidior, line 101: This cast is awkward but deliberate. These tax-clearance encrypted fields exist in stored runtime records and were omitted from the v191/v192 type trees. Trusting those older types here would stamp v193 while leaving exactly the legacy ciphertext this migration is supposed to remove.

api/src/db/migrations/v193_rotate_stranded_legacy_kms_fields.ts
```diff
@@ -0,0 +1,1368 @@
...
+    cigaretteLicenseData: business.cigaretteLicenseData
+      ? {
+          ...business.cigaretteLicenseData,
+          encryptedTaxId: cigaretteEncryptedTaxId,
+        }
+      : business.cigaretteLicenseData,
+  };
+};
+
+export interface v193IndustrySpecificData {
```
seidior, line 141: Most of the rest of this file is the repository's complete versioned type tree and generators, not new KMS logic. I considered broader migration-framework cleanup, but it would expand an incident repair across unrelated ownership boundaries. Keeping the conventional full v193 snapshot lets the existing schema tooling and migration checks continue to work.

api/src/functions/migrateUsersVersion/app.ts
```diff
@@ -62,17 +68,29 @@ export const handler = async (): Promise<void> => {
...
   const dynamoDataClient = DynamoDataClient(
     userDataClient,
     businessesDataClient,
     logger,
     isKillSwitchOn,
+    migrationDataClient,
   );
-  await dynamoDataClient.migrateOutdatedVersionUsers();
+  const result = await dynamoDataClient.migrateOutdatedVersionUsers({
+    canStartNextUser: () =>
```
seidior, line 92: The 30-second buffer is a successful soft stop, not an error. Already committed users stay migrated and the next scheduled invocation resumes the rest. This lets a slow but healthy KMS workload make progress without timing out and tripping the structural-failure alarm.

api/src/functions/migrateUsersVersion/app.ts
```diff
@@ -62,17 +68,29 @@ export const handler = async (): Promise<void> => {
...
+      !context || context.getRemainingTimeInMillis() > MIGRATION_SHUTDOWN_BUFFER_MS,
+  });
+  if (!result.success) {
+    throw new Error(result.error ?? "User migration failed");
```
seidior, line 96: This throw is needed even though the migration path already logged the failure. Lambda only increments its native `Errors` metric when the invocation rejects; resolving here would make the operator alarm and automatic kill switch miss a terminal migration failure.

api/cdk/lib/lambdaStack.ts
```diff
@@ -352,6 +352,9 @@ export class LambdaStack extends Stack {
         AWS_CRYPTO_TAX_ID_ENCRYPTION_KEY: awsCryptoTaxIdEncryptionKey,
         LEGACY_AWS_CRYPTO_TAX_ID_ENCRYPTION_KEY: legacyAwsCryptoTaxIdEncryptionKey,
         AWS_CRYPTO_CONTEXT_TAX_ID_ENCRYPTION_PURPOSE: awsCryptoContextTaxIdEncryptionPurpose,
+        AWS_CRYPTO_TAX_ID_HASHING_KEY: awsCryptoTaxIdHashingKey,
```
seidior, line 355: These hashing settings were already read by the handler but were not passed into this Lambda. That meant a record reaching v191/v192 hashing could fail with `Salt must be provided` even after KMS decryption was fixed. This configuration is part of making older records able to reach v193, not an unrelated Lambda cleanup.

api/cdk/lib/monitoringStack.ts
```diff
@@ -115,23 +109,14 @@ export class MonitoringStack extends Stack {
...
-        metricNamespace: "BFS/Navigator/Lambda",
-        metricName: "MigrateUserVersionsFailures",
-        filterPattern: logs.FilterPattern.literal('"MigrateUserVersions Failed"'),
-        metricValue: "1",
-        defaultValue: 0,
-      },
-    );
-
-    const migrationFailureMetric = migrationFailureMetricFilter.metric({
+    const migrationFailureMetric = new cloudwatch.Metric({
```
seidior, line 112: The previous alarm matched a log phrase the code never emitted, while the handler could return successfully after migration failure. I switched the signal to the native `AWS/Lambda` `Errors` metric so it follows invocation outcome rather than log wording, then routed the same failure to both operator notification and the existing kill-switch topic.

seidior (commented):
Self-Review

api/src/db/migrations/v191_rotate_new_kms_keys.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        newHashingClient: hashingClient,
+        logger,
+      }),
+    ).rejects.toThrow("terminal hashing failure");
+    expect(hashingClient.hashValue).toHaveBeenCalledTimes(3);
+    expect(logger.LogError).toHaveBeenCalledTimes(3);
+  });
+});
+
+describe("migrate_v191_to_v192 hashing", () => {
```
ashlsun, line 79: Should probably move this to a v192 test file

api/src/db/migrations/v192_fix_confirmation_email_sent_typo.ts
aadedejifearless: We can refactor this in a later PR for use across both v191 and v192

PR #12940, opened 2026-08-05
feat: [AB#17961] add callout to create an account on locked task pages

ashlsun:
## Description

As part of the login modal mission, we want to alert guest users that the task they're looking at requires an account.

This PR adds this yellow callout to the following tasks: 
- NAICS code
- Check Available Names and Form Your Business
- Get Your EIN
- Register for Taxes
- Cannabis Priority Status. 

Screenshot 2026-08-05 at 10 44 05 AM

### Ticket

This pull request resolves [#17961](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17961).

### Approach
I made a component that wraps the MiniCallout NJWDS component and does not appear if a user is already authenticated.

Some tasks have another alert (`UnlockedBy` component) that warns users of task prerequisites that haven't been completed. Because multiple alerts on the page look crowded and in most cases, lack of an account blocks users from completing prereqs anyway, I hide the prereq alert when the user doesn't have an account.

I added it only to the `needsAccount: true` tasks that show up in the required-only roadmap, which doesn't include non-required tasks that no longer show up in the roadmap (EIN for domestic employers) and License status tasks, which show the login modal immediately upon opening. Given the forthcoming roadmap redesign, these tasks may never be shown to users, so accounting for them is not worth the effort at this moment.

### Steps to Test

1. Onboard as a new business (while logged out)
2. Click into any task with a lock icon
3. The task page should show a callout that you need to create an account.

When logged in, follow the same steps, and the task page should not show this callout.

## Code author checklist

### DIFF
--- a/web/src/components/NeedsAccountMiniCallout.tsx
+++ b/web/src/components/NeedsAccountMiniCallout.tsx
@@ -1,0 +1,19 @@
+import { Content } from "@/components/Content";
+import { MiniCallout } from "@/components/njwds-extended/callout/MiniCallout";
+import { NeedsAccountContext } from "@/contexts/needsAccountContext";
+import { IsAuthenticated } from "@/lib/auth/AuthContext";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { ReactElement, useContext } from "react";
+
+export const NeedsAccountMiniCallout = (): ReactElement => {
+  const { isAuthenticated } = useContext(NeedsAccountContext);
+  const { Config } = useConfig();
+
+  return isAuthenticated === IsAuthenticated.FALSE ? (
+    <MiniCallout calloutType="warning">
+      <Content>{Config.taskDefaults.needsAccountCallout}</Content>
+    </MiniCallout>
+  ) : (
+    <></>
+  );
+};
--- a/content/src/fieldConfig/task-defaults.json
+++ b/content/src/fieldConfig/task-defaults.json
@@ -18,6 +18,7 @@
     "previousTaskButtonText": "Previous Task",
     "formNameText": "Form",
     "defaultCallToActionText": "Start Application",
-    "unlockedByPlural": "Before starting this task, we recommend you take a look at:"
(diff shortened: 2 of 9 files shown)


seidior (approved):
I like it. I wonder though if this is a good opportunity to use contextual info to explain to users what an account gets them, since that may be the first time they're seeing a request to create an account?

web/src/components/NeedsAccountMiniCallout.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { ReactElement, useContext } from "react";
+
+export const NeedsAccountMiniCallout = (): ReactElement => {
+  const { isAuthenticated } = useContext(NeedsAccountContext);
+  const { Config } = useConfig();
+
+  return isAuthenticated === IsAuthenticated.FALSE ? (
+    <MiniCallout calloutType="warning">
+      <Content>{Config.taskDefaults.needsAccountCallout}</Content>
```
seidior, line 14: Is there a world where we could do:

```jsx
<Content>
   :::miniCallout
   {Config.taskDefaults.needsAccountCallout}
   :::
</Content>
```

Instead of the direct `<MiniCallout>` code to reuse the MiniCallout code from our current markdown processing pipeline? 

(see `web/src/lib/cms/editors/mini-callout.tsx`)

aadedejifearless (approved):
I noticed that where we have multiple callouts  in the locked tasks,  these callouts are stacked vertically and can take up 
quite a bit of space. Would it be possible to combine them into a single 
callout when multiple exist? This might help reduce the vertical height.

web/src/components/NeedsAccountMiniCallout.tsx
ashlsun: Yes, this is possible and would be equivalent! Do you think there are benefits to routing it through markdown processing? I feel like this syntax is useful to give content team members the flexibility to use these elements through the CMS, but when hardcoded into a component like this, it is more opaque (more logic to trace through) than using the React component directly.

PR #12943, opened 2026-08-06
fix: harden KMS ciphertext migrations

seidior:
## Description

Encrypted user fields that were copied between environments, or that are still wrapped by a retired KMS key, caused migrations to fail hard and halt the scheduled migration run.

The crypto client now inspects ciphertext metadata before attempting a decrypt, so migrations can tell accepted ciphertext apart from foreign or unrecognized ciphertext. Foreign-environment values are reset in non-production environments, and unrecognized values are quarantined in production rather than cleared, so potentially valid user data is never destroyed.

### Ticket

This pull request continues work on [#17957](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17957).

### Approach

- Add `api/src/client/CiphertextMetadata.ts`, which deserializes the AWS Encryption SDK message header to read the encryption context and wrapping key ARNs without a KMS call, then validates both against the accepted contexts and allowed key IDs.

- Give `AWSCryptoFactory` an options argument carrying a `foreignEnvironmentPolicy`, and run header validation before decrypting when a policy is set.

- Add two typed errors in `api/src/domain/types.ts`: `ForeignEnvironmentCiphertextError` (ciphertext from another stage, safe to reset) and `QuarantinedCiphertextError` (unknown context or wrapping key, needs manual review), plus `isQuarantinedCiphertextError`, which walks the cause chain.

- Resolve the policy from stage in `api/src/functions/config.ts`: `prod` quarantines, other deployed stages reset, and `local` or an unset crypto context stage skip validation entirely.

- Extract the decrypt and re-encrypt helper that was duplicated across v191, v192, and v193 into a shared `rotateKmsField` in `api/src/db/migrations/kms_migration_utils.ts`. It returns `wasReset` so callers also clear the paired plaintext, masked, and hashed fields instead of leaving them stranded.

- Let the scheduled migration count quarantined users and keep going rather than failing the run and tripping the global kill switch. `migrateOutdatedVersionUsers` now reports `quarantinedCount`.

- Add a CloudWatch metric filter and alarm on the `migrateUsersVersion` log group so a quarantined record notifies through the existing migration error SNS topic.

- Add a read-only audit script, `api/scripts/audit-kms-key-usage.sh`, that scans a users table and reports current, legacy, foreign-context, unknown-key, and invalid ciphertext counts as CSV.

- Sanitize protected identifiers in `api/scripts/assume-user-dev.sh`, so copying a production record into dev no longer plants foreign ciphertext in the first place.

### Steps to Test

This work is backend and infrastructure only, with no direct change to a user's experience. To verify:

1. `yarn workspace @businessnjgovnavigator/api test` and confirm the `CiphertextMetadata`, `config`, `DynamoDataClient`, `v191`, and `v193` suites pass.

2. `yarn workspace @businessnjgovnavigator/api-cdk test` and confirm the `monitoringStack` test covering the quarantined-migration alarm passes.

3. `yarn typecheck` from the repo root.

4. In a deployed non-production environment, load a business whose encrypted tax ID was written under a different stage. The field should come back empty and be re-enterable, rather than erroring, and the migration log should record the reset.

5. Run `api/scripts/audit-kms-key-usage.sh --profile <profile> --table <users-table> --accepted-stages <stage>` against a non-production table and confirm the CSV summary rows.

### Notes

- No new migration file is added. v191 through v193 were strengthened in place, so the user data schema version is unchanged.

- The production and non-production policies are deliberately asymmetric. Resetting is safe in lower environments, where foreign ciphertext is almost always a copied record, but in production an unrecognized wrapping key could still be recoverable, so those records are quarantined for manual review instead.

- The alarm threshold is 1, so any single quarantined record raises a notification. Quarantined users stay on their old version until someone resolves them.

- Adds `@aws-crypto/serialize` and `@aws-crypto/material-management-node` (both 4.2.1, matching the existing `@aws-crypto/client-node` pin) to read message headers.

- Incidental housekeeping picked up along the way: `.playwright-mcp` added to `.eslintignore`, `ciphertext` added to the content spellcheck exceptions, and a corrected `userSchemaCompiler.ts` path in `scripts/generate-new-migration.sh`.

## Code author checklist


api/src/functions/config.ts
```diff
@@ -4,6 +4,8 @@ export const BUSINESSES_TABLE = process.env.BUSINESSES_TABLE || "businesses-tabl
 export const MESSAGES_TABLE = process.env.MESSAGES_TABLE || "messages-table-local";
 export const DYNAMO_OFFLINE_PORT = Number.parseInt(process.env.DYNAMO_PORT || "8000");
 export const STAGE = process.env.STAGE || "local";
+export const AWS_CRYPTO_TAX_ID_FOREIGN_ENVIRONMENT_POLICY =
+  STAGE === "local" ? undefined : STAGE === "prod" ? "quarantine" : "reset";
 
 export const AWS_CRYPTO_CONTEXT_ORIGIN = process.env.AWS_CRYPTO_CONTEXT_ORIGIN || "";
 export const AWS_CRYPTO_CONTEXT_STAGE = process.env.AWS_CRYPTO_CONTEXT_STAGE || "";
```
ashlsun, line 11: Claude suggests reordering these so that if we ever have a config error and `AWS_CRYPTO_CONTEXT_STAGE` happens to be the empty string, we can make sure the foreign environment policy is not `"reset"`, as otherwise  our migrations could wipe out all the encrypted values in one of our lower environments.
```
export const AWS_CRYPTO_CONTEXT_STAGE = process.env.AWS_CRYPTO_CONTEXT_STAGE || "";
export const AWS_CRYPTO_CONTEXT_ORIGIN = process.env.AWS_CRYPTO_CONTEXT_ORIGIN || "";
export const AWS_CRYPTO_TAX_ID_FOREIGN_ENVIRONMENT_POLICY =
    STAGE === "local" || !AWS_CRYPTO_CONTEXT_STAGE ? undefined : STAGE === "prod" ? "quarantine" : "reset";
````

File: api/src/db/DynamoDataClient.ts [hunk 2]
```diff
@@ -109,14 +127,21 @@ export const DynamoDataClient = (
           continue;
         }
         const errorMessage = error instanceof Error ? error.message : String(error);
+        if (isQuarantinedCiphertextError(error)) {
+          quarantinedCount += 1;
+          logger.LogError(
+            `Quarantined scheduled migration for user ${user.user.id} from version ${user.version}: ${errorMessage}`,
```
ashlsun, line 133: We should probably have an alert that catches this, since this lambda exits successfully (so it won't be caught by the Errors `migrationFailureMetric`)  and the message also doesn't match the "Dynamo User Migration Error" metric filter in `monitoringStack.ts`,

aadedejifearless: we could add this in `monitoringStack.ts` based on the above suggestion

``````suggestion

  const quarantinedMigrationMetricFilter = new logs.MetricFilter(
      this,
      "QuarantinedMigrationMetricFilter",
      {
        logGroup: dataMigrationLogGroup,
        filterName: `quarantinedMigrationCount-${props.stage}`,
        filterPattern: logs.FilterPattern.literal("Quarantined scheduled migration"),
        metricNamespace: "BFS/Navigator",
        metricName: `quarantinedMigrationCount`,
        metricValue: "1",
        defaultValue: 0,
      },
    );

     const quarantinedMigrationTopic = new sns.Topic(this, "QuarantinedMigrationTopic", {
        topicName: `bfs-navigator-${props.stage}-quarantined-migration-users`,
      });

      new sns.Subscription(this, "QuarantinedMigrationSubscription", {
        topic: quarantinedMigrationTopic,
        protocol: sns.SubscriptionProtocol.HTTPS,
        endpoint: "https://global.sns-api.chatbot.amazonaws.com",
      });

      quarantinedMigrationTopic.applyRemovalPolicy(RemovalPolicy.RETAIN);

      const quarantinedMigrationMetric = quarantinedMigrationMetricFilter.metric({
        statistic: "Sum",
        period: Duration.seconds(120),
      });

      const quarantinedMigrationAlarm = new cloudwatch.Alarm(
        this,
        "QuarantinedMigrationAlarm",
        {
          alarmName: `${props.stage}-bfs-navigator-quarantined-migration-users`,
          metric: quarantinedMigrationMetric,
          threshold: 1,
          evaluationPeriods: 1,
          datapointsToAlarm: 1,
          comparisonOperator:
            cloudwatch.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
          treatMissingData: cloudwatch.TreatMissingData.NOT_BREACHING,
          alarmDescription:
            "Triggered when a user record is quarantined during KMS migration due to a ciphertext that cannot be safely decrypted or re-encrypted. Quarantined records require manual review.",
        },
      );

      quarantinedMigrationAlarm.addAlarmAction(
        new cloudwatch_actions.SnsAction(quarantinedMigrationTopic),
      );
      quarantinedMigrationAlarm.applyRemovalPolicy(RemovalPolicy.RETAIN);
```

seidior: Added. I targeted `/aws/lambda/businessnjgov-api-v2-${stage}-migrateUsersVersion` because `LogError` writes the quarantine marker to `console.error`; the proposed `dataMigrationLogGroup` does not receive this handler's logs. The alarm fires at one quarantine and reuses the existing migration-operator SNS topic. It does not publish to `migrationLambdaTopic`, so it cannot activate the global kill switch.

api/src/functions/config.ts
seidior: Good catch. I reordered this and made the policy resolution explicit: local or an empty `AWS_CRYPTO_CONTEXT_STAGE` now returns `undefined`, production with a configured stage returns `quarantine`, and configured lower environments return `reset`. I also added a focused policy matrix test so an empty stage cannot regress back to destructive reset.

ashlsun (approved):

PR #12947, opened 2026-08-07
feat: [AB#17963] remove step numbers from onboarding pages

ashlsun:
## Description
In advance of further changes to the onboarding flow, we're removing the step numbers on each page of onboarding. This PR removes that logic and the tests related to it. It also renames some variables and test titles for clarity, now that we no longer show the user the "(Step X)" copy.

### Ticket

This pull request resolves [#17963](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17963).

## Code author checklist

### DIFF
--- a/web/src/pages/onboarding.tsx
+++ b/web/src/pages/onboarding.tsx
@@ -39,7 +39,6 @@ import {
 } from "@/lib/utils/analytics-helpers";
 import { getFlow, scrollToTop } from "@/lib/utils/helpers";
 import {
-  evalHeaderStepsTemplate,
   flowQueryParamIsValid,
   getAnimation,
   getTimeout,
@@ -318,9 +317,9 @@ const OnboardingPage = (props: Props): ReactElement => {
     setProfileData(newProfileData);
     setCurrentFlow(flowType);
 
-    /* The logic below used to skip the first step of onboarding for STARTING businesses,
+    /* The logic below used to skip the first page of onboarding for STARTING businesses,
     it has been temporarily commented out because it is incompatible with our current AB
-    test to skip the second onboarding step.
+    test to skip the second onboarding page.
 
     if (flowType === "OWNING") {
       setPage({ current: 1, previous: 1 });
@@ -341,8 +340,8 @@ const OnboardingPage = (props: Props): ReactElement => {
     setRegistrationDimension("Onboarded Guest");
 
     isAdditionalBusiness
-      ? analytics.event.onboarding_last_step_save_additional_business_button.click.finish_additional_business_onboarding()
-      : analytics.event.onboarding_last_step.submit.finish_onboarding();
+      ? analytics.event.onboarding_last_page_save_additional_business_button.click.finish_additional_business_onboarding()
+      : analytics.event.onboarding_last_page.submit.finish_onboarding();
 
     updateQueue.queueBusiness({
       ...updateQueue.currentBusiness(),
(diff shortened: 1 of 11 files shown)


seidior (commented):
The ADO ticket mentions that this should be under a feature flag, but I'm not seeing any current files changed around handling this under a feature flag. Did the requirements on the ticket change perhaps?

seidior:
Ashley's confirmed the feature flag is out of scope for this ticket.

web/src/pages/onboarding.tsx
```diff
@@ -502,9 +501,9 @@ const OnboardingPage = (props: Props): ReactElement => {
     return (
```
seidior, line 501: [pebble] I see the visible step suffix and most of the surrounding vocabulary are gone, but the completion path still exposes the old model through `onboarding_last_step...`. Some of those values then remain in `analytics.ts` and `analytics-legacy.ts`. For example:

- onboarding_last_step_save_additional_business_button
- onboarding_last_step

- "onboarding_last_step_save_additional_business_button"

- etc.

Could we rename the internal onboarding analytics API around completion/final-page behavior and explicitly resolve the legacy payload compatibility? If reporting requires the two legacy category strings to remain, keeping only those boundary values with a compatibility note would let the rest of the code stop modeling onboarding as steps.

web/test/pages/onboarding/onboarding-foreign.test.tsx
```diff
@@ -379,7 +355,7 @@ describe("onboarding - foreign business", () => {
       const { page } = renderPage({ userData });
       page.clickNext();
 
-      expect(screen.getByTestId("step-3")).toBeInTheDocument();
+      expect(screen.getByTestId("page-3-header")).toBeInTheDocument();
```
seidior, line 358: [pebble] This file still has the old onboarding-step model in its test names and selectors: the Step 2/3/4 descriptions, both `Nexus - ... step` suite names, and three `queryByTestId("step-4")` assertions.

The selectors are a behavioral problem as well as cleanup: production now emits only `page-<n>-header`, so each `step-4` assertion passes regardless of whether page 4 was rendered. Could we convert every remaining reference in this file to page terminology and query `page-4-header` so these tests continue protecting navigation? Or do you think that it's better to keep some of these references?

web/test/pages/onboarding/onboarding-shared.test.tsx
seidior, line 1: [pebble] Two old references remain here: “updates locally for each step” and the comment about “routing between steps.” What if we change both to your new onboarding-page terminology so this suite no longer references the old step-based model?

web/test/pages/onboarding/helpers-onboarding.tsx
```diff
@@ -294,9 +294,9 @@ export const composeOnBoardingTitle = (step: string, pageTitle?: string): string
         additional: "additional",
       },
     });
-    // return `${pageTitleDefault} ${step}`;
-    return `${[pageTitleDefault, step].join(" ")}`;
+    // return `${pageTitleDefault}`;
```
seidior: [sand] Curious to your thoughts on the value of keeping these comments now that they're direct commented-out copies of the return statements beneath them.

seidior (commented):
Just a couple of comments but nothing that needs an active re-review! Great work in addressing the needs in the ticket.

web/test/pages/onboarding/helpers-onboarding.tsx
ashlsun: Thanks for catching that, removed!

web/test/pages/onboarding/onboarding-foreign.test.tsx
ashlsun: Yes, there's no need to keep any references to steps here, and converted all the queries to page-4-header. I missed page numbers above 3 while doing find and replace, thanks!

web/src/pages/onboarding.tsx
ashlsun: OK, I will change the internal name in the analytics API but keep the legacy_event_category value the same because these GA4 values aren't mutable and I don't want break continuity in anybody's analyses of onboarding. I do think "steps" is fairly intuitive for a new engineer to grasp, even if everything else refers to these as "pages", so I don't think leaving this will cause any significant problem.

PR #12962, opened 2026-08-17
[AB#17904]: Add NJ feedback widget to static site

hinzed1127:
## Description

Adds the official [@newjersey/feedback-widget](https://github.com/newjersey/feedback-widget) web component to every static-site route.

### Ticket

This pull request resolves [#17904](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17904).

### Approach

**New env var:** `NEXT_PUBLIC_FEEDBACK_WIDGET_MODE` — `prod` | `mock` | `off`. Unset or unrecognized values fail closed to `off`. Added to `.env-template`, the Dockerfile, and the deploy workflows.

- `prod` — real submissions
- `mock` — widget renders and is fully interactive, but `fetch` to the feedback API is stubbed client-side, so nothing leaves the browser. Mirrors the `mockFetch` helper NJWDS ships for its own component docs.
- `off` — not rendered, and the widget bundle is never loaded

A few implementation notes worth knowing before reviewing:

- The widget bundle calls the browser `window` API with no `typeof window` guard, so it **must** load via a dynamic `import()` inside `useEffect`. A top-level import would crash `next build` with `window is not defined`.
- While the widget offers i18n integration via a ?lang=` query param, this app doesn't use that pattern. The component dispatches the widget's documented `changeLanguage` event instead so Spanish visitors get Spanish copy.
- `footer.css` adds a light amount of overrides so that the widget text and buttons still line up with the footer's `.grid-container` content instead of the widget's own smaller default padding.
- The package ships no types, hence the ambient `declare module` in`types/newjersey-feedback-widget.d.ts`. It has to be a separate file from `types/feedback-widget.d.ts`. The latter imports from `react`, which makes it a module and turns `declare module` into an augmentation rather than an ambient declaration.

### Steps to Test

From `packages/static-site`:

```bash
pnpm install
NEXT_PUBLIC_FEEDBACK_WIDGET_MODE=mock NEXT_PUBLIC_MULTILINGUAL_ENABLED=true pnpm dev
```

1. **Placement** — open http://localhost:4000/en-US and scroll to the bottom. The widget sits inside the footer, just below "Return to top" and above the "BUSINESS.NJ.GOV" logo row, asking "Did you find what you were looking for on this page?" with Yes/No buttons. Its gray background spans the full viewport width, while the prompt text and buttons line up with the footer's other content.
2. The widget should render and every route.
3. **Spanish** — open http://localhost:4000/es-US and validate the language switching mechanism is working.
4. **Mock mode sends nothing** — with DevTools Network open, click **Yes**. The widget advances to its comment step without any requests to `innovation.nj.gov`.
5. **Analytics still fires** — in the DevTools console, run `window.dataLayer.filter(e => e.event === "click_feedback_rating")` after clicking. You should see one entry.
6. **Off switch** — restart with `NEXT_PUBLIC_FEEDBACK_WIDGET_MODE=off pnpm dev` and reload. `document.querySelector("feedback-widget")` returns `null`.

Accessibility (the config now enables the widget automatically, so no env var needed):

```bash
pnpm test:accessibility
```

Three failures are pre-existing on `main` and unrelated: `not-found` en-US and es-US fail a `toHaveTitle` assertion, and one of `newsletter-signup` / `our-software-and-reuse` may fail intermittently from parallel workers racing dev-server route compilation (passes when run alone).

### Notes

1. **GitHub variables are already  set.** `NEXT_PUBLIC_FEEDBACK_WIDGET_MODE` is set per-environment. Prod should be ready to go live upon the merging and deployment of this PR.
2. Since this migrates from the previous webflow site, the Looker Studio dashboard already exists.

## Code author checklist

- [ ] ~~I have created and/or updated relevant documentation on the engineering documentation website~~ N/A


packages/static-site/domain/feedback/feedbackWidgetMode.ts
```diff
@@ -0,0 +1,38 @@
...
+ * the mock transport can recognize which requests to intercept.
+ */
+export const FEEDBACK_API_BASE_URL = "https://innovation.nj.gov/app/feedback/dev";
+
+/**
+ * - `prod` renders the widget and lets it submit for real.
+ * - `mock` renders the widget but stubs its network calls, so lower
+ *   environments stay interactive without writing to the shared dataset.
+ * - `off` does not render the widget at all.
+ */
```
hinzed1127, line 20: Debated between this and having 2 separate flags:

- `ENABLE_FEEDBACK_WIDGET`
- `MOCK_FEEDBACK_WIDGET`

Even though this feels slightly less conventional, it's one config value to deal with instead of 2 and just felt a little cleaner to me.

packages/static-site/components/feedback/FeedbackWidget.tsx
```diff
@@ -0,0 +1,90 @@
...
+    if (!isEnabled || !isDefined) {
+      return;
+    }
+
+    document.dispatchEvent(
+      new CustomEvent(CHANGE_LANGUAGE_EVENT, {
+        detail: toFeedbackWidgetLanguage(locale),
+        bubbles: true,
+      }),
+    );
```
hinzed1127, line 76: called out in the top-level comment, but [here's the official widget guidance on translations](https://github.com/newjersey/feedback-widget#spanish-content)

hinzed1127 (commented):
Self-review. The diff is slightly bloated due to test mocks. The core logic is in:
- `FeedbackWidget.tsx`, where the actual web component is injected with all of the various React/Next.js safeguards
- `feedbackWidgetMode.ts`, which is the feature flag type logic
- `mockFeedbackTransport.ts` - Response mocking for non-prod environments

packages/static-site/domain/feedback/feedbackWidgetLanguage.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * The widget ships copy for English and Spanish only, and identifies them by
+ * bare language subtags rather than the full locale tags this app routes on.
+ */
```
hinzed1127, line 4: we'd have to adapt our approach if we want to add support for Arabic or other potential languages in the future

williamavasquez (approved):

packages/static-site/components/feedback/FeedbackWidget.tsx
```diff
@@ -0,0 +1,90 @@
...
+      await import("@newjersey/feedback-widget/feedback-widget.min.js");
+
+      if (isActive) {
+        setIsDefined(true);
+      }
+    };
+
+    defineCustomElement().catch((error: unknown) => {
+      // biome-ignore lint/suspicious/noConsole: a failed third-party widget load should surface in the console without breaking the page.
+      console.error("Failed to load the NJ feedback widget.", error);
```
ashlsun, line 57: Not opposed to having this here, could see it being useful for devs actively testing the site, but if a user encounters this error and it gets logged in their browser without them noticing, how will we get visibility into these errors happening?

packages/static-site/playwright.a11y.config.ts
```diff
@@ -24,8 +24,17 @@ const playwrightAccessibilityConfig = defineConfig({
   },
   webServer: {
     command: "pnpm dev --port 3000",
-    // biome-ignore lint/style/noProcessEnv: playwright webServer inherits the host env and reads CI flag.
-    env: { ...process.env, NEXT_PUBLIC_MULTILINGUAL_ENABLED: "true" },
+    /**
+     * Setting the feedback widget mode here ensures the suite renders and checks
+     * the widget's markup on every run instead of relying on a one-time manual
+     * check. Uses "mock" so the suite never makes real network calls.
```
ashlsun: Dust: this comment feels a little redundant / verbose for what it's trying to communicate! Don't totally know what to make of "markup", "one-time manual check"

packages/static-site/types/feedback-widget.d.ts
```diff
@@ -0,0 +1,38 @@
...
+interface FeedbackWidgetAttributes
+  extends DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement> {
+  /** Destination for the widget's "contact us" link. */
+  "contact-link"?: string;
+  /** When `"true"`, ratings go to analytics only and skip the datastore. */
+  "only-save-rating-to-analytics"?: string;
+  /** When `"false"`, hides the disclaimer shown above the comment field. */
+  "show-comment-disclaimer"?: string;
+  /** When `"true"`, omits the email collection step. */
+  "skip-email-step"?: string;
```
ashlsun: Dust: possible to make these more explicitly "true" | "false"?

packages/static-site/messages/ar-US.json
```diff
@@ -6,6 +6,7 @@
   "layout": {
     "skipNavigationLabel": "مؤقت المنتج لاحقا سيتم",
     "mainContentId": "main-content",
+    "feedbackAriaLabel": "ملاحظات حول الصفحة",
```
ashlsun, line 9: Woah! I didn't know we had Arabic content on the static site. Is this file just placeholder for when we want to add it to the language switcher? (Just curious, not relevant to this PR)

packages/static-site/components/feedback/FeedbackWidget.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const { container } = await renderFeedbackWidget(englishMessages);
+
+    expect(container.querySelector("feedback-widget")).toBeNull();
+  });
+});
+
+describe("FeedbackWidget language announcements", () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+    vi.clearAllMocks();
```
ashlsun, line 121: Dust: It looks like each test adds their own languageEvents list and event listeners that do the same thing, and we don't remove the listeners after each test. IT might be good to remove them here, but the impact is prob negligible, so this is just dust!

packages/static-site/messages/ar-US.json
ashlsun: Woah! I didn't know we had Arabic content on the static site. Is this file just placeholder for when we want to add it to the language switcher? (Just curious, not relevant to this PR)

ashlsun (approved):
Looks great! 

Mocking so we can interact with it locally/on dev without actually submitting anything is a good idea! Feels like something the feedback-widget should try to support out of the box, maybe in the future. All comments I have are dust, nothing major, so feel free to ignore at your discretion.

packages/static-site/messages/ar-US.json
hinzed1127: >  Is this file just placeholder for when we want to add it to the language switcher? 

Yeah, exactly. This would still need to get vetted for accuracy, just LLM translations as a first pass placeholder

packages/static-site/components/feedback/FeedbackWidget.test.tsx
hinzed1127: Good call! Abstracted that to a `captureLanguageEvents` for the tests that do that

packages/static-site/playwright.a11y.config.ts
hinzed1127: good point. `mock` makes sense for these tests, so I'm just gonna remove it entirely

packages/static-site/components/feedback/FeedbackWidget.tsx
hinzed1127: Good callout. Right now, there's not any GTM set up within the static site. This is the first bit of "dynamic"-ish content that would probably be worth tracking something on, though. I'll flag it for follow up

hinzed1127: added a ticket [here](https://dev.azure.com/NJInnovation/BizX/_backlogs/backlog/BizX%20Team/Initiatives?workitem=17988)

PR #12963, opened 2026-08-17
feat: [AB#17964] Replace Starting Step 2 industry selection with intent selection

SirSamuelJoseph:
## Description

Enables a new onboarding flow where users who are starting a business can indicate whether they are "ready to start" or just "interested in learning." Right now that flow doesn't take you anywhere different, but that is future work within this mission. This PR enables this new flow and sets a new profileData field `learningBusiness` to indicate whether the user chose that option.

### Ticket

This pull request resolves [#17964](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17964).

### Approach

I created a new SwitchPage Component ( `StartingBusinessSwitchPage`). This will make it easier to transition fully to the new model when we are ready.

### Steps to Test
1. Set the feature flag `FEATURE_ENABLE_INTENT_SELECTION_FLOW` to true or enable it in the `StartingBusinessSwitchPage` component.
2. Onboarding as a new starting business. On step 2, you should not see the industry selection dropdown but two radio buttons.
3. You should not be able to proceed without selecting an option.
4. Selecting an option should take you to the dashboard as normal. In your userData, you should see the profileData field `learningBusiness` is true or false depending on your answer.

### Notes
There is a new feature flag added in this PR. I have added it to Bitwarden and Github values for all environments.

There is a migration to add a new field `learningBusiness` to profileData. The migration just adds that field and sets it to false for all users prior to this version. It will be set in the onboarding flow for new users

## Code author checklist


api/src/db/migrations/v195_add_learning_business_field_to_profile_data.ts
```diff
@@ -0,0 +1,1273 @@
...
+
+const migrate_v193Business_to_v194Business = async (
+  business: v193Business,
+): Promise<v194Business> => {
+  return {
+    ...business,
+    version: 194,
+    profileData: {
+      ...business.profileData,
+      learningBusiness: false,
```
SirSamuelJoseph: This is all that this migration is doing, setting it false for all existing users and adding it as an optional field for new users going forward.

web/src/components/data-fields/StartingBusinessIntent.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { FormControl, FormControlLabel, Radio, RadioGroup } from "@mui/material";
+import { ConfigType } from "@businessnjgovnavigator/shared/contexts";
+import { getProfileConfig } from "@/lib/domain-logic/getProfileConfig";
+import { useConfig } from "@/lib/data-hooks/useConfig";
+import { ProfileDataContext } from "@/contexts/profileDataContext";
+import { useFormContextFieldHelpers } from "@/lib/data-hooks/useFormContextFieldHelpers";
+import { DataFormErrorMapContext } from "@/contexts/dataFormErrorMapContext";
+import { FormContextFieldProps } from "@businessnjgovnavigator/shared/types";
+
+export const StartingBusinessIntent = <T,>(props: FormContextFieldProps<T>): ReactElement => {
```
SirSamuelJoseph, line 9: This is the main actual component being created. Its tests are in `onboarding-starting.test.tsx`

web/src/components/data-fields/StartingBusinessIntent.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const { Config } = useConfig();
+  const { state, setProfileData } = useContext(ProfileDataContext);
+  const contentFromConfig: ConfigType["profileDefaults"]["fields"]["learningBusiness"]["default"] =
+    getProfileConfig({
+      config: Config,
+      persona: state.flow,
+      fieldName: "learningBusiness",
+    });
+
+  const getValueFromState = (value?: boolean): string => {
```
ashlsun: dust: This function is only called once, right? It caught my eye because the name seemed a little opaque / generic, but once I read how it was used, I understood that it is meant to translate `state.profileData.learningBusiness` into either "LEARNING" or "READY". 

I wonder if instead of taking a param, it could just read directly from `state.profileData.learningBusiness`, since that makes the name make more sense on a first read.

web/src/components/onboarding/OnboardingFlows.tsx
```diff
@@ -49,16 +51,7 @@ export const onboardingFlows: Record<FlowType, OnboardingFlow> = {
       },
       {
         name: "industry-page",
```
ashlsun, line 53: Dust / note to clean up for later: this name will be inaccurate when the onboarding page is intent instead of industry

ashlsun (commented):
Will come back to finish reviewing after our meeting! Submitting comments I have so far!

One thing I'm trying to wrap my head around is whether `learningBusiness` should fit in `ProfileData` or a level higher. Each business has its own `profileData`, right?

My understanding of this intent selection page is to split users who wanted to get started on the interactive version of the tasks right away (so we make them sign up for MyNJ), vs users who just want to read our content in non-interactive guest mode. If someone has an account and is adding an additional business, then we could probably skip this intent question and take them to the interactive version. 

From what I'm seeing, it seems like this PR might allow people who already have an account to opt for the guest mode experience in each business they add? Not sure, will have to return and confirm!

web/.env-template
```diff
@@ -25,3 +25,4 @@ REDIRECT_URL=http://localhost:3000/loading
 SHOW_DISABLED_INDUSTRIES=
 USE_WIREMOCK_FOR_FORMATION_AND_BUSINESS_SEARCH=
 FEATURE_NAICS_INDUSTRY_DETECTION=
+FEATURE_ENABLE_INTENT_SELECTION_FLOW=
```
hinzed1127, line 28: [boulder] for the feature flag to work as constructed, we need to make sure it's also added to `env` block of the `nextConfig` in `next.config.ts`, i.e. like so:
```js
FEATURE_ENABLE_INTENT_SELECTION_FLOW: process.env.FEATURE_ENABLE_INTENT_SELECTION_FLOW ?? "false",
```
Alternatively, we could rename the flag, which is the more recent guidance [in the Next.js docs](https://nextjs.org/docs/pages/guides/environment-variables#bundling-environment-variables-for-the-browser), but the alternative is more in line with how we're already doing it

web/src/components/data-fields/StartingBusinessIntent.tsx
```diff
@@ -0,0 +1,90 @@
...
+          row
+        >
+          <div className={"max-width-half"}>
+            <FormControlLabel
+              style={{ alignItems: "center" }}
+              labelPlacement="end"
+              data-testid="starting-ready-business"
+              value={"READY"}
+              control={<Radio color="primary" />}
+              name={"ready-button"}
```
hinzed1127:
```suggestion

```
This should be deleted, as it currently causes a subtle accessibility bug where you can't navigate by keyboard between the 2 radio options, because it places the two radio buttons in separate radio groups. The name on the parent `RadioGroup` should be sufficient and this can be removed.

hinzed1127 (commented):
A couple blocking items, but both are easily fixed. Otherwise looks good!

Didn't dive super deeply into the question that Ashley raised, but probably worth validating the exact design there, as well.

web/src/components/data-fields/StartingBusinessIntent.tsx
SirSamuelJoseph: This makes sense to me, I'll update it

ashlsun (approved):
Read through again, I think the code looks good! 

I also set the feature flag and tested locally, and saw that this intent page shows for non logged in users and the industry page shows for logged in users.

I forgot, did we decide to not do the 2 column approach where the options are side by side? I see that they are vertical.

The next thing to do will be to make sure this selection place redirects to the right place, so Ready users get taken to sign up, and Learning users see the tasks. I think it makes sense to tackle that after this is merged and the page skeleton is in place.

PR #12980, opened 2026-09-04
chore: improve CSR generation script

seidior:
## Description

Updates `scripts/generate_csr.py` so the private key algorithm (RSA vs. ECDSA) is a real, documented CLI option instead of a commented-out code swap, and improves the script's developer experience end to end.

### Approach

- Added a `KEY_PRESETS` registry (`rsa-2048`, `rsa-3072`, `rsa-4096`, `ecdsa-p256`, `ecdsa-p384`), selectable via `--key` (default `rsa-2048`).
- Any subject flag not passed on the command line (country, state, locality, organization, organizational unit, common name, email) is now collected via an interactive prompt, instead of failing with a `required=True` argparse error.
- Renamed `--organizational_unit`/`--common_name` to the more conventional `--organizational-unit`/`--common-name`.
- Added `--output-dir` and `--force`. The script refuses to overwrite an existing private key or CSR unless `--force` is passed.
- Private key files are now written with owner-only (`0600`) permissions.
- Country codes are validated (ISO 3166-1 alpha-2) both at argument-parse time and at the interactive prompt, with a clear error message instead of a raw `cryptography` traceback.
- Filenames generated from the Common Name are sanitized against a broader set of unsafe characters (slashes, wildcards, etc).
- Expanded `scripts/test_generate_csr.py` from 2 to 28 tests, covering every key preset, CSR generation, country code validation, filename sanitization, interactive prompting (including re-prompt on invalid input), argument parsing, and end-to-end `main()` behavior (output directory, overwrite protection, `--force`, key file permissions).

No dependency changes; `cryptography` was already listed in `pyproject.toml`.

### Steps to Test

1. Run `python3 -m unittest discover -b -s scripts -p "test_generate_csr.py"` (or `yarn test:python`) and confirm all tests pass.
2. Run `./scripts/generate_csr.py --help` and review the new flags and key preset table.
3. Run `./scripts/generate_csr.py --common-name test.example.com --key ecdsa-p384 --output-dir /tmp/csr-test` and confirm a private key (mode `600`) and CSR are written to `/tmp/csr-test`.
4. Re-run the same command and confirm it refuses to overwrite the existing files without `--force`.

## Code author checklist

### DIFF
--- a/scripts/generate_csr.py
+++ b/scripts/generate_csr.py
@@ -1,13 +1,68 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
+"""Generate a private key and Certificate Signing Request (CSR).
 
-from cryptography import x509
-from cryptography.hazmat.primitives import hashes
-from cryptography.hazmat.primitives.asymmetric import ec, rsa
-from cryptography.hazmat.primitives import serialization
-from cryptography.x509.oid import NameOID
-from datetime import datetime
+Supports RSA and ECDSA keys via named `--key` presets. Any subject field not
+passed as a flag is collected with an interactive prompt, so the script can
+be run with zero arguments or fully scripted for automation. The private key
+and CSR are written to timestamped files (owner-only permissions on the key)
+and existing files are never overwritten unless `--force` is passed.
+
+Usage:
+    Fully interactive:
+        $ ./generate_csr.py
+
+    Fully scripted:
+        $ ./generate_csr.py --country US --state "New Jersey" --locality Newark \\
+              --organization MyOrg --organizational-unit IT \\
+              --common-name www.example.com --email admin@example.com \\
+              --san www.example.com example.com --key ecdsa-p384 \\
+              --output-dir ./certs
+
+    Mixed (prompts only for what's missing):
+        $ ./generate_csr.py --common-name www.example.com --key rsa-4096
+"""
+
+from __future__ import annotations
 
 import argparse
+import re
+import stat
+import sys
+from dataclasses import dataclass
+from datetime import datetime
+from pathlib import Path
+from typing import Callable, Final, NamedTuple, Union
+
+from cryptography import x509
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import ec, rsa
+from cryptography.x509.oid import NameOID
+
+__all__ = [
+    "KEY_PRESETS",
+    "DEFAULT_KEY_PRESET",
+    "PrivateKey",
+    "SubjectFields",
+    "generate_rsa_private_key",
+    "generate_ecc_private_key",
+    "generate_private_key",
+    "generate_csr",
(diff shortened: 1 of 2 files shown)


scripts/generate_csr.py
```diff
@@ -81,112 +173,285 @@ def generate_csr(
...
-            )
+    destination = Path(path)
+    destination.write_bytes(
+        private_key.private_bytes(
+            encoding=serialization.Encoding.PEM,
+            format=serialization.PrivateFormat.PKCS8,
+            encryption_algorithm=serialization.NoEncryption(),
         )
+    )
+    destination.chmod(_PRIVATE_KEY_FILE_MODE)
```
seidior, line 194: There's a small window between write_bytes creating the file (with whatever the process umask gives it) and this chmod locking it down to 0600. I could close that gap with os.open(path, os.O_CREAT | os.O_WRONLY | os.O_EXCL, 0o600) plus os.fdopen() instead, but for a script a dev runs once locally to produce a CSR, I didn't think the extra complexity was worth it. Calling it out in case anyone wants this hardened further before it's used somewhere with a stricter threat model.

scripts/generate_csr.py
```diff
@@ -81,112 +173,285 @@ def generate_csr(
...
+        organization=args.organization
+        or _prompt_required("Organization Name (e.g. company)"),
+        organizational_unit=args.organizational_unit
+        or _prompt_required("Organizational Unit Name (e.g. department or team)"),
+        common_name=args.common_name
+        or _prompt_required("Common Name (e.g. server FQDN or YOUR name)"),
+        email=args.email or _prompt_required("Email Address"),
+        san_dns_names=(
+            args.san
+            if args.san is not None
```
seidior, line 332: The None vs [] distinction here is doing double duty: --san omitted entirely leaves args.san as None so we prompt, but --san passed with zero names gives [] so we skip the prompt and use no SANs. That's why there's no default=[] on the --san argparse arg anymore like the old script had; a default would collapse both cases into the same value and remove the ability to explicitly opt out of the prompt.

scripts/generate_csr.py
```diff
@@ -81,112 +173,285 @@ def generate_csr(
...
+        print(f"Error: {error}", file=sys.stderr)
+        sys.exit(1)
+
+    private_key_path, csr_path = build_output_paths(
+        args.output_dir, subject.common_name
+    )
 
-    # Replace spaces and periods with underscores in the common name
-    formatted_common_name = args.common_name.replace(" ", "_").replace(".", "_")
+    if not args.force:
```
seidior, line 434: Added this guard mainly because build_output_paths() only encodes year and month in the filename, which I kept from the original script to match the naming convention used for already-issued certs. That granularity means two runs for the same common name in the same month used to silently overwrite the previous private key; --force is now required to do that on purpose.

seidior (commented):
Self-review.

PR #13001, opened 2026-09-15
feat: [AB#17983] Analytics tags for onboarding questions

SirSamuelJoseph:
### Ticket

This pull request resolves [#17983](https://dev.azure.com/NJInnovation/BizX/_workitems/edit/17983).

### Approach

This is the first PR in an attempt to update and fix analytics across the site. The new pattern is very similar to the old pattern but is strongly typed to make the events themselves cleaner- previously, every event had the same massive set of possible parameters in GA4 which made for confusing analysis.

The approach here is to start with onboarding, verify this approach works for the analytics team, and slowly spread it through the app. The old method is still here, and the events will remain as we slowly begin cleanup.

### Steps to Test
1. Go through onboarding flow locally. You should see events with the type `custom_analytics_event` being sent to GTM on:

- each page view (starting business only)
- form start
- form submission (starting business only)
- radio select
- next button click
- validation errors

### Notes

There still has to be a GTM deploy to have the events flowing to GA4. I will do that deploy once this is approved.

## Code author checklist


web/src/lib/utils/analytics.ts
```diff
@@ -346,6 +346,61 @@ export interface GTMEventData {
   status?: string;
 }
 
+type GTMCustomEventTypes =
```
SirSamuelJoseph, line 349: This file is where most of the substantive changes are. This section establishes the new pattern, with strongly typed events to ensure consistency.

web/src/lib/utils/analytics.ts
```diff
@@ -435,6 +490,13 @@ class GTMTracker {
       analytics.sendEvent(event);
     });
   }
+
+  custom_event_track(data: EventData): void {
```
SirSamuelJoseph, line 494: This is the new version of eventRunner.track(). It uses the same base layer but has the new event type `custom_analytics_event`. This allows us to use GTM triggers to send these to GA4 with the proper parameters

web/src/lib/utils/analytics.ts
```diff
@@ -553,20 +727,6 @@ export default {
         },
       },
     },
-    landing_page_hero_log_in: {
```
SirSamuelJoseph, line 556: The removed events weren't being called anywhere in the code, just some cleanup.

web/src/lib/utils/analytics-helpers.ts
```diff
@@ -199,13 +199,22 @@ const sendEssentialQuestionEvents = (newProfileData: ProfileData): void => {
...
+  const newProfileData = userData.businesses[userData.currentBusinessId].profileData;
   if (pageName === "industry-page" && newProfileData.industryId) {
     sendEssentialQuestionEvents(newProfileData);
   }
+
+  if (pageName === "persona-page" && newProfileData.businessPersona) {
+    analytics.event.onboarding.business_persona_selection_submit(newProfileData.businessPersona);
+  }
+
+  if (pageName === "business-intent" && userData.user.onboardedAsLearningUser !== undefined) {
```
ashlsun: If we deploy this before we turn on the feature flag for intent selection, we should account for the page being industry selection and make sure we don't send this event.

ashlsun (approved):
Tested with Tag Assistant and saw the custom_analytics_events come in at each step of onboarding. And the types make so much more sense this way. Great work! 

Is my understanding right that if we want to add more events in the future, we will need to set up new GTM tags and triggers to send everything to GA4?  It would be a good to add a how-to guide to our docs soon, to enable other devs to adopt this new approach.

web/src/lib/utils/analytics-helpers.ts
SirSamuelJoseph: Good catch! I'll add that

